PR #26, opened 2021-06-01
Change some ObservedObject to StateObject

PangMo5:
View is currently regenerated unwantedly under certain circumstances.
As a result, ViewModel, an ObservedObject, is regenerated and is making problems.
To resolve this issue, I changed it to StateObject, which was introduced in iOS 14.
ex. When the app enters the background and returns to the foreground.

See also: https://developer.apple.com/documentation/swiftui/managing-model-data-in-your-app


acvigue:
lgtm

acvigue (approved):

PR #43, opened 2021-06-10
Add Next Up Widgets

PangMo5:
## Changed
- Add Next Up Widgets (small, medium, large)
- Build a widget development environment
- Add App Groups
- Add Keychain Groups
- Add some Extensions

## Apply to the current project
App Groups and Keychain Groups are required to send and receive app data to iOS widgets.
These groups work with Apple Developer Identifiers and must be set up manually for your current developer account.
After setting each group, replace "*.dev.pangmo5.swiftfin.*" in the source to each ID.
App ID Prefix is required for Keychain group setup, like "4BHXT8RHFR.dev.pangmo5.swiftfin.keychainGroup"

## Side effects
Since the path of the Keychain and CoreData has been changed, the existing stored data in the app is destroyed.

## Known Issues
Currently, the app opens when the widget is clicked.
In the future, when a deep link is applied to the app, an action to the link to episode will be added.


acvigue:
Just a heads up - the CI now works and is a _bit_ more picky than Xcode. PRs are now set to be blocked unless all CI checks pass :)

Also - looks really good!

acvigue (approved):

WidgetExtension/WidgetEnvironment.swift
```diff
@@ -15,8 +15,8 @@ import UIKit
 final class WidgetEnvironment {
     static let shared = WidgetEnvironment()
     
-    var server: Server?
-    var user: SignedInUser?
+    var server: Server!
+    var user: SignedInUser!
```
PangMo5, line 19: Widgets can be created and accessed without a server and user.
This can cause a lot of crashes.
It would be better to explicitly declare this as "?".

acvigue: 93edcee7dd906649d9104ace65152af388eed830
resolved

PR #44, opened 2021-06-10
Apply SwiftFormat

PangMo5:
I applied the swiftformat rule currently used by my company.

Information about each rule is in the link and you can edit it yourself.
https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md

First, apply the `swiftformat .` command as a whole, and then

I think it would be good to use it for PR ci, etc.


acvigue (approved):

acvigue:
thanks! I took this and moved it into a ci action.

PR #78, opened 2021-06-19
Remove @Published used for input(keyboard) binding (side effect)

PangMo5:
Previously, input binding was received directly from ViewModel for TextField, etc.
As a result, there was a problem where the keyboard cursor was initialized because the view was reloaded when received input. 
and couldn't use the iOS remote keyboard in tvOS.

So I changed the related @Published to a combination of @State and CurrentValueSubject.

### DIFF
--- a/JellyfinPlayer tvOS/ConnectToServerView.swift
+++ b/JellyfinPlayer tvOS/ConnectToServerView.swift
@@ -10,12 +10,9 @@ import SwiftUI
 
 struct ConnectToServerView: View {
     @StateObject var viewModel = ConnectToServerViewModel()
-    @State
-    var username = ""
-    @State
-    var password = ""
-    @State
-    var uri = ""
+    @State var username = ""
+    @State var password = ""
+    @State var uri = ""
 
     var body: some View {
         VStack(alignment: .leading) {
--- a/JellyfinPlayer tvOS/ConnectToServerView.swift
+++ b/JellyfinPlayer tvOS/ConnectToServerView.swift
@@ -10,14 +10,20 @@ import SwiftUI
 
 struct ConnectToServerView: View {
     @StateObject var viewModel = ConnectToServerViewModel()
+    @State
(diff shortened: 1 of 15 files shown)


JellyfinPlayer tvOS/ConnectToServerView.swift
```diff
@@ -10,14 +10,20 @@ import SwiftUI
 
 struct ConnectToServerView: View {
     @StateObject var viewModel = ConnectToServerViewModel()
+    @State
+    var username = ""
```
acvigue, line 14: Could you put the property wrappers on the same line as the variables? This should fit with the rest of the project.

JellyfinPlayer/ConnectToServerView.swift
```diff
@@ -11,18 +11,25 @@ import KeychainSwift
 import SwiftUI
 
 struct ConnectToServerView: View {
-    @StateObject var viewModel = ConnectToServerViewModel()
+    @StateObject
+    var viewModel = ConnectToServerViewModel()
+    @State
```
acvigue, line 16: same here

JellyfinPlayer/LibrarySearchView.swift
```diff
@@ -12,7 +12,8 @@ import SwiftUI
 struct LibrarySearchView: View {
     @StateObject
     var viewModel: LibrarySearchViewModel
-
+    @State
```
acvigue, line 15: again

acvigue (changes requested):

JellyfinPlayer tvOS/ConnectToServerView.swift
PangMo5: Oh, I see.
It is probably modified when saving because of the SwiftFormat that I use personally.
Let me modify the related settings.

acvigue:
lgtm

PR #101, opened 2021-07-01
tvOS Video Player

stevebyatt10:
Up Next view in SwiftUI is going to be very difficult, initial views have been created but will come back to this when it is more important.

Changes:
- Polish now playing centre
- Fix media info panel design

Issues:
- Playing a transcoded video, VLC player struggles to obtain the remaining time.


acvigue:
Just did a checkout on this branch, some minor issues but I think the big issue (no remaining time for transcoded videos) could be solved by using the `RunTimeTicks` attr in the playback info response :)

PR #113, opened 2021-07-22
QOL - Set Theme

LePips:
Allows the user to set the app's appearance based upon "System", "Dark, or "Light".

"System" is default


PangMo5:
Thank you for the first contribution. 

Forcing the app to dark mode does not fit Human interface guideline.
https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/dark-mode

As you said, we need a setting to Dark Mode related.

If you want to update related PR, please update it.
If you don't, please close the PR and post it as an issue.

LePips:
While forcing the app into dark mode does go against the interface guidelines, it's just a temporary workaround for forcing Jellyfin's pattern of a dark theme and is generally acceptable to do.

I am able to do the work to supply the option in the settings as I started some work on it but I intend that to be its own PR/work.

PangMo5:
It's literally a recommendation, so it's not a reject to the AppStore, but it's enough to break the UX.

Anyway, I don't think it's a good idea to force dark mode so that users can't set it up.

we can a much better things, like add setting to Dark Mode related

LePips:
Apple understands if an app wants to run in a single mode and it does not necessarily break the UX in the meantime that we have it set to dark.

Anyways, I can look at doing the work to set the appearance in settings and update this PR.

acvigue:
but... the OS literally has a setting for light/dark/auto mode. What do you think Kwangmin?

LePips:
The OS does, however many users would like to use applications in a mode that differs from their OS. For example, using Discord in dark mode while having my phone use light mode everywhere else.

This is standard design in many apps.

PangMo5:
> but... the OS literally has a setting for light/dark/auto mode. What do you think @PangMo5

It doesn't exactly fit Apple's HIG. 🤔 
>Comply with the appearance mode people choose in Settings. If you offer an app-specific appearance mode option, >you create more work for people because they have to adjust more than one setting. Worse, they may think your >app is broken because it doesn't respond to their systemwide appearance choice.


But I think "it's okay" to give the user a choice, if unless forcing a certain Appearance mode.
And The default value should `Auto` that interlocking with system appearance mode.

LePips:
If you guys would like examples of apps that implement this:

Facebook Messenger, Reddit, Youtube, Twitch, Evernote, TikTok, Twitter

acvigue:
> But I think "it's okay" to give the user a choice, if unless forcing a certain Appearance mode.
And The default value should Auto(Naming may vary) that interlocking with system appearance mode.

I agree, a picker in the Settings tab such as "Dark"/"Light"/"System" where System would be default.

acvigue (approved):

PR #118, opened 2021-07-23
update Zh_Hans translation

charleypeng:
update Zh_Hans translation


acvigue:
Thank you for taking the time to translate SwiftFin! We normally prefer translations to come in through Weblate, did you run into any issues using our Weblate?

PR #122, opened 2021-07-31
Fix checkmark view

LePips:
Scales the background shape for the checkmark symbol such that it fills in the center transparent checkmark while staying inside the circle.

iOS 15 _looks like_ it will allow us to set the transparent checkmark white but I'm not entirely sure, we'll figure it out when we get there.

### DIFF
--- a/JellyfinPlayer/Components/PortraitItemView.swift
+++ b/JellyfinPlayer/Components/PortraitItemView.swift
@@ -46,8 +46,8 @@ struct PortraitItemView: View {
                             if item.userData?.played ?? false {
                                 Image(systemName: "checkmark.circle.fill")
                                     .foregroundColor(.accentColor)
-                                    .background(Color(.white).clipShape(Circle().scale(0.8)))
-                                    .clipShape(Circle())
+                                    .background(Color(.white))
+                                    .clipShape(Circle().scale(0.8))
                             } else {
                                 if item.userData?.unplayedItemCount != nil {
                                     Capsule()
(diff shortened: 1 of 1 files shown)


JellyfinPlayer/Components/PortraitItemView.swift
```diff
@@ -47,7 +47,7 @@ struct PortraitItemView: View {
                                 Image(systemName: "checkmark.circle.fill")
                                     .foregroundColor(.accentColor)
                                     .background(Color(.white))
```
acvigue, line 49: If we're thinking in the mindset of setting background/foreground colors directly on the icon, this icon supports multicolor mode from iOS 14 & up (i.e. the intended targets)
Screen Shot 2021-07-31 at 7 50 24 AM

acvigue: I'd rather do this then continue with my hacky way of doing this :p

LePips: The multicolor support in iOS 14 **only** allows for it to be that blue color with the white checkmark.

In iOS 15, we can change the palette of the colors for the symbol but it is only supported in iOS 15.

Look at section: _How to create a custom SF Symbols palette_
https://www.hackingwithswift.com/articles/237/complete-guide-to-sf-symbols

LePips: The hack you did was good and it's necessary here

acvigue: > The multicolor support in iOS 14 _only_ allows for it to be that blue color with the white checkmark.
> 
> In iOS 15, we can change the palette of the colors for the symbol but it is only supported in iOS 15.
> 
> Look at section: _How to create a custom SF Symbols palette_
> https://www.hackingwithswift.com/articles/237/complete-guide-to-sf-symbols

Oh, didn't know that. That wouldn't work as we use the purple accent color.
this lgtm!

acvigue (approved):

PR #126, opened 2021-08-05
Fix for Swift 5.4

LePips:
The compiler flag feature that was previously used is only in Swift 5.5 as noted here:
https://github.com/apple/swift-evolution/blob/main/proposals/0308-postfix-if-config-expressions.md

Please stop using beta Xcodes/languages for official development.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Check out our [Weblate instance](https://translate.jellyfin.org/projects/swiftfi
 
 ## ⚙️ Development
 
-Xcode 13.0 with command line tools.
+Xcode 12.5 with command line tools.
 
 ### Build Process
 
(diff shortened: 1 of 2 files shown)


acvigue:
lol i kind of have to use the beta bc my phone is on the iOS 15 developer preview, so I can't really change it.

acvigue:
and, there's really no reason that we *shouldn't* be using the beta frameworks, as we can plan for what's coming up & test for issues that may arise in the future.

LePips:
I use the beta too since my phone is also iOS 15, all I'm saying is:

Don't push changes that will break the project with current stable tech.

It's just good practice. For testing with beta tech, we can follow the practice of making a branch for it until official release.

LePips:
Also, this change is so small. If it were a larger feature I'd understand and go with it.

PR #127, opened 2021-08-06
SeriesItemView redesign

PangMo5:
# Changes
- SeriesItemView feature and UI redesign

Additionally, I think it seems better to use even number of padding for UI.
And recommended to use spacing param when using `VStack`, `HStack`, `LazyVStack`, and `LazyHStack`.
It is difficult to implement the desired UI when using the default value of `nil`.

These are met in this PR.
This will make it look more modern and regular UI.

# Before

# After

### DIFF
--- a/JellyfinPlayer/SeriesItemView.swift
+++ b/JellyfinPlayer/SeriesItemView.swift
@@ -9,27 +9,191 @@ import SwiftUI
 
 struct SeriesItemView: View {
     @StateObject var viewModel: SeriesItemViewModel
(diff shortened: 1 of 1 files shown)


PangMo5:
I don't think SonarCloud is suitable for SwiftUI lol

acvigue:
THANK YOU. ive been meaning to do this. SonarCloud is useful for certain things and I need to finish setting it up properly

acvigue (approved):

PR #129, opened 2021-08-06
UX - TvOS Rearrange Media Buttons

LePips:
Rearranges the media buttons for some better ux.

Previously instead of requiring users to click -> scroll to play -> click they can now click straight through to play.

Also breaks the shared row of buttons into its own view.


acvigue (changes requested):
Commit doesn't belong in this PR.

LePips:
Sorry, will be consolidated after the other PR is dealt with.

PR #143, opened 2021-08-15
Fix crash when appearing some ItemView

PangMo5:
- remove some force unwrapping

### DIFF
--- a/Shared/Extensions/APIExtensions.swift
+++ b/Shared/Extensions/APIExtensions.swift
@@ -12,70 +12,71 @@ import UIKit
 // 001fC^ = dark grey plain blurhash
 
 extension BaseItemDto {
-
     // MARK: Images
+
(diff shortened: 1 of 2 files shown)


Shared/Extensions/APIExtensions.swift
```diff
@@ -124,17 +129,18 @@ extension BaseItemDto {
             return formatter
         }()
 
-        let text = timeHMSFormatter.string(from: Double(self.runTimeTicks! / 10_000_000)) ?? ""
+        guard let runTimeTicks = runTimeTicks,
+              let text = timeHMSFormatter.string(from: Double(runTimeTicks / 10_000_000)) else { return "" }
```
PangMo5, line 133: Actually, this is the only place that's changed. :p

acvigue (approved):
LGTM!

PR #149, opened 2021-08-18
Add custom time skip durations

LePips:
Allows customization of time skip durations.

5 second forward/backward symbol is new in iOS 15. Once released, literally just uncomment the work for it.


Shared/Objects/VideoPlayerJumpLength.swift
```diff
@@ -0,0 +1,59 @@
...
+  * SwiftFin is subject to the terms of the Mozilla Public
+  * License, v2.0. If a copy of the MPL was not distributed with this
+  * file, you can obtain one at https://mozilla.org/MPL/2.0/.
+  *
+  * Copyright 2021 Aiden Vigue & Jellyfin Contributors
+  */
+
+import UIKit
+
+enum VideoPlayerJumpLength: Int32, CaseIterable {
```
PangMo5, line 12: In [Swift Docs](https://docs.swift.org/swift-book/LanguageGuide/TheBasics.html#ID319])

>Int
>In most cases, you don’t need to pick a specific size of integer >to use in your code. Swift provides an additional integer type, >Int, which has the same size as the current platform’s native word >size:
>
>On a 32-bit platform, Int is the same size as Int32.
>On a 64-bit platform, Int is the same size as Int64.
>Unless you need to work with a specific size of integer, always >use Int for integer values in your code. This aids code >consistency and interoperability. Even on 32-bit platforms, Int can store any value between -2,147,483,648 and 2,147,483,647, and is large enough for many integer ranges.

If you are using `Int32` because the value is fixed, `UInt8` is more suitable lol
However I recommend using `Int`.

PangMo5: And I recommend creating a property such as `length` rather than using rawValue.

```swift
var length: Int {
switch self {
	case thirty:
	return 30
	...
}
```

Shared/Extensions/DefaultsExtension.swift
```diff
@@ -17,4 +17,6 @@ extension Defaults.Keys {
     static let autoSelectSubtitlesLangCode = Key<String>("AutoSelectSubtitlesLangCode", default: "Auto")
     static let autoSelectAudioLangCode = Key<String>("AutoSelectAudioLangCode", default: "Auto")
     static let appAppearance = Key<String>("appAppearance", default: AppAppearance.system.rawValue)
+    static let videoPlayerJumpForward = Key<Int32>("videoPlayerJumpForward", default: VideoPlayerJumpLength.thirty.rawValue)
+    static let videoPlayerJumpBackward = Key<Int32>("videoPlayerJumpBackward", default: VideoPlayerJumpLength.thirty.rawValue)
```
PangMo5, line 21: https://github.com/sindresorhus/Defaults#enum-example
`Defaults` supports enum.
You can use this to remove some force unwrapping.

JellyfinPlayer/VideoPlayer.swift
```diff
@@ -76,7 +76,15 @@ class PlayerViewController: UIViewController, GCKDiscoveryManagerListener, GCKRe
...
     let playbackSpeeds: [Float] = [0.25, 0.5, 0.75, 1.0, 1.25, 1.5, 1.75, 2.0]
-
+    var jumpForwardLength: VideoPlayerJumpLength {
+        let storedJumpForwardLength = Defaults[.videoPlayerJumpForward]
+        return VideoPlayerJumpLength(rawValue: storedJumpForwardLength)!
+    }
+    var jumpBackwardLength: VideoPlayerJumpLength {
+        let storedJumpBackwardLength = Defaults[.videoPlayerJumpBackward]
+        return VideoPlayerJumpLength(rawValue: storedJumpBackwardLength)!
+    }
```
PangMo5, line 86: https://github.com/jellyfin/Swiftfin/pull/149/files#r690927393

Shared/Extensions/DefaultsExtension.swift
PangMo5: https://github.com/jellyfin/Swiftfin/pull/149#discussion_r690927465

Shared/Objects/VideoPlayerJumpLength.swift
LePips: It needs to be Int32 because that's what the media player needs, I wouldn't be using it otherwise.
I'm fine to change it to `length` variable.

Shared/Extensions/DefaultsExtension.swift
LePips: Thanks for pointing this out. Actually, in order to use this, it would be better to use the Int32 value instead of `length` var as I said I would change [here](https://github.com/jellyfin/Swiftfin/pull/149#discussion_r690931967)

Shared/Objects/VideoPlayerJumpLength.swift
PangMo5: > It needs to be Int32 because that's what the media player needs, I wouldn't be using it otherwise.

Oh, I understand.

Shared/Extensions/DefaultsExtension.swift
PangMo5: I prefer to use `String` rawValue, as in the example above. 
It's your choice.

LePips: Contextually, this is a length so it makes sense to be a number.

acvigue: > Contextually, this is a length so it makes sense to be a number.
I agree.

acvigue: Also, sorry, I've been swamped with schoolwork already. I'm lucky to get 10 minutes a day to devote to this app now. It may be different when my teachers decide to cool off lol.

acvigue (approved):
lgtm

PR #150, opened 2021-08-18
Temporarily remove widget

LePips:
Remove the widget from being shipped with the app. Right now, it doesn't have full functionality and I don't want something to be shipped that people will complain about since it doesn't work.

Just a temporary measure until it's in a good spot.

### DIFF
--- a/WidgetExtension/NextUpWidget.swift
+++ b/WidgetExtension/NextUpWidget.swift
@@ -337,7 +337,9 @@ struct NextUpWidget: Widget {
         }
         .configurationDisplayName("Next Up")
         .description("Keep watching where you left off or see what's up next.")
-        .supportedFamilies([.systemSmall, .systemMedium, .systemLarge])
+        .supportedFamilies([])
+        // Removed officially until working
+        // .supportedFamilies([.systemSmall, .systemMedium, .systemLarge])
     }
 }
 


acvigue:
@PangMo5

PangMo5:
What's the problem?

PangMo5:
If it's not Deeplink supported, you can create the function and post PR. Why remove the widget?
And for now, I think it's enough just to show the information.

LePips:
I just don't think it should be _shipped_ because right now it _doesn't work_. I'm not saying remove it all together, just remove pain points for people testing coming to us saying it doesn't work.

We should work towards a point for a Minimum Viable Product (MVP) which can be delivered on the App Store. We should come together and make this list.

PangMo5:
> I just don't think it should be _shipped_ because right now it _doesn't work_. I'm not saying remove it all together, just remove pain points for people testing coming to us saying it doesn't work.
> 
> We should work towards a point for a Minimum Viable Product (MVP) which can be delivered on the App Store. We should come together and make this list.

Swiftfin is still a lot left until uploaded to App Store, and deep link support will be before that.

There are a lot of bugs in the media player right now, are you going to remove media player too?

Just create an issue.

LePips:
It's all good. This just isn't core to the functionality of the app, like the video player. I'll create the issue.

PR #152, opened 2021-08-20
Video Player - Positive Time Ticker

LePips:
Adds the positive time (time spent watching media) to the video player.

From a UI perspective, this makes the video bar symmetrical with the positive on the left and the negative on the right, also centering the time slider. Making it look ✨ _nice_ ✨.

From a UX perspective, this allows the watcher to see how much they have seen already. This is especially handy with anime and shows with intros as Jellyfin doesn't have "Skip intro" support (...yet, it's being discussed here and there). This is overall a good feature when scrubbing to a specific section of the media.

This _also_ solves another problem: UI for scrubbing. Some players use thumbnails that pop up above the bar to show the picture at that current second (Youtube, when it works right) and some players pop up a view showing the seconds being scrubbed to (Quicktime I believe). I believe by just adding the positive time ticker we don't need anything to pop up.

The next steps for scrubbing are:
- try and implement high speed scrubbing, sounds tedious but doable
- better UI while scrubbing. My thoughts are to fade out the time jumps, play button, and dark overlay while scrubbing. This design mirrors some other players. However, if we want to implement thumbnails while scrubbing that's okay too but an entire feature of its own.


acvigue (approved):
constraints look good, merging.

PR #153, opened 2021-08-20
Settings - Advanced Server View

LePips:
Moves the Server and User information to the top of the Settings, which I argue are the most important settings.
Also adds a view to show the server public settings and refresh the library from within the app.

For the user row which just has the username, it is the beginning of work to show more user information. I'm looking at mirroring something like the iCloud row in the iOS Settings app. I started working on it but the JellyfinAPI isn't getting and caching images correctly from the server. Spent too much time on that.

Also removes the ability to "Switch user" because it straight up didn't work as intended. I will look at that as well when creating the user view.

I wanted this to update the `HomeView` as well. However, the source of truth for media in the app is in `HomeViewModel`. I kind of want to restructure how we use singletons and view models because we're breaking MVVM design paradigms a bit (my fault too). This can be addressed by pull down to refresh as another feature.


Shared/Extensions/ColorExtension.swift
```diff
@@ -0,0 +1,16 @@
...
+  * file, you can obtain one at https://mozilla.org/MPL/2.0/.
+  *
+  * Copyright 2021 Aiden Vigue & Jellyfin Contributors
+  */
+
+import SwiftUI
+
+extension Color {
+    
+    static let JellyfinPurple = Color(red: 172 / 255, green: 92 / 255, blue: 195 / 255)
```
PangMo5, line 14: Wouldn't it be better to use camelCase?

LePips: Yeah you're right

acvigue (approved):
looks ok to me. 
ty.
currently doing english hw 😵‍💫 :p

PR #157, opened 2021-08-25
Video Player - Refactor

LePips:
I'm going to refactor the video player with an explicit state system and view model for actual MVVM. The video player is the most important part of the app so it will be large but it doesn't have to be scary.

This will introduce some new visual design and UX elements and fix some current bugs.

Draft PR is up to track work.


LePips:
Removed for now because I won't get to this for a long time.

PR #165, opened 2021-09-01
Item View - Refactor

LePips:
I took some time to refactor how items are presented.

## Why?
The views for all the items are the same, except for `Season`. The original files for the views were large and ugly and would sometimes crash the type checker on my machine so I couldn't build for a little bit.

A lot of refactoring and consolidating code occurred and this just works nicely overall.

Enforced _better_ MVVM.

## Design/UX Changes
- Play button is shown when looking at a `Series`. Currently, this is enabled **only** when the specific series has a valid "Next Up" episode. I need to add implementation to check if an episode is currently being watched in the series and go to that, but that logic will come with some other work I will get to. The button will be visibly disabled in the state where there is not a valid "Next Up" episode
- Play button is shown when looking at a `Season`. This will play the next unwatched episode. If all episodes have been watched in the season, it will play the first episode of the season.
- Added `Cast & Crew`. This was previously just `Cast` but I added some valid crew as well and made them portraits. I also added logic to only grab the "first role" for a cast member because sometimes a cast member will have a super long role list. Also made the cards for the cast portrait because the small squares would awkwardly cut off faces. Overall, attempts were made to make text not truncate as much. Also, if a picture cannot be grabbed for the crew or cast member, it will display their initials instead of a blank card.
- Added pills. Recommended from the chat. `Genre` and other items we want will be rendered as little pills. This shows the user that these are selectable.
- Redid list design when looking at a `Season`. I think it looks better because episode titles aren't truncated as much which are often longer. Also looks a lot better on iPad.
- Made item title a little bigger, bringing more focus to the media.
- Moved item runtime to its own line (except for episode card), I think it's a very important metric when looking at media and needed to be more explicit.
- Customized the back button for the navigation bar. This was part of some work I attempted by making the navigation bar clear until you have scrolled down a certain offset and then appear blurred. Nothing currently works in SwiftUI for that to work.
- Removed item titles from navigation bar. Was redundant _but_ if I get the offset blur working as just previously described, I will also animate in/out the item title with the blur.

## Bug
This introduced a bug I haven't taken the time to look at _but_ want this work in and will file the bug to fix.

After playing media and exiting, when in portrait on an iPhone there will be a larger empty space at the top of the scroll view.

## Thoughts for Future Features
- Make the item view customizable. To me, I don't care about the cast or genres when looking at my media all of the time. I want the ability to just focus on my movie/show and just that. Would be customized through `Settings` and I have thoughts about how that would happen but this would low priority to me.


PangMo5:
That's really awesome! 😎

JellyfinPlayer/Components/PillHStackView.swift
```diff
@@ -0,0 +1,60 @@
...
+        VStack(alignment: .leading) {
+            Text(title)
+                .font(.callout)
+                .fontWeight(.semibold)
+                .padding(.top, 3)
+                .padding(.leading, 16)
+            
+            ScrollView(.horizontal, showsIndicators: false) {
+                HStack {
+                    ForEach(items, id: \.title) { item in
```
PangMo5, line 32: Wouldn't it be good to replace VStack and HStack that use `ForEach` with LazyStack and LazyHStack?
If there's no layout issue

JellyfinPlayer/Components/PortraitHStackView.swift
```diff
@@ -0,0 +1,87 @@
...
+        VStack(alignment: .leading) {
+            topBarView()
+            
+            ScrollView(.horizontal, showsIndicators: false) {
+                VStack {
+                    Spacer().frame(height: 8)
+                    HStack(alignment: .top) {
+                        
+                        Spacer().frame(width: 16)
+
```
PangMo5, line 46: https://github.com/jellyfin/Swiftfin/pull/165/files#r699876863

JellyfinPlayer/Components/CardVStackView.swift
```diff
@@ -0,0 +1,105 @@
...
+                        .foregroundColor(.jellyfinPurple)
+                }
+            }.padding(2)
+            .opacity(1)
+        }
+    }
+    
+    var body: some View {
+        VStack {
+            ForEach(items, id: \.id) { item in
```
PangMo5, line 47: https://github.com/jellyfin/Swiftfin/pull/165/files#r699876863

JellyfinPlayer/Components/PillHStackView.swift
PangMo5: There are many differences in performance with and without `Lazy`, even with simple-structured views. Because `HStack` and `VStack` loads all inner views when load a parent View.

acvigue:
will look at later today - at a volleyball game.

acvigue:
looking at this now.

LePips:
@acvigue 
Hey, sorry, I know that you're very busy with school and don't mean to be pressing. Could this be approved and merged?

acvigue:
Yes. I looked at it today and didnt see any problems with it

acvigue (approved):

G10vanni74:
Will the addition of Cast & Crew in the application be available in the betas of the application or after the release

Do you have screenshots of what it could look like on apple TV or is it the same on iPad ?

LePips:
Cast & Crew will be implemented before release.

G10vanni74:
Great. thank you

PR #169, opened 2021-09-21
iOS 15 todos and force iOS 15 minimum

LePips:
I went back to get some todos that I have left.

I also manually forced iOS 15 as minimum target. Since this year's release from Apple is different since people can still be on iOS 14 and get updates, this has ruffled the feathers of all the iOS developers I have talked to (even some working at Apple). Many came to the conclusion to just force iOS 15 as a minimum for future updates just like they were doing previously. This work is **necessary** for the 5-second skip SFSymbol and with other symbol work we need throughout the app. There are other technical features released in iOS 15 I'm sure that we will want to use as well. Building with iOS 14 and iOS 15 work also throws a lot of warnings throughout all the packages that we use, so this also cleans that up.

I am strongly on board with mandating iOS 15 minimum as done here.

If there is any disagreement on this, let's have a discussion here.


PangMo5:
I agree.
And I think Swiftfin is a free open-source project, not a commercial project, so we don't have to be restricted by the iOS version.

acvigue (approved):

PR #155, opened 2021-08-21
Apply Coordinator Pattern and Add Deep-Links

PangMo5:
With the Coordinator pattern,
- Helps to single-responsibility principle (SRP).
- It collects scattered navigation processing codes into one place.
- Makes the `View` lighter.
- etc...

# Coordinators
- [x] ConnectToServerCoodinator
- [x] HomeCoordinator
- [x] LibraryListCoordinator
- [x] MainCoordinator
- [x] MainTabCoordinator
- [x] SearchCoordinator
- [x] SettingsCoordinator
- [x] LibraryCoordinator
- [x] ItemCoordinator
- [x] FilterCoordinator
- [x] VideoPlayerCoordinator

# Deep-Links
- [x] jellyfin://Users/`{UserID}`/Items/`{ItemID}`

Feel free to leave any advice on the current structure!

PR #172, opened 2021-09-29
tvOS library work - add All Media tab

jhays:
- Adds tvOS version of LibraryListView under All Media tab. Allows access of TV Shows and Movies. 
- Adds item title to `PortraitItemElement`, increases height of `LatestMediaView` to fit added titles
- Adds a bunch of `.ignoreSafeArea()` to ensure background image & video playback fills full screen on tvOS
- Removed a force unwrap that caused a crash from `VideoPlayerViewController`
- Increased tvOS deployment target to 15


JellyfinPlayer tvOS/LatestMediaView.swift
```diff
@@ -42,15 +42,13 @@ struct LatestMediaView: View {
                 LazyHStack {
                     Spacer().frame(width: 45)
                     ForEach(items, id: \.id) { item in
-                        if item.type == "Series" || item.type == "Movie" {
```
jhays, line 45: I don't really know why, but I had to remove this because Xcode started choking on this view and would not build. I tried commenting out various parts of this until I figured out that this was the culprit. 

If this filtering is important, could we move it out of this view and filter down the items list before it gets here? Let me know its importance and if there is another strategy that could be used.

JellyfinPlayer tvOS/LibrarySearchView.swift
```diff
@@ -0,0 +1,119 @@
...
+            return viewModel.showItems
+        default:
+            return []
+        }
+    }
+}
+
+private extension ItemType {
+
+    var localized: String {
```
LePips, line 107: Could you move this to where `ItemType` is declared? This is good for now but we need to move to using the `ItemType` as declared in `BaseItemDtoExtensions`

LePips (changes requested):

JellyfinPlayer tvOS/LatestMediaView.swift
PangMo5: https://github.com/jellyfin/Swiftfin/pull/155/commits/a46ed4592539ad4e3fe1a71ae4136e342b8c7212

The same part of iOS has been modified in this commit. 
It has been modified to filter at the API call timing, so I don't think it matters if you just erase it.

LePips (approved):

PR #174, opened 2021-10-04
Add menu of Route function to ItemView

PangMo5:
Solved the inconvenience of moving to the series or season in EpisodeView or SeasonView.

https://user-images.githubusercontent.com/16532526/135886927-57c2f52a-7526-4548-a82f-728e23bd2b25.mp4

### DIFF
--- a/JellyfinPlayer/ItemView/ItemView.swift
+++ b/JellyfinPlayer/ItemView/ItemView.swift
@@ -55,6 +55,39 @@ private struct ItemView: View {
         }
     }
 
+    @ViewBuilder
+    var toolbarItemContent: some View {
+        switch viewModel.item.itemType {
+        case .season:
+            Menu {
+                Button {
+                    (viewModel as? SeasonItemViewModel)?.routeToSeriesItem()
+                } label: {
+                    Label("Show Series", systemImage: "text.below.photo")
+                }
+            } label: {
+                Image(systemName: "ellipsis.circle")
+            }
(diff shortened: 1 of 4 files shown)


JellyfinPlayer/ItemView/ItemView.swift
```diff
@@ -55,6 +55,39 @@ private struct ItemView: View {
         }
     }
 
+    @ViewBuilder
+    var toolbarItemContent: some View {
+        switch viewModel.item.itemType {
+        case .season:
+            Menu {
+                Button {
+                    (viewModel as? SeasonItemViewModel)?.routeToSeriesItem()
```
LePips, line 64: I understand why we need this pattern but I'm not that _comfortable_ with it. If this is the route we have to go that's okay but I want to see if I can do anything different. Would you mind making your branch public so that I can play with it for just a few minutes?

PangMo5: It would be nice if there was a better way 😆 
It is already public to 'PangMo5/ItemView-Navigation' in 'jellyfin/Swiftfin' and is feel free to fix.

LePips: Yeah, I'm too lazy to try anything. LGTM 👍 

In the future, I would recommend having work on your own branch and it's okay if it was an accident. This one was on the project's `upstream`.

LePips (approved):

PR #206, opened 2021-10-26
SwiftFormat

LePips:
Adds a large SwiftFormat ruleset to the project and also has the new changes from running.

This branch **has** to be merged before the Github Action is created. This is because the `--header "..."` rule is used and had some troubles with replacing our old header so I had to manually go in and delete the old header in almost all of the files.

The new header is as such and is also set in Xcode:
```//
// SwiftFin is subject to the terms of the Mozilla Public
//
// Copyright (c) 2021 Jellyfin & Jellyfin Contributors
//
```

SwiftFormat doesn't have support for (the more preferred way imo) `/* */` multiline comments.

A bunch of rules were explicitly mentioned just to make sure they were called, more rules can be added.

This is a draft at the moment because the `redundantSelf` may have caused some side effects. I will have to undo all formatting and then reformat again to fix this.

### DIFF
--- a/.swiftformat
+++ b/.swiftformat
@@ -1,5 +1,7 @@
 # version: 0.47.5
 
+--swiftversion 5.4
+
 --indent tab
 --tabwidth 4
 --xcodeindentation enabled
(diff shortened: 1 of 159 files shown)


LePips:
Closed for the updated PR: https://github.com/jellyfin/Swiftfin/pull/212

PR #211, opened 2021-11-03
Multi Server URL

LePips:
When I refactored the entire underlying data storage stack I deliberately made it impossible to add a server with the same server ID, even if you are attempting to attempt to the same server via different URLs (like a local and a public address).

This is the work that finishes what I had envisioned for accessing a server. You will be able to add multiple URLs to the same stored server and change which URL is being used for a connection.

This work will just implement it on iOS at first until I get to an entire tvOS re-design for the login flow.

This work will also change the underlying data storage model **which will remake the entire model and crash the app the first time it is opened after updating**. The package being used, CoreStore, allows creating migrations between old and updated models. I will not worry at all about migrations because this app is in TestFlight and not in production. I will announce the crash when the TestFlight build with this feature rolls out.

## Future Work
- Allow deleting URLs
- Go to User Login after adding a duplicate URL instead of popping
- Allow auto-detection via server pings to find suitable URL, then prompt via toast or banner that the URL was switched via auto-detection. Will have to discuss this design choice

I need to change how the HomeViewModel gathers media because the calls don't cancel correctly and doesn't show an error if something went wrong. This is something I have slightly started work for but haven't gotten far.

The ServerDetailView from the UserListView looks odd, but overall will have to change how this entire flow looks.


LePips:
To be merged after https://github.com/jellyfin/Swiftfin/pull/187 is merged and then main is pulled into here

LePips:
@acvigue, this is now ready to be merged

LePips:
@acvigue Could this be merged, so then I could do the final work on the Swiftformat and then that can be merged afterwards?

anthonylavado:
I'm almost done reviewing, will advise soon.

LePips:
Thanks, I just have to localize the new text

acvigue:
> Thanks, I just have to localize the new text

Looks good so far, once it's localized, I'll do a final review!

LePips:
@acvigue Sorry for the long wait, this is now finally ready.

PR #212, opened 2021-11-03
Actual SwiftFormat

LePips:
Had to recreate because of the `redundantself` rule made some side effects. Was too lazy to figure out rollback and everything so made a new branch.

Original: https://github.com/jellyfin/Swiftfin/pull/206

After it's merged I'll create the Github action for this.


PangMo5:
To avoid conflict, it would be nice to merge #187 PR into the main and then apply SwiftFormat again.

The work is almost done, so please wait a little bit.

LePips:
Will probably scrap this and just keep swiftlint

LePips:
Due to time constraints and big-implementations like live tv and player re-factor, we'll get to this after those are done ... sometime.

PR #224, opened 2021-11-12
[tvOS] Top Shelf Extension

stevebyatt10:
Bringing this over from some work I did a few months ago, adding a top shelf extension to display "Continue Watching" items.

Originally it worked using the old session manager but now using the new system it fails to retrieve a current user id. Draft PR until it can successfully work with the new session manager.

### DIFF
--- a/JellyfinPlayer tvOS/JellyfinPlayer tvOS.entitlements
+++ b/JellyfinPlayer tvOS/JellyfinPlayer tvOS.entitlements
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>com.apple.security.application-groups</key>
+	<array>
+		<string>group.swiftfinstore-defaults</string>
+	</array>
+</dict>
+</plist>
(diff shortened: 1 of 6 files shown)


LePips:
I think it may be better for you to start from scratch because this also does a lot of moving around the files. 

I mean at the time of writing this comment you only have one new file, so just create a branch off of main with that work.

stevebyatt10:
> Sorry, idk what I was thinking. Those are linking files to the extension.

Yeah all good haha, this has been branched from main and with a new target. 

The issue seems to be when initialising the session manager, it fails to retrieve the last logged in user from core data. Not sure if there are app groups or anything anymore which might solve it?

LePips:
I'll have to see exactly what app startup calls are made for a top shelf extension.

We initialize the data stack in the AppDelegate, and then the first time that the main SessionManager singleton is used it will initialize with the stored user.

LePips:
Maybe in the ContentProvider initializer you'll have to lazily setup the data stack like what happens in the AppDelegate because I don't believe the AppDelegate gets used here.

stevebyatt10:
> Maybe in the ContentProvider initializer you'll have to lazily setup the data stack 

I have tried what you mentioned but still didn't get any results, I enabled app groups and found that it did then retrieve the user id but ran into an issue with fetching from the datastack from outside the main thread.

Tried dispatching to the main thread but at that point just failed to retrieve the actual stored user object.

LePips:
It may be because the UserDefault used to store the ID of the current login may not be part of the App Group?

stevebyatt10:
```
 [🔵 DEBUG] ContentProvider#90:loadTopShelfContent(completionHandler:) Fetching top shelf content
 [🔵 DEBUG] ContentProvider#70:loadTopShelfContent(completionHandler:) main thread: true
❗ [CoreStore: Assertion Failure] DataStack+Querying.swift:107 fetchOne(_:_:)
  ↪︎ Attempted to fetch from a 'CoreStore.DataStack' outside the main thread.```
  
I've re connected to server and signed back in to make sure that it is stored within the app group, the user id is fetched from the defaults okay.

But looks like it could be a CoreStore issue, the assertion for main thread passes but still met with the error.

LePips:
This feature will have to be part of a larger discussion of how we allow development to continue. App Groups has to be set as part of your developer profile, so local development may be tricky. The widget is currently broken right now because of these same issues 🥲 but I kinda want to take a longer look at the entire widget and refactor it sometime, possibly taking what we learn here and applying it there.

Looking at this a bit more, we will have to somehow set the persistent store used in CoreStore to be a shared container. This is possible by just looking it up as it was a feature request a while ago so we'll have to see how it's officially done now.

LePips:
I'm going to close this because the project structure was renamed meaning it would be better just to start from scratch.

PR #248, opened 2021-12-28
iOS: hide home indicator on devices without a home button

knbx:
Swiftfin already contains a workaround to hide the home indicator, but this doesn't work any more.

The author of the workaround updated [their gist](https://gist.github.com/Amzd/01e1f69ecbc4c82c8586dcd292b1d30d) with a workaround for SwiftUI lifecycle. I've applied these changes to Swiftfin.

Some things to note:

- The new workaround uses [swizzling](https://nshipster.com/method-swizzling/) to point iOS to the right PreferenceUIHostController. Swizzling is generally considered a last resort and likely to break in future iOS versions.

- The gist uses a separate pod to do swizzling. Swiftfin seems light on dependencies, we could look into implementing the swizzling within the repo. Personally I wouldn't, since this swizzling should be replaced by a better solution when available. The pod dependency can help as a reminder to remove it.

- I'm new to SwiftUI, so I hope I didn't commit any basic mistakes. I added the new wrapper view in VideoPlayer.swift and had to update the edge insets to ignore the safe area.

Anyway, let me know what you think.


LePips (approved):

LePips:
There is some cleanup that can be done with this now but I can get to that.

buggerman:
Am I supposed to do anything for the workaround? Because the Home bar never hides itself for me in the version on AppStore right now.

knbx:
This fix pre-dates the recent big refactor. I haven't had the time to take another look at this

LePips:
The big refactor fixes it and just needs to be released after some things have been polished.

PR #259, opened 2022-01-06
tvOS - More Cinematic Views and Fixes

LePips:
Cinematic views are now a single toggle and removes remaining navigation link from old views


LePips:
@acvigue sorry to bug as I know you're busy, would it be possible to approve/merge this and then cut out a new TestFlight build soon with the new video player?

acvigue (approved):

PR #262, opened 2022-01-07
tvOS User Support

LePips:
Based off of https://github.com/jellyfin/Swiftfin/pull/259 so merge it first

Went through a roller coaster figuring this out so please forgive the commits. I attempted to remember the last JF user for the current tvOS user _however_ there's either a bug or deliberate behavior on the `TVUserManager.currentUserIdentifierDidChangeNotification` where it doesn't update for every _user_ change, only switching to the default user :/


LePips:
Revert the `rip everything out, again` commit to get to my latest attempt. Line `71` in `SessionManager` would sometimes return `nil`, meaning that we don't have a current user id when there should be one.

Issue #263, opened 2022-01-07
Cursor disappears on tvOS settings page


jorritolthuis:
### Describe the bug

1. Connect a server
2. Go to the settings tab
3. Go down one step
4. Notice that nothing is highlighted 
5. Go down another step
6. Notice that 'server' is highlighted

This seems to indicate that the cursor is hidden at 'user'. However, because the user cannot interact with this user entry, it does not make sense to have the cursor interact with it.

### Application version

6386bc9

### Device information

Apple TV 4K Gen 1

### OS version

tvOS 15.1.1 (19J581)

### Jellyfin server version

10.7.7

LePips:
Thanks for your PR, I'll actually make the user row a button (with no action) and that will give us a visual indicator when it's focused. It needs to be able to be focused so that the form can be scrolled correctly

jorritolthuis:
Thanks, that would also fix my confusion.

PR #264, opened 2022-01-07
Make the cursor not interact with the User field in tvOS settings

jorritolthuis:
Attempt to fix my own issue #263.

Removing the ability to focus on the 'User' entry in the tvOS settings page, will make the cursor skip over it.


LePips:
Sorry, this is actually deliberate as the form then won't scroll all the way back up if it was ever offset. I could add a focusing indicator like changing the color of a font

PR #275, opened 2022-01-09
tvOS Cinematic Home Screen View and Recently Added

LePips:
Based off of https://github.com/jellyfin/Swiftfin/pull/274, merge it first

Adds a big cinematic view to the home screen on tvOS. Resume items will be shown if there are any, otherwise Recently Added items will be shown. There isn't a good way to combine both Resume and Next Up items such that they are ordered by "last interacted with", however I have work in there such that Next Up items can be put in there as well for the future.

Adds the Recently Added to iOS as well.

`ContinueWatchingView` is still there even though it doesn't do anything because I have thoughts of making the home screen customizable in the future when I tackle a home screen redesign.


LePips:
Converted to draft as I want to adjust some things I thought about first

acvigue (approved):

acvigue:
merging 🎉

PR #276, opened 2022-01-09
Subtitle size options

knbx:
Default size remains as is.

Added 2 smaller and 2 larger options.

Added some documentation to clarify VLC doesnt officially support it & it wont work for tvOS until VLCKit 4.


LePips:
Good work! I say there should also be sizes available for iOS, use compiler flags on the textRendererFontSize property for it. I do believe that the font size changes on tvOS as I remember increasing the size of it for the tvOS video player implementation to make it smaller. I also can't remember which subtitle formats are affected by the font size change, this might be good to document (but not necessary).

acvigue (approved):

knbx:
> I say there should also be sizes available for iOS, use compiler flags on the textRendererFontSize property for it.

I don't think I follow, can you expand on this idea?

>  I do believe that the font size changes on tvOS

Ah, I saw it mentioned on the VLC forum that it wouldn't work for tvOS, but I never tried. 

> I also can't remember which subtitle formats are affected by the font size change, this might be good to document 

I had no idea this would be affected by subtitle formats. I totally agree, but wouldn't know where to start documenting this. I use Bazarr to find subtitles for my media, no clue what format I've tested lol.

LePips:
> expand on this idea?

Using pre-processor flags for iOS and tvOS would look something like this (just with dummy values), I would also ask you test them to make sure they look good (or "okay" at least with the larger/largest selections"):

```
#if os(tvOS)
  var textRendererFontSize: Int {
        switch self {
        case .smallest:
            return 24
        case .smaller:
            return 20
        case .regular:
            return 16
        case .larger:
            return 12
        case .largest:
            return 8
        }
    }
#else
  var textRendererFontSize: Int {
        switch self {
        case .smallest:
            return 1
        case .smaller:
            return 2
        case .regular:
            return 3
        case .larger:
            return 4
        case .largest:
            return 5
        }
    }
#endif
```

Shared/Objects/SubtitleSize.swift
```diff
@@ -0,0 +1,59 @@
...
+    }
+}
+
+// MARK: - sizing for VLC
+
+extension SubtitleSize {
+    /// Value to be passed to VLCKit (via hacky internal property, until VLCKit 4)
+    ///
+    /// note that it doesn't correspond to actual font sizes; a smaller int creates bigger text
+    var textRendererFontSize: Int {
```
LePips, line 45: Use pre-processor flags for iOS/tvOS difference and implement in iOS as well.

LePips (changes requested):

knbx:
Ah okay, I was confused by which OS you meant. 

So the sizes look good on iOS (imo). They range from fairly small to comically large, which seemed prudent to help the visually impaired. 

I'll take a look at tvOS to see if the size changes there and whether the options look good.

knbx:
So good news, subtitle sizing does in fact work on tvOS :D

I've added the option ot the settings view (made a new Accessibility section and moved poster labels into it, so it's consistent with the mobile app).

I haven't differentiated the sizing between iOS and tvOS because it looked good to me on simulator. As Ethan mentioned, it's easy enough to update later. Either way, I think some testing will be needed on different devices, because right now it's completely based on my subjective view.

LePips:
Sounds great! I'll check it out and approve

Swiftfin.xcodeproj/project.pbxproj
```diff
@@ -2023,6 +2043,40 @@
 			shellScript = "\"${PODS_ROOT}/Target Support Files/Pods-Swiftfin iOS/Pods-Swiftfin iOS-frameworks.sh\"\n";
 			showEnvVarsInLog = 0;
 		};
+		AB001A5552FE000244A447A4 /* [CP] Prepare Artifacts */ = {
```
LePips, line 2046: 1 - iOS isn't building because `VLCMediaPlayer.swift` was added to the extensions target as well, that must be removed.
2 - do a `pod install` so that these are removed as these shouldn't ever be added

LePips (changes requested):

knbx:
The widget extension compiles again. Not sure where that build phase came from, but I removed it manually.

LePips (approved):

PR #278, opened 2022-01-09
Refresh homeview automatically

knbx:
I noticed the refresh control sometimes doesn't trigger when pulling down. I haven't been able to reproduce it on the simulator yet, but I figured it's always nicer when the app refreshes automatically.

Small expansion to RefreshHelper so it will refresh the homeview on appear, if it's been at least a minute since last refresh.

### DIFF
--- a/Swiftfin/Objects/RefreshHelper.swift
+++ b/Swiftfin/Objects/RefreshHelper.swift
@@ -11,13 +11,30 @@ import UIKit
 
 // A more general derivative of
 // https://stackoverflow.com/questions/65812080/introspect-library-uirefreshcontrol-with-swiftui-not-working
-class RefreshHelper {
+final class RefreshHelper {
     var refreshControl: UIRefreshControl?
     var refreshAction: (() -> Void)?
+    private var lastAutomaticRefresh = Date()
 
     @objc func didRefresh() {
         guard let refreshControl = refreshControl else { return }
         refreshAction?()
         refreshControl.endRefreshing()
+    }
+}
+
+// MARK: - automatic refreshing
+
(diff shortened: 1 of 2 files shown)


LePips:
Thank you as this is something I've thought about as well for better home screen updating. However, I'll test this out to make sure the behavior is something that we want.

knbx:
Sure thing. 

It seemed a decent fix for my use case, where the home screen would still show the last episode when returning after playing an ep.
Ideally the detail screen would also automatically refresh I think, so you can back out of a movie and the app would update the last played position.

LePips:
I have an idea for the detail screen that I can get to soon.

I did try an implementation where the home screen would update upon closing playback but there was an issue where it would then pop the navigation back to the home screen. I sadly don't have that work indexed anywhere in git history.

knbx:
That does sound annoying. 

I guess from a performance standpoint it would be ideal if, after playback, the local data was updated to change the last position or mark the media as watched. 

But that could be difficult (I haven't looked at how data is saved in app). And eventually the app would still be outdated because of server side changes, such as newly added media. Seems overkill to me.

LePips:
Actually, more specifically, I made it so that it updated only the Resume and Next Up items on playback close. That's only a few items so it's just fine. I guess I can take another look at that implementation and work around the home screen popping-whatever with a trick.

Looks and works great, this will be merged after the chain from https://github.com/jellyfin/Swiftfin/pull/275 so there may be merge issues then.

LePips (approved):

knbx:
Sweet, I'll take a look after those PRs merge.

PR #282, opened 2022-01-10
Localize all (most) current strings

LePips:
Went through and localized strings except for some live TV and experimental labels (the current experimental feature might actually be removed with JF 10.8 around the corner which achieves the same feature but better).

Going to close the corresponding issue as a part of it, localization must be enforced for new strings that are part of final features.


PangMo5 (approved):
👍

PR #287, opened 2022-01-12
Spanish translation of newer changes

oxixes:
As Weblate doesn't seem to be updated (all of the localization changes made by @LePips weren't updated in Weblate) I decided to translate the `Localizable.strings` file manually to my mother tongue, Spanish (Spain).

All of the strings were translated and I made sure to check where strings were used to translate them appropriately.

Any suggestions before merging are highly appreciated.


LePips:
Thank you. I have surfaced the Weblate issue before to the people with access and I'll do it again.

LePips:
I apologize to say that we may not be able to accept these translation, as Weblate may cause merge errors with manually translated files (except for the base file which it doesn't touch, English). I'll just reject this for now and we can see later how to proceed. Weblate will be fixed in a bit.

LePips (changes requested):
Test later with new Weblate

oxixes:
I will close this pull request then, and submit the translations on Weblate once it's fixed. Thanks!

PR #294, opened 2022-01-13
Fix Transcoding

LePips:
Merge after https://github.com/jellyfin/Swiftfin/pull/291

Fixes transcoding now that I finally was able to test it. Something to know is that VLCKit takes a bit to start a `.m3u8` hls stream, so expect a very good delay when starting transcoded streams.


acvigue (approved):

acvigue (changes requested):
some of these changes conflict w/ main branch

acvigue (approved):

Issue #190, opened 2021-10-18
iPadOS - Purple Accent Color Not Working in Some Areas


LePips:
Something I've always noticed that on iPadOS the Jellyfin purple color isn't working in some areas. May be an accent color setting.

PR #298, opened 2022-01-14
iOS - LibraryView Improvements and Tidbits

LePips:
On iOS, moves the portrait items to the new portrait items and takes out pagination to move with lazy loading instead. I attempt to load enough items per page to fill the device's screen. It goes slightly over which is just fine, as 100 previously was kind of crazy.

Fixes https://github.com/jellyfin/Swiftfin/issues/190

Begins moving app-view customization options to a "Customize" settings, there will be more customization options and the settings will begin to feel crowded then. Might move "Missing Items" there in the future.

I had attempted to re-factor the filtering logic however I got kind of unmotivated as it works fine right now but should be changed later from from the binding that's passed around through 3 layers of views and viewmodels, consolidate filters to the `LibraryViewModel` or something.


Swiftfin/Views/LibraryView.swift
```diff
@@ -21,87 +22,75 @@ struct LibraryView: View {
...
+	}
+
+	@ViewBuilder
+	private var noResultsView: some View {
+		L10n.noResults.text
+	}
+
+	@ViewBuilder
+	private var libraryItemsView: some View {
+		DetectBottomScrollView {
```
PangMo5, line 44: As you wrote, it seems like a good way to create and use `DetectBottomScrollView`.

But wouldn't it be simpler to use `onAppear` for each cell and compare it with the last item of `viewModel.items` rather than that `DetectBottomScrollView`, and there will be less side effects?

```swift
PortraitItemButton(item: item) { item in
	libraryRouter.route(to: \.item, item)
}
.onAppear {
	  guard viewModel.hasNextPage,
	  	item == viewModel.items.last else { return }
	  viewModel.request~
}
```

like this

LePips: I actually did try that, however `ScrollView`'s lifecycle prevents this from acting as expected. It will render items twice in quick succession (once when added to the `ScrollView` and then again once the size of the `ScrollView` is calculated). Because of this, I experienced errors with the page calculations as `onAppear()` was run multiple times unexpectedly. 

`DetectBottomScrollView` is actually a lot better than `onAppear(...)` in this case with not side effects, what side effects are you pointing to?

PangMo5: If we use `onAppear` for cells in the current structure, doesn't it follow the lifecycle of `LazyVGrid`?
So I thought it would work normally.

Anyway, I don't know because I haven't used that code myself.
The overlapping use of A often had side effects in certain situations lol.

Once it works normally in the current situation, it will be fine.

PangMo5 (approved):

PR #300, opened 2022-01-14
Support Redirects

### LINKED ISSUE #290: Support HTTP Permanent and Temporary Redirect 301/307
The Jellyfin client should follow  HTTP/301 and HTTP/307 redirect responses from the server.

For Instance, a host listening on http://10.0.0.1 returns a redirect to http://10.0.0.2:8096, the client, configured to use http://10.0.0.1 should connect to the http://10.0.0.2:8096 Jellyfin server instead. 

```
> GET / HTTP/1.1
> Host: 10.0.0.1
> User-Agent: curl/7.81.0
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 307 Temporary Redirect
< Date: Wed, 12 Jan 2022 21:56:44 GMT
< Content-Length: 0
< Connection: keep-alive
< Location: http://10.0.0.2:8096
```

This feature would enable easily moving the back-end server without requiring users to modify the configuration and it requires less resources than having a reverse proxy.

LePips:
Add support for redirects, up to 3.

Closes #290


LePips:
Closed as this actually isn't done correctly. Re-directs should be an instance of the current call and not stored/given back to us. This may need to be addressed in the api layer.

LePips:
After thinking about this a bit more, this is the only way we can do it. This is a very niche issue so I don't really care if people have potential issues with it.

anthonylavado (dismissed):

anthonylavado (approved):

PR #310, opened 2022-01-17
Accessibility fixes

dcordero:
Some extra accessibility tweaks:

- [x] Avoid hiding player controls when VoiceOver is active
- [x] Add accessibility labels for the current position and remaining time labels in the player
- [x] Fix the background of each element in the server list when using bigger fonts with Dynamic Type


LePips:
The accessibility fix for the server list is fine for now, but that view eventually will be re-done. The list right now is just a temporary thing I first implemented.

LePips (approved):

Issue #308, opened 2022-01-17
Native Video Player


perrinpagess:
**Describe the feature you'd like**

I want an option to use AVPlayer for direct HLS instead of going through VLCKit. A toggle under "advanced settings" would be nice because it has some limitations.

**Additional context**

AVPlayer uses Apple's "EDR" tone mapping implementation for HDR playback on SDR and HDR displays. AVPlayer's HDR tone mapping is substantially better quality than VLCKit. AVPlayer natively supports surround audio for spatial audio processing and head tracking. Picture in picture support is native with AVPlayer as well. AVPlayer's surround to stereo mixdown is better than anything ffmpeg can do (from my testing).

I made a rudimentary video player in swift playgrounds to help demonstrate what I want (see screenshot). Instead of the url I used for the demo it would be the HLS url for a video on my Jellyfin server. 

Thanks!

![IMG_0034](https://user-images.githubusercontent.com/75100431/149742999-73f58283-91fb-4c39-9982-f2f33d887c85.PNG)

LePips:
I did implement the native player during my video player refactoring however I removed it due to navigation issue with selecting subtitles and such. It was just too much work at the moment.

I have thought to re-implement it and the work is in git history. This however made me think to create another setting where the AVPlayer layer as the rendering view for the video and we can use the same VLCKit overlay.

I'll first re-implement the native players as a part of this.

perrinpagess:
What was the issue with subtitles? My demonstration had the ability to pick subtitles with no extra work. Safari's current implementation of the native video player allows me to pick subtitles as well. HLS specification requires WebVTT subtitle format. 

What are direct play profiles required for? All my video files are sanitized and use Apple's HLS authoring approved codecs in MP4 container with proper headers. My MP4 files stream flawlessly in Safari when I copy a link from Jellyfin.

LePips:
There's a lot at play here. Most people's streams aren't nice like yours and the webvtt subtitle streams aren't consistent from the server, so we have to build the native player around that. For most, this would force embedding subtitles that they would choose before they play the media. Or in your case, you would be able to choose not to embed any subtitles and choose to play them from the stream instead.

perrinpagess:
I understand that most people don't want to learn how to use ffmpeg to make their files HLS compliant. I suggested hiding a toggle in "advanced settings" or whatever so that you wouldn't have to do all the extra work to make native player play nice with non-HLS compliant files. 

[...]

PR #320, opened 2022-01-19
Implement Experimental Native Player

LePips:
Some want the native video player, however it is still very experimental. The device builder doesn't work properly for native-only media so direct-played media won't necessarily work all of the time. I won't take any issues regarding this.

I will close https://github.com/jellyfin/Swiftfin/issues/308 because of this. I do not plan to put more work into this as all of the other parts of the apps are more important at the moment. PiP is not enabled as the view pops itself and we would have to implement custom views to handle that state. Other people are welcome to work and improve this.


LePips:
Cause some design issues on the overlay

anthonylavado (approved):

PR #325, opened 2022-01-20
Enable playback for LiveTV on tvOS

jhays:
Adds support to LiveTV to utilize the new VideoPlayer. 
Playback of LiveTV channels from both the Programs view and Channels view is now working. 

Minor adjustment to VLCPlayerViewController to remove usage of a force-unwrap on runTimeTicks, which was leading to a crash when attempting liveTV playback. If value is not found, 0 will be used.


Swiftfin tvOS/Views/LiveTVChannelsView.swift
```diff
@@ -55,7 +55,9 @@ struct LiveTVChannelsView: View {
 		let channel = item.channel
 		if channel.type != "Folder" {
 			Button {
-				self.router.route(to: \.videoPlayer, channel)
+                self.viewModel.fetchVideoPlayerViewModel(item: channel) { playerViewModel in
+                    self.router.route(to: \.videoPlayer, playerViewModel)
```
LePips, line 59: For now, literally copy the current video player and rename it to LiveTVPlayerViewController, copy the overlay and rename it to LiveTVOverlay and have the new view controller use it, make its corresponding SwiftUI LiveTVPlayer view, and fix the coordinators to use that view instead.

It's the beginning of work that we need.

LePips (changes requested):

LePips:
Also merge after https://github.com/jellyfin/Swiftfin/pull/320

Swiftfin tvOS/Views/LiveTVChannelsView.swift
LePips: Sorry, also add TODO comments at the top as well as descriptions saying what needs to happen next with these views.

Also add a TODO comment above VideoPlayerViewController and createVideoPlayerViewController stating that a new class must be created for LiveTV objects and then consumed by the other LiveTV views.

LePips:
Just something cool to note, this fixes a crash that I just received with the removal of the forced unwraps from `runtimeTicks`. Guess it was a false assumption, but media _should_ have a run time 🤷‍♂️

Swiftfin tvOS/Views/LiveTVChannelsView.swift
jhays: I've duplicated the player and overlay for use with LiveTV. 

Tested playback on a real AppleTV- while playback is successful, I'm getting no audio. So I'll have to investigate further on that.

jhays: I have to chase further to figure out why, but appears the `selectedAudioStreamIndex` from `VideoPlayerViewModel` is getting set to -1 .

LePips: Okay, that's expected as it doesn't have a default audio stream. At this point, you may actually have to copy over `createVideoPlayerViewModel()` and make a `createLiveTVViewModel()` method to accommodate for this and setting the correct values.

Eventually a `LiveTVViewModel` will be created and used there.

jhays: I've duplicated the create viewModel for now. 
But I'm not sure what to adjust to fix audio playback within that method. There is a stereo aac audio track there, but for some reason I don't get any audio playback. 
The same item plays fine with audio in Jellyfin web interface in chrome. Playback from there or the appleTV shows up in the dashboard as a remux. 
Any ideas on what to tweak to try to fix audio playback? Thanks.

LePips:
I think once the conflicts are resolved then the CI will get kicked in and then I can approve. For some reason, I _still_ cannot play live tv even on the web given whatever m3u playlist. However if it's working for you and will work for other, I'm more than happy to approve.

jhays:
@LePips , my latest commit resolves merge conflicts, and although this branch does successfully play live TV for me, I still can't seem to get audio playback to work. 
Do you have any suggestions on where I can investigate further to determine why the audio track is not playing? 
Thanks.

Shared/Extensions/JellyfinAPIExtensions/BaseItemDto+VideoPlayerViewModel.swift
```diff
@@ -166,4 +166,159 @@ extension BaseItemDto {
...
+					                                                subtitle: subtitle,
+					                                                directStreamURL: directStreamURL,
+					                                                transcodedStreamURL: transcodedStreamURL?.url,
+					                                                hlsStreamURL: hlsStreamURL,
+					                                                streamType: streamType,
+					                                                response: response,
+					                                                audioStreams: audioStreams,
+					                                                subtitleStreams: subtitleStreams,
+					                                                chapters: modifiedSelfItem.chapters ?? [],
+					                                                selectedAudioStreamIndex: defaultAudioStream?.index ?? -1,
```
LePips, line 305: This is where the audio stream index gets set and then is set in the media player. I assume that the channel stream doesn't have a default audio stream.

Swiftfin tvOS/Views/VideoPlayer/LiveTVPlayerViewController.swift
```diff
@@ -0,0 +1,900 @@
...
+		}
+	}
+}
+
+// MARK: PlayerOverlayDelegate
+
+extension LiveTVPlayerViewController: PlayerOverlayDelegate {
+
+	func didSelectAudioStream(index: Int) {
+		vlcMediaPlayer.currentAudioTrackIndex = Int32(index)
```
LePips, line 762: (2) Through a variety of mechanisms, this is where the audio track gets set. From (1), the track index was -1 which is the index for "no audio track" in VLCKit. You might just have to remove this line.

LePips:
@jhays see comments (1) and (2).

Shared/Extensions/JellyfinAPIExtensions/BaseItemDto+VideoPlayerViewModel.swift
jhays: If I put a breakpoint on 294 and take a look at this at the moment I select a liveTV channel item for playback, I am actually finding that the defaultAudioStream is set, to a stereo aac audio track. Also this track appears to have an index value of `-1`, which is how it matches up on line 202 where the default stream is selected from the audio streams array. In this case, that array also only has this same one audio stream object. 

Screen Shot 2022-01-24 at 6 14 38 PM

Swiftfin tvOS/Views/VideoPlayer/LiveTVPlayerViewController.swift
jhays: Bingo! Removing this made liveTV audio work for me. Thanks for the help.

Shared/Extensions/JellyfinAPIExtensions/BaseItemDto+VideoPlayerViewModel.swift
LePips: Interesting, as I don't know how live tv objects are stored in JF in respect to a `BaseItemDto` object. Over time, you'll have to do more investigation and tailor the video players to the data from these objects.

LePips (approved):

PR #351, opened 2022-01-26
Basic support for folder-type library items

mshockwave:
- Add a new BaseItemDto.ItemType member: .folder.
  - Use CollectionItemViewModel to display folder-type item.
  - For each episode in a folder, fallback to primary image if series
    primary image is not available.
  - (tvOS) fixed a bug in LibraryListView where proper parent ID was not passed
    to LibraryViewModel.

This PR also lifts some restricts that prevent non- _movies_ or _tvshows_ type libraries from being displayed.


Shared/ViewModels/LibraryViewModel.swift
```diff
@@ -94,10 +94,14 @@ final class LibraryViewModel: ViewModel {
...
 		let sortBy = filters.sortBy.map(\.rawValue)
+		let queryRecursive = filters.filters.contains(.isFavorite) ||
+			self.person != nil ||
+			self.genre != nil ||
+			self.studio != nil
 
 		ItemsAPI.getItemsByUserId(userId: SessionManager.main.currentLogin.user.id, startIndex: currentPage * pageItemSize,
 		                          limit: pageItemSize,
-		                          recursive: true,
+		                          recursive: queryRecursive,
```
LePips, line 104: This needs to be a customization option whether it should always be `true` or with this flag, as people organize their media within a "Movies"/other folder on their systems and want all their items to be flattened.

Swiftfin/Views/LibraryListView.swift
```diff
@@ -73,36 +73,32 @@ struct LibraryListView: View {
 					}
 
 					ForEach(viewModel.libraries, id: \.id) { library in
-						if library.collectionType ?? "" == "movies" || library.collectionType ?? "" == "tvshows" {
```
LePips, line 76: Don't re-implement this exact filter, but we need a filter of some kind so that we only show expected libraries (ie: not music libraries)

LePips (changes requested):

LePips:
Thanks for your contribution! I left a couple comments.

Swiftfin/Views/LibraryListView.swift
LePips: I also believe that you can delete the explicit check that we have for the "Collections" library, as this will catch it instead.

Shared/ViewModels/LibraryViewModel.swift
LePips: Even then, the web UI doesn't group folder items so it's good that we're doing something even better.

Have a toggle that belongs in settings under "Customization". If you don't know what I'm talking about, I can get this done.

mshockwave: oh okay, I actually just implemented that as an additional filter option before seeing your reply. But I agree putting it as a global option might be better so I'll change it.

Shared/SwiftfinStore/SwiftfinStoreDefaults.swift
```diff
@@ -41,6 +41,7 @@ extension Defaults.Keys {
 	// Customize settings
 	static let showPosterLabels = Key<Bool>("showPosterLabels", default: true, suite: SwiftfinStore.Defaults.generalSuite)
 	static let showCastAndCrew = Key<Bool>("showCastAndCrew", default: true, suite: SwiftfinStore.Defaults.generalSuite)
+	static let showFlattenView = Key<Bool>("showFlattenView", default: true, suite: SwiftfinStore.Defaults.generalSuite)
```
LePips, line 44: Let's change the string in English to `Flatten Library Items`, as we aren't toggling to show/hide something explicit like the poster labels.

LePips (changes requested):

LePips (approved):

PR #354, opened 2022-01-28
Add Keyboard Shortcuts

PangMo5:
- space: play / pause
- left/right keys: jump back/forwards
- escape: close
- Command+right key: next item
- Command+left key: previous item
- Command+]: next playback speed
- Command+]: previous playback speed
- Command+\\: normal playback speed

VideoPlayer code was neatly organized, so it was easy to work on 😄


LePips (approved):
Works great!

PR #356, opened 2022-01-29
Re-implement BlurHash

LePips:
Re-implements BlurHash that I commented out, turns out it just needed a resizing so it wasn't pixelated as much and some blurhash detection needed to be fixed.

Also adds a minus icon with "Remove from Resume".

Something that I noticed is that the blurhashes for episode items aren't found correctly (hence the forced first return). We probably just need to change these things, but I would like to ultimately clean up how all images are created throughout the app. Blurhashes still need to be implemented in some new views that I've created.

I also noticed the episode backdrop image sometimes snaps between transitions instead of a fade on iOS.

### DIFF
--- a/Shared/Extensions/BlurHashDecode.swift
+++ b/Shared/Extensions/BlurHashDecode.swift
@@ -8,6 +8,8 @@
 
 import UIKit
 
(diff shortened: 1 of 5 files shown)


PangMo5 (approved):
LGTM XD

PR #360, opened 2022-02-03
Added guide for joining without iOS device

WardPearce:


LePips:
Thank you for your contribution, however we won't be accepting this kind of documentation for any 3rd party services to access TestFlight links, all can be done with just TestFlight. If people use BrowserStack, it's expected they know how to use it.

PR #362, opened 2022-02-04
Pre-Pre-Alpha Downloading Items

### LINKED ISSUE #57: [FEATURE REQUEST] Download for OFFLINE viewing
Would love to be able to sync/download video for offline viewing.

LePips:
Developer notes discussion about what this is expected to have: https://github.com/jellyfin/Swiftfin/discussions/364

I'm going to close #57 because the base has been implemented and has a roadmap.

Right now, it's way messier than I intended due to evolving ideas. However, I know what needs to change but don't have the time. Will figure that out in Pre-Alpha.

As I typically have Alpha features in the Experimental section, the work for this is locked in much of the foundation of the app so for code maintainability it's not behind a flag. However it should still be treated as Experimental as it is very much just an experiment until Beta.

### Other Work
Other work that was a part of this:

- Refactored Notifications
- Refactored `ImageView` for async BlurHash and `init`s for provided failure views


LePips:
Every time that I think about and work on this I keep thinking about how I need to redo it so that I don't create a foundation I need to rewrite later on.

UnicornsOnLSD:
As someone who is currently rewriting the foundation that they created, here's what I'm thinking of making for Finamp:

Using some sort of relational database (I'm using [Isar](https://isar.dev/) for Flutter, there's probably something out there for Apple), have downloaded parents and downloaded children. Parents represent albums (in Swiftfin's case, shows/movies), and children represent individual songs (or episodes). Linking is possible with any type, so a parent could depend on a parent, a parent could depend on children, etc. This would allow a tree of dependencies flexible enough to model relations for stuff like `artist -> album -> songs`. With this, I can easily check which files to delete when a non-leaf is deleted. For example, when deleting an album, you may need to keep some songs as they're depended on by a playlist.

Hopefully this turns out to work, managing downloads is easily the most challenging part of making a Jellyfin client imo (alongside debugging peoples' weird playback errors due to odd formats).

dm3ch:
Is there any chance that work on this would be resumed anytime soon?

PR #369, opened 2022-02-08
(Re)Enable the play button on the season page

mshockwave:
Resurrecting some of the previously-commented code to enable the play button on the season page. Which plays the next unplayed episode in that season.

In addition, we no longer bail out on the `BaseItemDto.missing` flag because sometimes it shows false negatives even on legit season items. Instead, we rely on results (i.e. list of episodes) returning from the API call.

### DIFF
--- a/Shared/ViewModels/ItemViewModel/SeasonItemViewModel.swift
+++ b/Shared/ViewModels/ItemViewModel/SeasonItemViewModel.swift
@@ -30,6 +30,7 @@ final class SeasonItemViewModel: ItemViewModel, EpisodesRowManager {
 		getSeriesItem()
 		selectedSeason = item
 		retrieveSeasons()
+		requestEpisodes()
 	}
 
 	override func playButtonText() -> String {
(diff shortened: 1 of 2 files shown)


LePips:
I'm fine with removing the missing flag for now, the logic for that does need to be fixed and match how the web does it. By just removing it now, it won't necessarily behave properly for actual missing seasons with missing episodes. I'll take a look later.

mshockwave:
> Btw, the second commit swiftformat-s an unrelated file -- I have no idea why it was not previously formatted.

Nevermind, it turns out I need to upgrade my `swiftformat`. I'd reverted that particular commit (let me know if you prefer to rebase it instead).

mshockwave:
> it won't necessarily behave properly for actual missing seasons with missing episodes.

Yeah that's also one of my concerns. Though I haven't tested on a real missing season, _ideally_ if anything goes wrong, it will be handled by the existing API error handler and leaving the `episodes` list empty, which makes the play button un-clickable.

Shared/ViewModels/ItemViewModel/SeasonItemViewModel.swift
```diff
@@ -38,56 +39,51 @@ final class SeasonItemViewModel: ItemViewModel, EpisodesRowManager {
...
-//			playButtonItem = firstUnwatched
-//		} else {
-//			guard let firstEpisode = episodes.first else { return }
-//			playButtonItem = firstEpisode
-//		}
-//	}
+	private func setNextUpInSeason() {
+		guard !episodes.isEmpty else { return }
+
+		var firstUnwatchedSearch: BaseItemDto?
```
LePips, line 70: Back to me saying `isMissing` needs to be fixed: this will need to be visited later on after that is fixed to account for missing episodes as I say we want the functionality to be such that it will only consider episodes that the user has.

Shared/ViewModels/ItemViewModel/SeasonItemViewModel.swift
```diff
@@ -38,56 +39,51 @@ final class SeasonItemViewModel: ItemViewModel, EpisodesRowManager {
...
-//			}, receiveValue: { [weak self] response in
-//				self?.episodes = response.items ?? []
-//				LogManager.shared.log.debug("Retrieved \(String(self?.episodes.count ?? 0)) episodes")
-//
-//				self?.setNextUpInSeason()
-//			})
-//			.store(in: &cancellables)
-//	}
+	private func requestEpisodes() {
+		LogManager.shared.log
```
LePips, line 47: So, I actually found out that we shouldn't necessarily do it this way. I had the thought "what would happen if there were hundreds of episodes" (which happens) and looked to the web and found out that they use the `NextUp` endpoint and pass in the series ID. Sorry to say that I will initially reject this implementation for that but it is a simple endpoint and will end up being cleaner here.

LePips (approved):

Shared/ViewModels/ItemViewModel/SeasonItemViewModel.swift
LePips: re-review request changes comment

LePips (changes requested):

Shared/ViewModels/ItemViewModel/SeasonItemViewModel.swift
mshockwave: I fiddled around and conclude that `NextUp` doesn't work with a specific season id[1]. Namely, it only tells you nextup in the enclosing series but not nextup in a specific season. So right now I still leverage `NextUp`, but validate if the returning nextup items belong to the current season. If not, falling back to the old mechanism a.k.a picking the first unplayed episode as nextup.

[1]: Including providing season id to the `seriesId` argument, populating the `parentId` argument with either a season or series id.

mshockwave: > you can also fix this for the series as well if you want to. I can't remember if it works and if it does that it uses the same endpoint.

I think the series view model has been using the `NextUp` endpoint.

Shared/ViewModels/ItemViewModel/SeasonItemViewModel.swift
```diff
@@ -38,56 +39,73 @@ final class SeasonItemViewModel: ItemViewModel, EpisodesRowManager {
...
+				self.setNextUpInSeason()
+			})
+			.store(in: &cancellables)
+	}
+
+	private func setNextUpInSeason() {
+
+		TvShowsAPI.getNextUp(userId: SessionManager.main.currentLogin.user.id,
+		                     fields: [.primaryImageAspectRatio, .seriesPrimaryImage, .seasonUserData, .overview, .genres, .people],
+		                     seriesId: item.seriesId ?? "", enableUserData: true)
```
LePips, line 69: I'm fine to merge this for now since you do all the calculations for the season, however we should be able to use the `parentId` parameter to specify a season in the future.

LePips (approved):

PR #370, opened 2022-02-08
[tvOS] Add Top Shelf Extension

### LINKED ISSUE #335: Enable Top Shelf items on AppleTV
AppleTV apps that are in the top shelf row normally have media items that show on the main part of the screen.  It would be great if we could have Jellyfin show media items that are selectable here as well.

Right now when the Jellyfin application is in the top shelf and highlighted, it just shows the Jellyfin logo in the background.

This feature should also try to accommodate parental controls settings that are set in the AppleTV settings.

Top Shelf example:

Top Shelf Guidelines:
https://developer.apple.com/design/human-interface-guidelines/tvos/overview/top-shelf/

stevebyatt10:
This was originally created here #224 but has now been re-implemented.

This adds "Continue Watching" Top Shelf content for the tvOS app, and resolves #335 

This PR is set to draft as this requires app groups to be set up, which will need to be discussed further.


LePips:
For app groups, I'm fine for making one that's named suitably. I just request that you add documentation in the Contribution Guidelines outlining this which we can discuss later.

stevebyatt10:
Would someone with access to the release developer profile create one?

Issue #217, opened 2021-11-07
Clean up dependencies for native M1 Xcode


LePips:
Dependencies need to be cleaned up in order to use Xcode natively on M1 Macs.

Issues come from `google-cast-sdk` and `Puppy`. We may just need to replace `Puppy` as our logging entirely.

bestlem:
What can be done about google-case-sdk?

LePips:
For google-cast arm64 has to be an ignored architecture for the Debug target but Puppy has to be built in arm64, which is why we need to replace it in order to set the right flags

LePips:
Follow this issue for when this can be done: https://stackoverflow.com/questions/70574070/google-cast-sdk-not-working-in-apple-m1-pro-chipset/70618073

PR #392, opened 2022-03-18
Support Apple Silicon Mac's simulator

### LINKED ISSUE #391: Development Environment in Apple Silicon
Some existing frameworks did not support the arm64 simulator architecture.
Due to this it was impossible to build project for simulators in silicon Mac.

Previously, Google Cast framework, that was the most problematic.
But finally, arm64 simulator architecture support has been added, so I'll respond to Swiftfin.

https://developers.google.com/cast/docs/ios_sender#xcframework_beta

PangMo5:
- Replace Cocoapods with Carthage (VLCkit)
- Support Xcode 13.3
- add GoogleCastSDK xcframework (I don't think it's used, but I added it for future implement.)
- update contributing.md 

fixes #391, #217


PangMo5:
I think the `Package.resolved` has become version 2 from Xcode 13.3
However, because GithubAction does not support macos-12, it is impossible to build with Xcode 13.3.
So I think need to exclude build ci temporarily.
What do you think about it? @LePips

LePips:
Yeah, that looks like something we'll have to wait for. Thanks for this! Everything works well and we also get VLCKit 3.4.0 which updated the libvlc bindings, which I think helps with the `ass` subtitle crashes (which are like 99% of crashes).

LePips (approved):

oxixes:
> Yeah, that looks like something we'll have to wait for. Thanks for this! Everything works well and we also get VLCKit 3.4.0 which updated the libvlc bindings, which I think helps with the `ass` subtitle crashes (which are like 99% of crashes). We can just ignore the builds for a while until macos-12 is available.

You were right, now all my files with ASS subtitles play correctly. Thanks you all!

PR #393, opened 2022-03-19
Improve Image Views/Notifications

LePips:
Other work that I had from https://github.com/jellyfin/Swiftfin/pull/362 that moves blur hash decryption/rendering to the background thread, improves overall image view capacity, and improves our internal notifications.

### DIFF
--- a/Shared/Views/BlurHashView.swift
+++ b/Shared/Views/BlurHashView.swift
@@ -1,0 +1,61 @@
+//
+// Swiftfin is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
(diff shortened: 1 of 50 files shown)


Shared/Views/BlurHashView.swift
```diff
@@ -0,0 +1,62 @@
...
+	private let imageView: UIImageView
+
+	init(_ blurHash: String) {
+		let imageView = UIImageView()
+		imageView.translatesAutoresizingMaskIntoConstraints = false
+		self.imageView = imageView
+
+		super.init(frame: .zero)
+
+		computeBlurHashImageAsync(blurHash: blurHash) { blurImage in
```
PangMo5, line 34: Don't we need a weak capture here?

LePips: Yeah probably, would just be a small optimization.

PangMo5 (approved):
XD

PR #398, opened 2022-03-22
LiveTV respect Force Direct Play and Native Player settings

jhays:
Ensures LiveTV respects those settings. LiveTV / HLS works better in the Native player than in VLCKit. 

However, I wonder if it would make sense to break these out as separate settings specifically for LiveTV? Because it might make sense to use Force Direct Play / Native Player for liveTV, but maybe I wouldn't choose those settings when playing something from my library. 

I've also added usage of the loading indicator right upon selecting an item in the LiveTV Channels view, as there was previously a bit of a delay after selecting and item and the viewModel creation completing. This change makes it feel a little more responsive to selection. However, while the loading overlay is shown, the Tab bar takes focus. This could create some confusion, as you could change tabs while the loading is still in progress. I'd prefer to hide it, but I'm not exactly sure if that can easily be done with a Tab Bar on tvOS.

Lastly, I had to put a short delay on hiding the loading indicator after the view model is built just before player view appears. I don't love it, but without that delay, the loading completes and the channels grid view flashes momentarily before the player view appears, which didn't feel clean. 

Loading example: 

https://user-images.githubusercontent.com/125045/159558737-d90e2c8c-636e-4886-978b-7be0a0eb2ba8.mp4

### DIFF
--- a/Shared/ViewModels/VideoPlayerViewModel/VideoPlayerViewModel.swift
+++ b/Shared/ViewModels/VideoPlayerViewModel/VideoPlayerViewModel.swift
@@ -151,7 +151,8 @@ final class VideoPlayerViewModel: ViewModel {
 	}
 
 	func setSeconds(_ seconds: Int64) {
-		let videoDuration = item.runTimeTicks!
+        guard let runTimeTicks = item.runTimeTicks else { return }
+		let videoDuration = runTimeTicks
 		let percentage = Double(seconds * 10_000_000) / Double(videoDuration)
 
 		sliderPercentage = percentage
--- a/Shared/ViewModels/VideoPlayerViewModel/VideoPlayerViewModel.swift
+++ b/Shared/ViewModels/VideoPlayerViewModel/VideoPlayerViewModel.swift
@@ -151,7 +151,7 @@ final class VideoPlayerViewModel: ViewModel {
 	}
 
 	func setSeconds(_ seconds: Int64) {
-        guard let runTimeTicks = item.runTimeTicks else { return }
+		guard let runTimeTicks = item.runTimeTicks else { return }
 		let videoDuration = runTimeTicks
 		let percentage = Double(seconds * 10_000_000) / Double(videoDuration)
 
--- a/Swiftfin tvOS/Views/LiveTVChannelsView.swift
+++ b/Swiftfin tvOS/Views/LiveTVChannelsView.swift
@@ -55,15 +55,21 @@ struct LiveTVChannelsView: View {
 		let channel = item.channel
 		if channel.type != "Folder" {
 			Button {
+                self.viewModel.isLoading = true
 				self.viewModel.fetchVideoPlayerViewModel(item: channel) { playerViewModel in
 					self.router.route(to: \.videoPlayer, playerViewModel)
+                    DispatchQueue.main.asyncAfter(deadline: .now() + 2) {
+                        self.viewModel.isLoading = false
+                    }
 				}
 			} label: {
-				LiveTVChannelItemElement(channel: channel,
+                let progressPercent = item.program?.getLiveProgressPercentage() ?? 0
+                LiveTVChannelItemElement(channel: channel,
 				                         program: item.program,
 				                         startString: item.program?.getLiveStartTimeString(formatter: viewModel.timeFormatter) ?? " ",
 				                         endString: item.program?.getLiveEndTimeString(formatter: viewModel.timeFormatter) ?? " ",
-				                         progressPercent: item.program?.getLiveProgressPercentage() ?? 0)
+                                         progressPercent: progressPercent > 1.0 ? 1.0 : progressPercent
+                )
 			}
 			.buttonStyle(PlainNavigationLinkButtonStyle())
 		}
--- a/Swiftfin tvOS/Views/LiveTVChannelsView.swift
+++ b/Swiftfin tvOS/Views/LiveTVChannelsView.swift
@@ -54,24 +54,21 @@ struct LiveTVChannelsView: View {
 		let item = cell.item
 		let channel = item.channel
 		if channel.type != "Folder" {
-			Button {
-                self.viewModel.isLoading = true
-				self.viewModel.fetchVideoPlayerViewModel(item: channel) { playerViewModel in
-					self.router.route(to: \.videoPlayer, playerViewModel)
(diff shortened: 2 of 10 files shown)


LePips:
> separate settings specifically for LiveTV

This should be the case. We'll have to make just another menu for LiveTV settings.

> loading indicator

I'm fine that there's a loading indicator, however I think that instead of a fullscreen overlay that it should be a small loading indicator on the card, and then the player is viewed. This circumvents the issues you describe from putting in the full screen overlay.

Shared/ViewModels/VideoPlayerViewModel/VideoPlayerViewModel.swift
```diff
@@ -151,7 +151,8 @@ final class VideoPlayerViewModel: ViewModel {
 	}
 
 	func setSeconds(_ seconds: Int64) {
-		let videoDuration = item.runTimeTicks!
+        guard let runTimeTicks = item.runTimeTicks else { return }
```
jhays, line 154: prevents a crash that could occur here with the implicitly unwrapped optional. Ran into this with LiveTV.

Swiftfin tvOS/Views/LiveTVChannelsView.swift
```diff
@@ -55,15 +55,21 @@ struct LiveTVChannelsView: View {
 		let channel = item.channel
 		if channel.type != "Folder" {
 			Button {
+                self.viewModel.isLoading = true
 				self.viewModel.fetchVideoPlayerViewModel(item: channel) { playerViewModel in
 					self.router.route(to: \.videoPlayer, playerViewModel)
+                    DispatchQueue.main.asyncAfter(deadline: .now() + 2) {
```
jhays, line 61: Sorry about this. I don't like it, but without it, the loading view dismisses a fraction too early, so the LiveTV channels grid flashes visible before the player view loads. That experience felt janky. This code also feels janky... but it fixes it from the viewer's perspective.

jhays:
> > separate settings specifically for LiveTV
> 
> This should be the case. We'll have to make just another menu for LiveTV settings. If you want to do this now, put it also under Experimental. Make a separate section for LiveTV with the toggle for LiveTV and the settings (either in the section or to another page. It will eventually be in a separate page later on but idc rn).
> 
> > loading indicator
> 
> I'm fine that there's a loading indicator, however I think that instead of a fullscreen overlay that it should be a small loading indicator on the card, and then the player takes over. This circumvents the issues you describe from putting in the full screen overlay.

Understood, I can add the additional settings, and I like the idea of having the loading indicator on the card instead. I can make those adjustments.

Swiftfin tvOS/Views/LiveTVChannelsView.swift
```diff
@@ -55,15 +55,21 @@ struct LiveTVChannelsView: View {
...
+                        self.viewModel.isLoading = false
+                    }
 				}
 			} label: {
-				LiveTVChannelItemElement(channel: channel,
+                let progressPercent = item.program?.getLiveProgressPercentage() ?? 0
+                LiveTVChannelItemElement(channel: channel,
```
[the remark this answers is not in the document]
LePips: I'll take a look in a bit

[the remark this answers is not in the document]
jhays: I made some progress! I was able to move the button inside `LiveTVChannelItemElement`, just had to play around with FocusState stuff to get focus working again. I should be able to implement loading & spinner logic now. I'll get the updates pushed up here soon.

jhays:
Added LiveTV experimental settings, channel cell loading indicator, and some minor channel cell improvements.

Simulator Screen Shot - Apple TV 4K (at 1080p) (2nd generation) - 2022-03-25 at 22 50 36


https://user-images.githubusercontent.com/125045/160226030-649e9eea-ef40-43f6-a285-04e8af2496ad.mp4

Swiftfin tvOS/Views/SettingsView/ExperimentalSettingsView.swift
```diff
@@ -28,9 +33,19 @@ struct ExperimentalSettingsView: View {
...
+
 				Toggle("Live TV (Alpha)", isOn: $liveTVAlphaEnabled)
 
-				Toggle("Native Player", isOn: $nativePlayer)
+				Toggle("Live TV Force Direct Play", isOn: $liveTVForceDirectPlay)
+
+				Toggle("Live TV Native Player", isOn: $liveTVNativePlayer)
 
 			} header: {
 				L10n.experimental.text
```
LePips, line 51: Let's just call this section "Live TV" since it's the section for it, I'm fine for not localizing this as it will eventually be moved out.

LePips (changes requested):
Just this change then all good

Swiftfin tvOS/Views/SettingsView/ExperimentalSettingsView.swift
jhays: Sure- thought you might catch this :) 
But as it turns out, I'm not sure how to add new localized strings! Seems like there is a generated file, is there a tool I need to use? Or how does that work?

LePips:
Linting also required

jhays:
> Linting also required

Did you guys switch to SwiftLint or something? I ran swiftformat earlier, and then again after this last change and it didn't seem to make any additional format changes

LePips:
The action pulls the latest SwiftFormat version whenever it runs so you may need to update locally. I had to update mine recently.

jhays:
Hmm.. I definitely needed to update swift format. But after doing so and running it again, no changes were applied. Any other ideas what I might be missing? 

```
jhays$ swiftformat .
Running SwiftFormat...
Reading config file at /Users/jhays/Dev/orbosphere/Swiftfin/.swiftformat
SwiftFormat completed in 0.26s.
0/241 files formatted, 2 files skipped.
```

LePips:
When I run SwiftFormat I get a file linted, using version 0.49.6 from `swiftformat -v` (it also lints the file that the action is complaining about). Honestly I'm okay to just approve and merge and I'll get it next time I do some work.

LePips (approved):

jhays:
Thanks- I think in the end it was my issue, I don't think I fully updated swiftformat all the way somehow. I believe I have sorted it out now.

PR #405, opened 2022-03-29
Fix Seasons Sorting

### LINKED ISSUE #383: Number seasons in numerical order
Can we get seasons numbered in numerical order, and not by the starting digit?

How it is:
image

**Additional context**
Seasons are categorized by their starting digit, instead of the logical way which is 1-9,10-19, etc.

LePips:
Fixes seasons sorting for the small menu, small fix that we weren't able to catch because CI is off rn, and a linting fix I said that I would fix.

Closes #383

### DIFF
--- a/Swiftfin/Components/EpisodesRowView/EpisodesRowView.swift
+++ b/Swiftfin/Components/EpisodesRowView/EpisodesRowView.swift
@@ -29,7 +29,8 @@ struct EpisodesRowView<RowManager>: View where RowManager: EpisodesRowManager {
 					}
 				} else {
 					Menu {
-						ForEach(Array(viewModel.seasonsEpisodes.keys).sorted(by: { $0.name ?? "" < $1.name ?? "" }), id: \.self) { season in
+						ForEach(Array(viewModel.seasonsEpisodes.keys).sorted(by: { $0.indexNumber ?? 0 < $1.indexNumber ?? 0 }),
+						        id: \.self) { season in
 							Button {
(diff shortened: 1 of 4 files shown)


LePips:
@jhays this probably breaks the Livetv channel cards because they are no longer focusable. I honestly didn't notice that these views were in Shared, they should eventually be moved to the tvOS project and that focus modifier added back in.

PangMo5 (approved):

Swiftfin/Components/EpisodesRowView/EpisodesRowView.swift
```diff
@@ -29,7 +29,8 @@ struct EpisodesRowView<RowManager>: View where RowManager: EpisodesRowManager {
 					}
 				} else {
 					Menu {
-						ForEach(Array(viewModel.seasonsEpisodes.keys).sorted(by: { $0.name ?? "" < $1.name ?? "" }), id: \.self) { season in
+						ForEach(Array(viewModel.seasonsEpisodes.keys).sorted(by: { $0.indexNumber ?? 0 < $1.indexNumber ?? 0 }),
```
PangMo5, line 32: It may be better to implement a `Comparable` protocol or to be in charge of logic in ViewModel.

LePips: You're right, good catch. I'll do that now instead

PangMo5 (approved):

PR #412, opened 2022-04-16
Add Double Tap Jump Gesture, Gestures Lock

PangMo5:
- Replace swipe gestures for jump with doubleTapGesture (like Youtube)
- Add Gestures Lock (like nPlayer)


LePips:
I played around with double tap gesture a while ago however I didn't like how it worked with the overlay. I won't be able to test for a couple days, so I'll be able to look at it then.

PangMo5:
> I played around with double tap gesture a while ago however I didn't like how it worked with the overlay. I won't be able to test for a couple days, so I'll be able to look at it then.

Yeah, Check it out. Maybe there's no problem 😉

Swiftfin/Views/VideoPlayer/VLCPlayerViewController.swift
```diff
@@ -205,22 +230,24 @@ class VLCPlayerViewController: UIViewController {
...
+
 		view.addGestureRecognizer(singleTapGesture)
 		view.addGestureRecognizer(pinchGesture)
+		view.addGestureRecognizer(longPeessGesture)
 
 		if viewModel.jumpGesturesEnabled {
-			view.addGestureRecognizer(rightSwipeGesture)
-			view.addGestureRecognizer(leftSwipeGesture)
+			view.addGestureRecognizer(doubleTapGesture)
+			singleTapGesture.require(toFail: doubleTapGesture)
```
LePips, line 248: The "usability issue" that I mention is a long delay when tapping the _show_ the overlay when it isn't showing. I find it rather very annoying. My previous work had this exact logic of falling back gestures. I'm not against double tap, however I think that overall it's better to provide the option of double tap or swipe (as I actually prefer swipe). As long as there's the option I personally don't care about this delay, however it would be best to find the right logic to prevent the delay and mirror something like Youtube's double tap.

I think the lock gesture is nice, however I think it should ultimately be toggle-able from settings.

I'm okay to merge with:
- option for swipe/double tap
- option in Settings for gesture lock
- `longPeessGeature` -> `longPressGesture` spelling error

I know that I may be catering to my own preferences, however I just thought about how we would possibly want to implement 2x/3x/4x speed scrubbing which would be:
- tapping when you have swipe to skip
- swiping when you have tap to skip

Just a thought for work down the road.

LePips (changes requested):

Swiftfin/Views/VideoPlayer/VLCPlayerViewController.swift
PangMo5: Oh, I got you.
I didn't check the UX properly because I only works on the simulator.
First of all, I will try to improve UX as much as possible in the current situation and add only the gesture lock option.
If UX cannot be improved, I will consider the swipe option.

PangMo5: I removed the swipe because I wanted to control the slider with a pan gesture later. (like Infuse or etc..)

LePips: That sounds great! I also think that in the future that we can create "gesture profiles" that are a combination of gestures from different products to appease the most people.

PangMo5:
@LePips It's been complicated, but I've implemented it like YouTube.

PangMo5:
Wait a min. I'll improve a little more

Swiftfin/Views/VideoPlayer/VLCPlayerViewController.swift
```diff
@@ -237,22 +261,49 @@ class VLCPlayerViewController: UIViewController {
...
+		backgroundView.layer.zPosition = 1
+		backgroundView.alpha = 0
+		backgroundView.translatesAutoresizingMaskIntoConstraints = false
+		let button = UIButton(type: .custom, primaryAction: UIAction(handler: { [weak self] _ in
+			self?.isGesturesLocked = false
+			self?.hideLockedOverlay()
+			self?.didGenerallyTap()
+		}))
+		button.translatesAutoresizingMaskIntoConstraints = false
+		button.setImage(UIImage(systemName: "lock.open", withConfiguration: UIImage.SymbolConfiguration(pointSize: 48))?
```
LePips, line 281: I think just "lock", or "lock.circle.fill" would be a better indicator for when gestures are locked.

LePips (changes requested):

LePips:
Very impressive! Just that one request and fix with main

PangMo5:
@LePips approve plz. It's protected, so I can't force merge

LePips:
You need to merge your fork's main branch (that's up to date with upstream) into this branch to fix instead of a force merge.

PangMo5:
> You need to merge your fork's main branch (that's up to date with upstream) into this branch to fix instead of a force merge.

Is it solved now?

LePips:
It it fixed now, yes.

LePips (approved):

PR #425, opened 2022-04-30
Fixes compiler warnings for tvOS

TheNightmanCodeth:
Silenced some compiler warnings regarding OpenGL, non-constant range in a foreach, and a dangling pointer in `createSocket`. Hoping to contribute more to the tvOS side of things but thought I'd start here! Everything is building and running just fine on my end, but I'm honestly not sure of the implications of removing OpenGLES. It doesn't seem to be used anywhere in the project and I didn't get any dependency issues.


LePips:
Thanks for the contribution! The only thing is that OpenGLES is marked as required by VLCKit. While I was able to test with my device by removing OpenGLES, I don't want to remove a framework that it has marked as required so lets keep it.

Additionally, just a linting run is required.

LePips:
Your SwiftFormat isn't the latest version, which pulled in unnecessary changes.

LePips (approved):

PR #426, opened 2022-05-02
LiveTV support for iOS

jhays:
Adds LiveTV channels view to iOS. As well as custom LiveTV player / native player views, and LiveTV options to Experimental Settings.

Added display of additional "upcoming programs", and updated the tvOS LiveTV channel cells to display this data as well. 

iOS:
Simulator Screen Shot - iPhone 13 Pro Max - 2022-04-27 at 22 06 18
Simulator Screen Shot - iPhone 13 Pro Max - 2022-04-27 at 22 05 58
Simulator Screen Shot - iPhone 13 Pro Max - 2022-04-28 at 14 07 57

iPad:
Simulator Screen Shot - iPad Pro (9 7-inch) - 2022-04-28 at 14 27 11

tvOS:
Simulator Screen Shot - Apple TV 4K (2nd generation) - 2022-05-01 at 21 19 34


jhays:
@LePips do you have any idea what CI is having trouble with here? 

Logs seem to show: 

```
/Users/runner/work/Swiftfin/Swiftfin/Swiftfin/Views/VideoPlayer/LiveTVPlayerViewController.swift:813:1: error: type 'LiveTVPlayerViewController' does not conform to protocol 'PlayerOverlayDelegate'
[3225](https://github.com/jellyfin/Swiftfin/runs/6251539682?check_suite_focus=true#step:8:3225)
extension LiveTVPlayerViewController: PlayerOverlayDelegate {
[3226](https://github.com/jellyfin/Swiftfin/runs/6251539682?check_suite_focus=true#step:8:3226)
^
[3227](https://github.com/jellyfin/Swiftfin/runs/6251539682?check_suite_focus=true#step:8:3227)
/Users/runner/work/Swiftfin/Swiftfin/Swiftfin/Views/VideoPlayer/PlayerOverlayDelegate.swift:23:7: note: protocol requires function 'didGenerallyTap(point:)' with type '(CGPoint?) -> ()'
[3228](https://github.com/jellyfin/Swiftfin/runs/6251539682?check_suite_focus=true#step:8:3228)
        func didGenerallyTap(point: CGPoint?)
[3229](https://github.com/jellyfin/Swiftfin/runs/6251539682?check_suite_focus=true#step:8:3229)
             ^
[3230](https://github.com/jellyfin/Swiftfin/runs/6251539682?check_suite_focus=true#step:8:3230)
/Users/runner/work/Swiftfin/Swiftfin/Swiftfin/Views/VideoPlayer/PlayerOverlayDelegate.swift:24:7: note: protocol requires function 'didLongPress()' with type '() -> ()'
[3231](https://github.com/jellyfin/Swiftfin/runs/6251539682?check_suite_focus=true#step:8:3231)
        func didLongPress()
[3232](https://github.com/jellyfin/Swiftfin/runs/6251539682?check_suite_focus=true#step:8:3232)
             ^
```


But I am able to build Swiftfin iOS and tvOS locally without issue. 

And I cannot find matching method signatures `func didGenerallyTap(point: CGPoint?)` or `func didLongPress()` related to `PlayerOverlayDelegate`. So what is CI complaining about here?

LePips:
Changes were made in https://github.com/jellyfin/Swiftfin/pull/412. It looks like you need to merge main into your branch.

jhays:
> Changes were made in #412. It looks like you need to merge main into your branch.

Thanks, I think I sorted it

LePips:
Just as a heads up to set expectations, I will be able to get to this Saturday, May 7th.

Swiftfin/Views/LiveTVProgramsView.swift
```diff
@@ -10,7 +10,201 @@ import Stinsen
...
+									if let chanId = item.channelId,
+									   let chan = viewModel.findChannel(id: chanId)
+									{
+										self.viewModel.fetchVideoPlayerViewModel(item: chan) { playerViewModel in
+											self.programsRouter.route(to: \.videoPlayer, playerViewModel)
+										}
+									}
+								} label: {
+									#if os(iOS)
+									#elseif os(tvOS)
```
LePips, line 104: This file is only targeted for iOS, it will never reach tvOS, right? There are a few of these in this file.

Swiftfin/Views/LibraryListView.swift
```diff
@@ -16,7 +17,16 @@ struct LibraryListView: View {
 	@StateObject
 	var viewModel = LibraryListViewModel()
 
-	let supportedCollectionTypes = ["movies", "tvshows", "boxsets", "other"]
+	@Default(.Experimental.liveTVAlphaEnabled)
+	var liveTVAlphaEnabled
+
+	var supportedCollectionTypes: [String] {
+		if liveTVAlphaEnabled {
+			return ["movies", "tvshows", "livetv", "boxsets", "other"]
```
LePips, line 25: I recommend looking to see if these (and their usage through this file) could be changed to use the `itemType` property declared in `BaseItemDtoExtensions`.

Swiftfin/Views/VideoPlayer/LiveTVPlayerViewController.swift
```diff
@@ -0,0 +1,1033 @@
...
+		let view = UIView()
+		view.translatesAutoresizingMaskIntoConstraints = false
+		view.backgroundColor = .black
+
+		return view
+	}
+
+	// MARK: MainGestureView
+
+	private func makeMainGestureView() -> UIView {
```
LePips, line 202: With the gesture stuff being talked about in other PRs, you might have the responsibility of implementing whatever we decide in these tvOS views.

LePips (changes requested):
Since I was actually able to look at it, not as many comments as I thought.

Swiftfin tvOS/Views/LiveTVChannelsView.swift
```diff
@@ -11,6 +11,8 @@ import JellyfinAPI
 import SwiftUI
 import SwiftUICollection
```
LePips, line 12: I've been looking at collection views for other work and remember you chose to use this one. I would prefer to move to https://github.com/apptekstudios/ASCollectionView as it's largely more supported and (seemingly) maintainable as the target source for this package is no longer active (and also weirdly a competitor)

If you feel like switching over to that collection view is a lot of work I'm fine to let this in, however it will be a goal to switch over to what I prefer both here and in tvOS.

LePips (changes requested):

Swiftfin tvOS/Views/LiveTVChannelsView.swift
jhays: Understood, I can start looking into using `ASCollectionView` for Live TV stuff.

jhays: Sorry for the delay on this. I am in progress on the updates. I have to finish out the change to ASCollectionView but I think I should have it ready soon.

Swiftfin/Views/VideoPlayer/LiveTVPlayerViewController.swift
jhays: So, unfortunately it seems that `ASCollectionView` does not support tvOS, and for now there are no plans to add support: https://github.com/apptekstudios/ASCollectionView/issues/138

I discovered this after migrating my iOS LiveTV implementation over to `ASCollectionView` when I went to move on to move tvOS over to it as well. This is unfortunate because the `ASCollectionView` api was easier/simpler than `SwiftUICollection` and just generally felt better to work with. 

Options for moving forward:
1. Stick with `SwiftUICollection` for LiveTV on both iOS/tvOS
2. Add `ASCollectionView` and use it for LiveTV on iOS, while leaving the tvOS implementation on `SwiftUICollection`
3. Some other solution 

`ASCollectionView` definitely feels better to work with, but using it on iOS would then also mean further divergence between the implementations on iOS vs tvOS. 

@LePips What would be your preferred solution here?

Swiftfin tvOS/Views/LiveTVChannelsView.swift
jhays: I thought I had already left a comment about this, but I can't seem to see it now- 

@LePips it seems `ASCollectionView` does not support tvOS: 
https://github.com/apptekstudios/ASCollectionView/issues/138


How would you prefer I proceed? 

Would you rather see `ASCollectionView` for iOS and `SwiftUICollection` for tvOS? or Keep them both on the same with `SwiftUICollection` for now?

Swiftfin/Views/LibraryListView.swift
jhays: Addressed, but it required some minor changes to support creating `itemType` from raw string value.

Swiftfin tvOS/Views/LiveTVChannelsView.swift
PangMo5: I think it's ideal to use SwiftUI's LazyVGrid.
Is there still a performance problem?

PangMo5: Or we can look at SwiftUIX's CollectionView.
https://github.com/SwiftUIX/SwiftUIX/blob/95dfcbc789d7e04ee075b3b41296ee227aba6308/Sources/Intramodular/Collection%20View%20%26%20List/CollectionView.swift

LePips: Well, bummer. Thanks for looking into that and I apologize if it took a good amount of time.

Overall, we need to be moving to a collection view for these large lists of objects (channels, libraries). LazyVGrid + ScrollView is a performance problem and is the cause of the stutters. It is lazy loaded, but it doesn't dequeue cells when necessary and that's what we need.

I loaded up ASCollectionView in a tvOS project and see that most of the errors are about tvOS versioning and some elements that I can remove (refresh control elements, table view stuff, keyboard stuff) if I were to fork it. I can try that sometime. Because of this thought, I'm more inclined towards ASCollectionView on iOS and SwiftUICollection on tvOS until I take a look at that, especially if you've already redone the work for iOS in ASCollectionView. Sure it'll be a big difference but we can just change it back or I'm honestly fine just leaving it if this doesn't pan out.

Overall, I just want to move away from the original source of the package which is a fork with a single change to the original repo, https://github.com/defagos/SwiftUICollection, which PangMo also linked. Do you need the change that our current source has made?

In the end, I'm fine with either option and I'll work with whatever you choose. I just hope Apple announces an official SwiftUI CollectionView tomorrow.

jhays: Super cool, however I'd bet it will be iOS16+. What are your thoughts on that?

jhays:
Latest commit adds usage of `ASCollectionView` for iOS implementation of LiveTV channels view.

Swiftfin tvOS/Views/LiveTVChannelsView.swift
jhays: Wait, so you're saying `SwiftUICollection` is a fork of the SwiftUIX collection view that PangMo linked? 

I'm not even aware of what change the fork made, I wouldn't anticipate any specific issues if we wanted to move to the SwiftUIX collection view for tvOS. 

We can also look to move towards Grid API depending on os level supported

LePips: Oh sorry, I got my SwiftUICollection tabs mixed up. What I linked is the original SwiftUICollection and PangMo linked a different project. So again, you can make the decision to split or not.

Yes, the new Grid API will be iOS 16+, as well as the new NavigationStack that I want to use. We'll have this discussion elsewhere.

jhays: looking at it again, I don't think I need the change that the ABJC fork added. However, I think I grabbed that SPM package by accident. I just tried again, look at this screenshot... even though I enter the `defagos` git URL, the one that Xcode's SPM tool seems to grab says ABJC. I'm not sure what is happening here..

Screen Shot 2022-06-06 at 4 05 39 PM

jhays: Ok, I think I won't have any issue switching to the original source of `SwiftUICollection` rather than the ABJC fork I grabbed. 

I think I'll make that adjustment so tvOS is still using `SwiftUICollection` for now, and we could consider SwiftUI Grid API as a possible future replacement.

LePips (dismissed):

jhays:
@LePips Sorry, had to dismiss your review with one last commit, but I consider this PR finished now.

Final commit(s) added:

-  (mostly) fixed a bug where navigating to the end of content on tv or movies library caused a next page load, which would then refresh the entire collection view and return focus to the first item. This is now fixed at least so it doesn't reload the entire collection view, however, on page load / refresh, the focused item still jumps back a few rows. I'm not sure how to fix it and don't want to dive too deep as SwiftUICollection is a likely a temporary measure anyways. 
- Additionally reduces occurrence of this bug by doubling page size on tvOS
- Now prevents navigating back to the top level of movies/tv libraries view when only a single tv or movies library exists.

LePips (approved):

PR #428, opened 2022-05-03
Add Seek Slide Gesture

### LINKED ISSUE #427: double tap skip causes freezing UI
### Describe the bug

double tap skip causes freezing UI, and need 2s to recover.
I hope next build can go back to swipe gesture, more efficient.

### Application version

1.0.0 (75)

TestFlight

### Device information

iPhone 12 mini

### OS version

15.4.1

### Jellyfin server version

10.7.7

PangMo5:
resolve #427


LePips:
I apologize as I won't be able to get to this right away, however with the introduction of many gestures the player can start to get confusing (even with the option to remove them). I think overall things are getting a bit busy.

We should outline together exactly what actions we want the player to have and what form they take in gestures. While I do think that a feature-rich gesture set is impressive, simplicity (not less necessarily) and customizability leads to a better user experience. This will lead us to defining a "gesture profile" like I previously mentioned. I do think that being influenced by other players is a good thing, however I don't think we should implement them for the sake of implementing them.

PangMo5:
Personally, I'm not adding **any** gestures, added gestures that I usually use or that were very convenient.

And I don't know what kind of confusion it can cause to the user because there is no problem when all supported gestures are activated, and all gestures have activation options. Gesture lock function is also added so that you can focus on watching the video even when the gestures is enabled.

I think the gesture profile you are talking about is an additional convenience.
And I think this function should be taken care of separately with this issue, so it would be good to let me know the framework you think.
And I agree with the function of the gesture profile in the future, but I think it is overspec that is being added in current scope of gesture support.
I think it would be meaningful if a scope of gestures similar to [nPlayer](https://apps.apple.com/us/app/nplayer/id1116905928) were supported.

ex video

https://user-images.githubusercontent.com/16532526/166615707-81913145-9136-4642-835a-f3bbbe85d47c.MP4

LePips:
Just as a heads up to set expectations, I will be able to get to this Saturday, May 7th. This will also be merged after https://github.com/jellyfin/Swiftfin/pull/426 because I think that in the event of conflicts there would be less.

PangMo5:
> Just as a heads up to set expectations, I will be able to get to this Saturday, May 7th. This will also be merged after #426 because I think that in the event of conflicts there would be less.

okay 😄

LePips:
So my general review about this PR is that I think the gesture is okay however it would work better if we were to replicate how it works on Youtube where it shows the progress bar, where the time displays are shown down there as well. This would look far more professional but require a slight rework of the overlay. This behavior is a lot more friendly because it also shows the relative position of the scrub instead of a +/- and timestamp, as I don't entirely fancy the way that it looks now.

Thank you for showing how nplayer does their gestures as I think it's great to look up to!

PangMo5:
> So my general review about this PR is that I think the gesture is okay however it would work better if we were to replicate how it works on Youtube where it shows the progress bar, where the time displays are shown down there as well. This would look far more professional but require a slight rework of the overlay. This behavior is a lot more friendly because it also shows the relative position of the scrub instead of a +/- and timestamp, as I don't entirely fancy the way that it looks now.

When I have time, I will push the related overlay and additional UX UX improvement work.

PangMo5:
@LePips Please review it when you have time. And I uploaded a video on the element channel

LePips (approved):

PR #460, opened 2022-06-14
Modify External Subtitle Processing Logic

### LINKED ISSUE #417: External subtitles are not displayed
### Describe the bug

All external subtitles are not displayed.

No matter the format of external subtitles is srt or ass can not be displayed.

Embedded subtitles can be displayed.

The video formats tested are all mkv.

### Application version

1.0.0(74)

TestFlight

### Device information

iPhone 12 Pro

### OS version

IOS 15.4.1

### Jellyfin server version

10.8.0(docker nyanmisaka/jellyfin)

PangMo5:
The `deliveryUrl` of the external subtitles provided by Jellyfin 10.8 did not work properly, so I changed the logic by referring to the source.

Source: https://github.com/jellyfin/jellyfin-androidtv/blob/8c01a3d0ebd98355c4f068767f6ed4920a664767/app/src/main/java/org/jellyfin/androidtv/data/compat/StreamInfo.java

fix #417

### DIFF
--- a/Shared/Extensions/JellyfinAPIExtensions/MediaStreamExtension.swift
+++ b/Shared/Extensions/JellyfinAPIExtensions/MediaStreamExtension.swift
@@ -10,12 +10,12 @@ import Foundation
 import JellyfinAPI
 
 extension MediaStream {
-
-	func externalURL(base: String) -> URL? {
-		guard let deliveryURL = deliveryUrl else { return nil }
-		var baseComponents = URLComponents(string: base)
-		baseComponents?.path += deliveryURL
-
-		return baseComponents?.url
+	func externalSubtitleURL(base: String, item: BaseItemDto) -> URL? {
+		guard let id = item.id,
+		      let index = index,
+		      let format = codec else { return nil }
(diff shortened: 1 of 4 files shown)


LePips:
Unable to test for a bit. Does this break 10.7.7?

PangMo5:
In a different discussion, I wonder if it is okay to use playbackPositionTicks for startPositionTicks.

LePips:
I'll be able to test and let you know in a few days

LePips:
There is something incorrect in server 10.8.0 that incorrectly labels the subtitle format on the media stream, not with the delivery URL. At the same time, this change creates the same URL, just without the api key and some `-` that are handled by the server.

For proof that it is a server issue, I currently have `srt` subtitles that are incorrectly being labeled as `subrip` on my home server:
Screen Shot 2022-06-17 at 12 45 51 PM
Screen Shot 2022-06-17 at 12 51 03 PM

and with my development branch on `master` they are labeled correctly:
Screen Shot 2022-06-17 at 12 53 49 PM

I will have to find the logs and file a server issue.

PR #464, opened 2022-06-17
Update jellyfin-sdk-swift

PangMo5:
I checked 10.7.7 and 10.8 for simple actions (list, search, detail, play, playback reports).

- update Package ([jellyfin-sdk-swift](https://github.com/jellyfin/jellyfin-sdk-swift), etc..)
- update ci

### DIFF
--- a/Shared/ViewModels/HomeViewModel.swift
+++ b/Shared/ViewModels/HomeViewModel.swift
@@ -135,7 +135,7 @@ final class HomeViewModel: ViewModel {
 		                              	.people,
 		                              	.chapters,
 		                              ],
-		                              includeItemTypes: ["Movie", "Series"],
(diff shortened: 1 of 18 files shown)


Shared/ViewModels/HomeViewModel.swift
```diff
@@ -135,7 +135,7 @@ final class HomeViewModel: ViewModel {
 		                              	.people,
 		                              	.chapters,
 		                              ],
-		                              includeItemTypes: ["Movie", "Series"],
+		                              includeItemTypes: [.movie, .series],
```
LePips, line 138: Finally, this allows us to remove the `itemType` I created.

LePips (approved):

PR #470, opened 2022-07-03
Modify External Subtitle URL Logic

### LINKED ISSUE #417: External subtitles are not displayed

PangMo5:
Modify with reference to https://github.com/jellyfin/Swiftfin/issues/417#issuecomment-1173009298
Checked work normally in 10.7.7 and 10.8.1 (ass, srt)

### DIFF
--- a/Shared/Extensions/JellyfinAPIExtensions/MediaStreamExtension.swift
+++ b/Shared/Extensions/JellyfinAPIExtensions/MediaStreamExtension.swift
@@ -11,6 +11,10 @@ import JellyfinAPI
 
 extension MediaStream {
 	func externalURL(base: String) -> URL? {
+		var base = base
+		while base.last == Character("/") {
+			base.removeLast()
+		}
 		guard let deliveryURL = deliveryUrl else { return nil }
 		return URL(string: base + deliveryURL)
 	}
--- a/Shared/Extensions/JellyfinAPIExtensions/MediaStreamExtension.swift
+++ b/Shared/Extensions/JellyfinAPIExtensions/MediaStreamExtension.swift
@@ -10,12 +10,8 @@ import Foundation
 import JellyfinAPI
 
 extension MediaStream {
-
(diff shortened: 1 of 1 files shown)


Shared/Extensions/JellyfinAPIExtensions/MediaStreamExtension.swift
```diff
@@ -10,12 +10,8 @@ import Foundation
...
 
 extension MediaStream {
-
 	func externalURL(base: String) -> URL? {
 		guard let deliveryURL = deliveryUrl else { return nil }
-		var baseComponents = URLComponents(string: base)
-		baseComponents?.path += deliveryURL
-
-		return baseComponents?.url
+		return URL(string: base + deliveryURL)
```
LePips, line 15: Great find! However, we are not guaranteed that the base url does not end in a trailing `/`, and just simply appending it would break it. A check should be made first whether `base` ends in a trailing slash, removing it from the end of`base` or the first slash from `deliveryURL`, and then creating the URL.

LePips (changes requested):

LePips (approved):
However, please do not close #417 as we need to address ShadowGhost's latest comment in that issue about the track re-ordering. I experienced this on my own media where the external subtitle track and the list of available subtitle tracks on `vlcMediaPlayer` don't match. This throws the player into an infinite loop at `didSelectSubtitleStream(...)` since I check that the correct subtitle track has been selected almost every second. We could possibly remove the that check, however the track re-ordering still needs to be addressed.

PangMo5:
ok, I see.

PR #473, opened 2022-07-10
iOS - Add known public users to login screen

daullmer:
Add public users to login screen

https://user-images.githubusercontent.com/16425113/178153693-70f3f270-8a51-4b71-bdf1-ac4612a7b6b6.mov


Shared/Generated/Strings.swift
```diff
@@ -138,6 +138,8 @@ internal enum L10n {
   internal static func jumpLengthSeconds(_ p1: UnsafePointer<CChar>) -> String {
     return L10n.tr("Localizable", "jumpLengthSeconds", p1)
   }
+  /// Known users
+  internal static var knownUsers: String { return L10n.tr("Localizable", "knownUsers") }
```
LePips, line 142: Change the original localizable string to "Public Users", as "Known" means something different.

Shared/Views/UserLoginCellView.swift
```diff
@@ -0,0 +1,69 @@
...
+				           	Image(systemName: "person.circle")
+				           		.resizable()
+				           		.font(.system(size: 40))
+				           		.scaledToFit()
+				           		.frame(maxWidth: 50, maxHeight: 50)
+				           })
+				           .padding(.vertical, 4.0)
+
+				Text(user.name ?? "")
+					.padding(.leading, 4.0)
```
LePips, line 55: Please remove this padding.

Shared/Views/UserLoginCellView.swift
```diff
@@ -0,0 +1,69 @@
...
+				Spacer()
+			}
+		}
+	}
+
+	func getProfileImageUrl() -> URL? {
+		if let userId = user.id, let imageTag = user.primaryImageTag, let server = baseURL {
+			let url = URL(string: "\(server)/Users/\(userId)/Images/Primary?width=200&tag=\(imageTag)&quality=90")
```
LePips, line 63: Please use the API and these `nil` checks will not be necessary either:
```swift
ImageAPI.getUserImageWithRequestBuilder(userId: userId,
                                        imageType: .primary,
                                        width: 200,
                                        quality: 90)
```

LePips (changes requested):
Thank you for your contribution! Please see comments.

Swiftfin/Views/UserSignInView.swift
```diff
@@ -19,7 +19,19 @@ struct UserSignInView: View {
 	private var password: String = ""
 
 	var body: some View {
-		Form {
+		List {
+			#if !os(tvOS)
+				// DisclosureGroup not available on tvOS
+				if !viewModel.users.isEmpty {
```
LePips, line 25: Please put these below the original text fields. This is dynamic, there may be no public users or many public users and will push down the main part of the view and mirrors the way local servers are handled.

LePips (changes requested):

daullmer:
Thank you very much for your comments! I will change the things you requested and then work on the refresh functionality like in `ConnectToServerView`

Swiftfin/Views/PublicUserSignInCellView.swift
```diff
@@ -0,0 +1,44 @@
...
+		} label: {
+			HStack {
+				ImageView(viewModel.getProfileImageUrl(user: user)) {
+					Image(systemName: "person.circle")
+						.frame(width: 50, height: 50)
+				}
+				.frame(width: 50, height: 50)
+				.clipShape(Circle())
+
+				Text(user.name ?? "")
```
LePips, line 39: The "--" fallback should be used here when shown to the user.
```suggestion
				Text(user.name ?? "--")
```

Swiftfin/Views/PublicUserSignInCellView.swift
```diff
@@ -0,0 +1,44 @@
...
+			SecureField(L10n.password, text: $enteredPassword)
+			Button {
+				viewModel.login(username: user.name ?? "--", password: enteredPassword)
+			} label: {
+				L10n.signIn.text
+			}
+		} label: {
+			HStack {
+				ImageView(viewModel.getProfileImageUrl(user: user)) {
+					Image(systemName: "person.circle")
```
LePips, line 33: I apologize and was incorrect, `.resizable()` actually is required here.

LePips (changes requested):
Looks great! Just a couple small things and something I was wrong about.

LePips:
Would you like to handle the linting errors? I'm fine with leaving it and catching it later on.

PR #474, opened 2022-07-10
Use UDPBroadcastConnection library

daullmer:
Well, how about we use [the original](https://github.com/gunterhager/UDPBroadcastConnection) as a library and don't just copy stuff over.

This will prevent things like #432 from happening


LePips (approved):

LePips:
Thanks for the contribution!

PR #477, opened 2022-07-12
Add Quick Connect code entry on iOS in settings

daullmer:
Maybe a good preparation for adding Quick Connect Sign In for tvOS


Shared/Generated/Strings.swift
```diff
@@ -262,6 +262,12 @@ internal enum L10n {
   internal static var programs: String { return L10n.tr("Localizable", "programs") }
   /// Public Users
   internal static var publicUsers: String { return L10n.tr("Localizable", "publicUsers") }
+  /// Quick Connect
+  internal static var quickConnect: String { return L10n.tr("Localizable", "quickConnect") }
+  /// Authorize
+  internal static var quickConnectAuthorize: String { return L10n.tr("Localizable", "quickConnectAuthorize") }
```
LePips, line 268: We can just call the key "authorize" instead of "quickConnectAuthorize".

Shared/ViewModels/SettingsViewModel.swift
```diff
@@ -44,4 +55,26 @@ final class SettingsViewModel: ObservableObject {
 		}.sorted(by: { $0.name < $1.name })
 		self.langs.insert(.auto, at: 0)
 	}
+
+	func sendQuickConnect() {
+		QuickConnectAPI.authorize(code: self.quickConnectCode)
+			.sink(receiveCompletion: { completion in
+				switch completion {
+				case .failure:
+					LogManager.log.debug("Invalid Quick Connect code entered")
```
LePips, line 64: This failure isn't necessarily correct, as a failure here does not indicate an invalid code entered but a failure of the call due to HTTP errors and such. `handleAPIRequestError` also needs to be called to handle the entire completion in all API calls. You can just use `self.handleAPIRequestError(completion: completion)` in this closure, and for future reference the `displayMessage` parameter is to override the default, generic "Error" title and these custom errors need to be localized.

Shared/ViewModels/SettingsViewModel.swift
```diff
@@ -44,4 +55,26 @@ final class SettingsViewModel: ObservableObject {
...
+				default:
+					self.handleAPIRequestError(displayMessage: "Error", completion: completion)
+				}
+			}, receiveValue: { value in
+				if !value {
+					LogManager.log.debug("Invalid Quick Connect code entered")
+					self.validQuickConnect = false
+				} else {
+					LogManager.log.debug("Valid Quick connect code entered")
+					self.router?.dismissCoordinator()
```
LePips, line 75: I don't think we should dismiss the coordinator, so the router reference isn't necessary as well.

Swiftfin/Views/SettingsView/SettingsView.swift
```diff
@@ -83,6 +83,23 @@ struct SettingsView: View {
 				}
 			}
 
+			Section(header: L10n.quickConnect.text) {
```
LePips, line 86: Instead of making another section in the main settings page, make another `Button` in the first section that routes to a view to handle Quick Connect. You can see how this is done for many other views in `SettingsCoordinator`.

Shared/ViewModels/SettingsViewModel.swift
```diff
@@ -8,16 +8,27 @@
...
 	var langs: [TrackLanguage] = []
 
 	let server: SwiftfinStore.State.Server
 	let user: SwiftfinStore.State.User
 
+	@Published
+	var quickConnectCode = ""
+
+	@Published
+	var validQuickConnect = true
```
LePips, line 30: To indicate errors, we instead present an alert instead of this handling error state this way. `ViewModel` already has the error variable `errorMessage` to be modified and presented to the user using `.alert`. You can see a simple example in `UserSignInView` how an error is presented. 

You will also need to indicate loading, like done in your last PR about user sign in. You will need to attach `.trackActivity(loading)` to the API call, as done in `login` in `UserSignInViewModel` for reference. I forgot to request this change to your Public Users PR to do this for the `isLoadingUsers` as well instead of handling the boolean manually. Since I mention this comment here, I'm fine if you do this now, or later, or I'll probably get to it when I do another large clean up PR. You can then reference `loading` like you've done before and disable the "Authorize" button while loading.

LePips (changes requested):

Shared/ViewModels/SettingsViewModel.swift
LePips: I also just had the thought to keep the button disabled until a full 6 character code.

daullmer: I added the `trackActivity` for the Public Users as well

PR #479, opened 2022-07-13
Improve Chapters features of Player

PangMo5:
- Masked chapters position in seekbar
- Add current chapter display UI
	- tapped action: show chapters list(existing feature)
- Set `UDPBroadcast.xcframework` to Embed Frameworks

https://user-images.githubusercontent.com/16532526/178713525-6c087c8d-318f-446e-ae85-46c435eff124.mov


Swiftfin/Views/VideoPlayer/Overlays/VLCPlayerOverlayView.swift
```diff
@@ -339,59 +339,96 @@ struct VLCPlayerOverlayView: View {
...
+						Button {
+							viewModel.playerOverlayDelegate?.didSelectChapters()
+						} label: {
+							HStack {
+								Text(currentChapter.name ?? "--")
+								Image(systemName: "chevron.right")
 							}
+							.font(.system(size: 16, weight: .semibold, design: .default))
+						}
+						.padding(.leading, 16)
```
LePips, line 355: This padding isn't necessary.

LePips (changes requested):
Great! This is also a large visual change, please make in an option in `Settings -> Overlay` and I think it being on by default is good. I ultimately plan on making an overlay settings page in the player alongside playback information, so these settings will be more accessible.

I need to do a lot of cleanup in these overlays.

LePips:
This is a great implementation. I think the next step for this feature is instead of just creating the separators, each chapter has a view section and can be highlighted. Additionally, snapping to the beginning of chapters if you've scrubbed close enough as a behavior implementation. So, pretty much like Youtube. I don't know how specifically we would accomplish the highlighting feature, maybe with the `ZStack` stuff you've already done or contributing to the base sliders project (much harder).

PangMo5:
That's a good idea! Maybe we can make it similar enough with the current implementation. I will work on it when I have time later.

LePips (approved):

PR #484, opened 2022-07-14
Use homebrew SwiftGen instead of included library

LePips:
Clean up along with https://github.com/jellyfin/Swiftfin/pull/483. `contributing.md` updated with steps and some clarifications.

We already make people use homebrew for Carthage, include `swiftgen` and `swiftformat` as well.


PangMo5 (commented):
How about using Brewfile? It's gonna be a little easier.
Also In the case of swiftformat, we can control the version completely using [mint](https://github.com/yonaskolb/Mint) (Brew only supports the latest version of package).

LePips:
I'm fine with a Brewfile if we want one. The SwiftFormat issue is just an annoying bug that happened with the release yesterday. There's already an issue on the repo since it effects almost everybody and expect it to be fixed soon. After all, this is the first time this has happened since we've used SwiftFormat. I also don't want to default recommending using both Homebrew and Mint, especially just for one dependency. People can download them too through Mint if they know about it and want to.

PangMo5:
Then, I think it would be good to apply Brewfile.

PangMo5 (approved):
I think it's okay to apply the brewfile in another PR.
Do as you please.

LePips:
I mean, it's just one line `brew install` or one line `brew bundle` with a new file with 3 lines. I'm more inclined to not have one unless we have a more complex build.

PangMo5:
Well, okay. Think about it when we have more dependencies.

PR #482, opened 2022-07-14
No Tab Characters and Before First for Argument and Parameter Wrapping

LePips:
Remove `--indent tab` as it just made tab characters that weren't changing anything worthwhile. I realize that these were  causing small linting annoyances like in [this recent commit](https://github.com/jellyfin/Swiftfin/pull/473/commits/1e73e90d4f0eb92329711873661e8ca7c70b0522). This was also prevalent when any change was made on a line, even if the tabs were not.

Change to `before-first` for `--wraparguments` and `--wrapparameters`. This was brought up in [this comment](https://github.com/jellyfin/Swiftfin/discussions/451#discussioncomment-3144747) and after playing with it and seeing the new spacing, I like it. A noticeable change for all API calls and function declarations. Using this style is easy in existing calls when adding new arguments/parameters and for making new argument/parameter lists. You can still develop however you want and then the linter handles it. 

There is the question of the `)` being on the next line instead of being on the same line as the last parameter. After seeing the difference, I have come to prefer it being on the next line since we deal a lot with Combine operators and the indentation would cause confusion.

<details>
<summary>Old argument style</summary>

```swift
UserLibraryAPI.getLatestMedia(userId: SessionManager.main.currentLogin.user.id,
                              fields: [
                                  .primaryImageAspectRatio,
                                  .seriesPrimaryImage,
                                  .seasonUserData,
                                  .overview,
                                  .genres,
                                  .people,
                                  .chapters,
                              ],
                              includeItemTypes: [.movie, .series],
                              enableImageTypes: [.primary, .backdrop, .thumb],
                              enableUserData: true,
                              limit: 8)
```

</details>

<details>
<summary>New argument style</summary>

```swift
UserLibraryAPI.getLatestMedia(
    userId: SessionManager.main.currentLogin.user.id,
    fields: [
        .primaryImageAspectRatio,
        .seriesPrimaryImage,
        .seasonUserData,
        .overview,
        .genres,
        .people,
        .chapters,
    ],
    includeItemTypes: [.movie, .series],
    enableImageTypes: [.primary, .backdrop, .thumb],
    enableUserData: true,
    limit: 8
)
```
</details>

<details>
<summary>Old parameter style</summary>

```swift
    init(item: BaseItemDto,
         title: String,
         subtitle: String?,
         directStreamURL: URL,
         transcodedStreamURL: URL?,
         hlsStreamURL: URL,
         streamType: ServerStreamType,
         response: PlaybackInfoResponse,
[...]

PR #486, opened 2022-07-15
Add Quick Connect sign in to tvOS

### LINKED ISSUE #288: Support QuickConnect
Support signing in via QuickConnect

https://jellyfin.org/docs/general/server/quick-connect.html

daullmer:
Simulator Screen Shot - Apple TV 4K (at 1080p) (2nd generation) - 2022-07-15 at 20 31 42

Not sure if white is the correct color for the text.

Steps Strings were copied from the Android TV implementation

Fixes #288


PangMo5:
I think it's the same reason as https://github.com/jellyfin/Swiftfin/pull/482#issuecomment-1185043700

LePips:
Thank you for implementing this but I apologize, this implementation is not correct. `SessionManager` should and needs to handle all formal auth and the builder cannot be used, or builders of this nature really with only one specific purpose. All Quick Connect sign in functionality can and should be done in `UserSignInViewModel`. I know that you didn't know this, but I will be moving away from singletons and pushing them into the Environment for more proper SwiftUI practice. They were good for development speed when I had less time.

Please check out the work I replicated at my branch: https://github.com/LePips/SwiftFin/tree/quickconnect-tvos with a different implementation.

I changed the layout just a bit to name the section "Quick Connect", centering it and the code as well. Please also _default_ to using the `spacing: ...` argument of `_Stack` over `.padding(.vertical/.horizontal)`. After good default spacing has been used then additional padding can be used on some views where desired.

String change from "webbrowser" to "web browser". Android TV should fix that as well. I'm fine with the grammar of the steps now but will probably change them later on, long after this feature has been merged.

PR #487, opened 2022-07-16
Add Quick Connect sign in to tvOS (v2)

### LINKED ISSUE #288: Support QuickConnect

daullmer:
Reimplementation of #486 

Thanks @LePips for the help

Fixes #288

Simulator Screen Shot - Apple TV 4K (at 1080p) (2nd generation) - 2022-07-16 at 17 30 31


Shared/ViewModels/UserSignInViewModel.swift
```diff
@@ -79,4 +89,60 @@ final class UserSignInViewModel: ViewModel {
...
+    func startQuickConnect() {
+        QuickConnectAPI.getEnabled()
+            .sink(receiveCompletion: { completion in
+                self.handleAPIRequestError(completion: completion)
+            }, receiveValue: { enabled in
+                self.quickConnectEnabled = enabled
+            })
+            .store(in: &cancellables)
+
+        QuickConnectAPI.initiate()
```
LePips, line 102: Ah okay, the error I was getting on my remote server was because it wasn't enabled and I didn't know we had to check. `initiate()` is still called even when it's disabled, properly getting an error but we shouldn't show _that_ error to the user. So break out this initiate call into a private function `initiateQuickConnect` which is called in `getEnabled` if it's enabled.

LePips (changes requested):

LePips:
Just lint and then we can have this in the next release

LePips (approved):

Issue #489, opened 2022-07-19
Series Items are clipped when focussed in tvOS


adamjcampbell:
### Describe the bug

1. Go to a series
2. Select a season

Expected: Season is scaled to a size of `1.1` with all content fully visible
Actual: Content is clipped so that you cannot see number of unplayed episodes or the season and year text properly

![clipped](https://user-images.githubusercontent.com/5327203/179734921-08cb6467-15ae-49bd-be34-30f8a6133eaf.png)

### Application version

1.0.0 (75)

### Where did you install the app from?

TestFlight

### Device information

Apple TV 4K

### OS version

tvOS 15.4

### Jellyfin server version

10.8.1

adamjcampbell:
I'll throw up a PR for this even though it may be included in @LePips's PR as I've already done it

LePips:
I closed the PR, but I'll close this issue when I finish my work.

PR #490, opened 2022-07-19
Fix series item clipping

adamjcampbell:
Closes: #489 

Increases frame heights to allow for the scale effect.

**Note:** any less height than this and the shadow gets clipped resulting in a line artifact at the top edge of the horizontal scroll view.

| Before | After |
| --- | --- |
| clipped | clipped-fix |

The same fix has been applied to 'More Like This' such that the design now looks like:
Simulator Screen Shot - Apple TV 4K (at 1080p) (2nd generation) - 2022-07-19 at 21 14 41


adamjcampbell:
@LePips let me know if there's anything I can help with to get your PR though. I have started using Swiftfin daily and would love to contribute

PR #498, opened 2022-07-23
Support subtitle font selection

### LINKED ISSUE #497: Some Chinese subtitles dispaly as 口(▢ a hollow square)
### Describe the bug

Only a few of words show wrongly.
e.g.
IMG_1058
ARCANE S1E4 12:23, using Subrip subtitle. The 口 in the subtitle should be "稳", which display correctly on Jellyfin web client:
无标题
Missing font?

### Application version

1.0.0(76)

TestFlight

### Device information

Both iPhone XS MAX and iPad 2019

### OS version

IOS 15.6, iPad OS 15.6

### Jellyfin server version

10.8.1 unofficial docker(nyanmisaka/jellyfin) which adds addtional chinese support.

PangMo5:
- Add subtitle font selection in Settings (Default value: .systemFont)


Shared/Coordinators/FontPickerCoordinator.swift
```diff
@@ -0,0 +1,42 @@
...
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2022 Jellyfin & Jellyfin Contributors
+//
+
+import Foundation
+import Stinsen
+import SwiftUI
+
+final class FontPickerCoordinator: NavigationCoordinatable {
```
LePips, line 13: A separate coordinator is not needed here.

Shared/Coordinators/SettingsCoordinator.swift
```diff
@@ -71,6 +72,10 @@ final class SettingsCoordinator: NavigationCoordinatable {
             let viewModel = QuickConnectSettingsViewModel()
             QuickConnectSettingsView(viewModel: viewModel)
         }
+
+        func makeFontPicker(onFontPick: @escaping (UIFontDescriptor) -> Void) -> NavigationViewCoordinator<FontPickerCoordinator> {
```
LePips, line 76: Since no coordinator is needed, can just be:
```
        func makeFontPicker() -> some View {
            FontPickerView()
                .navigationTitle(L10n.subtitleFont)
        }
```

LePips (changes requested):
It's quite fun to play around with the fonts! However, I get a crash from opengl every other time playback is started. Probably from the `Selector(("setTextRendererFont:"))` as this is currently a workaround for VLCKit. I'm not comfortable shipping this because of these crashes.

I would also like to find out if this directly solves the issue. The author of the issue isn't necessarily correct that it's a "missing font" issue, it just looks like the font doesn't support that character. I know that some other fonts might have these missing characters but we would need to test.

Shared/Coordinators/SettingsCoordinator.swift
```diff
@@ -32,6 +31,8 @@ final class SettingsCoordinator: NavigationCoordinatable {
     #if !os(tvOS)
         @Route(.push)
         var quickConnect = makeQuickConnectSettings
+        @Route(.modal)
```
LePips, line 34: Since no coordinator is needed, can be `@Route(.push)`

LePips (changes requested):

PangMo5:
Yes, I think it can definitely lead to problems. Did you see the error on a device other than the simulator? If you saw it on a real device, we could move it to the experimental settings.

LePips:
Sorry, I haven't been able to test on a physical device because I've been out of town so I'll be able to do that later. I do have that exact episode so I might be able to load up the exact word. The crashing occurred on any font.

Swiftfin/Views/Representable/FontPicker.swift
```diff
@@ -0,0 +1,61 @@
...
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2022 Jellyfin & Jellyfin Contributors
+//
+
+import SwiftUI
+import UIKit
+
+// https://github.com/SwapnanilDhol/SUIFontPicker
+public struct FontPicker: UIViewControllerRepresentable {
```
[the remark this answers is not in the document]
PangMo5: I agree with the other parts and will fix them.

However, since FontPickerView is literally a picker, I think it's better to process it with the font information selected from the outside(ex: `SettingsView`) rather than setting the `Defaults`'s value inside
so I will keep 'onFontPick: (UIFontDescriptor) -> Void'.

Swiftfin/Views/SettingsView/SettingsView.swift
```diff
@@ -185,6 +187,22 @@ struct SettingsView: View {
                         Text(appearance.localizedName).tag(appearance.rawValue)
                     }
                 }
+
+                Button {
+                    settingsRouter.route(to: \.fontPicker) { font in
```
LePips, line 192: We still don't need to pass an `onFontPick` closure around and I don't think popping on a selection is a nice behavior while a user is looking around.

Swiftfin/Views/FontPicker.swift
```diff
@@ -0,0 +1,50 @@
...
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2022 Jellyfin & Jellyfin Contributors
+//
+
+import Defaults
+import SwiftUI
+import UIKit
+
+// https://github.com/SwapnanilDhol/SUIFontPicker
```
LePips, line 13: When the `onFontPick` is removed we can remove this attribution as it's different enough from that source.

LePips (changes requested):

Swiftfin/Views/SettingsView/SettingsView.swift
```diff
@@ -185,6 +187,22 @@ struct SettingsView: View {
...
+                        subtitleFontName = font.postscriptName
+                        settingsRouter.popLast()
+                    }
+                } label: {
+                    HStack {
+                        L10n.subtitleFont.text
+                            .foregroundColor(.primary)
+                        Spacer()
+                        Text(subtitleFontName)
+                            .foregroundColor(.gray)
```
LePips, line 202: Chevron needs to be added.

LePips (changes requested):

LePips:
Okay, testing on a real device hasn't crashed on changing fonts. All good for normal use.

LePips:
Sadly, I don't think this necessarily fixes the original issue. I've tested that same media on different fonts and still get the box character or another character in that space. Due to that, I think this is a VLCKit issue so we can close the issue. Nonetheless, this is still a really cool feature.

Swiftfin/Views/SettingsView/SettingsView.swift
PangMo5: hmm okay..

PangMo5:
> Sadly, I don't think this necessarily fixes the original issue. I've tested that same media on different fonts and still get the box character or another character in that space. Due to that, I think this is a VLCKit issue so we can close the issue. Nonetheless, this is still a really cool feature.

I see :d

LePips (approved):

PR #512, opened 2022-08-05
Add ContentView ViewBuilder to ImageView initializer

PangMo5:
I don't know if this is the best way to implement it
You can use it like the following code:

```swift
ImageView(
	URL(~,
	resizingMode: .aspectFit
 )
```

```swift
ImageView(
	URL(~,
	resizingMode: .aspectFit,
    failureView: {
	Color.red                        
		.frame(width: 30, height: 30)
})
```

```swift
ImageView(
	URL(~,
	resizingMode: .aspectFit
 ) { view in
	view
		.frame(width: 300, height: 300)
} failureView: {
	Color.red                        
		.frame(width: 30, height: 30)
}
```

### DIFF
--- a/Shared/Views/ImageView.swift
+++ b/Shared/Views/ImageView.swift
@@ -21,49 +21,12 @@ struct ImageSource {
     }
 }
 
-struct DefaultFailureView: View {
-
-    var body: some View {
-        Color.secondary
-    }
-}
-
-struct ImageView<FailureView: View>: View {
-
-    @State
-    private var sources: [ImageSource]
-    private var currentURL: URL? { sources.first?.url }
-    private var currentBlurHash: String? { sources.first?.blurHash }
-    private var failureView: () -> FailureView
-    private var resizingMode: ImageResizingMode
-
-    init(
-        _ source: URL?,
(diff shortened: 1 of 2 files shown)


LePips:
I came up an entirely different implementation and you can see a work in progress of it with the `RefactoredImageView` here: https://github.com/LePips/SwiftFin/tree/image-view-refactor. This creates a temporary refactored view with an additional custom `placeholderView` parameter. I also add `maxWidth` for the image URL generation, which we probably should have done the entire time. It still has some kinks to work out.

Secondly, our problem actually comes from the missing resizing modes from NukeUI. I have created [a PR](https://github.com/kean/Nuke/pull/582) for it. By testing locally with the `.bottomLeft` resizing mode, it will work as expected.

Finally, with images now having the `maxWidth/maxHeight` being passed in and scaled to the screen as the URL is being created we don't even need the `.frame` modifier as the image itself will be the correct dimensions. So this means I will just add a couple of other `init`s to that refactored image view so that you do not have to pass in the `ViewBuilder` if you do not want to.

PangMo5:
Good. I'll check it out when the PR comes up.

PR #515, opened 2022-08-07
Change condition of VideoPlayer's ToggleOverlay

### LINKED ISSUE #514: Screen looks like "locked" if clicking before the animation ends
### Describe the bug

1.Play any video
2.Click the screen
E7F1EB78-4976-4010-B78C-5CC82F456FF2
3.Now click left(right) of screen to hide the buttons
4.Click right(left) of screen after the fade out animation starts but before it ends.
5.The screen is "locked" for seconds.
6.If you return to home before it recovery, the whole app stuck untill you kill it.

### Application version

1.0.0(76)

_No response_

### Device information

iPhone XS Max

### OS version

IOS 15.6

### Jellyfin server version

10.8.2

PangMo5:
Fixes #514, #427

### DIFF
--- a/Swiftfin/Views/VideoPlayer/VLCPlayerViewController.swift
+++ b/Swiftfin/Views/VideoPlayer/VLCPlayerViewController.swift
@@ -728,9 +728,7 @@ extension VLCPlayerViewController {
     }
 
     private func toggleOverlay() {
-        guard let overlayHostingController = currentOverlayHostingController else { return }
-
-        if overlayHostingController.view.alpha < 1 {
+        if viewModel.isHiddenOverlay {
             showOverlay()
         } else {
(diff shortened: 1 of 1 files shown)


LePips:
I'll first test to see if this fixes it and everything behaves properly but the view model should not store whether a _view_ overlay is shown. There's a lot of cleanup in these views I can do and in my typical fashion: a lot of this will be rewritten as I look at downloads and ChromeCast support since I've had some thoughts.

PangMo5:
I didn't want to save overlay to ViewModel, but it was implemented like this because of gesture-related issues and animation issues between UIKit and SwiftUI.
If you can improve that part, it would be good to change it.

LePips (approved):

PR #517, opened 2022-08-08
`ImageView` Refactor

LePips:
- Refactors `ImageView` with new declarations and helper functions
- Uses the `BlurHashView` now provided in BlurHashKit
- Updated to use the content modes I added in NukeUI

New content modes are used to fix the placement of images.

Using `.resizingMode(.bottomLeft)` for example below. These are tvOS vs iPadOS but both had the same problem.
<details>
<summary>Before</summary>

182706047-4776f9ec-fcf5-4483-9180-77b01685bd87
</details>

<details>
<summary>After</summary>

after
</details>

### However

Even with the new content modes this does not fix the logos on the iOS logo and cinematic views. I cannot use a combination of `.bottom` and `.frame` to achieve this. The iPadOS and tvOS views can be easily done because we request the image sizes in the perfect pixel sizes relative to the screen but on iOS I grab the device width and then there's a padding, so with the previous combination the view would actually be clamped on the sides because the image is wider than its container. This can most likely be fixed with a `GeometryReader` so that we get the image rightly sized in the first place and then use `.bottom` but I don't do that here.


PangMo5 (approved):
I think it's much easier to use than before 👍

PR #522, opened 2022-08-09
iOS/iPadOS Quick Connect

LePips:
- Adds Quick Connect to iOS and iPadOS
- Refactors the underlying Quick Connect logic a bit

Design is bare-bones by being a modal as we shouldn't show the whole steps and code since there's less real estate on iOS*. I tried a basic alert but it isn't supposed to work this way and at least didn't work for me by throwing in a `VStack`. I may look at implementing a custom alert package in the future.

<details>
<summary>iOS Quick Connect</summary>

</details>


PangMo5 (dismissed):
I've used it and it seems to work very naturally. 👍

PangMo5 (approved):

LePips:
Oops sorry, right as you approve I also removed the old widget scheme. Thanks for you being quick.

Issue #494, opened 2022-07-20
Resume watching use title cards instead of series backdrop


jmvbrown:
Option to use the title card for the episode instead of the series backdrop

jmvbrown:
Yeah, so I have title cards I’ve created for episodes so it is like the series backdrops just episode specific.
![55A82A0E-4560-453C-A836-820E6174AB17](https://user-images.githubusercontent.com/45808940/180042356-2d7ec30c-ed67-4f63-8f69-fb035dabe9c9.png)

jmvbrown:
I also know this is an option on the other Jellyfin iOS app so just wanted to see if it was going to be an option on Swiftfin

LePips:
Oh, the portrait photos. This can be an option!

jmvbrown:
Thank you!

Issue #442, opened 2022-05-19
Horizontal thumbs


dannymichel:
I know it might sound stupid but a huge reason I switched to Emby and ultimately Jellyfin is because next up and continue watching are horizontal thumbnails rather than vertical posters. It'd be great if that were included in swiftfin.

pixselious:
An option to have 16:9 thumbnails in the libraries would be great too. I have a library for YouTube videos, and it just crops the thumbnails.

Issue #329, opened 2022-01-20
Device Dependent Item Sizes


cheetashock:
With the new collections integration, I was checking them out. It took quite a long time for me to figure out that I had to scroll horizontally to see more content. Horizontal scrolling isn't an issue, just make it more obvious that you need to scroll horizontally. Maybe make it so the next movie "peaks out"?

![image](https://user-images.githubusercontent.com/1431563/150439540-d65bec6b-7887-4f1d-8ed5-60fac4aff2c1.png)

LePips:
This definitely sounds like a design improvement. What device are you using? On my big iPhone 11 Pro Max, items peek just fine but we may need to adjust some sizing depending on device, which is fine.

cheetashock:
iPhone 12 Pro, so you are correct.

PR #526, opened 2022-08-12
iOS/iPadOS - Landscape/Thumb Posters

LePips:
- Closes https://github.com/jellyfin/Swiftfin/issues/442
  - 2nd comment is not applicable as we are not focusing on any Youtube video-focused design at the moment
- Closes https://github.com/jellyfin/Swiftfin/issues/329
- Closes https://github.com/jellyfin/Swiftfin/issues/494
- ~~Depends on https://github.com/jellyfin/Swiftfin/pull/517~~

Allows users to customize either portrait posters or landscape posters (thumbs) but only on the horizontal scrolling views. It won't look like much for most people as default content providers don't have thumbs but will look nice for people who do. 

Additional features:
- One can customize the home screen per Next Up, Recently Added, and Latest in Library views.
- Aggregate customization Settings to a single view
- Watched indicators for episodes
- Can customize whether episode or series backdrop for episodes for landscape posters

<details>
<summary>iOS Home Screen - all thumbs</summary>

</details>

<details>
<summary>iPadOS Home Screen - mixed</summary>

IMG_0159
</details>

### Developing

`PosterHStack` and `PosterButton` have been created which take a `.portrait/.landscape` type for what view it should be with a large API for customization. This uses a lot of generics which I think Swift 5.7 can help with.

- I stubbed out the customization option for recommended items as well. Didn't have the motivation to do it for the ton of item views that use it, but someone else can
- Collection views are ignored in this because I want it to more replicate library views
- Should refine poster sizing to probably the height instead of the width? Would need some tinkering
- General cleanup with our Defaults
- I did _not_ localize even though I say PRs need to be accepted with localizations. I want to wait until after the strings are cleaned up and I have stubbed out where to localize.

**Not on tvOS** due to issues with tvOS with the new item views. I port over the views to tvOS but leave them only as portrait. I will take a deep dive on tvOS for a bunch of stuff and this will be a part of it.

### Next Steps

- Proper loading/empty states. Right now you can tap on the loading/empty views and it will take you to a "not implemented yet" screen


Shared/Extensions/ViewExtensions/ViewExtensions.swift
```diff
@@ -46,12 +46,28 @@ extension View {
         }
     }
 
+    func poster(type: PosterType, width: CGFloat) -> some View {
```
PangMo5, line 49: If it's that name, I feel like It's going to return the `poster` itself.
What about naming like `posterStyle`?

PangMo5 (approved):

Shared/Extensions/ViewExtensions/ViewExtensions.swift
LePips: I'll do `posterStyle` in future work so that the change won't remove your approval

PR #530, opened 2022-08-18
Update JellyfinAPI

LePips:
This PR servers as the (long) transition to the new async/await provided in the new JellyfinAPI.

I am developing this transition with the following steps:
1. remove jellyfin-sdk-swift
2. have a local copy of jellyfin-sdk-swift at version 0.1.0, rename _folder_ to `jellyfin-sdk-swift-legacy`
3. rename the target to `JellyfinAPILegacy`
4. import as local package and update imports to `JellyfinAPILegacy`
5. add jellyfin-sdk-swift at version 0.2.0

This allows incremental updates instead of having to replace all of the calls at once before a successful build.

- [ ] make async/await wrappers
- [ ] inject some sort of session object into environment
    - this helps with resetting the app view on logout
- [ ] ActivityIndicator replacement
- [ ] refactor view models error handling
- [ ] replace calls...
    - might require slight adjustments to other views/callsites that require a full URL, like `ImageView`

### DIFF
--- a/Shared/Coordinators/HomeCoordinator.swift
+++ b/Shared/Coordinators/HomeCoordinator.swift
@@ -7,7 +7,7 @@
 //
 
 import Foundation
-import JellyfinAPI
+import JellyfinAPILegacy
 import Stinsen
 import SwiftUI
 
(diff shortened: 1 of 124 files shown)


LePips:
After starting this a bit, I'm not so inclined to expedite this so that other features can be worked on. This will be in the background.

LePips:
Will be done in #593.

PR #536, opened 2022-08-23
iOS/iPadOS - Transition library views to Collection Views

LePips:
Transitions the library views on iOS/iPadOS to my new [CollectionView](https://github.com/LePips/CollectionView). This moves us away from `Lazy_Grid` which has many performance problems. There is customization for choosing either portrait or thumb buttons for these library views as [previously implemented](https://github.com/jellyfin/Swiftfin/pull/526). It looks pretty good with poster labels disabled.

<details>
<summary>iOS portrait</summary>

</details>

<details>
<summary>iOS thumb</summary>

</details>

<details>
<summary>iPadOS thumb, landscape orientation</summary>

</details>

tvOS was transitioned as well but without the portrait/thumb customization as also previously mentioned in that PR. This will be done in the next "tvOS deep dive" I've mentioned here and there. I just implemented it as well while I was developing CollectionView. The tab bar does not hide on scrolling with CollectionView and this requires some manual work to manually hide it, which I have tested but is not included here.

<details>
<summary>tvOS portrait</summary>

tvos-portrait
</details>

<details>
<summary>tvOS tab bar not hiding</summary>

tvos-portrait-tab
</details>

There are _some_ bugs with CollectionView that I have observed, like sometimes items overlap as they are coming onto the screen while scrolling and then they move themselves as expected. However this was rare enough that I couldn't record it and this will have to be done in CollectionView.


Swiftfin tvOS/Components/PosterButton.swift
```diff
@@ -26,9 +28,9 @@ struct PosterButton<Item: Poster, Content: View, ImageOverlay: View, ContextMenu
     private var itemWidth: CGFloat {
         switch type {
         case .portrait:
-            return portraitPosterWidth * itemScale
+            return PosterButtonWidth.portrait * itemScale
```
LePips, line 31: This will match the iOS one-liner in the future

Swiftfin/Views/LiveTVChannelsView.swift
```diff
@@ -37,29 +36,30 @@ struct LiveTVChannelsView: View {
...
-                self.checkOrientation()
-            }
-            .onDisappear {
-                viewModel.stopScheduleCheckTimer()
-            }
-            .onReceive(NotificationCenter.default.publisher(for: UIDevice.orientationDidChangeNotification)) { _ in
-                self.checkOrientation()
-            }
+            Text("nothing")
+//            ASCollectionView(data: viewModel.channelPrograms, dataID: \.self) { channelProgram, _ in
```
LePips, line 40: I'm fine with this right now until I get to implementing CollectionView

PangMo5 (approved):

Issue #575, opened 2022-09-11
See all next up


chnaski:
The same way you click on ``see all`` for latest shows, it'd be nice to click on ``see all`` for next up.

PR #583, opened 2022-09-13
Generic Libraries

LePips:
- Closes https://github.com/jellyfin/Swiftfin/issues/575

Makes libraries more generic so that it can be used for more contexts.

### Details
- Next Up, Recently Added, Recommended, and Cast and Crew can be displayed as libraries
- Recently Added now uses items endpoint with the `DateAdded` parameter instead of `/Latest` for pagination. It does have some ordering changes but is negligible.
- Resume can also become a library, however I don't think it fits for resume items given its overall design
- Limit some `PosterHStack` to 20 items for performance, should happen everywhere else I didn't get to it. Can change back to arbitrary length if changed to `CollectionView`
- tvOS Movies and TV no longer just use the first found library, it just queries the corresponding item types
- Lots of other cleanup
- On iOS I might think about adding an option for changing whether "See All" is the button now or a poster like tvOS. Since "See All" is very prevalent I can find it annoying.

Some examples below.

<details>
<summary>iOS Home</summary>

</details>

<details>
<summary>iOS Item</summary>

</details>

<details>
<summary>iOS Cast - Poster</summary>

</details>

<details>
<summary>iOS Cast - List</summary>

</details>

<details>
<summary>tvOS Home</summary>

tvos-cast-library
</details>

<details>
<summary>tvOS Cast</summary>

tvos-cast-library
</details>


PangMo5 (approved):
👍

Issue #441, opened 2022-05-19
Save sort


dannymichel:
Right now I have to redo what I chose as a sorting option every time I go to the movies or tv libraries.

PR #585, opened 2022-09-14
Save Library Filters

LePips:
- Closes https://github.com/jellyfin/Swiftfin/issues/441

Save library filters if initially given no filters from the caller. If pushed from other contexts _with_ given filters, like `Latest`, don't save filter changes. Also removes "tags" filtering.


PangMo5 (approved):

dannymichel:
Thanks for this. Will this be updated in a test flight build any time soon?

LePips:
It's out when it's out

PR #587, opened 2022-09-15
Implement Factory

LePips:
Implements [Factory](https://github.com/hmlongco/Factory) for services instead of a us setting an explicit singleton. I will be working on refactoring playback next and this this will be used. This will also be used for the API refactor.

### DIFF
--- a/Shared/Coordinators/MainCoordinator/iOSMainCoordinator.swift
+++ b/Shared/Coordinators/MainCoordinator/iOSMainCoordinator.swift
@@ -8,6 +8,7 @@
 
 import Combine
 import Defaults
+import Factory
 import Foundation
 import Nuke
 import Stinsen
@@ -15,6 +16,10 @@ import SwiftUI
 import WidgetKit
 
 final class MainCoordinator: NavigationCoordinatable {
+
+    @Injected(LogManager.service)
+    private var logger
+
     var stack: NavigationStack<MainCoordinator>
 
     @Root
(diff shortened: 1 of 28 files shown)


PangMo5 (commented):
Oh, that's a pretty good library
So what effect can we expect from our project by using this library?

PangMo5 (approved):

LePips:
This changes how these dependencies are injected to where they are used. This is similar to `Environment/EnvironmentObject` for views, but allows dependency injection within non-view contexts. Notifications and logging just received a singleton-to-singleton-scope translation to just get started so there isn't any noticeable change there.

A good example for more purposeful usage would refactoring sessions, a small brainstorm:

```swift
extension Container.Scope {
    // indicates
    static var userSession = Cached()
}

extension Container {
    static let currentSession = Factory(scope: .userSession) { /* get current session from CoreStore */ }
}

// Usage in ViewModels/views
class MyViewModel {
    @Injected(Container.currentSession)
    var currentSession
}

// Example on signing out
func didSignOut() {
    Container.Scope.userSession.reset()
    // other signing out stuff
}
```

As stated above, a session lives within a scope that is determined by our login state. At any point I can reset its scope and set the new session and the app state will change, just like using `EnvironmentObject` in a view.

I've also given some more thought to testing. By injecting mock objects, like a mock API client and sessions, we can test view model and other manager behaviors.

PangMo5:
Thank you for your kind and good explanation 😄

Issue #610, opened 2022-10-20
Issue playing HDR content on Apple devices


sul4bh:
### Describe the bug

I am trying to play an HDR video with “Native Player” and “Force Direct play” experimental features turned on. I only get audio but no video. Same behavior on ios, ipados and tvos. Web client plays the video perfectly. See screenshot below:
![F06FC752-B948-4370-85FC-91B4EA9D4E65](https://user-images.githubusercontent.com/109486/197017626-66d3f206-3eaa-4087-883a-a541e9571293.jpeg)

Any thoughts on what is going on and if a fix is possible/planned?

### Application version

1.0.0 (78)

### Where did you install the app from?

TestFlight

### Device information

Iphone 14, ipad pro, apple tv 4k 1st gen

### OS version

Latest

### Jellyfin server version

10.8.5

LePips:
I will not be taking any issues regarding the native player: https://github.com/jellyfin/Swiftfin/discussions/324

This may be related: https://github.com/jellyfin/Swiftfin/issues/296

sul4bh:
@LePips 
Sorry for posting a message in a closed issue but I was able to fix this issue in a local build of Swiftfin and I have a proposal.

So I am wondering if you will take a PR that adds a new experimental setting "HDR profile" that simplifies profile generation in `DeviceProfileBuilder.buildProfile()` with a focus on supporting HDR videos? I know there is an ongoing effort to improve profile builder so this new setting can be a temporary fix. What are your thoughts?

sul4bh:
@LePips please review the PR to fix this issue at your convenience: https://github.com/jellyfin/Swiftfin/pull/623

n0thing2speak:
I have the same problem, I hope jellyfin can strengthen the ability of dirct streaming play, which can greatly reduce the cost of setting up the server.

holow29:
@sul4bh Have you had any luck using native player to play anthing HEVC or 4K on Apple TV? I'm trying to work on the device profiles, and it is so strange because iOS works fine, but tvOS does not for me.

sul4bh:
@holow29 I just checked and I have HDR running on iOS, macOS and tvOS. I have native player and fmp4 experimental features turned on. Server is on Version: 10.8.9.

holow29:
@sul4bh Thank you; that is helpful to know. If it isn't too much trouble, could you please:

- Switch your Apple TV to "4K SDR" mode under Settings->Video and Audio->Format
[...]

Issue #562, opened 2022-09-05
Fix CI for Ready to Review PRs


LePips:
I turned off running actions for draft PRs, https://github.com/jellyfin/Swiftfin/pull/527, however they don't run when PRs are open as Ready for Review.

LePips:
Forgot to tag, fixed in #778

PR #623, opened 2022-11-01
Adds "use fmp4" experimental option

sul4bh:
Fixes issue https://github.com/jellyfin/Swiftfin/issues/610

Adds a new experimental option that makes the native player use fmp4 segment container instead of ts. HEVC/HDR is only supported in fmp4 stream as documented here: "The container format for HEVC video MUST be fMP4.", https://developer.apple.com/documentation/http_live_streaming/http_live_streaming_hls_authoring_specification_for_apple_devices

FYI, fmp4 setting is already available in the web client:
Screenshot 2022-11-01 at 10 12 26 AM

**The new experimental option**
IMG_0046

**HDR now works as expected on my Apple TV**
IMG_0045


sul4bh:
Not sure why tests did not run for this PR.

LePips:
That's just due to https://github.com/jellyfin/Swiftfin/issues/562, which is fine. Carthage is broken right now and should anticipate a fix soon.

I'm glad that the change is small to get HDR with the native player. Since it's so small and technically a proven "feature-that-works" I'll most likely make it an option in my current work and it won't remain in experimental.

LePips (approved):

Issue #621, opened 2022-10-31
tvOS - Movies library empty - but not on iphone


psyciknz:
### Describe the bug

1. Go to TV Shows, populates
2. Go to Movies - library is empty.
3. Switch to jelly fin web site, movies library is populated.
4. Switch to ios jellyfin client, movies library is populated

### Application version

1.0.0(74) I think

### Where did you install the app from?

TestFlight

### Device information

Apple tv and apple tv 4k

### OS version

16.0

### Jellyfin server version

10.8.5

Bronson-Brown-deVost:
I also have this problem. I can find my movies individually if I search for them. Also my movie "collections" work fine. 

The problem started after I moved my server and possibly switched from one version of Jellyfin server to another. My current version of Jellyfin server is 10.8.5, I don't know what my previous version was, sorry.

ttonych:
The same problem began after upgrading Swiftfin to 1.0.0(83). The previous version worked fine. 
Jellyfin 10.8.5 and 10.8.7

MichaelAllenHardeman:
I am experiencing this problem as well. Fresh install of 1.0.0(83). I don't see my movie library 

[Here is the log](https://pastebin.com/9kQsmeM7) from the server side of the tvOS login and playing a video from my TV Show libary.

The only thing that looks suspicious is the Warning I see in the logs about a Slow Response form includeItemTypes=movie

[...]

PR #593, opened 2022-09-23
Refactor Video Playback

LePips:

PR #630, opened 2022-11-21
Log decoding errors

kcramer:
While debugging issue #621 I noticed the app does not log JSON decoding errors.  They are logged with an error code of 200 and description of "Unknown Error".

This patch logs the errors as a decoding error and includes the associated data to help debug such issues.  Perhaps it might also help to assign an error code (maybe -3) for such errors?


LePips:
Apologies that I haven't gotten to this. I am rewriting the entire `ViewModel` layer which this is a part of (#530, which is now #593), but I won't trouble you with that and just accept this as a note while I work on better error handling.

LePips (approved):

kcramer:
Thanks for the update.  If there is something I can do to help, let me know.

PR #633, opened 2022-11-23
Fix crashes when mediaStreams is nil

bzz445:


LePips:
I apologize, but this work will be removed with the work I am currently doing but this has made me put in proper error checking to prevent crashes.

PR #635, opened 2022-11-27
Wrong route type open search result item

bzz445:
If using .modal open ViewController it does not have a NavigationController so it is not possible to open child ViewControllers.
TestFlight version (78) uses .push and works fine

### DIFF
--- a/Shared/Coordinators/SearchCoordinator.swift
+++ b/Shared/Coordinators/SearchCoordinator.swift
@@ -17,7 +17,7 @@ final class SearchCoordinator: NavigationCoordinatable {
 
     @Root
     var start = makeStart
-    @Route(.modal)
+    @Route(.push)
     var item = makeItem
     @Route(.push)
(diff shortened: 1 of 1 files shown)


LePips:
Thank you, I've caught this in my current work as well. However, my mistake was removing the different route types per OS, which I have fixed in my current work:
```swift
    @Root
    var start = makeStart
    #if os(tvOS)
    @Route(.modal)
    var item = makeItem
    @Route(.modal)
    var library = makeLibrary
    #else
    @Route(.push)
    var item = makeItem
    @Route(.push)
    var library = makeLibrary
    @Route(.modal)
    var filter = makeFilter
    #endif
```

bzz445:
thx for explain. which branch is your current work?

LePips:
My work is on https://github.com/jellyfin/Swiftfin/pull/593, which has become a conglomerate of a lot of other stuff as well.

Issue #203, opened 2021-10-25
Home Screen "Latest" Context / Home Screen Improvements


Jinxyface:
I believe this was fixed before, but it must have came back with a recent update. For example, on my browser, my newest TV episodes:

![image](https://user-images.githubusercontent.com/19889486/138627911-9bc32393-b4ef-4e57-904a-5fd9464a54a3.png)

However, on the iOS app, it doesn't show the newest episodes of my TV series, but rather just the latest (newest) series that has been added:

![photo_2021-10-24_21-56-35](https://user-images.githubusercontent.com/19889486/138628117-e2d455e1-b9c7-40d5-9735-cebe95eaa274.jpg)

But when I do manual searches, the newest episodes have been added:

![photo_2021-10-24_21-56-39](https://user-images.githubusercontent.com/19889486/138628137-965200a0-2375-49ca-8f31-6d7613c985f6.jpg)

Not really a bad breaking issue or anything, just feel like it's a consistency issue that would make for a more polished experience.

let me know if there's any other info I can provide.

LePips:
I believe the main issue reported here was fixed by #895.

PR #659, opened 2022-12-31
Make the tvOS Home Screen look more like the one from the AppleTV app

rasmuslos:
I had nothing to do over the holidays so I decided to try to recreate the AppleTV app home screen ui using SwiftUI. I tried my best to make it feel like a native app, mostly by using a lot of subtle animations

https://user-images.githubusercontent.com/49640416/210133784-7f3a962e-cfcb-4c00-aaf5-0bb5c289b7e1.mov

If you want to know what the AppleTV app looks like screenshots can be found [here](https://www.apple.com/apple-tv-app/)

While doing this I noticed two things:
- Virtually no views use lazy stacks. I think changing this could improve scrolling massively, especially when the user has a large library. Using lazy stacks improved, at least for me, the performance on the Home Screen considerably
- SwiftUI has a `CardButtonStyle` wich adds this nice "wiggle" effect when you swipe just a little. I think it should be applied to more images

https://user-images.githubusercontent.com/49640416/210133920-70acedb3-cbba-4fe1-a38c-0def40728dd5.mp4

Maybe I will do a second PR with these changes


rasmuslos:
I have no idea why but I cannot add labels, it would be great if someone could add the "enhancement" label

LePips:
Thank you for these improvements! I will be able to clarify some things (why no lazy stack) and code review later. I do apologize in saying that I do have comments regarding how this would be done, and would most likely result in work being removed and done a different way.

As a heads up, my work is largely encompassing and your work will conflict, so we may have some issues.

LePips:
I do apologize to say that these changes cannot be accepted.

1. The home screen does need improvements, as first seen in #203. Currently we do some manual sections and then the libraries as set from user's server preferences. The long term goal is to have more customization in the form of our own local preferences which would allow many types of sections: genres, libraries, random, IMDb (which you have done), and so on. However there is a lot of work and planning that needs to be done before anything like that occurs.

2. I have outlined in the Contributing Guidelines: 
> User customizable UI/UX features are welcome and intended, however not all customization may be accepted for code maintainability and to also establish a distinct Swiftfin design. Taking inspiration, but not always copying, from other applications is encouraged.

While looking more like Apple TV is cool, it is not an intended explicit goal. It is also not necessarily possible due to technical limitations in SwiftUI. For example, you have the `See All` button _above_ the stack. This is a problem for stacks with few elements as it cannot gain focus until a button directly above or below the button itself is focused. There are a lot of quirks that need to be worked around the focus engine which I [have kind of done](https://github.com/jellyfin/Swiftfin/blob/main/Swiftfin%20tvOS/Objects/FocusGuide.swift), however this isn't a solution and there are many issues with it. This is why the `See All` button is a trailing button in the stack, which overcomes a technical limitation and is also the app's own design.

Other design changes made:
- The new hero element section deviates from the app's own implementation with the logo/name. Right now there is a bug that mistakenly uses the same image (mainly if a thumb has been set) and I am currently working on that.
- While Apple TV uses the `title-episode number` format for their poster labels, the app uses a title and subtitle for posters.

3. Using only landscape posters. Posters cannot be defaulted to landscape until thumb images are able to be retrieved by default by the server. This is why posters use the portrait primary image.

4. A lot of work is duplicated and doesn't use existing implementations, some include:
- a new `HStack` for posters was created in the new rows, but there's already [`PosterHStack`](https://github.com/jellyfin/Swiftfin/blob/main/Swiftfin%20tvOS/Components/PosterHStack.swift). This also already has the card effect applied.
- a new element for large hero elements was created, but there's already [`CinematicItemSelector`](https://github.com/jellyfin/Swiftfin/blob/main/Swiftfin%20tvOS/Components/CinematicItemSelector.swift)
- the [change to `HomeViewModel`](https://github.com/jellyfin/Swiftfin/pull/659/files#diff-9f7d048c28d0c2f9b0f483f03aa64b2a67bc71576b14597d2695977d06e78076R19) breaks what it's meant for. It's a small check for next up items which will then show the [`NextUpView`](https://github.com/jellyfin/Swiftfin/blob/main/Swiftfin%20tvOS/Views/HomeView/Components/NextUpView.swift) which will get the next up items. This also breaks on iOS. While previously the `HomeViewModel` did hold all of the items shown for each section this isn't good in the long term, especially for home screen customization as previously explained.

-----

**Why not `LazyHStack`?**

`LazyHStack` actually doesn't bring the performance improvement that one would think at first, as well as a small UX issue while the elements load in. While it does lazily _load_ the elements it doesn't _deque_ them. For example, the library views used to use `LazyVGrid` but performed terribly when a lot of elements were loaded, and this required the switch to a `UICollectionView` wrapper which I have created, `CollectionView`. Now, there is still a performance issue just using an `HStack`, however I have attempted to switch to using `CollectionView` but ran into an issue that I have a possible fix for. Overall, not using `HStack` is the long term goal.

-----

However, not all of the ideas tried here are unwelcome.

1- On iOS, [some types of sections allow customizing the poster type](https://github.com/jellyfin/Swiftfin/blob/6fe6fd6f8404b5a49fd1582c13740706473990a0/Shared/SwiftfinStore/SwiftfinStoreDefaults.swift#L35) This customization just hasn't made its way to tvOS yet. This is relevant to 3. from above.

2- The label color indicating the focus state and the label offset is something that I have attempted before, but I ran into performance issues. You are welcome to try it again in [`PosterButton`](https://github.com/jellyfin/Swiftfin/blob/main/Swiftfin%20tvOS/Components/PosterButton.swift).

I do appreciate the work and motivation for this work. However, a lot of just doesn't fit. If you would like to make other changes to the project, I recommend getting comfortable with the project for pre-existing implementations and coding style, as I would have also made a ton of code quality changes.

rasmuslos:
First of all, thank you for the comprehensive feedback. I mostly did this as a kind of DailUI-challenge and thought that some things, like not reusing PosterButton, could cause problems.
And if it is somethign that is usefull I will porbably take a look at the `PosterButton` when your PR is finished :)

Issue #258, opened 2022-01-06
tvOS mapping Apple TV users to Jellyfin users


jorritolthuis:
**Describe the feature you'd like**
The user switching in this app is already much better than many other similar applications. It would be even better if the Jellyfin users switches automatically when the Apple TV user changes.

**Additional context**
Apple provides a class to link the in-app user (they call this a _profile_) to the user of the Apple TV: https://developer.apple.com/documentation/tvservices/mapping_apple_tv_users_to_app_profiles#overview.

They also presented this at WWDC: https://developer.apple.com/videos/play/wwdc2019/211 at 31:40

acvigue:
This should already be implemented. I may be able to take a closer look later tonight.

LePips:
This used to be implemented but is a regression when I did my Session re-factoring. This will actually be a bit of work, adding another field to the StoredUser/User objects that will store the corresponding tvOS user profile, switching users on startup (and error handling if the JF user is signed out), and then being able to select which tvOS user maps to a JF user.

acvigue:
> Or, we can just remember the last JF user that was signed in with the current tvOS user, this seems a lot easier. Can just be done in Defaults with a mapped dictionary [tvOS user : JF user]

I think this was actually how I had it implemented earlier, however it makes sense that it was removed when the entire session system was redone.

LePips:
I have work for this and am almost done

LePips:
So I had an attempt on the linked PR with comments as to why this may not work. However, I have been thinking of a "Quick Switch User" little list that can pop up on the settings menu.

We could run the app with the `Get Current User` entitlement to achieve however I think that would cause more problems than what its worth since all JF users have to be re-logged in per tvOS user.

If somebody would like to take my work and try and fix it (like putting a retry-delay if the tvOS user id is nil or something, idk just something I thought about) that's fine.

LePips:
> your second paragraph

That way is easier just implementation-wise. Instead of dealing with CoreData everything is handled in defaults and by the `SessionManager`. My attempt includes all of updating on signing out and signing in. You're correct in that switching while watching media is a good case (and the most sensible one) but that also presents difficulties because we would have to send a playback report (actually not that difficult).

> your third paragraph

Sorry yes, I mixed them up. Running the app as `runs-as-current-user` instead would cause more problems than what its worth.

holow29:
[...]

PR #661, opened 2023-01-01
Native multi-user support on tvOS 16

awein:
The new entitlement introduced with iOS 16 allows full multi-user support: [`runs-as-current-user-with-user-independent-keychain`](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_developer_user-management)

### DIFF
--- a/Swiftfin tvOS/Swiftfin tvOS.entitlements
+++ b/Swiftfin tvOS/Swiftfin tvOS.entitlements
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>com.apple.developer.user-management</key>
+	<array>
+		<string>runs-as-current-user-with-user-independent-keychain</string>
+	</array>
+</dict>
+</plist>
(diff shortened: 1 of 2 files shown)


LePips:
I will play around with this for a while and see if this is the way that we should just finally handle it, see #258.

awein:
Valid point and I tend to agree that `runs-as-current-user` is enough and will allow support for this to be available on tvOS 14 and above (in contrast to tvOS 16 and above with the suggested change). 
That being said: Both variants will behave the same on tvOS 16 given the user specific keychain is used by default. The shared user-independent keychain is only used when specifically specified. E.g. it could be used to store the server URL to streamline the login flow for additional users but wouldn't be used to store the actual user credentials of course.

holow29:
Great points! I was forgetting that each item that used the shared keychain needed to have a declaration as such. However, currently Swiftfin supports tvOS 15. The tradeoff seems to be: do we care about support for this specific feature in tvOS 15 (given users can update to tvOS 16) or is it fine to keep in tvOS 16+ exclusive which would have the advantage of perhaps being the more supported way of doing it now? (It makes me think `runs-as-current-user` will be deprecated first.) I'd lean towards using `runs-as-current-user-with-user-independent-keychain` since it is basically a superset of the funcionality.

ShadowJonathan:
Maybe `runs-as-current-user` for now, and `runs-as-current-user-with-user-independent-keychain` when there's the slightest hint that the former will be deprecated (and minimal support has shifted up to tvOS 16)?

Issue #676, opened 2023-01-07
TRUEHD7.1 track has no sound


chengli0946:
### Describe the bug

TRUEHD7.1 track has no sound

### Application version

1.0.0

### Where did you install the app from?

None

### Device information

iphone13 IOS15.7.1

### OS version

iphone13 IOS15.7.1

### Jellyfin server version

10.8.6

kubik369:
I am experiencing the same issue on:
- Apple TV 4K 2nd gen running tvOS 16.2 with just the TV speakers
- my iPhone 13 mini running iOS 16.2 on both the speakers and Airpods

Whilst running on Jellyfin server version 10.8.8. When I switch the audio track in the file, sound works fine with the other track. I can provide logs if it would be helpful (although I haven't seen anything useful in them).

songweibiao:
I am experiencing the same issue on:
Application version：1.0.0
Device information：ipad 6th genrunning ipados 15.7.1
Jellyfin server version：10.8.8

mrmodius:
Same issue, Apple TV 4K 2022

LoganVessell:
I am experiencing the same issue on the following devices:
- Apple TV 4K (3rd generation) running tvOS 16.3 w/ TV Speakers & HomePod mini. Swiftfin v 1.0.0 (84)
- iPad mini (6th generation) running iPadOS 16.3 w/ AirPods & built-in speakers. Swiftfin 1.0.0 (79)
- iPhone 14 Pro Max running iOS 16.3 w/ AirPods & built-in speakers. Swiftfin 1.0.0 (79)
Running Jellyfin Server 10.8.9 on an M1 Mac mini w/ macOS Monterey 12.6.2.

holow29:
Unfortunately, Dolby TrueHD playback requires licensing (which is one reason it is only available in the Pro version of Infuse, for example). I think the best we can do for now is transcode it server-side to something like FLAC. I'm looking into that approach as I deal with the device profile (https://github.com/jellyfin/Swiftfin/issues/296) to see if it even makes sense given ffmpeg's truehd implementation (vs. transcoding to something like AC3).

kubik369:
Oh you are right. I didn't realise that none of the devices I tried support it natively - error on my part, especially regarding the  Apple devices I mentioned :/ I will check whether any of my devices support it and will report back if it works.

targos:
> Dolby TrueHD playback requires licensing 

[...]

Issue #643, opened 2022-12-11
No sound for TrueHD Atmos 7.1 content


JackJ74:
### Describe the bug

1. Play a video with TrueHD Atmos 7.1 audio format.
2. No sound playing even comfirming TureHD Atmos 7.1 audio tack is selected

Expect:
Sound playing 

### Application version

1.0.0(79)

### Where did you install the app from?

TestFlight

### Device information

iPhone 11

### OS version

iOS 16.2

### Jellyfin server version

10.8.6

lavavex:
I noticed this as well on my new AppleTV 4k, but it seems like Apple doesn't support TrueHD audio. Hopefully they can make it so it auto selects Dolby Digital instead of TrueHD

LePips:
If the system or VLCKit don't support it, there's nothing we can do. I'm also not inclined for an automatic selection until we figure out the device builder for supported formats.

PR #700, opened 2023-02-15
prevent direct play of Dolby TrueHD and DTS

schrock:
AFAIK, no Apple devices support Dolby TrueHD or any DTS audio format. Some do support Dolby Atmos, but only over Dolby Digital Plus (EAC3). This PR removes the code block for A12 and up devices which wrongly assumes that they support Dolby TrueHD and DTS. This should force the Jellyfin server to transcode these kinds of audio tracks to compatible formats. Currently, they are sent to the Swiftfin client as-is, which results in no audio.

This should fix the problems described in #643 and #676. I understand that there is currently a refactor for the Device Profile Builder underway in #519, but that has been going on for 6+ months. This is meant as a hot-fix until that effort is complete.

This code HAS NOT been tested.

### DIFF
--- a/Shared/Objects/DeviceProfileBuilder.swift
+++ b/Shared/Objects/DeviceProfileBuilder.swift
@@ -83,16 +83,6 @@ class DeviceProfileBuilder {
             )] // H.264/HEVC with Dolby Digital - No Atmos - Vision
         }
 
-        // Device supports Dolby Atmos?
-        if supportsFeature(minimumSupported: .A12) {
-            directPlayProfiles = [DirectPlayProfile(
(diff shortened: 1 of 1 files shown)


holow29:
I actually hope to push something to my PR tomorrow that can be used. Obviously up to the maintainers what to do from there, but I thought I'd let you know!
As an FYI, VLCKit player (not Native player) supports DTS lossy 5.1 core, so you should hopefully be able to get that to work as is. My PR will properly separate profiles between Native and VLCKit since the support is vastly different.

LePips:
Yes, I will close this in favor of the overall device builder overhaul.

PR #718, opened 2023-03-07
temp fixes

LePips:
Is a patch, for now, to account for the new `Data` media stream type and deal with the performance issues of focusing for tvOS. While this is a direct solution for the most outstanding issues, I will not close any here and will close them with #593 since it is a gigantic refactor.


PangMo5 (approved):
👍

PR #731, opened 2023-03-17
Refresh resume items when going back to Home

shadzik:
The CinematicResumeView is updated onAppear to always show all resume watching items

### DIFF
--- a/Swiftfin tvOS/Views/HomeView/Components/CinematicResumeItemView.swift
+++ b/Swiftfin tvOS/Views/HomeView/Components/CinematicResumeItemView.swift
@@ -17,6 +17,8 @@ extension HomeView {
         private var router: HomeCoordinator.Router
         @ObservedObject
         var viewModel: HomeViewModel
+        @State
+        var resumeItems: [BaseItemDto]
 
         private func itemSelectorImageSource(for item: BaseItemDto) -> ImageSource {
             if item.type == .episode {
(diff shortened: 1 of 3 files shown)


LePips:
Thank you for your interest and contributions! I see that you have a few other PRs (#728, #730) but I will leave my general comment here.

Sadly, all this work is rewritten in #593, which also selects the first item and changes how items are done on the tvOS home screen.

I am sorry that I will be closing these but I do hope to be done with #593 soon and we can pick up from there.

Swiftfin tvOS/Views/HomeView/Components/CinematicResumeItemView.swift
```diff
@@ -17,6 +17,8 @@ extension HomeView {
         private var router: HomeCoordinator.Router
         @ObservedObject
         var viewModel: HomeViewModel
+        @State
+        var resumeItems: [BaseItemDto]
```
LePips, line 21: We don't want the view to hold items, that is what the view model is for.

Swiftfin tvOS/Views/HomeView/Components/CinematicResumeItemView.swift
```diff
@@ -67,6 +69,13 @@ extension HomeView {
                 .onSelect { item in
                     router.route(to: \.item, item)
                 }
+                .onAppear {
+                    viewModel.refreshResumeItems { newItems in
```
LePips, line 73: Refreshing home screen(s) will be done in favor of a global refresh instead.

Issue #741, opened 2023-04-09
AirPlay should continue to work while app is backgrounded


slocook:
**Describe the feature you'd like**
When using the Native Player with AirPlay on iOS, the video should continue to play on the AirPlay receiver while the iPhone app is backgrounded. Currently, if I AirPlay to my TV and then switch to another app, the AirPlay session stops.

**Additional context**
There is a really simple fix for this by changing the Background Mode in capabilities in Xcode settings.

LePips:
Apologies for not merging your PR first, this was also implemented in #786.

PR #742, opened 2023-04-09
AirPlay should continue to work while app is backgrounded

slocook:
This change only affects the Native Player. When the app is backgrounded while a video is playing, one of two things happens:
1. If AirPlay is active, the video continues to play on the AirPlay receiver
2. If AirPlay is not active, the video continues to play in a picture-in-picture window

Fixes https://github.com/jellyfin/Swiftfin/issues/741


LePips:
I am not against background play, however the only requirement is that the playback reports are still sent. After #593, please test to make sure that they are.

LePips:
I apologize for not merging this one first, the background mode was implemented in #786.

PR #752, opened 2023-04-21
Fix building and LiveTV item layout

jhays:
Fixes a couple issues in `LiveTVChannelItemElement.swift` preventing build. Also fixes a layout issue for this view.

### DIFF
--- a/Swiftfin tvOS/Components/LiveTVChannelItemElement.swift
+++ b/Swiftfin tvOS/Components/LiveTVChannelItemElement.swift
@@ -92,12 +92,11 @@ struct LiveTVChannelItemElement: View {
                     programLabel(
                         timeText: currentProgramText.timeDisplay,
                         titleText: currentProgramText.title,
-                        color: Color("TextHighlightColor"),
+                        color: Color.primary,
                         font: Font.system(size: 20, weight: .bold, design: .default)
                     )
-                    if !nextProgramsText.isEmpty,
-                       let nextItem = nextProgramsText[0]
-                    {
+                    if !nextProgramsText.isEmpty {
+                        let nextItem = nextProgramsText[0]
                         programLabel(
                             timeText: nextItem.timeDisplay,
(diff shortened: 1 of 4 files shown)


jhays:
I'd like to look into the iOS build as well, but I'm currently blocked from running the iOS build as I cannot seem to get the `UDPBroadcast` framework to load via Carthage correctly.  @LePips, do you have any suggestions regarding that? What command do you usually use to build Carthage dependencies for this project?

Swiftfin tvOS/Components/LiveTVChannelItemElement.swift
```diff
@@ -177,3 +176,19 @@ struct LiveTVChannelItemElement: View {
         }
     }
 }
+
+struct LiveTVChannelItemElement_Preview: PreviewProvider {
```
LePips, line 180: I am trying to not do previews since they can just become a pain to maintain in the long run and are also funky to use when some views have them and some don't. Even though I have left some during my development in `tvOS/AboutViewCard` and `tvOS/MainOverlay`, they can be deleted as well.

LePips (changes requested):

LePips:
I just run the same Carthage command but I ran it lately with the `--platform iOS,tvOS` flags as well since I didn't download the WatchOS runtime. We should probably add that to the docs too and I see that you figured it out.

Don't worry about the build system as I found out why it's failing [here](https://github.com/jellyfin/Swiftfin/pull/593#issuecomment-1515731712). I checked out your changes and was able to build just fine. I thought that I had 5.8 installed but I didn't so I wasn't able to catch these errors since they improved result builders.

Swiftfin tvOS/Components/LiveTVChannelItemElement.swift
jhays: Ok! No problem. Preview removed in latest commit.

LePips (approved):

Issue #759, opened 2023-04-26
Playback does not start when requesting a transcode


holow29:
### Describe the bug
fullTranscodeURL did not contain the base server path as far as I could tell. This was making it so no transcodes would play successfully on VLCKit (Swiftfin) player and also making it so anything that requested playbackURL where streamType = .transcode (vs. requesting HLS playlist directly as Native player is set to do atm) would fail.

### Application version

a9db83d

### Device information

tvOS Simulator

### OS version

16.1

### Jellyfin server version

10.8.9 and 10.8.10

PR #760, opened 2023-04-27
Fix transcodingURL

holow29:
Add server base URL to fullTranscodeURL

Fixes https://github.com/jellyfin/Swiftfin/issues/759

I am not that familiar with JellyfinAPI, so I am not sure if this is the correct method, but this seems to work. Before this fix, the URL was of the form "/videos/..." whereas normally the playbackURL starts with "http://[server info]/Videos/..."


LePips:
I didn't test this but yep was missing the relative path and was just blank.

LePips (approved):

PR #765, opened 2023-05-04
Cleanup and UDPBroadcastConnection Package

LePips:
Some cleanup and now UDPBroadcastConnection is a package. `EnumPicker` was refactored to take an optional binding as well for an enum before I noticed that `Defaults` doesn't properly support optional values, but I'll leave it in here.


LePips:
Currently failing now because `macos-12` uses Xcode 14.2. `macos-13` is in beta and I don't care to use it right now. I'll either request that UDPBroadcastConnection moves to a lower tools version or fork it.

PangMo5 (approved):

Issue #684, opened 2023-01-23
Subtitles don't work on Apple TV's experimental player


lourenci:
### Describe the bug

1. Active both experimental features:
  - Use native player
  - Use fmp4 with HLS
2. Try to run any video with subtitles, it could be either embedded or external (SRT). 

### Application version

1.0.0 (84)

### Where did you install the app from?

App Store

### Device information

Apple TV 4K (3rd generation)

### OS version

tvOS 16.2

### Jellyfin server version

10.8.8

LePips:
Please see https://github.com/jellyfin/Swiftfin/discussions/324, the native player is very much just in development. At the same time, the server doesn't have the best HLS support that Apple expects so it will never really be as expected compared to watching other media on an iOS/tvOS device.

PR #766, opened 2023-05-06
Support for `.strm` files.

AhsanFazal:
# Rationale

My whole library consists of `.strm` files. This makes Swiftfin pretty much useless for me in the current state, which is a shame, because I really love the idea of Swiftfin.

# The Plan
The `.strm` file extension and file type is natively supported by Jellyfin. They are files that contain a single line with a streaming URL. Ideally, this streaming URL should be played back directly on the client device itself, without any involvement from the transcoding or remuxing engine of the Jellyfin server. This is why my focus will be on the VLC player, not the native player, as the VLC player can support many more file types out of the box.

The easiest solution seems to:
1. Check if the requested media is indeed in `.strm` format.
2. Plug the URL from that `.strm` file directly into the native player or VLCKit player (the `path` property of `MediaSourceInfo` contains this URL out of the box).

## VLC `VideoPlayer`
The VLC video player functionality just consists of checking if a file is a `.strm` file (by checking if the `path` property of `MediaSourceInfo` starts with `http(s)://` and playing it back directly.

## `NativeVideoPlayer`
Here, the only change needed is to set the `isStatic` property in `hlsPlaybackUrl` of the `VideoPlayerManager` to false when it is requesting an HLS stream of an `.strm` file. However, subtitles are a little more complicated, as subtitles cannot directly be injected into the `AVPlayer` instance. I will need to do a deeper dive into the Apple documentation to figure out how to combine the `master.m3u8` stream of the video with the `subtitles.m3u8` stream of the subtitles.

### DIFF
--- a/Shared/Coordinators/VideoPlayerCoordinator.swift
+++ b/Shared/Coordinators/VideoPlayerCoordinator.swift
@@ -33,9 +33,6 @@ final class VideoPlayerCoordinator: NavigationCoordinatable {
             Group {
                 if Defaults[.VideoPlayer.videoPlayerType] == .swiftfin {
                     VideoPlayer(manager: self.videoPlayerManager)
-                        .overlay {
-                            VideoPlayer.Overlay()
-                        }
                 } else {
                     NativeVideoPlayer(manager: self.videoPlayerManager)
                 }
--- a/Shared/Extensions/JellyfinAPI/MediaSourceInfo+ItemVideoPlayerViewModel.swift
+++ b/Shared/Extensions/JellyfinAPI/MediaSourceInfo+ItemVideoPlayerViewModel.swift
@@ -17,10 +17,12 @@ extension MediaSourceInfo {
     func videoPlayerViewModel(with item: BaseItemDto, playSessionID: String) throws -> VideoPlayerViewModel {
 
         let userSession = Container.userSession.callAsFunction()
-        let playbackURL: URL
+        let playbackURL: URL?
         let streamType: StreamType
 
-        if let transcodingURL, !Defaults[.Experimental.forceDirectPlay] {
+            playbackURL = URL(string: path)
+            streamType = .strm
+        } else if let transcodingURL, !Defaults[.Experimental.forceDirectPlay] {
             guard let fullTranscodeURL = URL(string: "".appending(transcodingURL))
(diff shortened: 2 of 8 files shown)


LePips:
I need/want `strm` support to be minimal. Almost to the point where all we do is grab the URL from the file and instead use that as the playback URL. If it doesn't work at that point, then we can look at what is required to accommodate it.

I don't have any intentions to maintain `strm` support if there is any refactoring in the future.

AhsanFazal:
@LePips I totally understand, let's keep it minimal. I've already committed that very same minimal functionality to this branch, so I will turn this draft PR into a real one. During review, two things need addressing:

1. The commit [[iOS] Initialize VideoPlayer.Overlay immediately](https://github.com/jellyfin/Swiftfin/pull/766/commits/638f653eed646577e8c12e11a91dfb6073c7051f) solves a bug I encountered. Do you want me to open a separate PR for this bugfix?
2. I will squash all other commits into a single commit. The most noteworthy thing here is that I added a third `StreamType`: `.strm`, which I set in the `MediaSourceInfo` API call and then use in conditional logic at building the HLS stream URL for `NativeVideoPlayer`. Are you okay with adding this third `StreamType`?

Then there is the issue of subtitles in the native player. I see you already addressed it in #684. The Jellyfin HLS API endpoint indeed does not return an Apple compliant `master.m3u8` file (as described in their documentation for AVKit). I have found a solution for this and will be opening a new PR to address this 🙂

LePips:
Don't worry about squashing, all PRs are squashed on merge. Per the overlay crashing issue, I'm looking into that as I think there was a change in Swift that is now causing the crash that didn't happen previously.

AhsanFazal:
I created `.strm` files for almost all modern video formats to test the playback capabilities on _both_ players. When testing and debugging, I realized that we need a solid player-based `DeviceProfileBuilder`. So I started work on that, before I saw #519. I really like their approach, as mine was almost identical. So I will convert this PR into a draft again, help out where needed with #519 and continue work on `.strm` file support after that PR is ready and merged 🙂

LePips:
Please see comment: https://github.com/jellyfin/Swiftfin/pull/767#issuecomment-2332452458

Much has changed architecturally since this PR (and will change soon atow) and `strm` support must be implemented in a different manner.

PR #774, opened 2023-05-15
Fix Video Player Overlay

### LINKED ISSUE #768: a08a92e introduced a crash on tvOS 16.4.1
### Describe the bug

1. Go to detail view of any video
2. Tap on "Play"
3. Application crashes

SwiftUI/EnvironmentObject.swift:82: Fatal error: No ObservableObject of type VideoPlayerViewModel found. A View.environmentObject(_:) for VideoPlayerViewModel may be missing as an ancestor of this view.
2023-05-07 18:31:42.624253+0100 Swiftfin tvOS[531:74054] SwiftUI/EnvironmentObject.swift:82: Fatal error: No ObservableObject of type VideoPlayerViewModel found. A View.environmentObject(_:) for VideoPlayerViewModel may be missing as an ancestor of this view.

No crashes if I checkout 87738af58744b1280da365aa94a29602172c7015 which is just before the a08a92e98ae3306f9464a5f34202534436f0def2

### Application version

a08a92e98ae3306f9464a5f34202534436f0def2

None

### Device information

Apple TV 4K

### OS version

tvOS 16.4.1

### Jellyfin server version

10.8.10

LePips:
- Closes #768

Fixes the crash when playing a video using the Swiftfin video player. I chose a bad name (`overlay`) and it instead uses the `SwiftUI.overlay` function. This worked in 5.7 and broke in 5.8, but it's good this happened anyways.

The original intent with the passed overlay was that we could use the video player in many contexts that aren't TV episodes or movies, like live tv (which is how it works today), and just use a different overlay for each. For example, it would be very cool if live tv had a different overlay for a live tv context (LIVE, guide, etc.)

However, the `VideoPlayer` implementation heavily depends on `VideoPlayer.Overlay` when being used for gestures and actions, which may not make sense in all contexts ... maybe. For now, everything will use the same overlay and when someone wants to make a custom overlay for live tv, the entire video player layer will be a separate view/implementation.

### DIFF
--- a/Shared/Coordinators/VideoPlayerCoordinator.swift
+++ b/Shared/Coordinators/VideoPlayerCoordinator.swift
@@ -33,9 +33,6 @@ final class VideoPlayerCoordinator: NavigationCoordinatable {
             Group {
                 if Defaults[.VideoPlayer.videoPlayerType] == .swiftfin {
                     VideoPlayer(manager: self.videoPlayerManager)
-                        .overlay {
-                            VideoPlayer.Overlay()
-                        }
                 } else {
                     NativeVideoPlayer(manager: self.videoPlayerManager)
                 }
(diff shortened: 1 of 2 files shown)


crobibero (approved):
I promise I know swift

LePips:
You foolishly approved a keylogger that sends credentials to my personal server, I now know everyone's login.

PR #794, opened 2023-06-17
Developer: Fix certain Swiftgen installations

SwiftCoderJoe:
When running scripts from an Apple Silicon Mac, the `$PATH` is different from your user's `$PATH` (see [this discussion](https://github.com/orgs/Homebrew/discussions/3035)). Since `contributing.md` suggests you install Swiftgen using Homebrew, we should definitely be supporting this use case as well as we possibly can.

Furthermore, a cryptic `Command PhaseScriptExecution failed with a nonzero exit code` appears when Swiftgen is not found, so I added a new error instructing developers to go to `contributing.md` for installation instructions.

### DIFF
--- a/Swiftfin.xcodeproj/project.pbxproj
+++ b/Swiftfin.xcodeproj/project.pbxproj
@@ -3009,7 +3009,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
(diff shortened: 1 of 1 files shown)


ericswpark:
Can confirm that this PR fixes the build failing on ASi due to Swiftgen.

LePips:
Thank you! I still have my old Homebrew installation before they migrated installation destinations so it's always worked for me.

LePips (approved):

PR #831, opened 2023-08-29
Handle app phases (auto pause/play on enter background/active phase)

sy6sy2:
Hi!

I noticed that when watching a movie on iOS, if I go back to Home Screen without pausing the movie then the movie juste continue playing in background.

This PR adds notification support for active, inactive and background phases of the app and also pause the video when app enter in background and resume the movie when enter in active phase.

This is still a first step, I guess other things can be done about that like stopping instead of pausing for live TV.

@LePips and @jhays if you can have a look on this, I'm a newbie on all the swift and Objective-C stuff...
Thank you!

### DIFF
--- a/Shared/Extensions/ViewExtensions/ViewExtensions.swift
+++ b/Shared/Extensions/ViewExtensions/ViewExtensions.swift
@@ -218,4 +218,8 @@ extension View {
             .ignoresSafeArea()
         }
     }
+
+    func onScenePhase(_ phase: ScenePhase, _ action: @escaping () -> Void) -> some View {
+        modifier(ScenePhaseChangeModifier(phase: phase, action: action))
+    }
 }
--- a/Swiftfin tvOS/Views/SettingsView/VideoPlayerSettingsView.swift
+++ b/Swiftfin tvOS/Views/SettingsView/VideoPlayerSettingsView.swift
@@ -21,6 +21,11 @@ struct VideoPlayerSettingsView: View {
     @Default(.VideoPlayer.resumeOffset)
     private var resumeOffset
 
+    @Default(.VideoPlayer.Transition.pauseOnBackground)
+    private var pauseOnBackground
+    @Default(.VideoPlayer.Transition.playOnActive)
+    private var playOnActive
+
     @EnvironmentObject
     private var router: VideoPlayerSettingsCoordinator.Router
 
@@ -56,6 +61,12 @@ struct VideoPlayerSettingsView: View {
                         }
                 } footer: {
                     Text("Settings only affect some subtitle types")
+                }
+
+                Section {
+
+                    Toggle("Pause on background", isOn: $pauseOnBackground)
+                    Toggle("Play on active", isOn: $playOnActive)
                 }
             }
             .navigationTitle("Video Player")
--- a/Swiftfin tvOS/Views/VideoPlayer/VideoPlayer.swift
+++ b/Swiftfin tvOS/Views/VideoPlayer/VideoPlayer.swift
@@ -19,6 +19,9 @@ struct VideoPlayer: View {
         case smallMenu
     }
 
+    @Environment(\.scenePhase)
+    private var scenePhase
+
     @EnvironmentObject
     private var router: VideoPlayerCoordinator.Router
 
@@ -99,6 +102,16 @@ struct VideoPlayer: View {
         .onChange(of: isScrubbing) { newValue in
             guard !newValue else { return }
             videoPlayerManager.proxy.setTime(.seconds(currentProgressHandler.scrubbedSeconds))
+        }
+        .onScenePhase(.active) {
+            if Defaults[.VideoPlayer.Transition.playOnActive] {
+                videoPlayerManager.proxy.play()
+            }
+        }
+        .onScenePhase(.background) {
+            if Defaults[.VideoPlayer.Transition.pauseOnBackground] {
+                videoPlayerManager.proxy.pause()
+            }
         }
     }
 }
--- a/Shared/Extensions/ViewExtensions/Modifiers/ScenePhaseChangeModifier.swift
+++ b/Shared/Extensions/ViewExtensions/Modifiers/ScenePhaseChangeModifier.swift
@@ -1,0 +1,26 @@
+//
+// Swiftfin is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2023 Jellyfin & Jellyfin Contributors
+//
+
+import SwiftUI
+
+struct ScenePhaseChangeModifier: ViewModifier {
+
+    @Environment(\.scenePhase)
+    private var scenePhase
+
+    let phase: ScenePhase
+    let action: () -> Void
+
+    func body(content: Content) -> some View {
+        content.onChange(of: scenePhase) { newValue in
+            if newValue == phase {
+                action()
+            }
+        }
+    }
+}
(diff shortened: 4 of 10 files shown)


Shared/ViewModels/VideoPlayerManager.swift
```diff
@@ -21,6 +21,24 @@ import VLCUI
 
 class VideoPlayerManager: ViewModel {
 
+    override init() {
```
LePips, line 24: I think I would rather have this be on the actual `NativeVideoPlayer`/`VideoPlayer` views like you do with the app instead of more notifications, the `VideoPlayerManager` objects are already there.

LePips (changes requested):

Shared/ViewModels/VideoPlayerManager.swift
sy6sy2: I tried but I have an error that I failed to fix 😕 
Capture d’écran 2023-08-30 à 20 52 05

LePips: Not like that, as that is also invalid Swift. Add the `Environment` and `onChange`, like you do in `SwiftfinApp`, to the video players.

sy6sy2: I tried moving the `.onChange(of: scenePhase)` but unfortunately the onChange block is not triggered anymore. Don't know if it's bug or if this onChange only works at the App level... 😕

sy6sy2: After some internet research it seems that the `.onChange(of: scenePhase)` is buggy and does not always trigger if using with a `View`.
Is it a problem to leave this stuff at the `App` level?

PangMo5 (commented):
It would be nice to be able to control this as an option in the settings.
In particular, the part that unpause when you return to the foreground can be an uncomfortable experience for some users.

sy6sy2:
Yes totally agree with you @PangMo5.
At this state, this PR it's just a POC to show that Jellyfin can handle app phases, but if @LePips is ok with this initial work I will add more commits to this PR before any merge in order to improve the user experience like you suggest.

Some example I think about:

* For each phase, possibility to set the desired action (e.g. nothing, pause, stop, play, ...).
* Maybe a different setting for Movies vs Live TV (e.g. on background, pause a movie but stop a live TV).
* ...

Shared/ViewModels/VideoPlayerManager.swift
LePips: This wasn't working for me on the iOS 16 simulator but now since the new iOS/Xcode RCs are out it works in the iOS 17 simulator and on device. I can only see a few things online about the phase not working and those are related to macOS or the simulator which isn't reliable.

I recommend implementing a variation of [this solution:](https://stackoverflow.com/a/62850923)

```swift

struct ScenePhaseChangeModifier: ViewModifier {
    
    @Environment(\.scenePhase)
    private var scenePhase
    
    let phase: ScenePhase
    let action: () -> Void
    
        content.onChange(of: scenePhase) { newValue in
            if newValue == phase {
                action()
            }
        }
    }
}

// ViewExtensions.swift

extension View {

    func onScenePhase(_ phase: ScenePhase, _ action: @escaping () -> Void) -> some View {
        modifier(ScenePhaseChangeModifier(phase: phase, action: action))
    }
}

// VideoPlayer.swift

struct VideoPlayer: View {

// ...

    var body: some View {
		// ...
                playerView
                    .onScenePhase(.active) {
                        // play player and send play report
                    }
                    .onScenePhase(.background) {
                        // pause player and send pause report
                    }
		// ...
}

```

The reason why I just want to pause on `.background` _for now_ instead of `.inactive` is because `inactive` triggers while using the Control Center while also in the app which I don't find favorable until we implement a more comprehensible solution.

LePips:
I agree that this should be customizable, so for now I want 2 new keys under a new section:

```swift
// SwiftfinDefaults.swift

extension Defaults.Keys {
	enum VideoPlayer {

	// ...

		enum Transition {
			static let pauseOnBackground // ...
			static let playOnActive // ...
		}
	}
}
```

This will go in a new `Transition` section in the `Video Player` settings, at the bottom as I sort the sections alphabetically in English.

I have already thought about a more comprehensible solution for actions that will completely replace this, but frankly I don't know when I will get to it and will probably need to think about it more. It also may be sooner than expected when I get to it as this is fresh on my mind so 🤷. So, you can either implement these settings for now or wait until I get to it, your choice. 

<details>
<summary>[what I thought of]</summary>

Similar to how the video player gestures and action buttons work, there will be a certain set of actions created (play, pause, PiP) that can be selected per `active`, `inactive`, and `backgrounded`. This just lays the foundation for more actions to be easily added and I think is better when it comes to how different media want to react to these transitions. Sounds simple but is a bit more to implement.

</details.

---

Additionally, we have to think about this behavior with the `Native` player - the main issue being PiP. Since this feature may eventually come to the other player (once I get back to compiling mpv...) it wouldn't hurt to think about this behavior now. Currently, PiP isn't customizable (but should be) and, while I haven't tested it myself, the background pause may conflict with wanting to watch in PiP. Frankly because of this, just leave the `Native` player alone for now and we can get to it later.

Shared/ViewModels/VideoPlayerManager.swift
sy6sy2: You are right! It seems to works now on real devices (don't if it's seems the iOS 17 update or not ... ) anyway. I will continue working on it now that it works!

Shared/Extensions/ViewExtensions/ViewExtensions.swift
```diff
@@ -12,6 +12,30 @@ import SwiftUI
 
 // TODO: organize
 
+struct ScenePhaseChangeModifier: ViewModifier {
```
LePips, line 15: Make this its own file under `Shared/Extensions/ViewExtensions/Modifiers`.

Shared/Extensions/ViewExtensions/ViewExtensions.swift
```diff
@@ -12,6 +12,30 @@ import SwiftUI
...
+            if newValue == phase {
+                action()
+            }
+        }
+    }
+}
+
+extension View {
+
+    func onScenePhase(_ phase: ScenePhase, _ action: @escaping () -> Void) -> some View {
```
LePips, line 34: Add this to the bottom of the already existing block of `extension View`.

Swiftfin tvOS/Views/SettingsView/VideoPlayerSettingsView.swift
```diff
@@ -57,6 +62,12 @@ struct VideoPlayerSettingsView: View {
                 } footer: {
                     Text("Settings only affect some subtitle types")
                 }
+
+                Section {
+
+                    Toggle("Pause on background", isOn: $pauseOnBackground)
```
LePips, line 68: I'm fine with the way the settings are for now but I may clean them up in the future to something different.

Swiftfin tvOS/Views/VideoPlayer/VideoPlayer.swift
```diff
@@ -100,6 +103,12 @@ struct VideoPlayer: View {
             guard !newValue else { return }
             videoPlayerManager.proxy.setTime(.seconds(currentProgressHandler.scrubbedSeconds))
         }
+        .onScenePhase(.active) {
+            Defaults[.VideoPlayer.Transition.playOnActive] ? videoPlayerManager.proxy.play() : ()
```
LePips, line 107: Please just do `if` instead of a ternary.

LePips (changes requested):

LePips (approved):

Issue #680, opened 2023-01-11
Library Filter by / Jump to Letter


JPKribs:
**Describe the feature you'd like**

While scrolling through a library, it would be a nice addition to include the ability to jump directly to a letter in the scroll or filter by only that letter. This currently exists in Jellyfin-Web and Jellyfin-Roku as a filter/view. Please see the screenshot for context:

![image](https://user-images.githubusercontent.com/37599297/211707996-3e10fd25-d138-4af2-b5e4-7546b29f0186.png)

**Additional context**

I haven't seen any requests for this so I thought I would just make an issue for it. This could fall under https://github.com/jellyfin/Swiftfin/issues/333 so please let me know if this is a duplicate!

Jpkribs:
I actually took this as an opportunity to start helping out with the project! I'm not 100% done just yet but I just wanted to include a screenshot of this and say (to anyone else interested in this feature) I'll be done soon! 

I ended up going the filter-on-letter method as it seemed less resource intensive than pulling in library items to determine a scroll-to position. Screenshots below:

Unfiltered:
![image](https://github.com/jellyfin/Swiftfin/assets/37599297/b9a0a613-d440-49c2-a8c0-a15528fe4304)

Filtered:
![image](https://github.com/jellyfin/Swiftfin/assets/37599297/1c609e6b-d3a6-414b-8d35-eebf2b0150be)

Filtered with no results:
![image](https://github.com/jellyfin/Swiftfin/assets/37599297/062eab8e-d9fa-410e-847f-8e92d560e0d2)

LePips:
I'm glad that you are working on this and welcome your first contribution! However, I am wary about this on all platforms. I have setup the entire filtering/ordering/etc. interface to be through the pill buttons along the top for iOS. For the entire English alphabet to be displayed, it looks like another vertical scroll view will be used and that probably isn't the best interaction along with the collection view, regardless of screen size.

Additionally, the underlying `CollectionView` layout for these library views needs to be fixed and I think that in combination with this new scroll view it will be difficult to account for. See: #557.

For tvOS, I've only briefly thought about how I would make the filtering layout. The focus engine is a nightmare to work with, and with this I don't expect the interaction to be good either. The same `CollectionView` layout isn't much of an issue on tvOS.

So, mainly due to SwiftUI gripes that I have, the view itself may be an issue but I very much welcome the improvement. Regardless, I want this to be toggle-able option in `CustomizeViewsSettings`.

[...]

PR #847, opened 2023-09-16
Filter Toggles [iOS] [iPadOS]

Jpkribs:
Added an option in Settings > Accessibility > Customization for enabling/disabling Library & Search Filters. When enabled, this removed the Filter Pill from the HStack for the particular selected filter. Filters can be reordered from the same screen. This should be a simple change but the goal is to create this section/setting so as future filters are added they can be enabled/disabled to reduce clutter and improve UX.

If all filters are disabled, I chose to hide the entire Filter HStack. Please see the example images below:

<details>
<summary>Settings Location</summary>

</details>

---------------

<details>
<summary>Adjusting Search Filters</summary>

</details>

<details>
<summary>Search Results</summary>

</details>

---------------

<details>
<summary>Adjusting Library Filters</summary>

</details>

<details>
<summary>Library Results</summary>

</details>

---------------

<details>
<summary>Disabling All of the Filters</summary>

</details>

<details>
<summary>Search Results</summary>

</details>

<details>
<summary>Library Results</summary>

</details>

---------------

The eventual goal is to enable the functionality from this: https://github.com/jellyfin/Swiftfin/issues/680 but I wanted to ensure that this option could be disabled if unwanted. Instead of just adding an option for this "Filter by Letter" filter, this will allow for any filter to be disabled or re-ordered.


LePips (changes requested):
I like the idea! 👍 

However, I would want this to have a similar implementation as the video player action buttons. I found it a lot more customizable and even more manageable than a bunch of toggles - it makes adding new stuff almost trivial.

Relevant view: https://github.com/jellyfin/Swiftfin/blob/main/Swiftfin/Views/SettingsView/VideoPlayerSettingsView/Components/ActionButtonSelectorView.swift

Additionally, you can implement the "general implementation" I note at the top for this kind of interaction, if you want.

---- 

<details>
<summary>PR Description Meta</summary>

I do really like having screenshots in PRs, however I like them a lot better formatted so they make things easier to see. I use the below format on my portrait screenshots.

```
<details>
<summary>Screenshot Title</summary>

<img src="<image url>" width="30%" height="30%">
</details>
```

After you drag the photo I take the image url out of the markdown and put it into the `src`. It does take some time but makes things a lot more manageable.

</details>

Jpkribs:
Sounds good! Sorry about the formatting, I've only ever used VS + TFVC for code so I'm still a bit of a git novice. 

Just to confirm, implement a filterSettingsView Enum that points back to the Defaults.Filters. Use this enum to pass in the labels and filters using a ForEach(_selectedButtons) back on the filterDrawerHStack?

Jpkribs:
> I like the idea! 👍
> 
> However, I would want this to have a similar implementation as the video player action buttons. I found it a lot more customizable and even more manageable than a bunch of toggles - it makes adding new stuff almost trivial.
> 
> Relevant settings view: https://github.com/jellyfin/Swiftfin/blob/main/Swiftfin/Views/SettingsView/VideoPlayerSettingsView/Components/ActionButtonSelectorView.swift
> 
> Additionally, you can implement the "general implementation" I note at the top for this kind of interaction, if you want.
> PR Description Meta
> 
> I do really like having screenshots in PRs, however I like them a lot better formatted so they make things easier to navigate/read. I use the below format on my portrait screenshots.
> 
> ```
> <details>
> <summary>Screenshot Title</summary>
> 
> " width="30%" height="30%">
> </details>
> ```
> 
> After you drag the photo I take the image url out of the markdown and put it into the `src`. It does take some time but makes things a lot more manageable.

Just to confirm, is this what you meant:

<details>
<summary>Changed the Filter Section to a Sub-Section of Library</summary>


</details>

<details>
<summary>Updated the Filters from Individual Enums to a Single Array</summary>


</details>

I also added those icons on the selection but my heart isn't sold on them. I was just trying to mirror the Video Player buttons.

LePips:
Yep just like that, but I don't think we need the icons either as they aren't shown anywhere else.

Jpkribs:
> Yep just like that, but I don't think we need the icons either as they aren't shown anywhere else.

I have this working now with one exception. I've moved the label, filter, and selectorType into the enum for the Filters. The only issue I have now is that the "Activated" flag is determined by:

`viewModel.CurrentFilters.<filter/genres> != []
`
I could just add a function for isActive but I don't think we want a function on the view nor the enum. What would you say is the best route for this?

<details>
<summary>My Revisions to FilterDrawerHStack. Highlighted the Activated call.</summary>


</details>


<details>
<summary>Original FilterDrawerHStack. Highlighted the Activated call.</summary>


</details>

LePips:
Ah, I see. I'm find with a function on the new enum that takes on an `ItemFilters` parameter and determines whether that filter is active or not.

Jpkribs:
> Ah, I see. I'm find with a function on the new enum that takes on an `ItemFilters` parameter and determines whether that filter is active or not.



I've made these changes! I added an enum for filterDrawerActionButton that contains all of the components for adding in new filters. Moving forward, this should enable new filters by just updating the enum here and it will be populated in the settings and HStack.

The only item that I feel warrants some discussion is how I am setting the Filter Pills to Active/Inactive:

```
func getFilterProperty<T>(from object: T, propertyName: String) -> [ItemFilters.Filter] {
        let mirror = Mirror(reflecting: object)

        for child in mirror.children {
            if let label = child.label, label == propertyName {
                if let filterProperties = child.value as? [ItemFilters.Filter] {
                    return filterProperties
                }
            }
        }

        return []
    }
```

The purpose of this function is to find the the ItemFilters.Filters using the String value of the Filter. In my testing it works well but it does add a ForLoop so I'm up for other ideas. The reason I chose this approach is that it enables all of the filter configuration to be done in the filterDrawerActionButton Enum. This prevents the need for additional updates to logic in the FilterDrawerHStack when adding new filters.

Let me know if you can see anything else that needs attention!

Shared/Coordinators/SettingsCoordinator.swift
```diff
@@ -118,6 +121,11 @@ final class SettingsCoordinator: NavigationCoordinatable {
     func makeVideoPlayerSettings() -> VideoPlayerSettingsCoordinator {
         VideoPlayerSettingsCoordinator()
     }
+    
+    func makefilterDrawerButtonSelector(selectedButtonsBinding: Binding<[FilterDrawerActionButton]>) -> some View {
+        FilterDrawerButtonSelectorView(selectedButtonsBinding: selectedButtonsBinding)
```
LePips, line 126: The file for this view must have been linked incorrectly, Xcode on my machine says that it is in your `Documents` directory.

Shared/Coordinators/SettingsCoordinator.swift
```diff
@@ -40,6 +40,9 @@ final class SettingsCoordinator: NavigationCoordinatable {
     @Route(.push)
     var videoPlayerSettings = makeVideoPlayerSettings
 
+    @Route(.push)
+    var filterDrawerButtonSelector = makefilterDrawerButtonSelector
```
LePips, line 44: For future reference, I like these with some sorting. As you can see in the section above, they are sorted alphabetically. The `DEBUG` is split because it's obviously a special case.

Swiftfin/Components/FilterDrawerHStack/FilterDrawerHStack.swift
```diff
@@ -7,18 +7,22 @@
...
+    
     @ObservedObject
     private var viewModel: FilterViewModel
 
     private var onSelect: (FilterCoordinator.Parameters) -> Void
 
     var body: some View {
         HStack {
-            if viewModel.currentFilters.hasFilters {
+            if !filterActiveDrawerButtons.isEmpty && viewModel.currentFilters.hasFilters {
```
LePips, line 25: We don't want to disable the view of the buttons if there are none here - we want to disable the viewing of the `FilterDrawerHStack` view itself. This would mean that the drawer itself isn't shown when used in the `navBarDrawer` `View` extension.

Swiftfin/Components/FilterDrawerHStack/FilterDrawerHStack.swift
```diff
@@ -85,4 +61,19 @@ extension FilterDrawerHStack {
     func onSelect(_ action: @escaping (FilterCoordinator.Parameters) -> Void) -> Self {
         copy(modifying: \.onSelect, with: action)
     }
+    
+    // Finds the Filters/Genres/SortBy/Etc Property from the the ItemsFilter & 
+    func getFilterProperty<T>(from object: T, propertyName: String) -> [ItemFilters.Filter] {
```
LePips, line 66: One should rarely have to use reflection, if at all. There is already the `ItemFilters` key path defined on the `enum`, we need to use that instead.

LePips (changes requested):
I left a few comments on what I could see after pulling this down. Since there was a reference issue with the file and I had some time, I went ahead and did a more general implementation below, but it isn't complete and I leave it to you to fill in the gaps.

- https://github.com/LePips/SwiftFin/tree/filter-drawer-customization

This was a great start, and I had to think also about how this is used in conjunction with search since that uses the filters too. So, this entire feature was split into its own section in the settings that controls how all the filters work. I don't really care for a Library vs Search filters customization - even though I think someone will ask for that in the future.

I did rename some things but I'm agnostic, you can change them back if you would like. However I don't think these represent "actions" like the video player buttons do. Additionally, the filters themselves have been broken for some time and I may redo the entire filter implementation (but this feature would have minor modifications to accommodate that).

Swiftfin/Components/FilterDrawerHStack/FilterDrawerHStack.swift
Jpkribs: Gotcha! I was worried the nesting looping was a bit over-engineered for this (plus a bit inefficient). Just for future reference, why is reflection a rarely-or-never function? I'm still pretty new to Swift.

Since I was calling this from viewModel.currentFilters I wasn't sure what was the best way to get this call done. Below is the reworked method that I put on the FilterDrawerActionButton enum instead. Sorry for the screenshot I wasn't sure how to reference the code directly. Is there a better way I should be making this call?

<details>
<summary>Re-Worked isItemFilterActive</summary>


</details>

<details>
<summary>Re-Worked FilterDrawerHStack ForEach</summary>


</details>

Swiftfin/Components/FilterDrawerHStack/FilterDrawerHStack.swift
Jpkribs: In iOS, I found 2 references to this FilterDrawerHStack in libraryView and searchView. Is the goal something like this:

<details>
<summary>Disable ScrollView Inside navBarDrawer</summary>


</details>

Since it looks like the FilterDrawerHStack is passed into the navBarDrawer, I wasn't sure how to handle this aside from removing the FilterDrawer inside. Let me know if there is a better route than the screenshot above!

Shared/Coordinators/SettingsCoordinator.swift
Jpkribs: Just to confirm, this is the Shared/Coordinators/SettingsCoordinator.swift file? That's the only file that's referenced like that or is it all of them?

I can take a look at this and try re-pushing if it's only the one file. It's everything then god help me haha

Swiftfin/Components/FilterDrawerHStack/FilterDrawerHStack.swift
Jpkribs: Moved Filter == Activated test to the enum. See previous comment with screenshot. This should be in the PR branch now.

Shared/Coordinators/SettingsCoordinator.swift
Jpkribs: Re-ordered to fit the original alphabetical layout. This is in the iOS section alphabetically.

Shared/Coordinators/SettingsCoordinator.swift
Jpkribs: I *believe* this is referenced correctly now! That being said, I'm not sure how this got out of sync in the first place so... I'll await from confirmation that this looks correct.

Swiftfin/Components/FilterDrawerHStack/FilterDrawerHStack.swift
Jpkribs: See previous screenshot. Called the searchFilterDrawerButton setting for SearchView and libraryFilterDrawerButton for libraryView. Validated that this removes this the filters from the NavBar. It works the same as before where there is still a NavBar but it's empty. This is in the PR now but I think it's valid to want to approach this from another route if there is a way for me to wrap this where the navbar doesn't exist at all opposed to just being empty. I was wrapping the navbar in the If/Then before but it through me a View error.

Shared/Objects/FilterDrawerButton.swift
```diff
@@ -0,0 +1,101 @@
...
+    var settingsSelectorType: SelectorType {
+        switch self {
+        case .filters, .genres:
+            return .multi
+        case .order, .sort:
+            return .single
+        }
+    }
+    
+    var settingsItemsFilterProperty: String {
```
LePips, line 58: This is now unused.

Shared/Objects/FilterDrawerButton.swift
```diff
@@ -0,0 +1,101 @@
...
+        case .genres:
+            return "genres"
+        case .order:
+            return "sortOrder"
+        case .sort:
+            return "sortBy"
+        }
+    }
+    
+    func isItemsFilterActive(viewModel: FilterViewModel) -> Bool {
```
LePips, line 71: We should not reference a view model, we want to reference the `ItemFilters` object instead using the key path defined above. I actually liked defining the "inactive/default" state that's defined and then checking that. Additionally, that state could be used to initialize the `ItemFilters` properties.

```swift
var genres: [Filter] = FilterDrawerButton.genres.default
var filters: [Filter] = // etc
var sortOrder: [Filter] = // etc
var sortBy: [Filter] = // etc
```

Swiftfin/Views/LibraryView.swift
```diff
@@ -68,13 +71,15 @@ struct LibraryView: View {
         .navigationTitle(viewModel.parent?.displayTitle ?? "")
         .navigationBarTitleDisplayMode(.inline)
         .navBarDrawer {
```
LePips, line 73: If you look at how the navigation bar drawer was implemented, the drawer is shown when this modifier is used no matter what the view passed in is. Instead, we need to use the `.if` modifier and if there are buttons, show the drawer.

This is done on my branch for reference.

Shared/Objects/FilterDrawerType.swift
```diff
@@ -0,0 +1,45 @@
...
+        }
+    }
+    
+    var id: String {
+        rawValue
+    }
+    
+    var settingsDefaults: [FilterDrawerButton] {
+        
+        @Default(.Customization.Filters.libraryFilterDrawerButtons)
```
LePips, line 31: For future reference, this isn't the intended use of this property wrapper.

- https://github.com/sindresorhus/Defaults#swiftui-support

Shared/Objects/FilterDrawerButton.swift
```diff
@@ -0,0 +1,101 @@
...
+        case .genres:
+            return \.genres
+        case .order:
+            return \.sortOrder
+        case .sort:
+            return \.sortBy
+        }
+    }
+    
+    var settingsSelectorType: SelectorType {
```
LePips, line 49: Just do `selectorType`.

Shared/Objects/FilterDrawerButton.swift
```diff
@@ -0,0 +1,101 @@
...
+        case .sort:
+            return L10n.sort
+        }
+    }
+    
+    var id: String {
+        rawValue
+    }
+
+    var settingsFilter: WritableKeyPath<ItemFilters, [ItemFilters.Filter]> {
```
LePips, line 36: I don't think "settings" in the name fits, let's find a different name.

LePips (changes requested):

Swiftfin/Components/FilterDrawerHStack/FilterDrawerHStack.swift
LePips: Er saying "if at all" is a bit dogmatic from me, but reflection just isn't necessary and is used in very purposeful cases. That just happens to be very rare.

Shared/Objects/FilterDrawerButton.swift
Jpkribs: Ah, yes it is. Removed. I realized that this was the same class name as the one in Swiftfin/Components/FilterDrawerHStack. I've renamed this to FilterDrawerButtonSelection. TBH, I am terrible at naming things and I'm unsure what the convention is for Swift. Let me know if this should be named something else.

Shared/Objects/FilterDrawerButton.swift
Jpkribs: Done!

Shared/Objects/FilterDrawerButton.swift
Jpkribs: I changed this to "itemFilter" since that is what it returns. To be more in line with that naming, i am using isItemFilterActive for the function and itemFilterDefault for the setting where the default value is assigned on the enum. Feel free to let me know if that naming is confusing!

Swiftfin/Views/LibraryView.swift
Jpkribs: Ah! Gotcha, okay that makes a lot of sense. I have implemented this the way that you were doing it.

Shared/Objects/FilterDrawerButton.swift
Jpkribs: I've re-added the defaults and am instead passing in the itemFilters. I am a little confused by your declaration idea. In the enum, they are a  case so I was getting an error when trying to convert them to var. That being said, I think I'm just looking in the wrong place. Below is what I currently have. It feels a little repetitive  but I think changing the setting in the enum oppposed to changing the function:

```
    var itemFilterDefault: [ItemFilters.Filter] {
        switch self {
        case .filters:
            return []
        case .genres:
            return []
        case .order:
            return [APISortOrder.ascending.filter]
        case .sort:
            return [SortBy.name.filter]
        }
    }
    
    func isItemsFilterActive(activeFilters: ItemFilters) -> Bool {
        switch self {
        case .filters:
            return activeFilters.filters != self.itemFilterDefault
        case .genres:
            return activeFilters.genres != self.itemFilterDefault
        case .order:
            return activeFilters.sortOrder != self.itemFilterDefault
        case .sort:
            return activeFilters.sortBy != self.itemFilterDefault
        }
    }
```
Let me know if there is a better way for me to be doing this. 100% this is just me putting this in the wrong place.

Swiftfin.xcodeproj/project.pbxproj
```diff
@@ -771,6 +775,9 @@
 
 /* Begin PBXFileReference section */
 		091B5A872683142E00D78B61 /* ServerDiscovery.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = ServerDiscovery.swift; sourceTree = "<group>"; };
+		4E5E48DC2AB58F39003F1B48 /* FilterDrawerButtonSelectorView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; name = FilterDrawerButtonSelectorView.swift; path = ../../../../../../../../../Documents/FilterDrawerButtonSelectorView.swift; sourceTree = "<group>"; };
```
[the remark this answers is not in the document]
Jpkribs: I've repulled the project completely from scratch and experienced the same issue. This **SHOULD** be mapped appropriately now. Committed it and repulled it and everything so hopefully this isn't just some weird environment issue I need to work out.

Jpkribs:
I believe all changes are in with the exception of potentially changing https://github.com/jellyfin/Swiftfin/pull/847#discussion_r1329302416 to be cleaner. Updated the original post with new screenshots and a more accurate description of what this PR does.

LePips (dismissed):

LePips (approved):

PR #852, opened 2023-09-19
Filter by Letter [iOS] [iPadOS]

Jpkribs:
Creation of a Filter to Filter by the Starting Letter of a media item. Resolves the request created here: https://github.com/jellyfin/Swiftfin/issues/680

Relies on: https://github.com/jellyfin/Swiftfin/pull/847

------------

**New Settings**

Defaults to "Disabled", resulting in no Lettered Scrollbar.

<details>
<summary>Alpha Picker Selection</summary>

</details>

<details>
<summary>Letter Filter</summary>

</details>

------------

**Letter Filter**

<details>
<summary>NameStartsWith Filter</summary>

</details>

<details>
<summary>Drop Down Letter List</summary>

</details>

<details>
<summary>Letter Selection</summary>

</details>

<details>
<summary>Filtered Library</summary>

</details>

<details>
<summary>Filtered Library on Non-Existing Letter</summary>

</details>

------------

**Alpha Picker**

<details>
<summary>Disabled (Default)</summary>

</details>

<details>
<summary>Right</summary>

</details>

<details>
<summary>Left</summary>

</details>

<details>
<summary>Selected Letter</summary>

</details>

<details>
<summary>iPad Vertical</summary>

</details>

<details>
<summary>iPad Horizontal</summary>

</details>

------------

**Done:**

- Created a new class for ItemLetterFilter as an Enum for all of the letters. 
- Added the ItemLetterFilter to the ItemFilters and FilterDrawerSelection to be usable as a Filter and FilterDrawerButton respectively. 
- Updated the Search & Library ViewModels to include the nameStartsWith parameter to the API Call using the selection from the Letters FilterDrawerButton.
- Created another catch to prevent endless loading when reaching the end of the page of a letter-filtered view.*
- Created a localization label for "Letter."
- Re-Base on Main after https://github.com/jellyfin/Swiftfin/pull/847
- Created the alphaPicker enum, alphaPicker view, and alphaPickerButton view
- Added the alphaPicker next to the libraryView (if enabled)

*Will need a second set of eyes to confirm this is best practice.


LePips:
> jumps to a letter opposed to filtering on a letter ... would require the entire library up to that letter be loaded in prior to scrolling to it.

I do agree that a jump mechanism would be nicer, however something like that would require a more advanced mechanism to how we load the library. Ideally, we would want to implement a custom `UICollectionView` for these libraries and refactor how we load from the server entirely and store the results in some kind of ordered set. That is pretty advanced stuff and would require a lot of dedication to it.

Jpkribs:
> > jumps to a letter opposed to filtering on a letter ... would require the entire library up to that letter be loaded in prior to scrolling to it.
> 
> 
> 
> I do agree that a jump mechanism would be nicer, however something like that would require a more advanced mechanism to how we load the library. Ideally, we would want to implement a custom `UICollectionView` for these libraries and refactor how we load from the server entirely and store the results in some kind of ordered set. That is pretty advanced stuff and would require a lot of dedication to it.

I've been able to somewhat re-create the experience of the AlphaPicker from Jellyfin-Web where it just filters on nameStartsWith then deactivates when you select the same letter. For now, my outstanding issue is I want the alphaPicker to only exist on screen orientation where the screen size is tall enough where all of the letters can fit. Right now, I have it where when onAppear() it makes that check, but if you change the orientation of your screen, it won't update. I have some downtime at the airport next week where I was going to look at cleaning this up more.

Do you think it makes sense to proceed with the AlphaPicker route as a stopgap to "mirror" Jellyfin-Web? Or, would it make more sense for me to spend some time buckling down on creating our own version of collectionView? Or, I mean, if you think this is also some thing that doesn't make sense for someone at my level to look at, I wouldn't be offended either.

Jpkribs:
Working on this, I'm trying to get the alphaPicker to be hidden if it will exceed the size of the screen. I am using:

`NotificationCenter.default.addObserver(forName: UIDevice.orientationDidChangeNotification, object: nil, queue: .main)`

Then using this as an observable object. Is this the best way to achieve this?

LePips:
Only briefly looking, I'd have to see how I feel about hiding it. Frankly, I'm fine with people enabling it on small devices even if they know the size problems.

But until I think about that, you can take a look at the `onSizeChanged/onFrameChanged` modifiers instead of listening to notifications (which I would prefer).

Jpkribs:
> Only briefly looking, I'd have to see how I feel about it not allowing to be shown on smaller screens. Frankly, I'm fine with people enabling it on small devices even if they know the size problems.
> 
> But until I think about that, you can take a look at the `onSizeChanged/onFrameChanged` modifiers instead of listening to notifications (which I would prefer).

At least on the emulator, it looks like the full letter bar shows up in portrait mode, even on the mini/SE. I am primarily concerned with horizontal phones since there's really no spacing/sizing I could do to make it fit there. 

I could turn it into a scrollView? I agree with a comment you made earlier about putting a scrollview next to the libraryview feels a little clunky... that being said, that could be a better solution than checking screen resolution? I personally prefer the static letters that are hidden but what do you think @LePips?

I think my only other outstanding item I wanted to vet is potentially adding a setting to expand the size of the letters. Just because, on the iPad, it can be a little small comparatively (see original post images).

LePips:
> horizontal phones

I'm not worried about that since we should be making the app portrait only on iOS except for the video player. If you say that all of the letters show in portrait, then we shouldn't need a scroll view then.

> adding a setting to expand the size of the letters

I'm against this, as now we're just getting way too granular. I'm fine with checking at runtime the device type (iPhone, iPad) and having a different font size.

> buckling down on creating our own version of collectionView

Let's not do that right now, I'm exhausted just even thinking about how much work that would/will take.

---

*Localization*

I haven't mentioned this but it was first on my mind when I saw this feature: localization. While I know that this mirrors the web alphabet picker, I'm unsure how this interacts with other non-English alphabets (Japanese, Korean, etc.). Now, I'm pretty uneducated in this topic of accessibility but want to eventually get into it. I only bring this up to say that we shouldn't be surprised if this solution eventually (years from now at the least tbh) gets replaced by a more general solution.

Jpkribs:
> > horizontal phones
> 
> I'm not worried about that since we should be making the app portrait only on iOS except for the video player. If you say that all of the letters show in portrait, then we shouldn't need a scroll view then.
> 
> > adding a setting to expand the size of the letters
> 
> I'm against this, as now we're just getting way too granular. I'm fine with checking at runtime the device type (iPhone, iPad) and having a different font size.
> 
> > buckling down on creating our own version of collectionView
> 
> Let's not do that right now, I'm exhausted just even thinking about how much work that would/will take.
> 
> _Localization_
> 
> I haven't mentioned this but it was first on my mind when I saw this feature: localization. While I know that this mirrors the web alphabet picker, I'm unsure how this interacts with other non-English alphabets (Japanese, Korean, etc.). Now, I'm pretty uneducated in this topic of accessibility but want to eventually get into it. I only bring this up to say that we shouldn't be surprised if this solution eventually (years from now at the least tbh) gets replaced by a more general solution.

Sounds good. Personally, locking to portrait on iPhone is a much cleaner solution so I've removed my orientation management class and logic that was trying to remove/hide the bar. If it's enabled it will always be enabled.

For now, I have them both using the same layout but I've added the option to pass in a font so that should make making an iPad vs iPhone easy.
 
**Localization**

I completely agree with you! I was asking around when I was first trying to get my head around Swift just to understand what localization for this looks like. My understanding is that this nameStartsWith parameter on the API only does Romantic letters and numbers. So, at this time there is no localization available for this (I assume this is mirror on Jellyfin-Web).

TBH, when the day comes I have no idea how this could even be localized. I suppose if we had a localization for "Alphabet" that was just [A,B,C,D...] then we could parse it? I know those localization are just strings so I don't know how that would work.

Without having to look at handling orientation changes/Alpha-Picker exceeding the screen I think I am at a good for a second set of eyes.

Shared/Objects/ItemLetterFilter.swift
```diff
@@ -0,0 +1,43 @@
...
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2023 Jellyfin & Jellyfin Contributors
+//
+
+import Foundation
+import JellyfinAPI
+
+public enum ItemLetterFilter: Character, Codable, CaseIterable {
```
LePips, line 12: I took some time to figure out the alphabet localization because I didn't like defining the alphabet manually like this.

The following API returns the locale alphabet + English alphabet + `#`, so we get this for free: https://developer.apple.com/documentation/uikit/uilocalizedindexedcollation

For example, in a Korean locale app the following would be returned:

```swift
print(UILocalizedIndexedCollation.current())

// ["ㄱ", "ㄴ", "ㄷ", "ㄹ", "ㅁ", "ㅂ", "ㅅ", "ㅇ", "ㅈ", "ㅊ", "ㅋ", "ㅌ", "ㅍ", "ㅎ", "A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R", "S", "T", "U", "V", "W", "X", "Y", "Z", "#"]
```

I know that I had previously mentioned that "we can do this later", but I would now like to do the localization since I would find it a bit less work than manually defining alphabets. As a part of this, I would like to change the contexts/names of everything to match Apple's, which I find more preferable and generic. For example, we would change the context of "letter" to "index" and even "alpha picker" to "index title selector". These aren't hard names so you may think of something better.

The `#` would be a new feature and that works with the `NameLessThan=A` parameter.

I anticipate that this would require refactoring and I would like consistent naming across the new feature both in code and that is displayed (`nameStartsWith` key vs `Letter`).

Shared/Components/AlphaPicker/AlphaPickerButton.swift
```diff
@@ -0,0 +1,105 @@
...
+                            viewModel.currentFilters.nameStartsWith = []
+                        }
+                    }
+                }
+            ) {
+                Text(String(letter))
+                    .font(letterFont)
+                    .frame(width: letterFrame, height: letterFrame)
+                    .foregroundColor(activated ? Color(UIColor.white) : accentColor)
+                    .opacity(1.0)
```
LePips, line 67: Unnecessary opacity.

Shared/Components/AlphaPicker/AlphaPickerButton.swift
```diff
@@ -0,0 +1,105 @@
...
+                    .padding(.vertical, letterPadding)
+                    .fixedSize(horizontal: true, vertical: true)
+                    .background {
+                        RoundedRectangle(cornerRadius: 5)
+                            .frame(width: letterHighlight, height: letterHighlight)
+                            .foregroundColor(activated ? accentColor.opacity(0.5) : Color(UIColor.clear).opacity(0.0))
```
LePips, line 73: `Clear` doesn't need an opacity.

Shared/Components/AlphaPicker/AlphaPickerView.swift
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    var body: some View {
+        VStack(spacing: 0) {
+            ForEach(ItemLetterFilter.allCases, id: \.self) { letter in
+                AlphaPickerButton(
+                    letter: letter.rawValue, letterFont: Font.caption.weight(.bold),
+                    letterFrame: 15, letterPadding: 2, letterHighlight: 20,
+                    activated: viewModel.currentFilters.nameStartsWith.contains {
+                        $0.displayTitle == String(
```
LePips, line 31: We shouldn't use `displayTitle` for verification.

Shared/Components/AlphaPicker/AlphaPickerButton.swift
```diff
@@ -0,0 +1,105 @@
...
+
+extension AlphaPickerView {
+
+    struct AlphaPickerButton: View {
+
+        @Default(.accentColor)
+        private var accentColor
+
+        private let letter: Character
+        private let letterFont: Font
```
LePips, line 23: The font, frame, padding, and highlight properties aren't how UI values are passed in SwiftUI. Instead, use the `.font()` and `.padding()` modifiers which are passed into the button's environment. I also don't see value in the manual padding and highlight padding.

For font sizes, please use the defined font sizes as defined by the HIG, only in a few places are the fonts manually defined but should be moved over (a place in LiveTV).

- https://developer.apple.com/design/human-interface-guidelines/typography#Specifications

LePips (changes requested):
I want this to go in a different direction now since I actually took some time to see how we could localize this and found out it was more accessible than expected. I won't be all encompassing on the SwiftUI changes that I want until after some refactoring.

Shared/Strings/Strings.swift
```diff
@@ -28,6 +28,8 @@ internal enum L10n {
   internal static let allGenres = L10n.tr("Localizable", "allGenres", fallback: "All Genres")
   /// All Media
   internal static let allMedia = L10n.tr("Localizable", "allMedia", fallback: "All Media")
+  /// Lettered Scrollbar
+  internal static let alphaPickerTitle = L10n.tr("Localizable", "alphaPickerTitle", fallback: "Lettered Scrollbar")
```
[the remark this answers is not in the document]
Jpkribs: Agreed. I think AlphaPicker makes sense as the "behind the scenes" name since it matches Jellyfin-Web but the only reference to "Lettered Scrollbar" should be the String. So changing that should be easy so long as we're fine with AlphaPicker as the variable name?

Any thoughts on other names? Lettered Scrollbar was my Hail Mary because I'm drawing a blank...

Shared/Objects/ItemLetterFilter.swift
Jpkribs: Sounds good! I originally had # included but it gets a tiny bit messier with the logic. I can re-add this back in with the localization sometime this week!

Shared/Objects/AlphaPickerSelection.swift
```diff
@@ -0,0 +1,28 @@
...
+// Copyright (c) 2023 Jellyfin & Jellyfin Contributors
+//
+
+import Defaults
+import Foundation
+
+enum AlphaPickerSelection: String, CaseIterable, Defaults.Serializable, Displayable {
+
+    case disabled
+    case left
```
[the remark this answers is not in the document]
Jpkribs: Do we still want the labels to be Left/Right or should those be Leading/Trailing as well?

Shared/Components/AlphaPicker/AlphaPickerButton.swift
Jpkribs: A lot of this came from an older version of this I was working on with the goal of reusing this class for both iOS/iPadOS & tvOS. Looking back at it, this would likely create a subpar experience on 2 platforms. I'll clear these out!

Shared/Objects/AlphaPickerSelection.swift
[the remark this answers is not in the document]
Jpkribs: Changed to None/Leading/Trailing. Labels are now None/Left/Right.

Shared/Strings/Strings.swift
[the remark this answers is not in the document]
Jpkribs: Temporary change to Letter Slicer??? I dunno, kind of does the same thing as an Excel Slicer? But that's just where my head is today... 100% open to change that naming but I'll need someone a little more creative to provide a good name.

Shared/Objects/ItemLetterFilter.swift
Jpkribs: This is done. I am handling that switch over from 'nameStartsWith: #' -> 'nameLessThan: A' on the libraryModelView. Let me know if that isn't where that belongs.

Since I am using the full list of characters to populate the Filter the HStack Pill and the AlphaPicker, I made an email that just returns all characters sorted. I put my phone in Korean and this does work. 

The only issue/compromise that I made was that, with English+Other Language, this will likely exceed the static AlphaPicker that fit on all screens. My compromise was to turn the AlphaPicker into a ScrollView to better fit additional Characters. This isn't the end of the world to have it this way, especially since your currently scroll on the ScrollView isn't reset anymore because of your recommendations here: https://github.com/jellyfin/Swiftfin/pull/852#discussion_r1336236717

Let me know if the ScrollView is too detrimental (I personally think the static AlphaPicker is ideal). The other alternative is to only take the localization letters into account. That is, Korean localization would ONLY have Korean letters and no Romantic Letters. That's obtainable by using .IsLetter() since non-Romantic characters return false for that test. End of the day, the ScrollView isn't a bad solution but I don't love the move part right next to the libraryView being another scrolling, moving part.

Jpkribs:
I believe I've made the changes requested! I have three outstanding "maybes" that I would appreciate a second set of eyes on. 

1. The creation of AlphaPickerCharacters to hold the letters/characters from the localized language. Is this the best way to do this? Should this be an enum or a function? 

> The goal is to eventually be able to swap this out. The reason I would want to swap this out is to be more dynamic and instead of "Filter based on localization" it should be "Filter based on the first found characters for the media present." I'm not sure how to achieve that at this point without being VERY performance heavy but I think this would be the ideal solution.


2. Should I declare another filter for nameLessThan or am I fine with the logic on the libraryViewModel? Instead of a second filter, I just have logic on the Model View for for "IF # THEN nameLessThan ELSE nameStartsWith" The current logic works but I just want to make sure there isn't a better way. I personally like this solution since there are less items to maintain but the alternative solution is a boolean filter for nameLessThan but then I would need additional logic to ignore the nameStartsWith... I like the current solution better but let me know if there is a better way to approach this.

3. In the AlphaPicker I have turning into a ScrollView when > 27 characters. This is because 27 characters will be the Romantic + #. Since the API referenced returns Localized + Romantic + #, this will always exceed 27 on non-Romantic languages. In which case, I change this to a ScrollView to fit additional letters. Is this an okay solution or do we want to only include localized characters and not Romantic + localized? I lean in favor of Romantic + localized since I know there is a lot media created in western countries so non-romantic Language speakers will still likely utilize Romantic Languages for their media. This is just my thought for now unless we have a route to achieve #1.

Swiftfin/Components/AlphaPicker/AlphaPickerButton.swift
```diff
@@ -0,0 +1,90 @@
+//
+// Swiftfin is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2023 Jellyfin & Jellyfin Contributors
+//
+
+// Joe Kribs: joseph@kribs.net 02/06/2023
```
LePips, line 9: ?

Swiftfin/Components/AlphaPicker/AlphaPickerView.swift
```diff
@@ -0,0 +1,63 @@
...
+        .frame(width: 20)
+    }
+
+    var body: some View {
+        Group {
+            if AlphaPickerCharacters.letters.characters.count > 27 {
+                ScrollView(showsIndicators: false) {
+                    alphaPickerBody
+                }
+                .padding(.top, 5)
```
LePips, line 46: Playing around with it for a second, it would be more useful to have a `.padding(.vertical)` on the `VStack`. For future reference, setting `padding` on the contents of a `ScrollView` is better to avoid ugly clipping.

Swiftfin/Components/AlphaPicker/AlphaPickerView.swift
```diff
@@ -0,0 +1,63 @@
...
+                        onSelect(.init(
+                            title: L10n.letter,
+                            viewModel: viewModel,
+                            filter: \.nameStartsWith,
+                            selectorType: .single
+                    ))
+                }
+            }
+        }
+        .frame(width: 20)
```
LePips, line 37: Don't set the `frame` here if you're also setting the `frame` when using the view.

Swiftfin/Components/AlphaPicker/AlphaPickerCharacters.swift
```diff
@@ -0,0 +1,30 @@
...
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2023 Jellyfin & Jellyfin Contributors
+//
+
+import Defaults
+import Foundation
+import SwiftUI
+
+enum AlphaPickerCharacters: String, CaseIterable, Displayable, Defaults.Serializable {
```
LePips, line 13: This type is unnecessary. A better implementation would be to pass in the characters to the `AlphaPickerSelection`.

Swiftfin/Components/AlphaPicker/AlphaPickerButton.swift
```diff
@@ -0,0 +1,90 @@
...
+
+        @Default(.accentColor)
+        private var accentColor
+
+        private let nameStartsWith: String
+        private let activated: Bool
+        private let viewModel: FilterViewModel
+        private var onSelect: () -> Void
+
+        private init(
```
LePips, line 27: This `init` is unnecessary.

Shared/Strings/Strings.swift
```diff
@@ -28,6 +28,8 @@ internal enum L10n {
   internal static let allGenres = L10n.tr("Localizable", "allGenres", fallback: "All Genres")
   /// All Media
   internal static let allMedia = L10n.tr("Localizable", "allMedia", fallback: "All Media")
+  /// Letter Slicer
+  internal static let alphaPickerTitle = L10n.tr("Localizable", "alphaPickerTitle", fallback: "Letter Slicer")
```
LePips, line 32: Let's just name it `Letter Selector`, as I don't think a `Slicer` makes sense.

LePips (changes requested):
After having looked at my filters implementation in these PRs, I will definitely think of refactoring them. I don't like the type erasure (`Filters`) and having multiple layers (`FilterViewModel` + `ItemFilters`).

Swiftfin/Components/AlphaPicker/AlphaPickerButton.swift
Jpkribs: Whoops! That's from some X-Code defaults I was messing around with when I first started. Hence the February (8 months ago). 

It'll be removed on the next commit.

Jpkribs:
> After having looked at my filters implementation in these PRs, I will definitely think of refactoring them. I don't like the type erasure (`Filters`) and having multiple layers (`FilterViewModel` + `ItemFilters`).

Would be the same route/rework for implementing filters on tvOS? Or is that a whole other beast?

LePips:
The refactoring would change how filters works on both platforms, but I wouldn't be implementing them on tvOS with that work.

Swiftfin/Components/AlphaPicker/AlphaPickerCharacters.swift
Jpkribs: Just so I understand. I should be calling UILocalizedIndexedCollation.current().sectionTitles.sorted() from the alphaPickerSelection in Default settings? I was originally looking at something like that but I didn't know if 1) that would work or if 2) calling a method there was allowed.

LePips: We just don't need this intermediary type for a single static function that returns a `[String]`. The selector can take a `[String]` or it can be defined in the selector itself.

Swiftfin/Components/AlphaPicker/AlphaPickerButton.swift
```diff
@@ -0,0 +1,90 @@
...
+            viewModel: FilterViewModel,
+            onSelect: @escaping () -> Void
+        ) {
+            self.nameStartsWith = nameStartsWith
+            self.activated = activated
+            self.viewModel = viewModel
+            self.onSelect = onSelect
+        }
+
+        var body: some View { Button(
```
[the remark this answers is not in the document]
Jpkribs: Would that be this?

```
some View {

Button( Nested {

...

) {
```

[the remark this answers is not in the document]
LePips: Buttons should be declared like below, this is the [multiple trailing closure](https://www.hackingwithswift.com/swift/5.3/multiple-trailing-closures) syntax.

```swift
Button {
// action
} label: {
// label
}
```

Jpkribs:
All reviews should be resolved in this latest commit. I will likely have some linting changes that need to be addressed but at this time I believe the code portion of this is ready.

Swiftfin/Components/AlphaPicker/AlphaPickerButton.swift
```diff
@@ -0,0 +1,76 @@
...
+        private var accentColor
+
+        private let nameStartsWith: String
+        private let activated: Bool
+        private let viewModel: FilterViewModel
+        private var onSelect: () -> Void
+
+        var body: some View { 
+            Button {
+                action: do {
```
LePips, line 26: I believe you may be confused on trailing syntax, this is a [labeled statement](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/statements/#Labeled-Statement) to help with control flow. While the syntax itself looks similar to a [trailing closure label](https://www.fuckingmultipletrailingclosuresyntax.com/), they are completely different.

Remove the `action: do` block.

LePips (changes requested):
Syntax and fix merge errors.

Swiftfin/Components/AlphaPicker/AlphaPickerButton.swift
Jpkribs: Gotcha! This should be updated now and I've test it to confirm this does not change the functionality.

PR #857, opened 2023-09-21
Minor UI fixes, force orientation

tonyd33:
# Purpose

1. X button to close video player had too small of a clickable area
2. "Chapters" and "Current" text in chapter select menu had too little padding on iPad
3. Portrait orientation was allowed during video playback
  a. UI barely fit on small screens; even worse when trying to open "Advanced" menu
  b. On iPad, changing from landscape to portrait caused major problems that forced closing the app

## Related images/videos

image

orientation_breaking

# Changes

1. Add `contentShape(Rectangle())` to the X button
2. Add padding to the HStack surrounding "Chapters" and "Current"; refactor safe area padding to outermost stack
3. Lock orientation to landscape during playback — mostly uncommented existing code

# Testing

- Verifying X button clickable by physically pressing it with thumb
- Verifying rotations and chapters text look good on iOS 14 and 16 on iPhone and iPad

| Before | After |
|--------|--------|
| [can't click X](https://matrix.to/#/!UGqxMjsElfCqgeDJBO:matrix.org/$oW-D3AuDVDaIWh5hRsbNbPzGC1ZkUH6-Tiar5Zwtf4g?via=bonifacelabs.ca&via=t2bot.io&via=matrix.org) | [can click X](https://matrix.to/#/!UGqxMjsElfCqgeDJBO:matrix.org/$6U60Zb3jmFc6t4-eMZLN7aSXuiwpDOytlpeKpYYNezs?via=bonifacelabs.ca&via=t2bot.io&via=matrix.org) |
| image | image |
| image | image |
| image | orientation_force |

### DIFF
--- a/Swiftfin/App/AppDelegate.swift
+++ b/Swiftfin/App/AppDelegate.swift
@@ -38,17 +38,37 @@ class AppDelegate: NSObject, UIApplicationDelegate {
         AppDelegate.orientationLock
     }
 
-    static func changeOrientation(_ orientation: UIInterfaceOrientationMask) {
+    static func enterPlaybackOrientation() {
+        AppDelegate.changeOrientation(.landscape)
+    }
 
-//        guard UIDevice.isPhone else { return }
-//
-//        Self.orientationLock = orientation
-//
-//        if #available(iOS 16, *) {
-//            let windowScene = UIApplication.shared.connectedScenes.first as? UIWindowScene
-//            windowScene?.requestGeometryUpdate(.iOS(interfaceOrientations: orientation))
-//        } else {
-//            UIDevice.current.setValue(UIInterfaceOrientation.landscapeRight.rawValue, forKey: "orientation")
-//        }
+    static func leavePlaybackOrientation() {
+        if UIDevice.isIPad {
+            AppDelegate.changeOrientation(.allButUpsideDown)
+        } else {
+            // On iPhone, users likely want to return to portrait mode after leaving playback.
+            // However, we don't want to lock them into portrait mode, so switch back after a delay.
+            AppDelegate.changeOrientation(.portrait)
+
+            // 0.25 seconds is about the time to switch from landscape to portrait.
+            // Changing orientation again too early will cause the top time/battery bar to remain
+            // at the side instead of moving up top, like it should.
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.25) {
+                AppDelegate.changeOrientation(.allButUpsideDown)
+            }
+        }
+    }
+
+    private static func changeOrientation(_ orientation: UIInterfaceOrientationMask) {
+        guard UIDevice.isPhone || UIDevice.isIPad else { return }
+
+        orientationLock = orientation
+
+        if #available(iOS 16, *) {
+            let windowScene = UIApplication.shared.connectedScenes.first as? UIWindowScene
+            windowScene?.requestGeometryUpdate(.iOS(interfaceOrientations: orientation))
+        } else {
+            UIDevice.current.setValue(UIInterfaceOrientation.landscapeRight.rawValue, forKey: "orientation")
+        }
     }
 }
--- a/Swiftfin/Views/VideoPlayer/Overlays/ChapterOverlay.swift
+++ b/Swiftfin/Views/VideoPlayer/Overlays/ChapterOverlay.swift
@@ -65,8 +65,7 @@ extension VideoPlayer.Overlay {
                             .foregroundColor(accentColor)
                     }
                 }
-                .padding(.leading, safeAreaInsets.leading)
-                .padding(.trailing, safeAreaInsets.trailing)
+                .padding(.horizontal)
 
                 ScrollViewReader { proxy in
                     ScrollView(.horizontal, showsIndicators: false) {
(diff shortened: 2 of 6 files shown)


Swiftfin/App/AppDelegate.swift
```diff
@@ -38,17 +38,37 @@ class AppDelegate: NSObject, UIApplicationDelegate {
...
+            // 0.25 seconds is about the time to switch from landscape to portrait.
+            // Changing orientation again too early will cause the top time/battery bar to remain
+            // at the side instead of moving up top, like it should.
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.25) {
+                AppDelegate.changeOrientation(.allButUpsideDown)
+            }
+        }
+    }
+
+    private static func changeOrientation(_ orientation: UIInterfaceOrientationMask) {
```
tonyd33, line 62: I know @LePips had been trying an implementation with `PreferenceUIHostingController` and I tried to get that working, but I couldn't. If it had been working and you guys want to enforce that design, I'll revert these changes and let you guys implement it again.

I think changing orientation imperatively as opposed to declaratively on the Views is fine for now, since there are so few entry and exit points for the video player and not enforcing orientation is causing major problems on iPad.

Additionally, if we are going with this approach, I can clean up and delete the other commented-out code from the other approach if you'd like.

Swiftfin/Views/VideoPlayer/Overlays/ChapterOverlay.swift
```diff
@@ -65,8 +65,7 @@ extension VideoPlayer.Overlay {
                             .foregroundColor(accentColor)
                     }
                 }
-                .padding(.leading, safeAreaInsets.leading)
```
LePips, line 68: This padding rework isn't right because it breaks iOS. The entire view is padded so the linear gradient is as well and it makes the buttons when outside of the view not usable (as expected). This padding is somewhat considerable nowadays and we don't want that.

Simulator Screenshot - iPhone 15 Pro - 2023-09-22 at 22 26 31

What I am fine with is a check whether it's an iPad and adding more padding. In fact, I do that a lot if you search `.if(UIDevice.isIPad)`.

Swiftfin/App/AppDelegate.swift
LePips: I'm fine with this for now if you want to keep it in, however it isn't the final behavior that we want. We want iOS to be in portrait only except for the video players and maybe some other views down the road? iPadOS should always be anything but upside down. This more closely mirrors the Apple TV app.

I actually do intend to make the video player overlay friendlier on portrait, meaning revamping the Advanced Settings stuff entirely and making things pretty on the overlay, but I never got to that.

I do want to take another shot at the preference orientation solution sometime. You're correct that there are only a few points where this stuff matters so it is fairly maintainable but I guess I just like how things are more declarative 🤷‍♂️ .

Again, I'm fine with this now but don't be surprised if it eventually gets reworked a bit.

LePips (changes requested):

Swiftfin/Views/VideoPlayer/Overlays/ChapterOverlay.swift
tonyd33: Who watches on aspect fill though :P

Thanks for noticing. These padding changes have been reverted except adding default padding on iPad:

image
Maybe could use a little bit more padding, but eh, it doesn't look like the text might be cut off anymore which is what I originally wanted to solve so I'm happy with it.

And here's a screenshot of chapter menu open with aspect fill on iPhone with these changes:
image

Swiftfin/App/AppDelegate.swift
tonyd33: I'm totally fine with these changes getting overridden. I just want to make it so that the main branch of this repo is in good enough condition that someone like me could come by, build the app, and just start using it as an end user happily.

I slightly prefer the preference orientation approach here too and I would've implemented it here if I were more experienced in Swift (and if a clean solution was possible) so I'm looking forward to seeing your take on it.

I'll let you guys handle the rest of the orientation behavior. I mean for these changes to be a quick and dirty fix to clunky iOS portrait UI and app-breaking iPadOS rotation issues, not a long term solution — especially to the iPad rotation issues.

Swiftfin/Views/VideoPlayer/Overlays/ChapterOverlay.swift
```diff
@@ -65,8 +65,8 @@ extension VideoPlayer.Overlay {
...
-                .padding(.trailing, safeAreaInsets.trailing)
+                .padding(.leading, UIDevice.isIPad ? nil : safeAreaInsets.leading)
```
LePips, line 68: I would prefer the way I have done this in `PillHStack/PosterHStack`, for example:

```swift
MyView()
	.padding(.leading) // padding on all platforms
	.if(UIDevice.isIPad) { view in
		view.padding(.leading)
	}
```

Additionally, I would like the headers and the buttons to align, so this extra padding should be put on the chapter `HStack` as well.

LePips (changes requested):

Swiftfin/Views/VideoPlayer/Overlays/ChapterOverlay.swift
tonyd33: Ah sorry, I didn't read your comment closely enough. To be sure, this is what we want right? Lmk if not

| Before | After |
|--------|--------|
| Simulator Screenshot - iPad Air 4th - 2023-09-24 at 09 17 00 | image |

LePips: Yep, just like that 👍

LePips (approved):

Issue #711, opened 2023-02-28
playback speed control saved


marouamghar:
**Describe the feature you'd like**
when playing a new episode, the playback speed resets to 1. It would be nice to be able to have it saved

PR #859, opened 2023-09-22
Add audio, subtitle, playback select for tvOS

tonyd33:
# Purpose

- Audio/subtitle/playback selection was completely missing in tvOS
- Opening up the audio/subtitle/playback selection menu opened to a button that said "None"
- Toggling autoplay reset focus

# Changes

- Add audio, subtitle, and playback selection in tvOS
- Move `playbackSpeed` property into `VideoPlayerManager` to keep consistency with audio and subtitle property
- Make unselected label have unfilled circle icon, make selected label have checkmark circle filled icon

# Testing

- Check audio, subtitle and playback select work in tvOS
- Tested both iOS and tvOS to make sure the `playbackSpeed` move didn't break anything
- Removed chapter select from the enum in `SmallMenuOverlay` because chapter selection already exists outside of that menu

| Before | After |
|--------|--------|
| [track select before](https://imgur.com/IR8BR5Y)  | [track select after](https://imgur.com/xM2Wntc) |
| image  | image |
| image | image |
| button_reset | button_reset_fix |
| none_button | none_button_fix |

### iOS playback speed works

ios_playbackspeed

# Notes

While working on the fix, I noticed:

- Audio/subtitle were not set when loading video on tvOS in `VideoPlayerManager`, so opening the selection menu shows that no audio/subtitle track is selected until manual selection
- ~~Opening the `SmallMenuOverlay` for the first time makes the default `Text("None")` button because `lastFocusedSection` is not set by default~~ — fixed in this PR now

**Do we want these problems fixed in this PR?**

## TODO

- [x] Attach before and after images
- [x] ~~(Maybe) change `playbackSpeed` in `VideoPlayerManager` to `Float` in case we ever want to apply a nonstandard speed~~ — lmk if we want this.
- [x] ~~(Maybe) refactor out shared code among the selection menus in `SmallMenuOverlay`~~ — lmk if we want this

### DIFF
--- a/Swiftfin tvOS/Views/VideoPlayer/Overlays/Overlay.swift
+++ b/Swiftfin tvOS/Views/VideoPlayer/Overlays/Overlay.swift
@@ -30,72 +30,73 @@ extension VideoPlayer {
         @StateObject
         private var overlayTimer: TimerProxy = .init()
 
-        var body: some View {
-            ZStack {
-
-                MainOverlay()
-                    .visible(currentOverlayType == .main)
-
-                ConfirmCloseOverlay()
-                    .visible(currentOverlayType == .confirmClose)
-
-                SmallMenuOverlay()
-                    .visible(currentOverlayType == .smallMenu)
-
+        @ViewBuilder
+        private var currentOverlay: some View {
+            switch currentOverlayType {
+            case .chapters:
                 ChapterOverlay()
-                    .visible(currentOverlayType == .chapters)
+            case .confirmClose:
+                ConfirmCloseOverlay()
+            case .main:
+                MainOverlay()
+            case .smallMenu:
+                SmallMenuOverlay()
             }
-            .visible(isPresentingOverlay)
-            .animation(.linear(duration: 0.1), value: currentOverlayType)
-            .environment(\.currentOverlayType, $currentOverlayType)
-            .environmentObject(overlayTimer)
-            .onChange(of: currentOverlayType) { newValue in
-                if [.smallMenu, .chapters].contains(newValue) {
-                    overlayTimer.pause()
-                } else if isPresentingOverlay {
+        }
+
+        var body: some View {
+            currentOverlay
+                .visible(isPresentingOverlay)
+                .animation(.linear(duration: 0.1), value: currentOverlayType)
+                .environment(\.currentOverlayType, $currentOverlayType)
+                .environmentObject(overlayTimer)
+                .onChange(of: currentOverlayType) { newValue in
+                    if [.smallMenu, .chapters].contains(newValue) {
+                        overlayTimer.pause()
+                    } else if isPresentingOverlay {
+                        overlayTimer.start(5)
+                    }
+                }
+                .onChange(of: overlayTimer.isActive) { isActive in
+                    guard !isActive else { return }
+
+                    withAnimation(.linear(duration: 0.3)) {
+                        isPresentingOverlay = false
+                    }
+                }
+                .onSelectPressed {
+                    currentOverlayType = .main
+                    isPresentingOverlay = true
                     overlayTimer.start(5)
                 }
-            }
-            .onChange(of: overlayTimer.isActive) { isActive in
-                guard !isActive else { return }
+                .onMenuPressed {
 
-                withAnimation(.linear(duration: 0.3)) {
-                    isPresentingOverlay = false
-                }
-            }
-            .onSelectPressed {
-                currentOverlayType = .main
-                isPresentingOverlay = true
-                overlayTimer.start(5)
-            }
-            .onMenuPressed {
+                    overlayTimer.start(5)
+                    confirmCloseWorkItem?.cancel()
 
-                overlayTimer.start(5)
-                confirmCloseWorkItem?.cancel()
-
-                if isPresentingOverlay && currentOverlayType == .confirmClose {
-                    proxy.stop()
-                    router.dismissCoordinator()
-                } else if isPresentingOverlay && currentOverlayType == .smallMenu {
-                    currentOverlayType = .main
-                } else {
-                    withAnimation {
-                        currentOverlayType = .confirmClose
-                        isPresentingOverlay = true
-                    }
-
-                    let task = DispatchWorkItem {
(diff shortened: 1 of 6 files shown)


Swiftfin tvOS/Views/VideoPlayer/Overlays/Overlay.swift
```diff
@@ -30,72 +30,73 @@ extension VideoPlayer {
...
             }
-            .onChange(of: overlayTimer.isActive) { isActive in
-                guard !isActive else { return }
+        }
 
-                withAnimation(.linear(duration: 0.3)) {
-                    isPresentingOverlay = false
+        var body: some View {
+            currentOverlay
+                .visible(isPresentingOverlay)
```
tonyd33, line 49: Everything below should just be from `swiftformat`

Swiftfin/Views/VideoPlayer/Overlays/Components/ActionButtons/PlaybackSpeedActionButton.swift
```diff
@@ -30,19 +26,19 @@ extension VideoPlayer.Overlay.ActionButtons {
...
-                        if Float(speed.rawValue) == playbackSpeed {
+                        if speed == videoPlayerManager.playbackSpeed {
                             Label(speed.displayTitle, systemImage: "checkmark")
                         } else {
                             Text(speed.displayTitle)
                         }
                     }
                 }
 
-                if !PlaybackSpeed.allCases.map(\.rawValue).contains(where: { $0 == Double(playbackSpeed) }) {
```
LePips, line 44: I can't remember why this exists (checking for a non-defined playback speed), so we can remove it.

LePips (changes requested):
Thank you for working on this. The playback speed being pulled into the manager is great as it is related to #711, but I haven't made a good mechanism for having values "transfer" between different media items.

I see that the small overlay was expanded to fill the entire screen, which we don't want, but I'm fine with for now to fix later. I'm fine with the circles on the buttons, but I may change that later on.

Per the "lmk if we want this" comments:

1 - keep the `playbackSpeed` as the `PlaybackSpeed` type
2 - I'm always for some cleanliness, so I'm welcome to that.

In fact, I think that the entire overlay on tvOS needs a redesign, both in terms of design itself and implementation. I've just had to deal with SwiftUI and its lack of support on tvOS and I don't think it's the prettiest thing in the world. /rant

tonyd33:
> Thank you for working on this. The playback speed being pulled into the manager is great as it is related to #711, but I haven't made a good mechanism for having values "transfer" between different media items.
> 
> I see that the small overlay was expanded to fill the entire screen, which we don't want, but I'm fine with for now to fix later. I'm fine with the circles on the buttons, but I may change that later on.
> 
> Per the "lmk if we want this" comments:
> 
> 1 - keep the `playbackSpeed` as the `PlaybackSpeed` type 2 - I'm always for some cleanliness, so I'm welcome to that.
> 
> In fact, I think that the entire overlay on tvOS needs a redesign, both in terms of design itself and implementation. I've just had to deal with SwiftUI and its lack of support on tvOS and I don't think it's the prettiest thing in the world. /rant

Just noticed the "Small" in "SmallMenuOverlay", haha.

For the circles on the buttons, I added it because changing from `Text` (no checkmark/image) to `Label` caused weird issues with the button width, making the text clip out of the button's bounds. Of course, these changes are likely going to be placeholders until the tvOS overlay overhaul.

Seeing as the playback slider is completely broken on tvOS among the other design issues, I'd be happy to see a rework on tvOS and help implement it. If you have mocks or even a detailed enough description of design goals, I'd love to incrementally work on it over the next while. We can discuss more on element

Swiftfin tvOS/Views/VideoPlayer/Overlays/Overlay.swift
```diff
@@ -30,72 +30,73 @@ extension VideoPlayer {
...
+            switch currentOverlayType {
+            case .chapters:
+                ChapterOverlay()
+            case .confirmClose:
                 ConfirmCloseOverlay()
-                    .visible(currentOverlayType == .confirmClose)
-
+            case .main:
+                MainOverlay()
+            case .smallMenu:
```
tonyd33, line 42: Also, so that this is attributable in the future, this was changed from a `ZStack` because

1. We only show one overlay at a time
2. With multiple overlays mounted, each of them fight for focus. Specifically, having the `MainOverlay` mounted with the `SmallMenuOverlay` caused focus conflicts that I couldn't resolve.

Swiftfin tvOS/Views/VideoPlayer/Overlays/SmallMenuOverlay.swift
```diff
@@ -170,5 +189,22 @@ extension VideoPlayer {
...
+        struct MenuStyle: ViewModifier {
+            var focusGuide: FocusGuide
+
+            func body(content: Content) -> some View {
+                content
+                    .frame(height: 80)
+                    .padding(.horizontal, 50)
+                    .padding(.top)
+                    .padding(.bottom, 45)
+                    .focusGuide(
```
LePips, line 202: This ordering causes the spacing issue. The `focusGuide` should come before the `frame(height: 80)`.

LePips (changes requested):

Swiftfin tvOS/Views/VideoPlayer/Overlays/SmallMenuOverlay.swift
tonyd33: Nice, this looks a lot better now.

simulator_screenshot_1544093D-2A3C-41CB-8AF5-9368A75ACC57

LePips (approved):

lavavex:
This menu seems to have broken in tvos 17 and later, still works fine when run in tvos 15.4 simulator

tonyd33:
> This menu seems to have broken in tvos 17 and later, still works fine when run in tvos 15.4 simulator

Yeah, it seems like tvOS 17+ broke the whole overlay menu stuff in general, even before this commit. I'll see if there are any bandages I can come up with quickly, but we had been discussing revamping the overlay anyhow so this might get pushed until then

PR #874, opened 2023-10-11
Fix quick connect

tonyd33:
# Purpose

Quick connect encountered a race condition where the task to monitor fails before the secret was properly fetched. This meant that even after authorization, nothing was monitoring for the authorization status and you'd have to close the quick connect menu and reopen it to authorize.

Additionally on iOS, after authorization, `QuickConnectView` uninitializes (as it should), but then reinitializes and uninitializes immediately after for some reason. This causes a strange state in `UserSignInViewModel` where an extra monitor task is spawned and keeps repeating forever. 

XCode gives a "Publishing changes from background threads is not allowed..." in `NavigationCoordinatable:600` after authorization, so I'm guessing it's something to do with the `router.dismissCoordinator()` in `QuickConnectView`, but wrapping that with `MainActor.run` didn't fix it.

## Bug showcase

(Poll rate was increased here)

quick_connect_keep_polling

# Fix

- Add `quickConnectState` property to properly block and signal to `checkAuthStatus` when we're ready to poll
  - It now stores the quick connect secret & code since we wait for authorization if and only if we have the secret & code
  - Display these states
- Add `quickConnectMonitorTaskID` property to enforce that only one monitor task should be running at a time, and a for a single source of truth for the task status. 
  - Otherwise, letting the task recurse onto itself over and over with no oversight gets out of hand really easily.
  - Cancelling the task would be nice, but I've had issues with that
- Add `quickConnectMaxRetries` property as a failsafe. In case `UserSignInViewModel` does enter a weird state, we don't want to silently poll for authentication during the entire session.
- Move quick connect stuff to its own section, add docs for the quick connect lifecycle

# Testing

- Check that authorization works after first time entering code
- Checked that polling no longer happens after authorization
- Checked that the new loading & error screens look fine

<details><summary>iOS</summary>
<p>

Simulator Screen Recording - iPhone 14 Pro - 2023-11-09 at 13 31 21

</p>
</details> 

<details><summary>tvOS</summary>
<p>

Simulator Screen Recording - Apple TV - 2023-11-09 at 13 23 46

</p>
</details> 

# Notes

I did try just using `Task.cancel` before the `taskID` approach, but it's been extremely unreliable and unpredictable for me. It didn't stop `checkAuthStatus` from spawning more tasks unless it was cancelled extremely early. I can elaborate if needed.

### DIFF
--- a/Shared/ViewModels/UserSignInViewModel.swift
+++ b/Shared/ViewModels/UserSignInViewModel.swift
@@ -25,9 +25,26 @@ final class UserSignInViewModel: ViewModel {
     let client: JellyfinClient
     let server: SwiftfinStore.State.Server
 
-    private var quickConnectTask: Task<Void, Never>?
-    private var quickConnectTimer: RepeatingTimer?
+    private var quickConnectMonitorTask: Task<Void, Never>?
+    // We want to enforce that only one monitoring task runs at a time, done by this ID.
+    // While cancelling a task is preferable to assigning IDs, cancelling has proven
+    // unreliable and unpredictable.
+    private var quickConnectMonitorTaskID: UUID?
+    // We want to signal to the monitor task when we're ready to poll for authentication.
+    // Without this, we may encounter a race condition where the monitor expects
+    // the secret before it's been fetched, and exits prematurely.
+    private var quickConnectStatus: QuickConnectStatus = .neutral
+    // If, for whatever reason, the monitor task isn't stopped correctly, we don't want
+    // to let it silently run forever.
+    private let quickConnectMaxRetries = 200
     private var quickConnectSecret: String?
+
+    enum QuickConnectStatus {
+        case neutral
+        case fetchingSecret
+        case fetchingSecretFailed
+        case awaitingAuthentication
+    }
 
     init(server: ServerState) {
         self.client = JellyfinClient(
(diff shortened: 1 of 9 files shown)


Shared/ViewModels/UserSignInViewModel.swift
```diff
@@ -73,59 +63,176 @@ final class UserSignInViewModel: ViewModel {
...
+    /// poll task continues to live with this ID.
+    private var quickConnectPollTaskID: UUID?
+    /// If, for whatever reason, the monitor task keeps going, we don't want to let it silently run forever.
+    private let quickConnectMaxRetries = 200
+
+    enum QuickConnectStatus {
+        case fetchingSecret
+        // After secret and code is fetched from the server, store it in the associated value as:
+        //                         (secret, code)
+        case awaitingAuthentication(String, String)
```
LePips, line 142: Just have labels: `case awaitingAuthentication(secret: String, code: String)`

Swiftfin tvOS/Views/UserSignInView.swift
```diff
@@ -146,11 +140,48 @@ struct UserSignInView: View {
             }
             .padding(.vertical)
 
-            Text(viewModel.quickConnectCode ?? "------")
+            Text(quickConnectCode)
```
LePips, line 143: Let's instead make it so that the steps are always on screen and the code area shows the loading/error/code state.

LePips (changes requested):
The backend work looks fine to me, just feels a bit overengineered to try and avoid re-authorizing/polling on that weird second start which I did observe only a few times but works great! Only design changes.

Thanks for the comprehensive comments!

Shared/ViewModels/UserSignInViewModel.swift
```diff
@@ -73,59 +63,176 @@ final class UserSignInViewModel: ViewModel {
...
         }
     }
 
-    private func checkAuthStatus(continuation: AsyncStream<QuickConnectResult>.Continuation) {
-
+    private func checkAuthStatus(continuation: AsyncStream<QuickConnectResult>.Continuation, id: UUID, tries: Int = 0) {
         let task = Task {
-            guard let quickConnectSecret else { return }
+            // Don't race into failure while we're fetching the secret.
+            while case .fetchingSecret = quickConnectStatus {
```
LePips, line 184: Why is this necessary? We should always have the secret before checking auth.

Shared/ViewModels/UserSignInViewModel.swift
```diff
@@ -73,59 +63,176 @@ final class UserSignInViewModel: ViewModel {
...
+
+    private let quickConnectPollTimeoutSeconds: UInt64 = 5
+
+    /// We don't use Timer.scheduledTimer because checking quick connect status is async. We only want to start the next
+    /// poll when the current has finished.
+    private var quickConnectPollTask: Task<Void, Never>?
+    /// When two quick connect tasks are started for whatever reason, cancelling the repeating task above seems to fail
+    /// and the attempted cancelled task seems to continue to spawn more repeating tasks. We ensure only a single
+    /// poll task continues to live with this ID.
+    private var quickConnectPollTaskID: UUID?
```
LePips, line 134: Actually having thought about this a bit more, remove this and add another `QuickConnectStatus.authorized` and don't start quick connect again if authorized. We aren't running into the issue that we are starting quick connect during auth checks and if we are, something is instead wrong with the views. If we ever run into the problem where we want to start quick connect again while authorized, we can deal with that later?

That's just a bit cleaner and we don't have to worry about tracking an ID, and instead we can cancel the `quickConnectPollTask` when stopping. In the task we can check if the task was canceled before we poll again.

LePips (changes requested):

Shared/ViewModels/UserSignInViewModel.swift
tonyd33: This was actually what was causing the bug: we _didn't_ have the secret before `checkAuthStatus`, so the poll loop would immediately terminate. A typical failure case would look like:
1. Call `startQuickConnect`
2. Initiate fetching secret
3. Call `checkAuthStatus`, which terminates immediately because secret isn't fetched

tonyd33:
> The backend work looks fine to me, just feels a bit overengineered to try and avoid re-authorizing/polling on that weird second start which I did observe only a few times but works great! Only design changes.

I agree it was overengineered. At the time, I wanted to preserve the existing interfaces as much as possible, but I didn't know them well, which lead to that solution. I addressed all your comments, but I ended up just rewriting the quick connect flow because I felt like there was an easier and simpler way to do it. Sorry for making you do a review from (mostly) scratch again. 

Here's a quick summary of key changes, apart from your comments:
- Polling is done in a much more sane way now – just read `pollForAuthSecret`. 
  - No more gimmicks with ID's, max retries, and cancelling the task works properly now
  - Polling can only be initiated when we have the initial secret
- No `AsyncStreams` anymore – from what I understand, this isn't a good use case of it. Just use `async` functions instead
- `startQuickConnect` was replaced with `signInWithQuickConnect` which gets the secrets and signs in automatically while updating state for views using `quickConnectStatus`

Shared/ViewModels/UserSignInViewModel.swift
```diff
@@ -73,6 +63,110 @@ final class UserSignInViewModel: ViewModel {
...
+    /// 5. Authentication poll succeeds with another secret, use secret to log in
+
+    @Published
+    private(set) var quickConnectEnabled = false
+    /// To maintain logic within this view model, we expose this property to track the status during quick connect execution.
+    @Published
+    private(set) var quickConnectStatus: QuickConnectStatus?
+
+    /// How often to poll quick connect auth
+    private let quickConnectPollTimeoutSeconds: UInt64 = 5
```
LePips, line 126: Just use `Int`

tonyd33:
The `Stateful` framework is a pretty natural extension of what I wrote, so I'm all for it.

LePips:
Hey! Is there interest to keep working on this? If not, no worries. I'll be doing a refactor of the server/user handling/flows soon and would overlap with this work.

Since I've been doing some work in the SDK for the 10.9.0 release the thought of having this in the SDK has been on my mind. I would very much like a quick connect manager in the SDK and we use it in here. It may take a different form than what's implemented here for general client use.

I may take this over if this isn't worked on soon.

LePips:
Actually, I may have been a little too hasty in thought since I've now taken a look at everything done here. Since this requires all the view work as well and I want to start on my work fairly soon, I'll accept this and will be able to work on top of it.

LePips (approved):

LePips:
I seem to have done a lot with `Stateful` since this work was last done and would have had probably more problems down the road, so I actually am glad that I got to this now.

Thank you for the contribution!

PR #878, opened 2023-10-13
Localization strings

chickdan:
I went through all of the views in the Swiftfin folder (not tvOS or Shared) and added strings to the localization file for translation once they are available. Luckily some translations are already done and filled in a few gaps.

### DIFF
--- a/Swiftfin/Views/DownloadTaskView/DownloadTaskContentView.swift
+++ b/Swiftfin/Views/DownloadTaskView/DownloadTaskContentView.swift
@@ -116,7 +116,7 @@ extension DownloadTaskView {
                     Text(L10n.dismiss)
                 }
             } message: {
-                Text(L10n.downlodsOnlyPlayableInSwiftfin)
+                Text("Downloaded items are only playable through the Swiftfin video player.")
             }
         }
     }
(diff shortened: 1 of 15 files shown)


Swiftfin/Views/DownloadTaskView/DownloadTaskContentView.swift
```diff
@@ -113,10 +113,10 @@ extension DownloadTaskView {
                 Button {
                     isPresentingVideoPlayerTypeError = false
                 } label: {
-                    Text("Dismiss")
+                    Text(L10n.dismiss)
                 }
             } message: {
-                Text("Downloaded items are only playable through the Swiftfin video player.")
+                Text(L10n.downlodsOnlyPlayableInSwiftfin)
```
LePips, line 119: I actually don't want download-related stuff localized right now, as I still see it very much a work in progress and don't want them publicized.

LePips (changes requested):

Swiftfin/Views/DownloadTaskView/DownloadTaskContentView.swift
chickdan: Understood, I'll get those reverted.

LePips (approved):

PR #881, opened 2023-10-13
iOS: Increase library list item size

ericswpark:
The current library list view seems a bit crowded:

Screenshot 2023-10-13 at 12 24 08 PM

With the increased size, we have three columns with a bit of padding between the items:

Screenshot 2023-10-13 at 12 25 12 PM

Unfortunately testing on a smaller screen (13 mini simulator) results in overlap with no padding. I'm not sure how to fix this, so leaving this PR as a draft for now:

Screenshot 2023-10-13 at 12 26 02 PM

### DIFF
--- a/Swiftfin/Components/LibraryItemRow.swift
+++ b/Swiftfin/Components/LibraryItemRow.swift
@@ -17,13 +17,15 @@ struct LibraryItemRow: View {
     private let item: BaseItemDto
     private var onSelect: () -> Void
 
+    private let posterWidth: CGFloat = 60
+
     var body: some View {
         Button {
             onSelect()
         } label: {
             HStack(alignment: .bottom) {
-                ImageView(item.portraitPosterImageSource(maxWidth: 60))
-                    .posterStyle(type: .portrait, width: 60)
+                ImageView(item.portraitPosterImageSource(maxWidth: posterWidth))
+                    .posterStyle(type: .portrait, width: posterWidth)
                     .posterShadow()
 
                 VStack(alignment: .leading) {
--- a/Swiftfin/Components/PagingLibraryView.swift
+++ b/Swiftfin/Components/PagingLibraryView.swift
@@ -27,7 +27,7 @@ struct PagingLibraryView: View {
         if libraryGridPosterType == .landscape && UIDevice.isPhone {
             return .fixedNumberOfColumns(2)
         } else {
-            return .adaptive(withMinItemSize: libraryGridPosterType.width + (UIDevice.isIPad ? 10 : 0))
+            return .adaptive(withMinItemSize: libraryGridPosterType.width + 10)
         }
     }
 
--- a/Swiftfin/Components/PagingLibraryView.swift
+++ b/Swiftfin/Components/PagingLibraryView.swift
@@ -23,11 +23,13 @@ struct PagingLibraryView: View {
 
     private var onSelect: (BaseItemDto) -> Void
 
+    private let portraitPosterScale = 1.25
+
     private var gridLayout: NSCollectionLayoutSection.GridLayoutMode {
         if libraryGridPosterType == .landscape && UIDevice.isPhone {
             return .fixedNumberOfColumns(2)
         } else {
-            return .adaptive(withMinItemSize: libraryGridPosterType.width + 10)
+            return .adaptive(withMinItemSize: libraryGridPosterType.width * portraitPosterScale + 10)
         }
     }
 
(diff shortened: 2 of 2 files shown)


tonyd33:
`PagingLibraryView.gridLayout` returns `.adaptive(withMinItemSize: libraryGridPosterType.width + (UIDevice.isIPad ? 10 : 0))`, which binds the min size to exactly the poster width. Giving some padding to both iPad and iOS should work.

On a side note, I'm not a main contributor, but these views seem like they warrant a SwiftUI rework anyway so this might get overridden when it happens

ericswpark:
@tonyd33 thank you, that was it! Now it shows in a two-column layout on the 13 mini:

Screenshot 2023-10-13 at 3 53 26 PM

Regarding the rework I don't really mind if this change gets overwritten, just something in the interim so that it looks a bit nicer.

LePips:
This is a fairly complicated and annoying issue (with SwiftUI), so I'll give a fairly descriptive overview.

---

The library view is just a wrapper around `UICollectionView`, which is required because `LazyVStack + ScrollView` isn't a solution for large collections of items because it doesn't implement the idea of dequeuing views. So, I made my own basic wrapper that attempts to find the size that fits the given by the SwiftUI view, which I referenced from other SwiftUI `UICollectionView` wrappers:

- https://github.com/LePips/CollectionView/blob/main/Sources/CollectionView/CollectionViewCell.swift

The overlap is caused as the collection view hasn't yet determined the size required as the view/cell is becoming rendered, which can happen if you scroll fast.

At the same time, I was pretty ignorant on how `UICollectionViewLayout` worked so I went with something that "just worked". I've since then worked a bit on `UICollectionView`s and have become comfortable with the layout system. The `PosterButton`'s [width is ultimately done staticly](https://github.com/jellyfin/Swiftfin/blob/043a95de00126306567b97f9c1cbfb6a40519888/Swiftfin/Components/PosterButton.swift#L93), and can further be changed [through scaling the item](https://github.com/jellyfin/Swiftfin/blob/main/Swiftfin/Components/PagingLibraryView.swift#L65). This causes the "spacing" issues between screen devices, because the widths are the same across devices and the underlying `UICollectionView` spaces them out because there is space to do so.

Frankly, I now see this as a dumb implementation, which was implemented by yours truly.

---

A solution we could implement "right now" is to [change the layout](https://github.com/jellyfin/Swiftfin/blob/main/Swiftfin/Components/PagingLibraryView.swift#L71) to `CollectionView` and either scale the `PosterButton` per the detected width of the device. We can just do this "right now" because it uses the existing `CollectionView`, but I am also unsure whether it would actually fix the overlap issue, I don't believe it would. This also would require figuring out the sizes manually and maintaining it which I find annoying.

The proper solution is to rewrite the library view (and frankly some other views...) into a `UICollectionView` in the project. This would frankly increase performance and fix these layout issues. While it is annoying that we would be reworking some things into UIKit away from SwiftUI, SwiftUI just doesn't have what we need and this is a general consensus among other iOS developers I know.

---

I am fine for having this "in the interim" as you say, but yes these views will be rewritten at some point in the future.

Shared/Objects/PosterType.swift
```diff
@@ -40,7 +40,7 @@ enum PosterType: String, CaseIterable, Displayable, Defaults.Serializable {
 
         static let landscape = 350.0
         #else
-        static var portrait = 100.0
+        static var portrait = 120.0
```
LePips, line 43: This changes the portrait poster size _everywhere_ and we don't want that. Instead, you'll have to do some logic with the poster type and scale with the `scaleItem` modifier:

https://github.com/jellyfin/Swiftfin/blob/18b06ff2972ad61d3da34cb112d0eceaf301d91c/Swiftfin/Components/PagingLibraryView.swift#L65

LePips (changes requested):

Shared/Objects/PosterType.swift
ericswpark: Reworked to use scaling with the new commit!

ericswpark:
@LePips I've modified the commit so that it now applies the scaling like you mentioned. I had to refactor out the magic number while changing it because not applying it to the `withMinItemSize` caused the overlap issue again on smaller screens.

I kind of wanted to move the entire ternary statement out to a variable, but this didn't work:

```
private let posterScale = libraryGridPosterType == .landscape && UIDevice.isPhone ? 0.85 : 1.25
```

I'm pretty sure there is a way to set a variable like this so I can reference it in both `gridLayout` and `libraryGridView`, but again, not very good with Swift. Please let me know if there's anything else I need to change or add!

LePips (approved):
As stated, I'm fine with this for now.

PR #883, opened 2023-10-16
Remove iOS `PosterButtonType` + cleanup

LePips:
This removes my terrible, no good, very bad `PosterButtonType` when I attempted to make a mechanism for loading and no results in contexts of a `PosterButton`. This just isn't the way to go.

Other stuff that I just picked up along the way:
- cleans up `posterStyle(...)` into a single method and the `frame` is determined by the caller. This adds a weird zoom effect to the `SeriesEpisodeSelector` but that will be changed to use `UICollectionView` so it will be dealt with.
- TODO comments
- change the multiple of `padding` (`padding2`)
- add `RatioCornerRadiusModifier` so that corner radii are set as a ratio of the view's size


Shared/Extensions/ViewExtensions/ViewExtensions.swift
```diff
@@ -50,57 +50,32 @@ extension View {
...
     func padding2(_ edges: Edge.Set = .all) -> some View {
         padding(edges).padding(edges)
     }
 
+    /// Applies the default system padding a number of times with a multiplier
+    func padding(multiplier: Int, _ edges: Edge.Set = .all) -> some View {
+        precondition(multiplier > 0, "Multiplier must be > 0")
+
+        return modifier(PaddingMultiplierModifier(edges: edges, multiplier: multiplier))
+    }
```
PangMo5, line 77: Are you doing this to use the default padding length provided by each OS?

Swiftfin/Views/ItemView/iOS/ScrollViews/CompactPortraitScrollView.swift
```diff
@@ -173,8 +173,8 @@ extension ItemView.CompactPosterScrollView {
                     // MARK: Portrait Image
 
                     ImageView(viewModel.item.imageSource(.primary, maxWidth: 130))
-                        .posterStyle(type: .portrait, width: 130)
-                        .accessibilityIgnoresInvertColors()
+//                        .posterStyle(type: .portrait, width: 130)
```
PangMo5, line 176: Is this an intended comment?

PangMo5 (approved):

Swiftfin/Views/ItemView/iOS/ScrollViews/CompactPortraitScrollView.swift
LePips: No, thank you for catching that 👍

PangMo5 (approved):

PR #887, opened 2023-10-21
show episodes instead of series in recently added view

elliots:
Previously it listed the recently added *series* which is pretty unuseful i think. If anyone wants that, they're still listed in the "Latest X" sections

image

### DIFF
--- a/Swiftfin tvOS/Views/HomeView/HomeContentView.swift
+++ b/Swiftfin tvOS/Views/HomeView/HomeContentView.swift
@@ -24,7 +24,7 @@ extension HomeView {
 
                     if viewModel.resumeItems.isEmpty {
                         CinematicRecentlyAddedView(viewModel: .init(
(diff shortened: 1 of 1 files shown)


elliots:
It would be nice if it combined multiple episodes of the same series together, like plex does.

LePips:
It's actually a conscious decision to make it the series and not episodes, as not only is it consistent with the web but also in the case new _entire_ series or seasons are added the episodes will populate individually.

The "better" UX would be to go to the first unwatched item when a series is selected, at least by my brief thought.

elliots:
The web version orders the shows by the latest episode to be released.

Swiftfin is ordering by the latest TV shows to be added. So, currently when a new episode is downloaded, there's no where you can see that happen.

elliots:
To make it just like the web version, I could change the "Latest X" sections, instead of the "Recently Added" (which I don't think is on web)

PR #894, opened 2023-10-24
Don't build non-existent image URLs

### LINKED ISSUE #884: Only get item images if they exist

Currently, anywhere we need an image from an item we request one even if the image doesn't exist, making a lot of unnecessary `404`s. We can tell whether the image exists on `BaseItemDto.imageTags` beforehand, which the web does.

Originally noticed in #690.

chickdan:
I believe this fixes #884 as an image URL will only be returned if there are image tags, and without a URL LazyImage should result in the failure case. I don't have a Jellyfin server setup so I could be wrong with my changes so guidance would be appreciated.

### DIFF
--- a/Shared/Extensions/JellyfinAPI/BaseItemDto+Images.swift
+++ b/Shared/Extensions/JellyfinAPI/BaseItemDto+Images.swift
@@ -105,6 +105,17 @@ extension BaseItemDto {
         return client.fullURL(with: request)
     }
 
+    private func getImageTag(for type: ImageType) -> String? {
+        switch type {
+        case .backdrop:
+            return backdropImageTags?.first
+        case .screenshot:
+            return screenshotImageTags?.first
+        default:
+            return imageTags?[type.rawValue]
+        }
+    }
+
     fileprivate func _imageSource(_ type: ImageType, maxWidth: Int?, maxHeight: Int?) -> ImageSource {
         let url = _imageURL(type, maxWidth: maxWidth, maxHeight: maxHeight, itemID: id ?? "")
         let blurHash = blurHash(type)
--- a/Shared/Extensions/JellyfinAPI/BaseItemDto+Images.swift
+++ b/Shared/Extensions/JellyfinAPI/BaseItemDto+Images.swift
@@ -19,7 +19,7 @@ extension BaseItemDto {
         _ type: ImageType,
         maxWidth: Int? = nil,
         maxHeight: Int? = nil
-    ) -> URL {
+    ) -> URL? {
         _imageURL(type, maxWidth: maxWidth, maxHeight: maxHeight, itemID: id ?? "")
     }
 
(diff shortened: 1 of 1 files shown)


Shared/Extensions/JellyfinAPI/BaseItemDto+Images.swift
```diff
@@ -80,11 +80,13 @@ extension BaseItemDto {
...
         itemID: String
-    ) -> URL {
+    ) -> URL? {
+        guard let imageTags = imageTags else { return nil }
+        
         // TODO: See if the scaling is actually right so that it isn't so big
         let scaleWidth = maxWidth == nil ? nil : UIScreen.main.scale(maxWidth!)
         let scaleHeight = maxHeight == nil ? nil : UIScreen.main.scale(maxHeight!)
-        let tag = imageTags?[type.rawValue]
+        let tag = imageTags[type.rawValue]
```
LePips, line 89: Checking the `imageTags` is the first step, the next is checking if this tag has a value and if not also `return nil`

LePips (changes requested):

Shared/Extensions/JellyfinAPI/BaseItemDto+Images.swift
LePips: Additionally, we will have to make sure the fields for the tags are requested along with the items.

LePips:
So currently, that fix would actually break backdrop images. I took a look again at these fields because I forgot they're split, `imageTags` and `backdropImageTags` to say a few. Backdrop image tags aren't in `imageTags`, so this will always fail for them. We will have to switch on the type of image being requested and retrieve + check in the right place. In the case for backdrops, we will always take the first (primary) image. This has always worked before because calling for an item backdrop image without an index/tag will just return the primary image.

chickdan:
> We will have to switch on the type of image being requested and retrieve + check in the right place.

Added a new function that switches on `ImageType`, most of the types do not have corresponding <>ImageTag fields so I default to `imageTags` as that's also how the behavior was previously.

> Backdrop image tags aren't in `imageTags`, so this will always fail for them... In the case for backdrops, we will always take the first (primary) image.

For this, I return `backdropImageTags.first` from the new `getImageTag()` function.

> This has always worked before because calling for an item backdrop image without an index/tag will just return the primary image.

Considering the existing function always worked with a `nil` tag as long as the type was `.backdrop` I added a check to allow the request to be made.

Hopefully, I understood the requirements 🙂 let me know if further adjustments are required.

Shared/Extensions/JellyfinAPI/BaseItemDto+Images.swift
```diff
@@ -80,11 +80,16 @@ extension BaseItemDto {
...
-    ) -> URL {
+    ) -> URL? {
+
         // TODO: See if the scaling is actually right so that it isn't so big
         let scaleWidth = maxWidth == nil ? nil : UIScreen.main.scale(maxWidth!)
         let scaleHeight = maxHeight == nil ? nil : UIScreen.main.scale(maxHeight!)
-        let tag = imageTags?[type.rawValue]
+        let tag = getImageTag(for: type)
+
+        if tag == nil && type != .backdrop {
```
LePips, line 90: Can remove this and just have:
```swift
guard let tag else { return nil }
```

We don't need the check for backdrop images, if there aren't any then they won't populate the image tags. I only mentioned the "lack of index returning the primary image" as a lucky coincidence with the API.

LePips (changes requested):

LePips (approved):

PR #895, opened 2023-10-25
update LatestInLibraryView to use getLatestMedia

elliots:
This now matches what the web player does.

image

Before this change, it displayed the latest *shows* added.

image

### DIFF
--- a/Shared/ViewModels/LatestInLibraryViewModel.swift
+++ b/Shared/ViewModels/LatestInLibraryViewModel.swift
@@ -1,0 +1,68 @@
+//
+// Swiftfin is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2023 Jellyfin & Jellyfin Contributors
+//
+
+import Combine
+import Foundation
+import JellyfinAPI
+
+final class LatestInLibraryViewModel: PagingLibraryViewModel {
+    
+    let parent: LibraryParent
+
+    init(parent: LibraryParent) {
+        self.parent = parent
+        
+        super.init()
+
+        _requestNextPage()
+    }
+
+    override func _requestNextPage() {
+        Task {
+
+            await MainActor.run {
+                self.isLoading = true
+            }
+
+            let parameters = Paths.GetLatestMediaParameters(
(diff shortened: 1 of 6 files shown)


elliots:
I've removed the trailing `SeeAllPosterButton`, as it wasn't happy with the new `LatestInLibraryViewModel`. But I'm sure I can work that out.

elliots:
The `getLatestMedia` doesn't appear to support paging, so I'm grabbing 3x the number so the trailing "Show All" button has a bit more in it. Not very nice. Is there a better way?

LePips:
This breaks the "random" selection. I admit I didn't take a good look at how the parameter creation was refactored when that work was done but it should be redone.

elliots:
I've fixed the linting issues, and implemented `getRandomItemFromLibrary`. The implementation grabs a random item from the lists and wraps it in `BaseItemDtoQueryResult`. Also not brilliant but i think better than getting a random one from the whole library. 

Is using `JellyfinAPIError` if it hasn't been loaded yet ok? 

I couldn't see where the toolbar with the `RandomItemButton` is displayed, so I haven't actually tested it yet.

Shared/ViewModels/LatestInLibraryViewModel.swift
```diff
@@ -0,0 +1,75 @@
...
+
+            await MainActor.run {
+                self.isLoading = false
+                self.items.append(contentsOf: items)
+            }
+        }
+    }
+
+    override public func getRandomItemFromLibrary() async throws -> BaseItemDtoQueryResult {
+        if items.isEmpty {
```
LePips, line 55: Sadly I let this function pass as is, since it should just return the item, but this just needs:
```swift
BaseItemDtoQueryResult(items: items.elements)
```

since right now the button handles choosing a random item. I will change that.

LePips (changes requested):

Shared/ViewModels/LatestInLibraryViewModel.swift
elliots: Done...

LePips: Looks like you still need to lint.

elliots: Done

LePips (approved):

PR #907, opened 2023-11-10
Use `jump` with native double tap gesture instead of single tap gesture in Video Player 

xiifain:
The problem with the current video player is that it is using single tap gesture with a custom timeout for `jump` action. This has multiple disadvantages

1. The overlay screen will always show in the first tap
2. Since it's not a native double tap gesture, I've experienced a scenario that this PR will address.

Scenario
1. first, tap on the player screen ( this will always shows the overlay )
2. since the overlay is shown, I want to hide it by tapping again.
3. The player recognizes the second tap within the timeout period and causes the player to perform `jump` instead.

What this PR will address
1. The first tap on the player will not show the overlay anymore. (the overlay will be shown only if the double tap gesture is not performed).
2. Will address the behavior scenario mentioned above.

### DIFF
--- a/Shared/Objects/GestureAction.swift
+++ b/Shared/Objects/GestureAction.swift
@@ -29,27 +29,13 @@ enum LongPressAction: String, GestureAction {
     }
 }
 
-enum MultiTapAction: String, GestureAction {
-
-    case none
-    case jump
-
-    var displayTitle: String {
-        switch self {
-        case .none:
-            return L10n.none
-        case .jump:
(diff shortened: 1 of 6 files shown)


LePips:
This is actually behavior I don't want and have addressed before: https://github.com/jellyfin/Swiftfin/pull/412#discussion_r854444133. In short, I like the way Youtube does it.

Instead, we can look at modifying the timeout to detect a jump action.

PR #911, opened 2023-11-11
Make Individual Settings Resettable

LePips:
TODO

- [ ] Figure out what to do about selector-type settings
- [ ] How would this interact with better user management?


LePips:
Gonna refactor settings a bit in the near future and this would either be refactored or would be a huge pita to merge.

PR #920, opened 2023-11-18
Fix transcode URL construction bug

ShadowJonathan:
I found this while mucking around the code; When the url is `/video/` etc., it gets incorrectly added onto the current server URL.

I have no idea why, but the way below seems to work properly, and after I added this method, it seems to work correctly on my Apple TV HD.

Otherwise, the transcode URL will have no host and no scheme, causing the player to exit immediately without error.

### DIFF
--- a/Shared/Extensions/JellyfinAPI/MediaSourceInfo+ItemVideoPlayerViewModel.swift
+++ b/Shared/Extensions/JellyfinAPI/MediaSourceInfo+ItemVideoPlayerViewModel.swift
@@ -21,9 +21,7 @@ extension MediaSourceInfo {
         let streamType: StreamType
 
         if let transcodingURL, !Defaults[.Experimental.forceDirectPlay] {
-            guard let fullTranscodeURL = URL(string: transcodingURL, relativeTo: userSession.server.currentURL)
-            else { throw JellyfinAPIError("Unable to construct transcoded url") }
-            playbackURL = fullTranscodeURL
+            playbackURL = userSession.client.fullURL(with: transcodingURL)
             streamType = .transcode
         } else {
 


holow29:
Can you explain more about what you mean? The current implementation should result in userSession.server.currentURL with the /video/... transcodeURL appended (though it looks different in the console/log). The one you suggest would result in userSession.client.fullURL with the /video/... transcodeURL appended. I think they should be the same, and I see no difference in my testing; I haven't had issues since #760.

ShadowJonathan:
I have a server url that goes like `http://flyingdutchman.fritz.box:8096/jellyfin`, and it resulted in the URL becoming `/video/...`, I have no idea why, but the new approach works with it, it seems.

holow29:
> I have a server url that goes like `http://flyingdutchman.fritz.box:8096/jellyfin`, and it resulted in the URL becoming `/video/...`, I have no idea why, but the new approach works with it, it seems.

Are you saying it becomes something like `http://flyingdutchman.fritz.box:8096/video/...` instead of `http://flyingdutchman.fritz.box:8096/jellyfin/video/...` ?

If not and you look at the playbackURL or fullTranscodeURL you might see it start with `/video/...`, but this is just how this URL builder displays it. It has the server information listed afterwards and AFAIK should be parsed the same way as .fullURL. I would be interested if this isn't working for you, though.

ShadowJonathan:
I have no idea why, but it is specifically `/video`, not `http://flyingdutchman.fritz.box:8096/video/`, and certainly not `http://flyingdutchman.fritz.box:8096/jellyfin/video/`, which is supposed to be the correct one

holow29:
How are you checking the playbackURL? The URL object should look like:
`/videos/etc -- http://flyingdutchman.fritz.box:8096/jellyfin`
The `-- http://flyingdutchman.fritz.box:8096/jellyfin` at the end of this object is what makes it equivalent to the fullURL string. It is simply in a different format, but AFAIK should be parsed the same way.

ShadowJonathan:
I don't know, but;
- I saw this URL (`/video/`) pop out in the debugger for the `playbackURL`, it caused the video player to exit immidiately
  - I did this by removing all directplayback profiles in another part of the code, if you'd like to test
- Adding in this piece of code seemed to fix it

Semantically they're meant to be equivalent, but apparently they're not

I don't know what `URL` does, but all i know is that this variant works for my setup, and the older one didn't

LePips:
I didn't read the thread, but to clear up some confusion the transcode URL for videos is of the format:

`/video/<path>?<parameters>`

as seen here: https://github.com/jellyfin/jellyfin/blob/000ccaa6dbf948aa427a8daac7d550dd598b3b5b/MediaBrowser.Model/Dlna/StreamInfo.cs#L681-L686

The current implementation is incorrect, tbh I'm not too familiar with URL-relativity and it works different than I read it online and as implemented but 🤷 and this makes me think that I used to be conscious about servers at subdirectories but must have forgotten about it.

For the `fullURL` function, we should actually change its contents as well. It assumes that a URL won't end with `/`, which a server technically can - subdirectory or not. Luckily we can make force the assumption that paths into the function start with a `/`, which our generated request objects do as well. We _could_ make a more robust path-appending function but I'm fine with forcing the assumption with the well formed URLs/paths that we use.

Here is a fiddle showcasing the different methods, how they interact with different forms of URLs, and my preferred solution: https://swiftfiddle.com/xpvkl7c5mneidejcqbrghgkqnq

LePips (changes requested):
Forgot to request changes based on: https://github.com/jellyfin/Swiftfin/pull/920#issuecomment-1841707148

holow29:
> The current implementation is incorrect, so this needed to be done, and tbh I'm not too familiar with URL-relativity and it works different than I read it online and as implemented but 🤷 and this makes me think that I used to be conscious about servers at subdirectories but must have forgotten about it.

Ah, my test setup does not have subdirectory structure like this, and I was unaware that relativeTo would strip it. Thank you for correcting me. This explains why it worked for me and not ShadowJonathan.

> For the `fullURL` function, we should actually change its contents as well.

Is there a difference or preference between userSession.server.currentURL and userSession.client.fullURL that you know of?

LePips:
Use the URL from the session's client. They ultimately will (should) be set to the same URL at all times but it just feels more correct to make sure that all networking (at least when you have a session client) uses the same source of truth.

LePips:
Taken over by #990, for sake of pushing this fix forward.

PR #933, opened 2023-12-12
[tvOS] Fix play/pause button of siri remote

sy6sy2:
Restore the Siri Remote play/pause button handling in main.

### DIFF
--- a/Swiftfin tvOS/Views/VideoPlayer/Overlays/Overlay.swift
+++ b/Swiftfin tvOS/Views/VideoPlayer/Overlays/Overlay.swift
@@ -26,6 +26,8 @@ extension VideoPlayer {
         private var confirmCloseWorkItem: DispatchWorkItem?
         @State
         private var currentOverlayType: VideoPlayer.OverlayType = .main
+        @State
+        private var isPlaying = true
 
         @StateObject
         private var overlayTimer: TimerProxy = .init()
@@ -69,6 +71,15 @@ extension VideoPlayer {
                     isPresentingOverlay = true
                     overlayTimer.start(5)
                 }
+                .onPlayPauseCommand(perform: {
+                    if isPlaying {
+                        proxy.pause()
+                        isPlaying = false
+                    } else {
+                        proxy.play()
+                        isPlaying = true
+                    }
+                })
                 .onMenuPressed {
 
                     overlayTimer.start(5)


LePips (changes requested):
This actually requires a lot more state handling.

1 - Don't make a new state handler, the player state is held in the environment `VideoPlayerManager` object. The behavior should be the same as the play buttons based on the player state for iOS.
2 - We need to set the state of the overlays. When the play/pause button is pressed, the main overlay should be presented if it isn't and poke the overlay timer.

Swiftfin tvOS/Views/VideoPlayer/Overlays/Overlay.swift
```diff
@@ -26,6 +26,8 @@ extension VideoPlayer {
         private var confirmCloseWorkItem: DispatchWorkItem?
         @State
         private var currentOverlayType: VideoPlayer.OverlayType = .main
+        @State
+        private var isPlaying = true
```
jhays, line 30:
```suggestion
@EnvironmentObject
private var videoPlayerManager: VideoPlayerManager
```

Swiftfin tvOS/Views/VideoPlayer/Overlays/Overlay.swift
```diff
@@ -69,6 +71,15 @@ extension VideoPlayer {
...
                     overlayTimer.start(5)
                 }
+                .onPlayPauseCommand(perform: {
+                    if isPlaying {
+                        proxy.pause()
+                        isPlaying = false
+                    } else {
+                        proxy.play()
+                        isPlaying = true
+                    }
```
jhays, line 81:
```suggestion
                    if !isPresentingOverlay {
                        isPresentingOverlay = true
                        overlayTimer.start(5)
                    }
                    switch videoPlayerManager.state {
                    case .playing:
                        proxy.pause()
                    default:
                        proxy.play()
                    }
```

Swiftfin tvOS/Views/VideoPlayer/Overlays/Overlay.swift
jhays: Probably best to place this `@EnvironmentObject` above with the other environment objects, its just that I couldn't find a way to use GitHub's suggestion feature on lines that have not been changed already.

Swiftfin tvOS/Views/VideoPlayer/Overlays/Overlay.swift
jhays: @LePips does this solve your change request?

Swiftfin tvOS/Views/VideoPlayer/Overlays/Overlay.swift
LePips: Yes, ti should go with the other `EnvironmentObject`s.

Swiftfin tvOS/Views/VideoPlayer/Overlays/Overlay.swift
LePips: Not just seeing if the overlay is being presented but also set it to the main overlay.

LePips:
This will be overridden by #1066 with the change to not use the SwiftUI press handlers and use `PreferencesView` instead.

See why necessary:
- https://github.com/jellyfin/Swiftfin/pull/1066#discussion_r1614035275

PR #942, opened 2023-12-29
Swiftformat on build

chickdan:
IMO basic formatting/linting rules should be forced at an earlier stage than CI. Adding a build phase forces SwiftFormat to run on every build so developers should be getting auto-corrected code much sooner than seeing a failed build on CI. It also removes the cognitive overhead of remembering to run the formatter before committing.

### DIFF
--- a/Swiftfin.xcodeproj/project.pbxproj
+++ b/Swiftfin.xcodeproj/project.pbxproj
@@ -3044,6 +3044,7 @@
 		};
 		BD0BA2282AD64BB200306A8D /* Run Swiftformat */ = {
 			isa = PBXShellScriptBuildPhase;
+			alwaysOutOfDate = 1;
 			buildActionMask = 2147483647;
 			files = (
 			);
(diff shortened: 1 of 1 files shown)


LePips:
I'd have to think about this more, I'm more for it being a pre-commit hook than code changing while building - mostly because when fighting Xcode with frustrating panic clean + build or other scenarios I don't necessarily like my code looking different. Yeah a small gripe but 🤷‍♂️

chickdan:
For what it's worth, the projects where I work have been set up with the build phase option for 6 years and it has not been an issue. I also set up my personal projects the same way (well, for JavaScript projects I opt for format on **save**).

Using a hook would require every contributor to change the hooks path for the local git instance while in the working directory: `git config --local core.hooksPath .githooks/`

Let me know what you think, I can update accordingly!

LePips:
Yeah I'm okay with it. If you could also make the changes on the tvOS side that'd be great! Also thanks for renaming the `R.swift` file, must have been a relic if/when this project used it.

PR #947, opened 2024-01-09
[tvOS][QOL] Supports back button click when in the native player.

### LINKED ISSUE #925: [ATV] Cannot exit playing session with native player
### Describe the bug

1. Login/setup Swiftfin \w Jellyfin server
2. Go to settings, toggle "player" to "native"
3. Go to any media, play
4. Press "return", see that it doesn't work, except the main-menu long-press return

### Application version

Latest main, eb17300061315ff310749871f5b79032385ba55a

None

### Device information

Apple TV HD

### OS version

tvOS 17.1

### Jellyfin server version

10.8.12

nokeeo:
Conforms `UINativeVideoPlayerViewController` to `AVPlayerViewControllerDelegate` and implements `playerViewControllerShouldDismiss`.  This method is called by AV player view controller when the player should be dismissed.  This function is only called when the playback overlay is not displayed otherwise a back button press will dismiss the overlay.

Fixes #925

### DIFF
--- a/Shared/Coordinators/VideoPlayerCoordinator.swift
+++ b/Shared/Coordinators/VideoPlayerCoordinator.swift
@@ -45,7 +45,7 @@ final class VideoPlayerCoordinator: NavigationCoordinatable {
             AppDelegate.enterPlaybackOrientation()
         }
 
-        #else // os(iOS)
+        #else
         if Defaults[.VideoPlayer.videoPlayerType] == .swiftfin {
             PreferenceUIHostingControllerView {
                 Group {
@@ -56,6 +56,6 @@ final class VideoPlayerCoordinator: NavigationCoordinatable {
         } else {
             NativeVideoPlayer(manager: self.videoPlayerManager)
         }
-        #endif // os(iOS)
+        #endif 
     }
 }
--- a/Shared/Coordinators/VideoPlayerCoordinator.swift
+++ b/Shared/Coordinators/VideoPlayerCoordinator.swift
@@ -3,7 +3,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, you can obtain one at https://mozilla.org/MPL/2.0/.
 //
-// Copyright (c) 2023 Jellyfin & Jellyfin Contributors
+// Copyright (c) 2024 Jellyfin & Jellyfin Contributors
 //
 
 import Defaults
@@ -45,19 +45,17 @@ final class VideoPlayerCoordinator: NavigationCoordinatable {
             AppDelegate.enterPlaybackOrientation()
         }
 
-        #else
-
-        PreferenceUIHostingControllerView {
-            Group {
-                if Defaults[.VideoPlayer.videoPlayerType] == .swiftfin {
+        #else // os(iOS)
+        if Defaults[.VideoPlayer.videoPlayerType] == .swiftfin {
+            PreferenceUIHostingControllerView {
+                Group {
                     VideoPlayer(manager: self.videoPlayerManager)
-                } else {
-                    NativeVideoPlayer(manager: self.videoPlayerManager)
                 }
             }
+            .ignoresSafeArea()
+        } else {
+            NativeVideoPlayer(manager: self.videoPlayerManager)
         }
-        .ignoresSafeArea()
-
-        #endif
+        #endif // os(iOS)
     }
 }
--- a/Shared/Coordinators/VideoPlayerCoordinator.swift
+++ b/Shared/Coordinators/VideoPlayerCoordinator.swift
@@ -48,14 +48,12 @@ final class VideoPlayerCoordinator: NavigationCoordinatable {
         #else
         if Defaults[.VideoPlayer.videoPlayerType] == .swiftfin {
             PreferenceUIHostingControllerView {
-                Group {
-                    VideoPlayer(manager: self.videoPlayerManager)
-                }
+                VideoPlayer(manager: self.videoPlayerManager)
             }
             .ignoresSafeArea()
         } else {
             NativeVideoPlayer(manager: self.videoPlayerManager)
         }
-        #endif 
+        #endif
     }
 }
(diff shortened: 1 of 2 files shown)


nokeeo:
ping @LePips for review or please assign to whomever is appropriate.

Swiftfin tvOS/Views/VideoPlayer/NativeVideoPlayer.swift
```diff
@@ -57,7 +56,9 @@ struct NativeVideoPlayerView: UIViewControllerRepresentable {
     func updateUIViewController(_ uiViewController: UINativeVideoPlayerViewController, context: Context) {}
 }
 
-class UINativeVideoPlayerViewController: AVPlayerViewController {
+// TODO: Refactor such that this does not subclass AVPlayerViewController. Subclassing is not
+// supported according to the apple docs.
+class UINativeVideoPlayerViewController: AVPlayerViewController, AVPlayerViewControllerDelegate {
```
LePips, line 61: Thank you for your contribution, however this isn't correct. We don't need to subclass from anything for this to work. In `VideoPlayerCoordinator` we create the view like so:

```swift
PreferenceUIHostingControllerView {
    Group {
        if Defaults[.VideoPlayer.videoPlayerType] == .swiftfin {
            VideoPlayer(manager: self.videoPlayerManager)
        } else {
            NativeVideoPlayer(manager: self.videoPlayerManager)
        }
    }
}
.ignoresSafeArea()
```

where the tvOS implementation of `PreferenceUIHostingControllerView` is a view that overrides button presses that we can detect in the Swiftfin video player. As we see above, the native player is wrapped in this view but we don't want it to be. This should be changed to the following, where the native player can handle all of its own button presses:

```swift
if Defaults[.VideoPlayer.videoPlayerType] == .swiftfin {
    PreferenceUIHostingControllerView {
        VideoPlayer(manager: self.videoPlayerManager)
    }
    .ignoresSafeArea()
} else {
    NativeVideoPlayer(manager: self.videoPlayerManager)
}
```

Luckily I was going to work on PRs and other work today, but please don't ping me in the future for reviews. I do see everything, but only get to stuff when I can.

LePips (changes requested):

Swiftfin tvOS/Views/VideoPlayer/NativeVideoPlayer.swift
nokeeo: >  please don't ping me in the future for reviews

Apologies.  It wasn't clear how to get a review and noticed that in an accepted PR someone pinged a maintainer.  Won't do in the future.

Incorporated these suggestions.  PTAL.

Shared/Coordinators/VideoPlayerCoordinator.swift
```diff
@@ -45,19 +45,17 @@ final class VideoPlayerCoordinator: NavigationCoordinatable {
             AppDelegate.enterPlaybackOrientation()
         }
 
-        #else
-
-        PreferenceUIHostingControllerView {
-            Group {
-                if Defaults[.VideoPlayer.videoPlayerType] == .swiftfin {
+        #else // os(iOS)
```
LePips, line 48: Don't need these comments

Shared/Coordinators/VideoPlayerCoordinator.swift
```diff
@@ -45,19 +45,17 @@ final class VideoPlayerCoordinator: NavigationCoordinatable {
...
                 }
             }
+            .ignoresSafeArea()
+        } else {
+            NativeVideoPlayer(manager: self.videoPlayerManager)
         }
-        .ignoresSafeArea()
-
-        #endif
+        #endif // os(iOS)
```
LePips, line 59: and here

Swiftfin tvOS/Views/VideoPlayer/NativeVideoPlayer.swift
```diff
@@ -37,7 +37,6 @@ struct NativeVideoPlayer: View {
             if let _ = videoPlayerManager.currentViewModel {
                 playerView
             } else {
-//                VideoPlayer.LoadingView()
```
LePips, line 40: Actually leave this, this is just left over from forever ago for when this needs to be properly implemented.

LePips (changes requested):
Just some cleanup, good to have the refactor comment about subclassing from `AVPlayerViewController`. Yeah it says subclassing isn't supported but we can generally get away/lazy with it, but we should wrap that into a parent view/VC that interacts with the underlying `AVPlayerViewController`.

Swiftfin tvOS/Views/VideoPlayer/NativeVideoPlayer.swift
nokeeo: Added a TODO instead of leaving commented out code to make the intention a little bit more explicit.

Shared/Coordinators/VideoPlayerCoordinator.swift
nokeeo: I personally like this style as I find it makes it easier to visually parse preprocessor expressions especially if they are nested, but reverted this change.

Shared/Coordinators/VideoPlayerCoordinator.swift
```diff
@@ -46,18 +46,16 @@ final class VideoPlayerCoordinator: NavigationCoordinatable {
         }
 
         #else
-
-        PreferenceUIHostingControllerView {
-            Group {
-                if Defaults[.VideoPlayer.videoPlayerType] == .swiftfin {
+        if Defaults[.VideoPlayer.videoPlayerType] == .swiftfin {
+            PreferenceUIHostingControllerView {
+                Group {
```
LePips, line 51: Since you have to lint anyways, this `Group` can be removed.

Swiftfin tvOS/Views/VideoPlayer/NativeVideoPlayer.swift
```diff
@@ -37,7 +37,7 @@ struct NativeVideoPlayer: View {
...
+                // TODO: Implement more polished loading screen.
```
LePips, line 40: To me, the commented out view is better because that's the actual view that needs to be implemented. (see the file/view)

LePips (changes requested):

nokeeo:
Done.  PTAL.

LePips (approved):

PR #949, opened 2024-01-12
[tvOS][Bug] Fixes stale home content.

### LINKED ISSUE #904: Refresh main page
A main page refresh on closing the watched series and a way to refresh it manually.

**Additional context**
It would be great if the app would refetch the main page when stopping to watch a series and when the screen is pulled down on the main page.  When you start to watch a series on your device. Stop and start again the app will pick up where you last started the app. When you did something like add new movies they won't show up without restarting the app. A way to fix that would be to add a reload feature. I thought of something like this: 
[Image](https://images.macrumors.com/t/unQL9NQx5FxQ0sHXWf-0DFaVN9I=/1600x0/article-new/2021/09/ios-15-pull-to-refresh-how-to.jpg) I think it’s this [Stackoverflow](https://stackoverflow.com/questions/24475792/how-to-use-pull-to-refresh-in-swift#24476087)

nokeeo:
When the tab bar controller appears on the screen the home view's content is refreshed.  This refreshes the content when it might be stale, like the user returning to the main page from the player.

Attaches an action on the tab bar coordinator's view.  Exposes the home coordinator's view model to refresh from the tab coordinator.

An alternative is to attach the onAppear to `HomeView` directly, however this action is fired on tab switch.  At the moment the content does not update on tab switch.

Fixes #904

### DIFF
--- a/Shared/Objects/AppNotifications.swift
+++ b/Shared/Objects/AppNotifications.swift
@@ -1,0 +1,14 @@
+//
+// Swiftfin is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2024 Jellyfin & Jellyfin Contributors
+//
+
+import Foundation
+
+enum AppNotifications {
+    // Posted when the now playing data has likely become stale.
+    static let nowPlayingDidChange = Notification.Name(rawValue: "nowPlayingDidChange")
+}
--- a/Shared/ViewModels/HomeViewModel.swift
+++ b/Shared/ViewModels/HomeViewModel.swift
@@ -3,7 +3,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, you can obtain one at https://mozilla.org/MPL/2.0/.
 //
-// Copyright (c) 2023 Jellyfin & Jellyfin Contributors
+// Copyright (c) 2024 Jellyfin & Jellyfin Contributors
 //
 
 import Combine
@@ -12,7 +12,6 @@ import Factory
 import Foundation
 import JellyfinAPI
 import UIKit
-
 final class HomeViewModel: ViewModel {
 
     @Published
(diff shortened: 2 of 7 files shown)


LePips (changes requested):
Apologies, but this is not correct. We should not handle refreshing in the coordinators, only in the views. Please change to an implementation that does that.

nokeeo:
Updated.  After using this for a few days it was annoying after returning to the home view when browsing my library, for it to update and lose my spot.

Changed it so the home model subscribes to an invalidation notification.  I imagine there might be other cases where we want to invalidate, like returning from the background.

PTAL.

Shared/Objects/AppNotifications.swift
```diff
@@ -0,0 +1,14 @@
...
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2024 Jellyfin & Jellyfin Contributors
+//
+
+import Foundation
+
+enum AppNotifications {
+    // Posted when the now playing data has likely become stale.
+    static let nowPlayingDidChange = Notification.Name(rawValue: "nowPlayingDidChange")
```
LePips, line 13: Don't create a new object to hold notification definitions, that's already in `SwiftfinNotifications.swift`.

Shared/ViewModels/HomeViewModel.swift
```diff
@@ -26,15 +25,39 @@ final class HomeViewModel: ViewModel {
     @Published
     var resumeItems: [BaseItemDto] = []
 
+    /// Indicates if the view the model is backing is visible. Setting this value may trigger a refresh.
+    var isVisible = false {
```
LePips, line 29: A view model should never hold a value whether a view is presented or not.

Swiftfin tvOS/Views/HomeView/HomeView.swift
```diff
@@ -34,5 +34,11 @@ struct HomeView: View {
         }
         .edgesIgnoringSafeArea(.top)
         .edgesIgnoringSafeArea(.horizontal)
+        .onAppear(perform: {
```
LePips, line 37: Remove based on comment in the view model. Additionally, this is attached to a `Group` which would require more testing when changing states since I think the `onAppear/onDisappear` would be in an undeterministic with all of the views.

LePips (changes requested):
I was going to review this further, however this is still incorrect. I've always needed to fix how the state of the view model is handled and we should wait until then to implement this.

nokeeo:
> I've always needed to fix how the state of the view model is handled and we should wait until then to implement this.

@LePips I have some free time.   Happy to help with this if given direction.

PR #905, opened 2023-11-09
Refactor `PosterButton` and good `UICollectionView`s

LePips:
wip

PR #950, opened 2024-01-13
[tvOS][Bug] Fixes season selector only showing a single season.

nokeeo:
Upon initial view load, only the first season is loaded into `menuSections` when the episode list is resolved.

This change stores the list of all seasons, which contain their display name.  When changing the series selector, `menuSections` is updated with season details.

Also when changing seasons in the picker, the containing scroll view scrolled to the top of the page.  This was jarring.  This change also removes that behavior.

### DIFF
--- a/Shared/ViewModels/ItemViewModel/SeriesItemViewModel.swift
+++ b/Shared/ViewModels/ItemViewModel/SeriesItemViewModel.swift
@@ -3,7 +3,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, you can obtain one at https://mozilla.org/MPL/2.0/.
 //
-// Copyright (c) 2023 Jellyfin & Jellyfin Contributors
+// Copyright (c) 2024 Jellyfin & Jellyfin Contributors
 //
 
 import Combine
@@ -20,9 +20,13 @@ final class SeriesItemViewModel: ItemViewModel, MenuPosterHStackModel {
     var menuSections: [BaseItemDto: [BaseItemDto]]
     var menuSectionSort: (BaseItemDto, BaseItemDto) -> Bool
 
+    /// A list of seasons in this series. Resolved once. Populated when `menuSelection` is published.
+    var seasons: [BaseItemDto]
+
     override init(item: BaseItemDto) {
         self.menuSections = [:]
         self.menuSectionSort = { i, j in i.indexNumber ?? -1 < j.indexNumber ?? -1 }
+        self.seasons = []
 
         super.init(item: item)
 
@@ -136,6 +140,7 @@ final class SeriesItemViewModel: ItemViewModel, MenuPosterHStackModel {
             if let firstSeason = seasons.first {
                 self.getEpisodesForSeason(firstSeason)
                 await MainActor.run {
+                    self.seasons = seasons
                     self.menuSelection = firstSeason
                 }
             }
--- a/Shared/ViewModels/ItemViewModel/SeriesItemViewModel.swift
+++ b/Shared/ViewModels/ItemViewModel/SeriesItemViewModel.swift
@@ -117,7 +117,7 @@ final class SeriesItemViewModel: ItemViewModel, MenuPosterHStackModel {
     func select(section: BaseItemDto) {
         self.menuSelection = section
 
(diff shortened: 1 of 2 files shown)


LePips (changes requested):
I seem to have broken this in: https://github.com/jellyfin/Swiftfin/pull/883/files#diff-1462c1160e5edfcfdbf96fa2622aeaead21bbc319987b8bfb6ace9cd32e66382L145

We shouldn't create a new variable to hold what seasons are there, that's what keys for the `menuSections` dictionary are. For now, what we should do is mimic the work that was done before where we instead set an empty array to each season before we get the episodes for the first season and when a season is selected and if it is empty, get the episodes for the season.

I have done work in #905 that lays the foundation for an eventual refactor of this entire view model so that episodes can be lazily loaded, mainly for the case of large seasons (eg: FMAB is one season with ~60 episodes).

nokeeo:
Thanks for the review!  PTAL.

Shared/ViewModels/ItemViewModel/SeriesItemViewModel.swift
```diff
@@ -131,11 +131,13 @@ final class SeriesItemViewModel: ItemViewModel, MenuPosterHStackModel {
             let request = Paths.getSeasons(seriesID: item.id!, parameters: parameters)
             let response = try await userSession.client.send(request)
 
-            guard let seasons = response.value.items else { return }
-
+            let seasons = response.value.items ?? []
```
LePips, line 134: Please use the `guard`.

Shared/ViewModels/ItemViewModel/SeriesItemViewModel.swift
```diff
@@ -131,11 +131,13 @@ final class SeriesItemViewModel: ItemViewModel, MenuPosterHStackModel {
...
             if let firstSeason = seasons.first {
                 self.getEpisodesForSeason(firstSeason)
                 await MainActor.run {
+                    for season in seasons {
```
LePips, line 138: Move this instead to right below checking whether we have seasons, like in: https://github.com/jellyfin/Swiftfin/pull/883/files#diff-1462c1160e5edfcfdbf96fa2622aeaead21bbc319987b8bfb6ace9cd32e66382L143

Shared/ViewModels/ItemViewModel/SeriesItemViewModel.swift
```diff
@@ -117,7 +117,7 @@ final class SeriesItemViewModel: ItemViewModel, MenuPosterHStackModel {
     func select(section: BaseItemDto) {
         self.menuSelection = section
 
-        if !menuSections.keys.contains(section) {
+        if let seasonData = menuSections[section], seasonData.isEmpty {
```
LePips, line 120:
```suggestion
        if let episodes = menuSections[section], episodes.isEmpty {
```

LePips (changes requested):

Shared/ViewModels/ItemViewModel/SeriesItemViewModel.swift
nokeeo: I am not super familiar with MainActor, but if implemented as before doesn't this task wait for the dispatch to the main thread to execute before continuing execution?  If so this wouldn't that introduce latency.

nokeeo:
Done.  PTAL.

Shared/ViewModels/ItemViewModel/SeriesItemViewModel.swift
LePips: That little latency is fine. In fact, `getEpisodesForSeason` should be migrated to `async` so we would have to `await` that before setting the seasons at all with this<sup>1</sup>. Instead, we would want a loading state while retrieving the current season's episodes. This is what I did (very badly) before #883 and in #905 (along with the foundation for lazily loading) this will be done again.

In fact _again_ we should actually probably create the empty-state dictionary and then set it so that the view updates only once in response to the `Publisher` emitting a change but 🤷 

---

1 - The basic `Task` wrapped function bodies were just to make the old `Publisher` -> `async/await` transition feasible.

LePips (approved):

PR #955, opened 2024-01-25
Enable Swift Symbol Generation for Assets Catalogs

chickdan:
This setting was introduced in Xcode 15 and allows for Swift symbols to be generated for Color and Image asset catalogs so that they can be validated by the compiler. Changing an asset name will result in a compiler error rather than a runtime crash 🙂

There is an asset named `git.commit` that doesn't appear to be used, is that safe to remove or are there plans to use it?


chickdan:
CI runs on Xcode 14.2 causing the build to fail. I'll open a PR to update the CI pipelines.

LePips:
As a note, the runners now use Xcode 15 by default.

chickdan:
> As a note, the runners now use Xcode 15 by default.

I was still getting errors on my fork after they switched the default and I haven't had time to dig into it. I'll hopefully get back to this soon 🙂

LePips (approved):

PR #957, opened 2024-01-26
Stateful POC

LePips:
POC for refactoring view models into proper stateful objects.

Note:
- Finish #905 first since that refactors libraries. While trying to refactor `HomeViewModel`, changed how items are kept on that object and that would conflict.


LePips:
Replaced by #905

PR #964, opened 2024-02-04
Minor Fix for Search View and Poster Button Text

tstraus13:
Fixes the Search View by adding padding/margin to the results. Before the search "keyboard' would overlap the results but now the results will appear below the keyboard. I am guessing the "keyboard" is not really a item in the view/stack from what I could tell and is a built in element that will overlap items if not accounted for. If there is a better way to handle this feel free to correct it but this was a fairly simple fix. It uses a percentage (65%) of the screen geometry to place the results, so in theory it should work on any screen but I tested in the simulator with 4k and 1080p and it seems to work alright. Maybe on more vertical spaced screens it might be off but I am thinking the keyboard will always take up approximately the same amount of space. Maybe the percentage could be based on screen size ratios?

I also added padding to the poster buttons text/title. When you would hover an item it would slightly cover the text/title. Now it should always be completely visible.


tstraus13:
After looking at the code once more and looking up what .ignoresSafeArea() does. I removed it from the code and it does what I was trying to do from the begginning. So I removed my code changes there and just removed the ignoresSafeArea call. Maybe it was placed there by mistake.

Swiftfin tvOS/Components/PosterButton.swift
```diff
@@ -149,6 +149,7 @@ extension PosterButton {
                         .font(.footnote)
                         .fontWeight(.regular)
                         .foregroundColor(.primary)
+                        .padding(.top, 1)
```
LePips, line 152: I know that it's annoying that the text is slightly covered when focused, but this doesn't really help either and we shouldn't add this. What I ideally want is something like [TVPosterView](https://developer.apple.com/documentation/tvuikit/tvposterview) from `TVUIKit`. These views can't be wrapped and I had attempted to make a recreation however I ran into horrendous performance issues with `FocusState`. However, I wonder if my work in #905 would make a recreation viable again.

Apple has left tvOS + SwiftUI in the dust - it's very frustrating.

Swiftfin tvOS/Views/SearchView.swift
```diff
@@ -45,7 +45,6 @@ struct SearchView: View {
                 }
             }
         }
-        .ignoresSafeArea()
```
LePips, line 48: We still need to ignore the safe area on the bottom, left, and right edges.

Re-add: `.ignoresSafeArea(edges: [.bottom, .horizontal])`

LePips (changes requested):

Swiftfin tvOS/Components/PosterButton.swift
tstraus13: So why exactly does adding padding not help? From my testing it does allow for the focused item to not cover the text without any major side effect besides the text being a little lower that before. Is that the issue? Maybe it's not a final solution but it fixes the issue and is easily removed when a better solution is found. Thanks.

Swiftfin tvOS/Views/SearchView.swift
tstraus13: Fair enough. I can add that back in. For my enlightenment why do we need it on the bottom and horizontal (left/right side)?

Swiftfin tvOS/Components/PosterButton.swift
LePips: The issue is the total design, it makes additional padding that doesn't look nice. Sure it fixes the button from overlapping the text as I said the ideal solution is a design that accommodates both focused and non-focused states.

Swiftfin tvOS/Views/SearchView.swift
LePips: We only need to ignore the safe area on the top, where the keyboard/tab bar are located. By not ignoring the horizontal and bottom sides, there will be padding from the safe area.

<details>
<summary>Not ignoring safe area sides</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) (at 1080p) - 2024-02-14 at 17 03 09

</details>

<details>
<summary>Ignoring safe area sides</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) (at 1080p) - 2024-02-14 at 17 02 52

</details>

tstraus13: I see. I don't remember seeing that padding in my testing but agree with from your screenshots it makes sense to ignore. I have added it back in for the horizontal (left/right) and bottom. Thanks.

Swiftfin tvOS/Components/PosterButton.swift
tstraus13: I think it looks nice but agree that a design that accommodates both focused and unfocused would be best. I have removed that code.

tstraus13:
Looks like the build failed. The only thing I see is the line "error: value of type '_UnsafeBitset' has no member 'isEmpty'" during the build process.

tstraus13:
Also it looks like the PR is stating that 1 changes is still requested but I think I "resolved" both changes. If there is something else your are looking for let me know. Otherwise I think its good to go. Thanks.

tstraus13:
I realized on the Search View I placed the ignoreSafeArea on the wrong thing. It was on the VStack when originally it was on the View.

LePips (approved):
This may be annoying, but try pulling in `main` since I have updated the packages and I think something changed with the runner images that I haven't paid attention to.

tstraus13:
Will do. I'll give it a shot today at some point.

tstraus13:
I have merged from upstream master. There were just two commits missing. Should be good to go?

PR #967, opened 2024-02-08
Minor UI Update for ConnectToServer View

alasclar:
Adds a picker before the text field to select 'http://' or 'https://', meaning default 'http://' text is not needed.
Original method just seemed a bit clunky when using the app, however there may be some reason why something other than http/s needs to be entered that I don't know about, so please let me know.

https://github.com/jellyfin/Swiftfin/assets/156718015/f5fa8dc9-0771-4929-824b-87a7f60d0efc

This is my first PR, so please let me know if I have done anything incorrectly. Thanks

### DIFF
--- a/Swiftfin/Views/ConnectToServerView.swift
+++ b/Swiftfin/Views/ConnectToServerView.swift
@@ -33,7 +33,13 @@ struct ConnectToServerView: View {
     @State
     private var isPresentingError: Bool = false
     @State
-    private var url = "http://"
+    private var url = ""
+    @State
+    private var urlPrefix = "http://"
+    @State
+    private var urlPostfix = ""
+
+    private let prefixOptions = ["http://", "https://"]
 
     private func connectToServer() {
         let task = Task {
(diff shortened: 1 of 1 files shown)


LePips:
I had thought about this a long time ago but I would have gone with a different implementation. However, asking people to make sure they input the right URL is not a big ask at all, it's networking 101. Instead, I would rather just remove the `http://` and by default use `http` if not provided and use the `https` redirect if properly configured.

alasclar:
@LePips 
Fair enough, could I get your opinion on something similar to what you described?

If the user doesn't specify a port, and no server is found, it could be good to automatically look for a server at the 8096 port of the same URL?

Had a couple times where I just forgot to add the port where this would've saved some time. I think this might be a feature on another jellyfin client but I'm not sure.

I understand we're assuming competency from the user but that kinda functionality would be a nice to have.
Thanks

LePips:
No, I don't think that's necessarily a good assumption to make. We shouldn't be going against the basic `http` convention on port 80 and any other port should be manually specified.

alasclar:
Understood. However (upon checking), this is how the regular Jellyfin app handles it.



I think using that app on an android device is what made me start dropping the port number in Swiftfin.

PR #977, opened 2024-02-18
Automatically hides home bar

erfansamandarian:


erfansamandarian:
Uhhh, I think the App Store version of this app is outdated, I got too excited when I saw my movies without the home bar when I made this change. Deploying this on my device, the home bar hides with or without this. I think this was confirmation bias. Sorry.

PR #980, opened 2024-02-22
Create Library Alpha Picker

JPKribs:
This feature creates a new filter in the ItemFilters for AlphaPicker. The goal is to (eventually) mirror the functionality of the [Jellyfin-Web AlphaPicker](https://github.com/jellyfin/jellyfin-web/tree/master/src/components/alphaPicker) which filters the library based on the letter selected. This PR focuses on the creation of the Filter as to split out the underlying Filtering components that exist within the existing ItemFilter / FilterDrawer objects. 

Letters for the filter are pulled from the localized language on the device. The filter sets the API Parameter nameStartsWith to the letter selected or uses nameLessThan: 'A' for symbols. This reflects the behavior seen in other implementations of this feature. To assist in this, I have an AlphaPicker class that returns the appropriate nameStartsWith and nameLessThan parameters based on the AlphaPicker.selection.

When testing, I found that the current unstable/development branch of Swiftfin has an issue where the Genre & Filter filters were only referenced in the _getDefaultParameters call. This means that when you updated a genre filter, it would not be reflected until you next hit the requestItems() call. This issue did not exist with the SortBy and SortOrder parameters since these are written during the requestItems() call. To resolve this, I moved the filter objects outside of the defaultParameters since 1) the default call for parameters doesn't need to be filtered and these parameters are being created and set to nil and 2) this resolves the issue and the filters now apply directly after being set. Let me know if there is a better way to handle this!

If this PR is merged, I will work on adding the AlphaPickerBar to the LibraryView to mirror Roku/Web.

**Potential Concern:** https://github.com/jellyfin/jellyfin-roku/pull/1716
- The Roku App uses the same nameLessThan: 'A' for filtering on symbols but there are some symbols that come after Z alphabetically. The Roku App only uses #,A-Z so this works for their implementation since it will only ever be Latin Letters & Symbols. The filter I created takes the localized alphabet from the device. As a result, non-Latin Characters will exist as a selection and anything nameGreaterThan: 'Z' will also include these characters.
- Roku & Jellyfin-Web both only allow Latin Letters. I do not know if the API works with non-Latin letters.

_This is a rewrite of something I started lasted year. I chose to take some time to learn Swift better and hopefully rewrite this feature with a little more direction. https://github.com/jellyfin/Swiftfin/pull/852_


LePips:
I have completely refactored how filters are implemented in #905. I'm almost done and adding this work should be easier to implement, so I ask that this work be redone once that's in, apologies.

> The filter I created takes the localized alphabet from the device.

This is the extent of what we can do given that the server doesn't give a list of section titles. However, that would be a great addition to collection folder/folder `BaseItemDto`s.

> Roku & Jellyfin-Web both only allow Latin Letters. I do not know if the API works with non-Latin letters.

By my n=1 testing of playing with a movie with Korean characters, it does. So as far as I can tell, we have full UTF-8 support for names.

LePips:
#905 has now been merged, so this feature can continue. We don't need to make another class for holding the letters, we only need to add another `ItemFilterType` and hopefully everything should statically make sure that the case it handled correctly.

Jpkribs:
First and foremost, #905 resolves my only pain point I've ran into with Swiftfin! The change in the library scrolling is awesome. Literally 0 hitching even on remote loading. It looks like a lot of hard work went into making this and it really shows. I'm excited for when this makes its way to the App Store!

Regarding this PR, I'm going to take a look at this tonight/this weekend. I might just nuke my branch and start over with the changes. PR#1 I'll add the filter itself then, with the library changes, I think I need to go back to the drawing board for the AlphaPicker itself. I personally like the idea of mirroring other clients/web with the letters down the side but I'd need to figure out how to avoid cramping into the library. I'll ping once I have something worth looking at!

LePips:
Any interest in continuing this work? If not, I can implement the letter filter and leave the library picker for later work.

Jpkribs:
Hey, sorry work has caught up to me as of late. I had a chance to review your new filters. It looks a lot cleaner/intuitive but when I took a crack at it I ran into some trouble understanding how to populate the filter values using the section titles (#, A-Z). 

I want to say I'll be able to knock this out soon but soon could mean early May just with how things are going as of late...

I don't want to hold anything back so if this is holding anything back I wouldn't be offended if you did it. Otherwise, I can work on getting this wrapped up soon (May) from my end so I'm not adding to your plate for Swiftfin. Your call!

LePips:
No worries at all. I just happen to have a lot of time on my hands to get things done and think this would be a great addition. I've implemented the letter filter in the PR below and I would anticipate the picker to be quick when you get to it. You are free to continue this PR or close it.

- https://github.com/jellyfin/Swiftfin/pull/1024

Jpkribs:
That sounds great! Thanks for doing that. I think the work that I have for the letter bar should be pretty easy to port over to a new filter once the filter exists. I was just having trouble trying to understand how to populate a filter with a static list of letters opposed to an enum/returned api list.

When the letter filter exists, it should only take me an hour or so to point the letter bar to that instead of the old version I was working with. The only quirk I am unsure about is how it will interact with the new library/list view. In particular, with how you are determining the size of posters and how much available screen space there is. But, we can get there when we get there. I'll keep an eye out for the new PR you made I will update this branch, and see about implementing the letter bar. From there, there might be a point where we need to have a discussion about spacing, but I can ping you when I get to that point!

Jpkribs:
@LePips Thank you for setting up the filter! Getting this migrated to the new filter was a cakewalk. Please see the filter below set to the Letter X. Re-selecting X disables the filter.

simulator_screenshot_559191DE-EA04-4DC0-92B4-1CCDE91F9CA4

My next step is I want to re-add a setting to enable, disable, and choose Left vs Right side of the screen. That shouldn't take too long but I wanted to make sure that:

1. PagingLibraryView was the right place to put this
2. Does that spacing look okay with this AlphaPicker there?
3. For the setting, Letter Bar an okay name for it? AlphaPicker? I'm terrible with naming so whatever sounds best to you.

Jpkribs:
With this last push, I think I have all of the items I am looking for but there are some rough parts that I want to address before pushing.

### Completed Items:

<details>
  <summary>Created an Alpha Picker Setting</summary>
  
simulator_screenshot_AC9DC1B9-52C2-4D05-976E-D09B916337C8

</details>

<details>
  <summary>Created New Labels for these Elements (Alpha Picker **Title** so if that changes we can just update the title)</summary>
  
Screenshot 2024-04-19 at 9 20 30 PM

</details>

<details>
  <summary>Functioning Filters from Alpha Picker Selection</summary>
  
simulator_screenshot_66E1DA6D-BD46-4738-83B5-11EA604272E5

</details>

<details>
  <summary>Flip from Static View to ScrollView if there are more than 27 Characters</summary>
  
simulator_screenshot_1FB2A528-C0ED-4873-9E99-C65B7117C08A

</details>


### Outstanding Considerations

1. The flip from View to ScrollView for Non-English looks a tad cramped?
2. Is the Paging Library View the best place for this to be? Is the alphaPickerWrapperView() a solution to this or does that just add noise?
3. I removed the .ignoresSafeArea() on the Paging Library View to keep the AlphaPicker with Non-English characters in view. Let me know if that will cause any issues or if there is a better implementation.
4. The setting is Left, Right, None with the Default to Right to mirror Jellyfin-Web. The backend names for these are .leading, .trailing, and .none. Let me know if that is the right way to use that or if the label should match the case.

Let me know!

Swiftfin.xcodeproj/project.pbxproj
```diff
@@ -1490,6 +1496,16 @@
...
 			sourceTree = "<group>";
 		};
+		4EE6535F2BD3456900064B5E /* AlphaPicker */ = {
+			isa = PBXGroup;
+			children = (
+				4EE6535D2BD3456900064B5E /* AlphaPickerBar.swift */,
+				4EE6535E2BD3456900064B5E /* AlphaPickerButton.swift */,
+			);
+			name = AlphaPicker;
+			path = ../../../../../Desktop/AlphaPicker;
```
LePips, line 1506: I was just about to take a look at this but the actual work isn't checked in and lives somewhere on your machine.

At the same time, let's just keep this in `Swiftfin/Components` and probably rename to `LetterPickerBar` just to keep the name consistent with the filter. I can't remember if I said earlier "rename to alpha" so I apologize if we're back to square one.

I would like it in the following structure:
```
Swiftfin
- /Components
  - /LetterPickerBar
    - /Components
      - LetterPickerButton.swift
    - LetterPickerBar.swift
```

LePips (changes requested):

Swiftfin.xcodeproj/project.pbxproj
JPKribs: Sorry about the delay on this! I have some time this weekend where I'll make the changes you were talking about. Do we want the name in code to be LetterPickerBar? I only ask because AlphaPicker seems to be the terminology used for both Web and Roku clients. I agree that the name that the end-user sees should probably be Letter Picker but I wasn't sure if we wanted everything under the hood to somewhat mirror what we were seeing on other clients? I personally have no preference so whatever you think is best!

Overall, the number of changes here is pretty small. I'll pull everything new down and copy it over. This time, actually **in** the PR... haha

LePips: Yeah I'd like `LetterPickerBar`. I'm doing my own namings independent of what other clients do.

JPKribs: This should now all exist in the PR. Sorry about that. I've changed all the naming over to LetterPicker / LetterPickerBar.

Original items still apply that I mentioned here: https://github.com/jellyfin/Swiftfin/pull/980#issuecomment-2067532843

The only new item is that last time I added the LetterPicker next to the LibraryView but I wasn't seeing that anymore in iOS? Feel free to let me know if this should be located somewhere else. It's working in testing added to PagingLibraryView but let me know if that should be somewhere else instead.

Swiftfin/Components/LetterPickerBar/LetterPickerBar.swift
```diff
@@ -0,0 +1,74 @@
...
+
+    @ViewBuilder
+    private var letterPickerBody: some View {
+        VStack(spacing: 0) {
+            ForEach(ItemLetter.allCases, id: \.self) { filterLetter in
+                LetterPickerButton(
+                    filterLetter: filterLetter.value,
+                    activated: viewModel.currentFilters.letter.contains { $0.value == filterLetter.value },
+                    viewModel: viewModel
+                )
```
LePips, line 44: As said in another comment, instead use the environment `isSelected`:

`.environment(\.isSelected, viewModel.currentFilters.letter.contains(filterLetter)`

Swiftfin/Components/LetterPickerBar/LetterPickerBar.swift
```diff
@@ -0,0 +1,74 @@
...
+//
+
+import Defaults
+import SwiftUI
+
+struct LetterPickerBar: View {
+
+    @ObservedObject
+    var viewModel: FilterViewModel
+    private var onSelect: (FilterCoordinator.Parameters) -> Void
```
LePips, line 16: Not used, can remove.

Swiftfin/Components/LetterPickerBar/LetterPickerBar.swift
```diff
@@ -0,0 +1,74 @@
...
+            case .trailing:
+                return L10n.right
+            }
+        }
+    }
+
+    @ViewBuilder
+    private var letterPickerBody: some View {
+        VStack(spacing: 0) {
+            ForEach(ItemLetter.allCases, id: \.self) { filterLetter in
```
LePips, line 39:
```suggestion
            ForEach(ItemLetter.allCases, id: \.hashValue) { filterLetter in
```

Swiftfin/Components/LetterPickerBar/LetterPickerBar.swift
```diff
@@ -0,0 +1,74 @@
...
+import Defaults
+import SwiftUI
+
+struct LetterPickerBar: View {
+
+    @ObservedObject
+    var viewModel: FilterViewModel
+    private var onSelect: (FilterCoordinator.Parameters) -> Void
+
+    enum Orientation: String, CaseIterable, Defaults.Serializable, Displayable {
```
LePips, line 18: This was replaced by `LetterPickerOrientation` and can be removed

Shared/Components/LetterPickerOrientation.swift
```diff
@@ -0,0 +1,28 @@
...
+//
+// Copyright (c) 2024 Jellyfin & Jellyfin Contributors
+//
+
+import Defaults
+import SwiftUI
+
+enum LetterPickerOrientation: String, CaseIterable, Defaults.Serializable, Displayable {
+
+    case none
```
LePips, line 14: Instead of `none`, for this I'd like there to be a `LetterPickerEnabled` bool option and this is just the orientation.

I would like the default to be `false`.

Swiftfin/Components/LetterPickerBar/Components/LetterPickerButton.swift
```diff
@@ -0,0 +1,67 @@
...
+                )
+                .font(.headline)
+                .frame(width: 15, height: 15)
+                .foregroundColor(activated ? Color.white : accentColor)
+                .padding(.vertical, 2)
+                .fixedSize(horizontal: false, vertical: true)
+                .background {
+                    RoundedRectangle(cornerRadius: 5)
+                        .frame(width: 20, height: 20)
+                        .foregroundColor(activated ? accentColor.opacity(0.5) : Color.clear)
```
LePips, line 43:
```suggestion
                        .foregroundStyle(activated ? accentColor.opacity(0.5) : Color.clear)
```

LePips (changes requested):
This looks great! I've only been able to test on a single device in the simulator and after some cleanup should be good to implement. I will come back to it after I do some refactoring on my side with `scroll(ifLargerThan:)` (or a different modifier) for many characters and may change the spacing/sizes a bit when I get time to tinker.

Shared/Components/LetterPickerOrientation.swift
JPKribs: How should I lay this out? Does this look like what you were expecting or should this be more of a cascading menu like the Library/Search Filters below it? Disregard the duplicate labels, this would be Letter Picker and Orientation respectively.

simulator_screenshot_A3D1648F-DF3D-423E-9C8A-463C39DC8140

Swiftfin/Views/PagingLibraryView/PagingLibraryView.swift
```diff
@@ -251,6 +253,35 @@ struct PagingLibraryView<Element: Poster>: View {
         .proxy(collectionVGridProxy)
     }
 
+    private func contentLetterBarView(content: some View) -> some View {
```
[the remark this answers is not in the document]
JPKribs: This has been 'changed' but I won't be offended if you end up cleaning this up from your end. I split this out a little more than I would like to get rid of a "TypeChecker" error. From the TODO, it looks like that was something others have seen as well but I'm sure there is a more logical way to split out the functions.

Swiftfin/Views/PagingLibraryView/PagingLibraryView.swift
```diff
@@ -260,18 +291,18 @@ struct PagingLibraryView<Element: Poster>: View {
...
-                errorView(with: error)
+                contentLetterBarView(content: errorView(with: error))
             case .initial, .refreshing:
-                DelayedProgressView()
+                contentLetterBarView(content: DelayedProgressView())
             }
         }
         .animation(.linear(duration: 0.1), value: viewModel.state)
-        .ignoresSafeArea()
+        // .ignoresSafeArea()
```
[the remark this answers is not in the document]
JPKribs: This currently only impacts non-romantic languages (Korean, Japanese, etc.) where I change the Letter Picker from a static VStack to a Scroll. I think this may lead into https://github.com/jellyfin/Swiftfin/pull/980#discussion_r1614395534

I'm having a bit of trouble getting this to work. I have a version that works where I move the .ignoresSafeArea() to each of the views that would be contained in the contentLetterBarView(). I don't think that is what we want to do.

What would be my best approach to make this work in a way that doesn't get locked like this when there are more than 27 characters:

<details>
  <summary>Korean Letter Picker</summary>
Screenshot 2024-05-25 at 1 01 16 PM
</details>

Swiftfin/Components/LetterPickerBar/LetterPickerBar.swift
JPKribs: Just the onSelect or both the onSelect AND the viewModel? I have the onSelect removed but let me know if that viewModel shouldn't be there. I have it as part of my init so I would need a bit of a rework.

Swiftfin/Components/LetterPickerBar/Components/LetterPickerButton.swift
```diff
@@ -0,0 +1,67 @@
...
+import SwiftUI
+
+extension LetterPickerBar {
+
+    struct LetterPickerButton: View {
+
+        @Default(.accentColor)
+        private var accentColor
+
+        private let filterLetter: String
```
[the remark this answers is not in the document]
JPKribs: I made this ItemLetter. I assume that's what you meant but I wanted to just note that in case that's something on my end.

Swiftfin/Components/LetterPickerBar/Components/LetterPickerButton.swift
```diff
@@ -0,0 +1,67 @@
...
+        private let activated: Bool
+        private let viewModel: FilterViewModel
+        private var onSelect: () -> Void
```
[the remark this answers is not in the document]
JPKribs: Same as https://github.com/jellyfin/Swiftfin/pull/980#discussion_r1614851400. Should all of 19-22 be removed or just 20 & 21? Those are the ones I could remove without a rework but also, if a rework is necessary, I'm never opposed!

JPKribs:
@LePips I've resolved *most* of the items you mentioned. I have a questions about some of the items you said I could remove just to make sure I understood. Lastly, the only outstanding issue is this item: https://github.com/jellyfin/Swiftfin/pull/980#discussion_r1614816506

Non-English/Romantic characters create an issue where the scrollbar starts in the safeArea and as a result the letters at the top and bottom cannot be scrolled to for smaller iPhones. iPad and Max iPhones are not impacted but there is the added issue that those devices, I incorrectly turn it into a scroll when a static bar works better and the scroll is not required. So I wasn't sure if this item was more incorporated into a scroll(ifLargerThan:) change or if that would still be impacted. I've included a screenshot.

Swiftfin/Components/LetterPickerBar/LetterPickerBar.swift
LePips: Just `onSelect` and everything that refers to it.

Swiftfin/Components/LetterPickerBar/Components/LetterPickerButton.swift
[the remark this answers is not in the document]
LePips: Just `onSelect` and everything that refers to it.

Swiftfin/Views/PagingLibraryView/PagingLibraryView.swift
```diff
@@ -251,6 +256,28 @@ struct PagingLibraryView<Element: Poster>: View {
...
 
+    @ViewBuilder
+    private func contentLetterBarView(content: some View) -> some View {
+        if letterPickerEnabled {
+            switch letterPickerOrientation {
+            case .trailing:
+                HStack(spacing: 0) {
+                    content
+                        .frame(maxWidth: .infinity)
+                    LetterPickerBar(viewModel: viewModel.filterViewModel!)
```
LePips, line 267: One thing I missed earlier is that not all libraries _have_ filters, which is why the `FilterViewModel` is optional. For example, the libraries reached from `HomeView` don't have filters and this crashes.

LePips (changes requested):
Also need to account for no filters.

For the safe area issue, see this commit (left to complete to you):
- https://github.com/jellyfin/Swiftfin/commit/3964493fc3348f3b9ba6abe571a236d87a905fae

Swiftfin/Views/PagingLibraryView/PagingLibraryView.swift
JPKribs: Shoot, I had something in there for that. I'm not sure why I got rid of that but I had:

if letterPickerEnabled && viewModel.filterViewModel != nil {...

I've re-added this.

JPKribs:
This should now work without filters. This works for non-English/Romantic characters and stays in screen but I think it's a little hacky since I am running the check for non-Romantic but skipping it for Romantic since it seems to loop forever with English. I'm 100% up for changes on that! SwiftUI is definitely my weakpoint here haha.

Swiftfin/Views/PagingLibraryView/PagingLibraryView.swift
```diff
@@ -282,179 +321,187 @@ struct PagingLibraryView<Element: Poster>: View {
...
 
-            if UIDevice.isPhone {
-                if defaultDisplayType == .list {
-                    collectionVGridProxy.layout()
-                } else {
+    // MARK: body
+
+    // TODO: becoming too large for typechecker during development, should break up somehow
+
+    var body: some View {
```
LePips, line 330: I would actually ask that we don't perform this refactoring here. I will take a look at it sometime.

Swiftfin/Views/PagingLibraryView/PagingLibraryView.swift
```diff
@@ -251,27 +258,59 @@ struct PagingLibraryView<Element: Poster>: View {
...
                 errorView(with: error)
             case .initial, .refreshing:
-                DelayedProgressView()
+                contentLetterBarView(content: DelayedProgressView())
             }
         }
         .animation(.linear(duration: 0.1), value: viewModel.state)
         .ignoresSafeArea()
+        .onSizeChanged { _, safeArea in
+            if LetterPickerBar.isScrolling {
```
LePips, line 308: Why the `if` and difference check?

Swiftfin/Views/PagingLibraryView/PagingLibraryView.swift
```diff
@@ -251,27 +258,59 @@ struct PagingLibraryView<Element: Poster>: View {
         .proxy(collectionVGridProxy)
     }
 
-    // MARK: body
-
-    // TODO: becoming too large for typechecker during development, should break up somehow
+    @ViewBuilder
+    private func contentLetterBarView(content: some View) -> some View {
+        if letterPickerEnabled && viewModel.filterViewModel != nil {
```
LePips, line 263:
```suggestion
        if letterPickerEnabled, let filterViewModel = viewModel.filterViewModel {
```

and use `filterViewModel` instance instead of force unwrapping below.

LePips (changes requested):

Swiftfin/Views/PagingLibraryView/PagingLibraryView.swift
JPKribs: At least when I was using the Simulator, when I was using English, the onSizeChanged would recursively call until Swiftfin crashed. The safeArea was changing by like 0.0001 and it would re-call this directly after setting it. I tried with a couple different devices in the Simulator and they all had that same issue for English but not when it was a Scroll when I was using Korean as the language. 

This could totally just be something on my end but this resolved it. I have no idea why the safe area was changing every time it got set with the static HStack but not with the Scroll?? I can take another look / record what I am seeing tomorrow if it's still happening.

Swiftfin/Views/PagingLibraryView/PagingLibraryView.swift
JPKribs: I can reset this back to where it was before. This was just my fix because I hit the typechecker errors when I added the extra if in https://github.com/jellyfin/Swiftfin/pull/980#discussion_r1615015997. If I can avoid that extra if the typerchecker doesn't cause issues on build.

JPKribs:
Alright! I think I finally have everything cleaned up @LePips. I think there is one final outstanding item that I can see and that's that the position of the scroll is reset when the letter is selected. I've hit a bit of a wall. I know I've resolved something like that in the past but I'm drawing a blank tonight. I can take a look at that another day.

The only thing that is new/not suggested by you is the LetterPickerOverflow class to handle the non-romantic characters. This was modeled off of what I read here: https://stackoverflow.com/questions/62463142/make-scrollview-scrollable-only-if-it-exceeds-the-height-of-the-window. Let me know if there is anything there I should be changing. This does the scroll based on screen size now. In testing this was my finding:

iPhone SE: Always Scrolls even in English
iPhone 15 Pro: Static in English but Scrolls in Korean
iPad 13": Always Static even in Korean

Swiftfin/Views/PagingLibraryView/PagingLibraryView.swift
LePips: Interesting, sorry I didn't test with that but glad you were able to figure it out!

LePips (approved):
Thank you for going through all of my requests and this has ended in a good state!

PR #1018, opened 2024-04-09
chore: remove refreshable prop from media and paging view

ClarkeCDC:
## Description
Pulling to refresh would freeze the app on all non Home views.

I don't think this is most optimal solution, I am just picking up swift and this does solve the issue but I think it indicates a larger code smell somewhere


LePips:
Bummer. I had experienced this during development but I thought it had fixed itself after some work so I left it in. Just playing around I can't repro it anymore so just have to be lucky to avoid the deadlock. The source of the issue actually comes from my `CollectionVGrid` package using the environment `RefreshAction`, which these views use. I'll have to take a closer look at what to do with my package as just wrapping in a `Task` shouldn't be necessary since the `send` function is just on the main actor anyways, but I'll take a look at that as well if that's just the fix.

If you would like, you can remove these `refreshable` instances instead.

ClarkeCDC:
> Bummer. I had experienced this during development but I thought it had fixed itself after some work so I left it in. Just playing around I can't repro it anymore so just have to be lucky to avoid the deadlock. The source of the issue actually comes from my `CollectionVGrid` package using the environment `RefreshAction`, which these views use. I'll have to take a closer look at what to do with my package as just wrapping in a `Task` shouldn't be necessary since the `send` function is just on the main actor anyways, but I'll take a look at that as well if that's just the fix.
> 
> If you would like, you can remove these `refreshable` instances instead.

Thanks for getting back to me @LePips , that explanation is useful thank you!

Wrapping it in a Task did seem like a dirty win. I added some logs and a few breakpoints and found that the `try await retry()` Was just hanging and the Refresh awaited was then never firing after pulling to refresh.


```swift
// Shared/ViewModels/MediaViewModel/MediaViewModel.swift#L43
case .refresh:
            cancellables.removeAll()

            Task {
                do {
                    logger.log(level: .info, "awaiting refresh")
                    try await refresh()
                    logger.log(level: .info, "Refresh awaited")
                    await MainActor.run {
                        logger.log(level: .info, "Updating state")
                        self.state = .content
                    }
                } catch {
                    await MainActor.run {
                        self.state = .error(.init(error.localizedDescription))
                    }
                }
            }
            .store(in: &cancellables)
            logger.log(level: .info, "return refrehsing")
            return .refreshing
        }
```
I'll update the PR to remove `refreshable` later on today

LePips:
Any interest on removing the `refreshable` instances in here? For the purpose of getting a release out, I may do this if not done fairly soon.

ClarkeCDC:
https://github.com/jellyfin/Swiftfin/issues/1026
Created Issue so the bug fix and re-introduction of the feature are tracked

LePips (approved):

Issue #665, opened 2023-01-02
Live tv upgrades


joshstubble:
Listing live tv better to see the channel title names. Additionally, when searching from Home Screen, searching channel names as well.

PR #1037, opened 2024-04-19
Search Channels and Programs

LePips:
- Implements basic search for channels and programs, matching the web, however I may want to do things differently and pull the recommended programs instead in the future or have an option for it, or both.
- Adds `TODO`s for how I want to handle that as well as view handling.
- Adds a reference implementation for how we should probably handle launching a program and getting the channel media source when launching the video player.
- Closing #665 since now everything has been handled in that request.

### DIFF
--- a/Swiftfin tvOS/Views/SearchView.swift
+++ b/Swiftfin tvOS/Views/SearchView.swift
@@ -16,6 +16,8 @@ struct SearchView: View {
     private var searchPosterType
 
     @EnvironmentObject
+    private var videoPlayerRouter: VideoPlayerWrapperCoordinator.Router
+    @EnvironmentObject
     private var router: SearchCoordinator.Router
 
     @StateObject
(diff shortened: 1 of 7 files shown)


PangMo5 (approved):

Swiftfin tvOS/Views/SearchView.swift
```diff
@@ -58,15 +60,35 @@ struct SearchView: View {
...
                     itemsSection(title: L10n.people, keyPath: \.people, posterType: .portrait)
                 }
+
+                if viewModel.programs.isNotEmpty {
+                    itemsSection(title: L10n.programs, keyPath: \.programs, posterType: .landscape)
+                }
+
+                if viewModel.channels.isNotEmpty {
+                    itemsSection(title: L10n.channels, keyPath: \.channels, posterType: .portrait)
+                }
```
ClarkeCDC, line 70: Are `programs` and `channels` more important than people? If so I think they should go above them

Swiftfin/Views/SearchView.swift
```diff
@@ -68,16 +72,36 @@ struct SearchView: View {
...
                     itemsSection(title: L10n.people, keyPath: \.people, posterType: .portrait)
                 }
+
+                if viewModel.programs.isNotEmpty {
+                    itemsSection(title: L10n.programs, keyPath: \.programs, posterType: .landscape)
+                }
+
+                if viewModel.channels.isNotEmpty {
+                    itemsSection(title: L10n.channels, keyPath: \.channels, posterType: .portrait)
+                }
```
ClarkeCDC, line 82: Same goes here

Swiftfin tvOS/Views/SearchView.swift
LePips: I have just attached them onto the bottom, even though it doesn't match the web. I will be introducing order customization in the future.

PR #1040, opened 2024-04-20
get latest items instead of recently added

ryanh2o3:
Retrieves items from each library and sort them by premiere date.

### DIFF
--- a/Shared/ViewModels/LibraryViewModel/LatestInLibraryViewModel.swift
+++ b/Shared/ViewModels/LibraryViewModel/LatestInLibraryViewModel.swift
@@ -13,20 +13,30 @@ final class LatestInLibraryViewModel: PagingLibraryViewModel<BaseItemDto>, Ident
 
     override func get(page: Int) async throws -> [BaseItemDto] {
 
-        let parameters = parameters()
-        let request = Paths.getLatestMedia(userID: userSession.user.id, parameters: parameters)
+        let parameters = parameters(for: page)
+        let request = Paths.getItemsByUserID(userID: userSession.user.id, parameters: parameters)
         let response = try await userSession.client.send(request)
 
-        return response.value
+        return response.value.items ?? []
     }
 
-    private func parameters() -> Paths.GetLatestMediaParameters {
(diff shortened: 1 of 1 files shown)


LePips:
I used to have exactly what you did before someone fixed it in #895 and I do agree that the grammar itself is somewhat confusing and somebody brought that up before in #203. However, we should be doing it the way it currently is to match the web.

I have ideas to implement a way to customize what sections+ordering to have on the home screen by the user and more. However, the user/server slight refactor I mentioned in PM is a prerequisite for that.

ryanh2o3:
> I used to have exactly what you did before someone fixed it in #895 and I do agree that the grammar itself is somewhat confusing and somebody brought that up before in #203 (but that was before the fix, so it is actually as expected now). However, we should be doing it the way it currently is to match the web.
> 
> 
> 
> I have ideas to implement a way to customize what sections+ordering to have on the home screen by the user and more. However, the user/server slight refactor I mentioned in PM is a prerequisite for that.

Ah ok. So what's the benefit of the 'recently added' section. Just having both film and tv show in one list? Would you be against adding a 'Recently Released Movies' section?

LePips:
Yeah you're right that the Recently Added is out of place since that doesn't have a web equivalent. That was added forever ago and only within the last little bit that I have been doing the "match the web" philosophy. Since the web doesn't have a type-specific section option it would also feel out of place.

I have been working on the exact specifications of how the customization feature would/should work and will report on that soon. As a part of it, the Recently Added section _will_ be removed (when wanting to use the _server_ defined home screen layout) but you'll see how the Recently Added and type-specific sections will be customizable.

ryanh2o3:
> Yeah you're right that the Recently Added is out of place since that doesn't have a web equivalent. That was added forever ago and only within the last little bit that I have been doing the "match the web" philosophy. Since the web doesn't have a type-specific section option it would also feel out of place.
> 
> I have been working on the exact specifications of how the customization feature would/should work and will report on that soon. As a part of it, the Recently Added section _will_ be removed (when wanting to use the _server_ defined home screen layout) but you'll see how the Recently Added and type-specific sections will be customizable.

Ah ok, makes sense.

Sounds good, looking forward to it then.

PR #1044, opened 2024-04-21
remove transition from image view

ryanh2o3:
Removed linear transition from ImageView which has fixed the issue of the header image sliding in from the right

Maybe this has an intended usage somewhere else.


Shared/Components/ImageView.swift
```diff
@@ -47,7 +47,7 @@ struct ImageView: View {
 
     var body: some View {
         if let currentSource = sources.first {
-            LazyImage(url: currentSource.url, transaction: .init(animation: .linear)) { state in
```
LePips, line 50: This does seem to fix the slide in issue, but this removes the nice fade transition between the placeholder (blur hash) and final image. Another solution needs to be figured out with this removed to get the transition _or_ figure out the sliding elsewhere.

My assumption is the issue is in the `BackgroundParallaxHeaderModifier` with how the header is being placed (which is the image) and if there are size differences between the blur hash view and the final image view.

LePips (changes requested):

Shared/Components/ImageView.swift
```diff
@@ -66,6 +69,10 @@ struct ImageView: View {
                 }
             }
             .pipeline(imagePipeline)
+            .opacity(isImageLoaded ? 1 : 0)
```
LePips, line 72: Sorry, but this still isn't correct and is also a big change in behavior. This just fades in the entire `ImageView` on appearance (which should probably actually happen with a `.transition`) and isn't what we want.

I assume you're working with a local server so that images are loading quite fast, which means that the lack of transition between the placeholder blur hash images and the final image isn't noticeable. With a server that takes a few seconds to load an image, the lack of a fade is still noticeable.

I took some time to take a look at this to help point to what the problem actually is, which is one I assumed in my last comment:

> size differences between the blur hash view and the final image view.

Here are steps below that cause the issue:
- For these header views, the aspect ratio is defined to `.fill`, which will cause the `ImageView` to aspect fill its container. Note: `ImageView` will have the same size of its contents (placeholder, failure, image).
	- https://github.com/jellyfin/Swiftfin/blob/384e80805e6c8949e29f67a4324e005d70602c72/Swiftfin/Views/ItemView/iOS/ScrollViews/CompactLogoScrollView.swift#L32

- For the placeholder view, this would create a size that matched the height of the defined header and with a width of the header, for example: `width: 393, height: 393`.
	- https://github.com/jellyfin/Swiftfin/blob/384e80805e6c8949e29f67a4324e005d70602c72/Shared/Extensions/ViewExtensions/Modifiers/BackgroundParallaxHeaderModifier.swift#L42

- When the image finally loads, it will also aspect fill the header but will get a different size, for example: `width: 530, height: 393` (the heights will always be the same due to rules of aspect fill)
- Since the `ImageView` will have the same size of its contents, and its contents changed, it needs to resize and recenter itself. That behavior is the "slide" that we see.

### Solution

The solution is to make the `ImageView` always have the same aspect ratio. If all of the views have the same aspect ratio - which means that they will size the same and correctly. This is used on the `.posterStyle` modifier to make all the posters, which is kind of why we don't see any "slide" behavior with them.

If you want to see this behavior yourself, I would recommend adding the modifier below in some places (like on the placeholder and image in `ImageView` or the header in each item view) to see the size differences before and after the fix:

```
.onSizeChanged { size in
	print(size)
}
```

### Merge

I did the work on my branch below, as well as left a `TODO` for some polish regarding the entire transition between a blur hash and the image view. You are welcome to merge that branch into this and I can approve or I can make the PR instead.

- https://github.com/jellyfin/Swiftfin/compare/main...LePips:SwiftFin:fix-item-header-image-slide

LePips (changes requested):

Shared/Components/ImageView.swift
ryanh2o3: Ah ok, I was thinking it felt kinda different but wasn't sure. Yeah my server is local so everything is pretty instant. 

Ah that makes sense. Thanks for the explanation, much appreciated.

I'll merge into mine and work on the todo.

Shared/Components/ImageView.swift
```diff
@@ -25,6 +25,7 @@ private let imagePipeline = {
 //       - instead of removing first source on failure, just safe index into sources
 // TODO: currently SVGs are only supported for logos, which are only used in a few places.
 //       make it so when displaying an SVG there is a unified `image` caller modifier
+// TODO: probably don't need both `placeholder` modifiers
```
LePips, line 28: I don't think you merged correctly as these aren't on main/my branch.

LePips (changes requested):

Shared/Components/ImageView.swift
LePips: That `TODO` doesn't need to be done immediately and is just a UX polish. That may actually take a lot of work/a refactor and may take a lot of work.

Shared/Components/ImageView.swift
ryanh2o3: yep. noticed. currently trying to figure out what happened

ryanh2o3: Fixed it. Very odd.

Shared/Components/ImageView.swift
ryanh2o3: Ah ok. I'll have a look anyway

Shared/Components/ImageView.swift
```diff
@@ -79,7 +85,7 @@ extension ImageView {
             sources: [source].compacted(using: \.url),
             image: { $0 },
             placeholder: nil,
-            failure: { EmptyView() }
+            failure: { DefaultFailureView() }
```
LePips, line 88: Need to remove the changes to these inits as well.

LePips (changes requested):

Shared/Components/ImageView.swift
ryanh2o3: done

LePips (approved):

PR #1066, opened 2024-05-23
tvOS - Show and interact with the video menu

MatsMoll:
Made it possible to use the Siri remote to play / stop the video.
This also shows the video menu. Therefore, making it possible to set subtitles etc.

The `PreferencesView` was also removed when showing the `VideoPlayer` as it added some unwanted padding to the video content.


Shared/Coordinators/VideoPlayerCoordinator.swift
```diff
@@ -52,9 +52,7 @@ final class VideoPlayerCoordinator: NavigationCoordinatable {
 
         #else
         if Defaults[.VideoPlayer.videoPlayerType] == .swiftfin {
-            PreferencesView {
```
LePips, line 55: Thanks for taking a look at this, however I'll take a while because I don't have my Apple TV with me personally and I'll have to think/see if this is the route to go because SwiftUI `onExitCommand/on_Command` has specific rules for how they get called. For example for what you've done in this PR it actually crashes because the view, without the overlay, is not in focus so SwiftUI will just exit the player without calling our `onExitCommand`. Focus on tvOS+SwiftUI is a mess and atow, none of what you've done works in the simulator for me because the view isn't in "focus", so SwiftUI handles all of the commands itself.

The purpose of `PreferencesView` is to intercept remote presses and not let SwiftUI handle them (but this isn't implemented yet). I would have also looked at implemented detecting swipes from the 1st gen Siri Remote at that layer.

Here's a super old commit when the video player was in UIKit where we handled the key presses:

- https://github.com/jellyfin/Swiftfin/blob/5d0f933a2c083fd27ae4579960042cc51cf22450/Swiftfin%20tvOS/Views/VideoPlayer/VLCPlayerViewController.swift#L206

Swiftfin tvOS/Views/VideoPlayer/VideoPlayer.swift
```diff
@@ -83,6 +96,10 @@ struct VideoPlayer: View {
                     .scrubbedSeconds = Int(CGFloat(videoPlayerManager.currentViewModel.item.runTimeSeconds) * newValue)
             }
         }
+        .onDisappear(perform: {
+            videoPlayerManager.proxy.stop()
+            router.dismissCoordinator()
```
LePips, line 101: This causes a crash because you are attempting to dismiss a coordinator when it has already been dismissed.

Swiftfin tvOS/Views/VideoPlayer/VideoPlayer.swift
```diff
@@ -83,6 +96,10 @@ struct VideoPlayer: View {
                     .scrubbedSeconds = Int(CGFloat(videoPlayerManager.currentViewModel.item.runTimeSeconds) * newValue)
             }
         }
+        .onDisappear(perform: {
```
LePips, line 99: For the future, please use trailing closure syntax.

Swiftfin tvOS/Views/VideoPlayer/Overlays/Overlay.swift
```diff
@@ -69,34 +74,33 @@ extension VideoPlayer {
...
-//                                overlayTimer.stop()
-//                            }
-//                        }
-//
-//                        confirmCloseWorkItem = task
-//
-//                        DispatchQueue.main.asyncAfter(deadline: .now() + 2, execute: task)
-//                    }
-//                }
+                .onExitCommand {
```
LePips, line 77: See other comment for why this isn't called.

Swiftfin tvOS/Views/VideoPlayer/VideoPlayer.swift
```diff
@@ -73,6 +73,19 @@ struct VideoPlayer: View {
                 .environmentObject(videoPlayerManager.proxy)
                 .environment(\.isPresentingOverlay, $isPresentingOverlay)
                 .environment(\.isScrubbing, $isScrubbing)
+                .onPlayPauseCommand(perform: {
```
LePips, line 76: See other comment for why this isn't called.

LePips (changes requested):

MatsMoll:
Thanks for the reference, it helped me implement an alternative version.
I have not added support for swipe recognition, but it is a start at least

PreferencesView/Sources/PreferencesView/PressCommandAction.swift
```diff
@@ -0,0 +1,42 @@
...
+}
+
+extension PressCommandAction: Equatable {
+
+    public static func == (lhs: PressCommandAction, rhs: PressCommandAction) -> Bool {
+        lhs.press == rhs.press
+    }
+}
+
+struct PressCommandsPreferenceKey: PreferenceKey {
```
LePips, line 35: Please move the `PreferenceKeys` and only make available to tvOS

PreferencesView/Sources/PreferencesView/PreferencesView.swift
```diff
@@ -19,8 +19,25 @@ public struct PreferencesView<Content: View>: UIViewControllerRepresentable {
     }
 
     public func makeUIViewController(context: Context) -> UIPreferencesHostingController {
-        UIPreferencesHostingController(content: content)
+        let controller = UIPreferencesHostingController(content: content)
+        self.addGesture(press: .menu, view: controller.view, coordinator: context.coordinator)
```
LePips, line 23: Let's do all of this in `UIPreferencesHostingController.viewDidLoad` instead

PreferencesView/Sources/PreferencesView/PressCommandBuilder.swift
```diff
@@ -0,0 +1,37 @@
...
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2024 Jellyfin & Jellyfin Contributors
+//
+
+import Foundation
+
+@resultBuilder
+public enum PressCommandsBuilder {
```
LePips, line 12: Since we now have two of these similar array-building result builders, let's instead make a generic `ArrayBuilder<Content>` and typealias this `PressCommandsBuilder` and `KeyCommandsBuilder`.

```swift

typealias KeyCommandsBuilder = ArrayBuilder<KeyCommandAction>
typealias PressCommandsBuilder = ArrayBuilder<PressCommandAction>

@resultBuilder
public enum ArrayBuilder<Content> {

    public static func buildBlock(_ components: [Content]...) -> [Content] {
        components.flatMap { $0 }
    }

    public static func buildExpression(_ expression: Content) -> [Content] {
        [expression]
    }

    public static func buildOptional(_ component: [Content]?) -> [Content] {
        component ?? []
    }

    public static func buildEither(first component: [Content]) -> [Content] {
        component
    }

    public static func buildEither(second component: [Content]) -> [Content] {
        component
    }

    public static func buildArray(_ components: [[Content]]) -> [Content] {
        components.flatMap { $0 }
    }
}
```

PreferencesView/Sources/PreferencesView/UIPreferencesHostingController.swift
```diff
@@ -27,6 +27,10 @@ public class UIPreferencesHostingController: UIHostingController<AnyView> {
                 .onPreferenceChange(SupportedOrientationsPreferenceKey.self) {
                     box.value?._orientations = $0
                 }
+            #else
+                .onPreferenceChange(PressCommandsPreferenceKey.self) {
```
LePips, line 31: Only make available to tvOS

PreferencesView/Sources/PreferencesView/UIPreferencesHostingController.swift
```diff
@@ -40,6 +44,16 @@ public class UIPreferencesHostingController: UIHostingController<AnyView> {
         fatalError("init(coder:) has not been implemented")
     }
 
+    private var _pressCommandActions: [PressCommandAction] = []
```
LePips, line 47: Only make available to tvOS

PreferencesView/Sources/PreferencesView/ViewExtensions.swift
```diff
@@ -10,6 +10,10 @@ import SwiftUI
 
 public extension View {
 
+    func pressCommands(@PressCommandsBuilder _ commands: @escaping () -> [PressCommandAction]) -> some View {
```
LePips, line 13: Only make available to tvOS

LePips (changes requested):
Thank you for the changes. However, let's only make press changes available on tvOS for now and make available to other platforms if applicable.

LePips (approved):
I will be running up against this in #1203 soon and have addressed my comments.

MatsMoll:
Thanks for completing it @LePips! I completely forgot about it, as I focused on some other projects

PR #1067, opened 2024-05-24
Fixed incorrectly formatted media descriptions.

Alexk2309:
Video descriptions often had HTML tags that weren't parsed correctly for example.

image

This change corrects the formatting so it now looks like
image

### DIFF
--- a/Swiftfin/Components/HTMLFormattedText.swift
+++ b/Swiftfin/Components/HTMLFormattedText.swift
@@ -1,0 +1,60 @@
+//
+// Swiftfin is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2024 Jellyfin & Jellyfin Contributors
+//
+
+struct HTMLFormattedText: UIViewRepresentable {
+    let text: String
+    private let textView = UITextView()
+
+    init(_ content: String) {
+        self.text = content
+    }
+
+    func makeUIView(context: UIViewRepresentableContext<Self>) -> UITextView {
+        textView.widthAnchor.constraint(equalToConstant: UIScreen.main.bounds.width).isActive = true
+        textView.isSelectable = false
+        textView.isUserInteractionEnabled = false
+        textView.translatesAutoresizingMaskIntoConstraints = false
+        textView.isScrollEnabled = false
+        return textView
+    }
+
+    func updateUIView(_ uiView: UITextView, context: UIViewRepresentableContext<Self>) {
+        DispatchQueue.main.async {
+            if let attributeText = self.converHTML(text: text) {
+                textView.attributedText = attributeText
+            } else {
+                textView.text = ""
+            }
+        }
+    }
+
+    private func converHTML(text: String) -> NSAttributedString? {
+        guard let data = text.data(using: .utf8) else {
+            return nil
+        }
+
+        if let attributedString = try? NSMutableAttributedString(
+            data: data,
+            options: [.documentType: NSAttributedString.DocumentType.html],
+            documentAttributes: nil
+        ) {
+            let range = NSRange(location: 0, length: attributedString.length)
+            attributedString.enumerateAttribute(.font, in: range, options: []) { value, range, _ in
+                if let oldFont = value as? UIFont {
+                    let fontSize = UIScreen.main.bounds.width * 0.05 // Adjust the multiplier as needed
+                    let dynamicFont = UIFont.systemFont(ofSize: fontSize)
+                    let newFont = oldFont.withSize(dynamicFont.pointSize)
+                    attributedString.addAttributes([.font: newFont], range: range)
+                }
+            }
+            return attributedString
+        } else {
+            return nil
+        }
+    }
+}
--- a/Swiftfin/Components/HTMLFormattedText.swift
+++ b/Swiftfin/Components/HTMLFormattedText.swift
@@ -6,6 +6,8 @@
 // Copyright (c) 2024 Jellyfin & Jellyfin Contributors
 //
 
+import SwiftUI
+
 struct HTMLFormattedText: UIViewRepresentable {
     let text: String
     private let textView = UITextView()
--- a/Shared/Extensions/String.swift
+++ b/Shared/Extensions/String.swift
@@ -114,6 +114,23 @@ extension String {
 
         return s
     }
+
+    var htmlToAttributedString: NSAttributedString? {
+        guard let data = data(using: .utf8) else { return nil }
+        do {
+            return try NSAttributedString(
+                data: data,
+                options: [.documentType: NSAttributedString.DocumentType.html, .characterEncoding: String.Encoding.utf8.rawValue],
+                documentAttributes: nil
+            )
+        } catch {
+            return nil
+        }
+    }
+
+    var htmlToString: String {
+        htmlToAttributedString?.string ?? ""
+    }
 }
 
 extension CharacterSet {
--- a/Shared/Extensions/String.swift
+++ b/Shared/Extensions/String.swift
@@ -114,23 +114,6 @@ extension String {
 
         return s
     }
-
-    var htmlToAttributedString: NSAttributedString? {
-        guard let data = data(using: .utf8) else { return nil }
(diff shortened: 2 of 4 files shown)


Swiftfin/Components/HTMLFormattedText.swift
```diff
@@ -0,0 +1,62 @@
...
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2024 Jellyfin & Jellyfin Contributors
+//
+
+import SwiftUI
+
+struct HTMLFormattedText: UIViewRepresentable {
+    let text: String
+    private let textView = UITextView()
```
LePips, line 13: Created views within `makeUIView`, not at the top level.

Swiftfin/Components/HTMLFormattedText.swift
```diff
@@ -0,0 +1,62 @@
...
+// Swiftfin is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2024 Jellyfin & Jellyfin Contributors
+//
+
+import SwiftUI
+
+struct HTMLFormattedText: UIViewRepresentable {
```
LePips, line 11: I asked our team and we "officially" support HTML *and* Markdown, so this view would have to accommodate *both*.

Swiftfin/Components/HTMLFormattedText.swift
```diff
@@ -0,0 +1,62 @@
...
+    func makeUIView(context: UIViewRepresentableContext<Self>) -> UITextView {
+        textView.widthAnchor.constraint(equalToConstant: UIScreen.main.bounds.width).isActive = true
+        textView.isSelectable = false
+        textView.isUserInteractionEnabled = false
+        textView.translatesAutoresizingMaskIntoConstraints = false
+        textView.isScrollEnabled = false
+        return textView
+    }
+
+    func updateUIView(_ uiView: UITextView, context: UIViewRepresentableContext<Self>) {
```
LePips, line 28: Change `UIViewRepresentableContext<Self>)` to just `Context`.

Swiftfin/Components/HTMLFormattedText.swift
```diff
@@ -0,0 +1,62 @@
...
+
+struct HTMLFormattedText: UIViewRepresentable {
+    let text: String
+    private let textView = UITextView()
+
+    init(_ content: String) {
+        self.text = content
+    }
+
+    func makeUIView(context: UIViewRepresentableContext<Self>) -> UITextView {
```
LePips, line 19: Change `UIViewRepresentableContext<Self>)` to just `Context`.

Swiftfin/Components/HTMLFormattedText.swift
```diff
@@ -0,0 +1,62 @@
...
+
+        if let attributedString = try? NSMutableAttributedString(
+            data: data,
+            options: [.documentType: NSAttributedString.DocumentType.html],
+            documentAttributes: nil
+        ) {
+            let range = NSRange(location: 0, length: attributedString.length)
+            attributedString.enumerateAttribute(.font, in: range, options: []) { value, range, _ in
+                if let oldFont = value as? UIFont {
+                    let fontSize = UIScreen.main.bounds.width * 0.05 // Adjust the multiplier as needed
```
LePips, line 51: We can't determine a font size based on the screen width. We should be using a provided system font size and the San Francisco font, not whatever serif font is being rendered.

Swiftfin/Components/HTMLFormattedText.swift
```diff
@@ -0,0 +1,62 @@
...
+        textView.widthAnchor.constraint(equalToConstant: UIScreen.main.bounds.width).isActive = true
+        textView.isSelectable = false
+        textView.isUserInteractionEnabled = false
+        textView.translatesAutoresizingMaskIntoConstraints = false
+        textView.isScrollEnabled = false
+        return textView
+    }
+
+    func updateUIView(_ uiView: UITextView, context: UIViewRepresentableContext<Self>) {
+        DispatchQueue.main.async {
```
LePips, line 29: This isn't necessary. Nothing is being updated externally that requires this view to update itself.

Swiftfin/Components/HTMLFormattedText.swift
```diff
@@ -0,0 +1,62 @@
...
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2024 Jellyfin & Jellyfin Contributors
+//
+
+import SwiftUI
+
+struct HTMLFormattedText: UIViewRepresentable {
+    let text: String
```
LePips, line 12:
```suggestion
    private let text: String
```

Swiftfin/Components/HTMLFormattedText.swift
```diff
@@ -0,0 +1,62 @@
...
+    private let textView = UITextView()
+
+    init(_ content: String) {
```
LePips, line 15: Change `content` to `text`.

Swiftfin/Components/HTMLFormattedText.swift
```diff
@@ -0,0 +1,62 @@
...
+        self.text = content
+    }
+
+    func makeUIView(context: UIViewRepresentableContext<Self>) -> UITextView {
+        textView.widthAnchor.constraint(equalToConstant: UIScreen.main.bounds.width).isActive = true
```
LePips, line 20: `UITextView`s in SwiftUI are very tricky to get right. I actually just helped someone make a `UITextView` wrapper so this came at the right time. For sizing we shouldn't set constraints and instead have to go to `setContentCompressionResistancePriority`.

Here is the code that I helped the other person with, so you'll have to decipher it a bit, but it is an example of how to wrap a `UITextView` in SwiftUI. I am unsure how it works with HTML or Markdown, so you will have to test that. This work may be bigger than you expected.

- https://pastebin.com/nbiTbwcN

Swiftfin/Components/HTMLFormattedText.swift
```diff
@@ -0,0 +1,62 @@
...
+        DispatchQueue.main.async {
+            if let attributeText = self.converHTML(text: text) {
+                textView.attributedText = attributeText
+            } else {
+                textView.text = ""
+            }
+        }
+    }
+
+    private func converHTML(text: String) -> NSAttributedString? {
```
LePips, line 38: We will have to determine whether the text has HTML and then "render" it as HTML.

LePips (changes requested):
I apologize but this work may be a lot more than you probably had anticipated. We "officially" support HTML and Markdown for item descriptions (and the server disclaimer).

*However* if you find a package that does what we need, we can use that instead.

Alexk2309:
> I apologize but this work may be a lot more than you probably had anticipated. We "officially" support HTML and Markdown for item descriptions (and the server disclaimer).
> 
> 
> 
> *However* if you find a package that does what we need, we can use that instead.

Hi if this is the case meaning html is supported how come the tags aren't being rendered correctly. Also does that mean i would have to find a package that converts html to markdown ? Because i know swift already supports markdown .

LePips:
My comment was that HTML and Markdown support was implemented, but that since we "support both", we need to implement support for both.

SwiftUI sadly only supports basic inline Markdown attributes (bold, italic) but not stuff like headers or bullet points. So, technically, even Markdown needs proper implementation.

LePips:
I'm actually going to have to think about this a bit more. While we/other clients "officially" support HTML, I think that this would be too large of a hassle for all clients to support and would advocate for stripping HTML on the server.

chickdan:
I tested the solution in [this SO post](https://stackoverflow.com/questions/56892691/how-to-show-html-or-markdown-in-a-swiftui-text) (`NSAttributedString`) and it worked. `Text` can also handle Markdown natively so maybe if there was a function to detect html (regex aught to do it) and return the overview as `NSAttributedString` otherwise return as a `String`. That would save having to use `UIVIewRepresentable` to make a `UITextView` and resolve the issue of supporting both formats.

LePips:
I'm closing this in favor of just using an `AttributedString` solution.

PR #1076, opened 2024-05-28
tvOS 17 as minimum

LePips:
For some reason I have never actually checked what devices each tvOS supported, so I looked it up and tvOS 17 supports the same devices as tvOS 15. I am so gladdened, elated, and so much more to see this. I cannot begin to list the problems that this minimum alone solves.

- https://en.wikipedia.org/wiki/TvOS#Supported_OS_releases

## Why

SwiftUI has so many features and developments on tvOS 16/17 that I've needed and tried to make workarounds for. This is a necessity.

## What about jailbreaks?

I'm sure this is a question that will be asked and I am not active on the jailbreaking scene to even know if this is applicable. Same reasoning for iOS 15 minimum: I'm not tailoring this client towards jailbreak users.

## iOS 15 dropped?

I am deferring this to a later time but I have had to think about this.


PangMo5 (approved):

holow29:
Jailbreaking not a concern regardless. No jailbreak for tvOS 15/16/17. IIRC TrollStore works on tvOS 17.0 as well. In other words, there are no downsides.

holow29:
tvOS 18 will support same devices as tvOS 17 unless something changes during beta period, which is doubtful.

holow29:
Just to circle back to this: tvOS 18 supports all same devices as tvOS 17, so IMO if there is anything that would be useful there, we can make tvOS 18 minimum.

Now, if we were concerned with jailbreaking, there is now a jailbreak for tvOS up to 17.5.1 (only for Apple TV HD and Apple TV 4k Gen 1 though) but not tvOS 18. TrollStore also supports tvOS 17.0.

LePips:
I have been perusing through some of the new tvOS APIs lately and thought about setting tvOS 18 as the minimum as well but we just haven't had a need for it (yet). I have also thought about doing the same for iOS when we eventually drop iOS 16, but we'll get there when we get there.

ceedveed:
I have devices on tvOS 16 due to TrollStore support. Is it possible to make the minimum tvOS 16 instead? If not, that's okay just wanted to know.

holow29:
> I have devices on tvOS 16 due to TrollStore support. Is it possible to make the minimum tvOS 16 instead? If not, that's okay just wanted to know.

If you have a device that supports it (checkm8 vulnerable - Apple TV HD or Apple TV 4K Gen 1), you can jailbreak on tvOS 18 using palera1n, though it is a semi-tethered jailbreak. I see now that while Trollstore should technically support tvOS 17.0, there might not be a current installation vector like misaka.

Regardless, this commit was merged almost a year ago, and I believe the ship has sailed on tvOS 16 support. The current version on the App Store still supports it, though, AFAIK.

ceedveed:
> > I have devices on tvOS 16 due to TrollStore support. Is it possible to make the minimum tvOS 16 instead? If not, that's okay just wanted to know.
> 
> If you have a device that supports it (checkm8 vulnerable - Apple TV HD or Apple TV 4K Gen 1), you can jailbreak on tvOS 18 using palera1n, though it is a semi-tethered jailbreak. I see now that while Trollstore should technically support tvOS 17.0 without jailbreak due to CoreTrust vulnerability, there might not be a current installation vector like misaka.
> 
> Regardless, this commit was merged almost a year ago, and I believe the ship has sailed on tvOS 16 support. The current version on the App Store still supports it, though, AFAIK.

Thanks. I have a couple Apple TV 4K Gen 3 on tvOS 16, so I can't JB. Is it possible to make the IPA available on Github once released?

PR #1080, opened 2024-05-28
Add Swift 6 Feature Flags

### LINKED ISSUE #1053: Enable Some Swift 6 Flags
In preparation for Swift 6, I would like to enable some flags ahead of time and fix if necessary for any existential and concurrency stuff. The thread below doesn't list all of the flags (since some were implemented after it was made) so there may be more.

- https://forums.swift.org/t/progress-toward-the-swift-6-language-mode/68315

While `ExistentialAny` won't be enabled by default, I'd also like to take a look at enabling it.

- https://github.com/apple/swift-evolution/blob/main/proposals/0335-existential-any.md

chickdan:
This branch is in preparation for Swift 6. These flags introduce a number of build errors so this branch should not be merged until the appropriate changes are complete.

Flags enabled:
- StrictConcurrency
- ExistentialAny
- ForwardTrailingClosures
- GlobalConcurrency

Closes #1053

### DIFF
--- a/Swiftfin.xcodeproj/project.pbxproj
+++ b/Swiftfin.xcodeproj/project.pbxproj
@@ -4784,6 +4784,7 @@
 					"@executable_path/Frameworks",
 				);
 				MARKETING_VERSION = 1.0.0;
+				OTHER_SWIFT_FLAGS = "$(inherited)";
 				PRODUCT_BUNDLE_IDENTIFIER = org.jellyfin.swiftfin;
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SDKROOT = appletvos;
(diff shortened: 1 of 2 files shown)


chickdan:
@LePips looking into some of these errors and many of them are around `@ViewBuilder` or other SwiftUI related code. Given that the SwiftUI framework is not up to date for Swift 6, some of these errors may resolve themselves. 

Now I'm not sure if Swift 6 changes to SwiftUI will be locked behind iOS 18 or if they will back port, but for now I think that instead of using these compiler directives you can change `Swift Concurrency Checking` from `minimal` to `complete`. That way instead of build errors there are warnings and we can see what comes out of WWDC before making any changes.

Thoughts?

LePips:
I had only seen the SwiftUI errors passively on the forums, but didn't know it was entirely unusable lol. Setting concurrency checking to `complete` still gets me build errors for some `UIScreen` and some `View` stuff though.

Welp, I was just curious about it for this project since WWDC soon and Swift 6 should be expected any time now. Looks like we'll just have to wait for some "official" transition instructions.

PR #1084, opened 2024-06-06
Cleanup video player settings

chickdan:
Completes the organization `TODO` of the `VideoPlayerSettingsView` plus two other small items.

PR #1085, opened 2024-06-06
Refactor sign in policy

chickdan:
Simple refactor to resolve `TODO`.

~~However, one of the views is timing out during the type check and is throwing an error for me so I added a `TODO`.~~
Edit:
Ok it's compiling for me now. If it continues to fail CI I can add some explicit typing to remove inferencing during compilation.

### DIFF
--- a/Swiftfin/Views/UserSignInView/UserSignInView.swift
+++ b/Swiftfin/Views/UserSignInView/UserSignInView.swift
@@ -104,7 +104,7 @@ struct UserSignInView: View {
 
     private func signInUplicate(user: UserState, needsPin: Bool = true, replace: Bool) {
         Task {
-            switch user.signInPolicy {
+            switch user.accessPolicy {
             case .none: ()
             case .requireDeviceAuthentication:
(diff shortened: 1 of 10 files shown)


LePips:
Thanks for doing more housekeeping. Yeah I'm pushing the type checker to the limit in some places, not ideal.

I'll be able to take a look at this soon.

Swiftfin/Views/UserSignInView/UserSignInView.swift
```diff
@@ -388,4 +371,25 @@ struct UserSignInView: View {
             Text("Set pin for new user.")
         }
     }
+
+    func handleSignIn(_ event: UserSignInViewModel.Event) {
```
LePips, line 375: Similar to https://github.com/jellyfin/Swiftfin/pull/1084#discussion_r1632113562, make `private` and move to above, probably best above all of the QuickConnect/sign in stuff.

LePips (changes requested):

LePips (approved):

PR #1090, opened 2024-06-08
[tvOS] Change password to be a SecureField and autosubmit on password comple…

petermcneil:
…tion

### DIFF
--- a/Shared/ViewModels/SelectUserViewModel.swift
+++ b/Shared/ViewModels/SelectUserViewModel.swift
@@ -44,7 +44,6 @@ class SelectUserViewModel: ViewModel, Eventful, Stateful {
 
     var events: AnyPublisher<Event, Never> {
         eventSubject
-            .receive(on: RunLoop.main)
             .eraseToAnyPublisher()
     }
 
--- a/Swiftfin tvOS/Views/UserSignInView/UserSignInView.swift
+++ b/Swiftfin tvOS/Views/UserSignInView/UserSignInView.swift
@@ -17,7 +17,7 @@ import SwiftUI
 
 struct UserSignInView: View {
 
-    enum FocusField: Hashable {
+    private enum FocusField: Hashable {
         case username
         case password
     }
--- a/Swiftfin tvOS/Views/UserSignInView/UserSignInView.swift
+++ b/Swiftfin tvOS/Views/UserSignInView/UserSignInView.swift
@@ -16,6 +16,10 @@ import SwiftUI
 // TODO: change public users from list to grid
 
 struct UserSignInView: View {
+    enum FocusField: Hashable {
+        case username
+        case password
+    }
 
     @Default(.accentColor)
     private var accentColor
@@ -24,7 +28,7 @@ struct UserSignInView: View {
     private var router: UserSignInCoordinator.Router
 
     @FocusState
-    private var focusedTextField: Int?
+    private var focusedTextField: FocusField?
 
     @State
     private var duplicateUser: UserState? = nil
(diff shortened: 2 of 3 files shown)


Swiftfin tvOS/Views/UserSignInView/UserSignInView.swift
```diff
@@ -49,20 +53,19 @@ struct UserSignInView: View {
     @ViewBuilder
     private var signInSection: some View {
         Section {
-            TextField(L10n.username, text: $username)
+            TextField(L10n.username, text: $username, prompt: Text("Required"))
```
LePips, line 56: The first parameter is already the prompt, which I'm fine with just being the username label. So, the new prompt is no longer required. I don't have a design for any indicators for what fields are required except for the sign in button which _should_ have dynamic colors for whether it can be selected or not (but it just hasn't been implemented because I needed to take a break from tvOS). I'm not worried about people being confused about knowing if username is required, because that's basic requirement of the entire Jellyfin user system.

Swiftfin tvOS/Views/UserSignInView/UserSignInView.swift
```diff
@@ -16,6 +16,10 @@ import SwiftUI
 // TODO: change public users from list to grid
 
 struct UserSignInView: View {
+    enum FocusField: Hashable {
```
LePips, line 19: Please put an empty line above `FocusField` as well, please.

```suggestion
    private enum FocusField: Hashable {
```

Swiftfin tvOS/Views/UserSignInView/UserSignInView.swift
```diff
@@ -49,20 +53,19 @@ struct UserSignInView: View {
...
                 .autocorrectionDisabled()
                 .textInputAutocapitalization(.never)
-                .focused($focusedTextField, equals: 0)
+                .focused($focusedTextField, equals: .username)
+
+            SecureField(L10n.password, text: $password)
+                .focused($focusedTextField, equals: .password)
                 .onSubmit {
-                    focusedTextField = 1
+                    if username.isEmpty {
```
LePips, line 64: Use `guard` instead:

`guard username.isNotEmpty else { return }`

LePips (changes requested):
Thank you for your contribution, just some changes.

Swiftfin tvOS/Views/UserSignInView/UserSignInView.swift
petermcneil: Okay, cool. I've removed the prompt - too be honest I'm not sure why I added it. It made the field just say "Required" which is really confusing to anyone.

Swiftfin tvOS/Views/UserSignInView/UserSignInView.swift
```diff
@@ -74,9 +77,13 @@ struct UserSignInView: View {
             .foregroundStyle(.red, .red.opacity(0.2))
         } else {
             Button(L10n.signIn) {
-                focusedTextField = nil
```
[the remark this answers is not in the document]
petermcneil: Good points! I've removed them now

petermcneil:
I've added in a fix for the SelectUserViewModel. When I was on the user selection page, I would click on the button for my user and it would do nothing. After following the flow with logging I found that the message just stopped when sending back via that events publisher. 

Removing `.receive(on: RunLoop.main)` made it work. I'm not really sure why, something to do with the scheduling. I'm obviously new to Swift, let me know if that was the wrong thing to do. Once I removed it I was able to login and do the rest of the actions in the tvOS app.

I got the "solution" from here - https://forums.swift.org/t/combine-receive-on-runloop-main-loses-sent-value-how-can-i-make-it-work/28631/24

Shared/ViewModels/SelectUserViewModel.swift
```diff
@@ -44,7 +44,6 @@ class SelectUserViewModel: ViewModel, Eventful, Stateful {
 
     var events: AnyPublisher<Event, Never> {
         eventSubject
-            .receive(on: RunLoop.main)
```
LePips, line 47: Thank you for finding that material for the fix. I currently run all event subjects through the main thread generally as a practice but I guess I'll have to think about it more. This view is unfinished and the reason why this doesn't break on iOS is because in `SelectUserView` it will eventually send the action in a `Task { @MainActor in`, which then will send its corresponding event on the main actor for us to listen to.

I'm fine with this for now but think there's a different concurrency issue with `Stateful`.

LePips (approved):

PR #1097, opened 2024-06-12
Migrate to the latest version of Factory

chickdan:
It looks like everything is migrated over. I kept the same scope levels with the exception of the `DownloadManager` which I don't think needs to be a singleton given it's limited scope (in terms of usage), although I'm sure others can be switched to `shared` or `cached` as well.

Additionally, I've moved all of the declarations into the `Container` extension as suggested by the library - otherwise what you have are standard singletons - but they are still scattered. Should I put all of the declarations in one place for easier maintenance?


LePips (changes requested):
Looks good, only request is that I would prefer the `callAsFunction` usage, which is just a basic replace `.resolve() -> ()`

chickdan:
@LePips I can make the change but I would like to understand your reasoning, as it isn't immediately clear to me. Thanks!

Before:
`Container.shared.currentUserSession.resolve()`
After:
`Container.shared.currentUserSession.callAsFunction()`

LePips:
`callAsFunction` is just syntactic sugar for things being callable as a function. The actual result of the renaming would be:

`Container.shared.currentUserSession.resolve() -> Container.shared.currentUserSession()`

https://github.com/swiftlang/swift-evolution/blob/main/proposals/0253-callable.md

chickdan:
Ah, I misunderstood the documentation. Thank you for clearing that up!

LePips (approved):

PR #1107, opened 2024-06-13
Update watched indicator color

Isaiah-Hamilton:
## Changes made

Updated the watched indicator color to use the users accent color

## Alternative Option

Instead of using the users accent color just make the indicator color white

## Feedback

Let me know which one is best and what you think.


LePips (approved):
Thank you for your contribution. I'm fine with this for now as I actually did want the watched indicator to have its own color and we aren't just sticking the user-defined accent color everywhere. However, I'll think about that later when I take a look at the indicators again.

Issue #1118, opened 2024-06-18
feature: Player Overlay Lock Button


thearyadev:
**Describe the feature you'd like**
A button in the overlay top bar to lock the player's overlay to avoid unintentional activation of the overlay and prevent the media from being obstructed. 
<img width="326" alt="image" src="https://github.com/jellyfin/Swiftfin/assets/87589047/f15ebde9-9e74-4dc7-b448-82cee4c77307">

**Additional context**
The user can tap the button to activate the overlay lock. Once activated, the player controls (play, pause, skip), scrubbing bar, and tinted overlay will be hidden, and the top bar (media title and other top-menu controls) will remain visible on interaction.

edit: While exploring the settiings of Swiftfin I've discovered the "Compact Playback Buttons" option. This partially solves the problem with obstructing the media playback.

thearyadev:
Aside: I've created a functioning prototype of the feature, I can open a PR soon.

PR #1120, opened 2024-06-18
feature: Player Overlay Lock Button

thearyadev:
# Overview
This PR adds a player overlay lock button to the top action bar (menu and action buttons) as a non-default menu option. 

Feature Outlined in #1118.

This feature solves the problem of unintentional activation of the overlay, preventing the overlay from obstructing the media playback. The button will allow the user to enable a lock which prevents the `Overlay.BottomBarView` and `Overlay.LargePlaybackButtons` from displaying on the screen. The lock does not affect the top overlay view. 

image

# Changes
- Add `LockOverlayButton` component
- Add LockOverlayButton to settings options for overlay top bar action button and menu
- Conditionally display components in MainOverlay


benonymity:
Doesn't there already exist an overlay lock function if you hold down on the screen? 

Settings>Video Player>Gestures>Long Press>Gesture Lock

thearyadev:
> Doesn't there already exist an overlay lock function if you hold down on the screen?
> 
> Settings>Video Player>Gestures>Long Press>Gesture Lock

oh, well you'd be right haha. 

I tested this while exploring an overlay lock but the long-press didn't do anything. I just tested it again but did a much longer long press and it worked.

LePips:
Yes this is already a feature, but I wouldn't mind having a menu bar button either.

Issue #678, opened 2023-01-09
Swiftfin doesn't respect download Mbs limit


michpas:
### Describe the bug

1. Goto jellyfin admin /dashboard/users/ set Internet streaming bitrate limit to 2
2. Logout admin. 
3. Login as user / settings / playback  / set home network quality & Chromecast quality to 480p -1.5 mbs
4. Play a 1080p 4+ Mbps video on user in swiftfin
5. Swiftin plays the file at original size (jellyfin states also: Device supports codecs

My expectations is Swiftin to respect the bitrate limit and not stream the original full size file

Use case: Ipad connecting to hotspot from mobilephone, to allow video streaming whilst traveling.

### Application version

1.0.0 (79)

### Where did you install the app from?

App Store

### Device information

Ipad Air 3rd Gen

### OS version

Ios 15.5

### Jellyfin server version

10.8.8

shimi-ke:
I can confirm my experience is the same. But I suspect the problem might be elsewhere (probably on a server). I think so because I have the same experience not only with Swiftin but with Infuse Pro as well. Both applications seem to ignore playback settings set on server.

When I tried to inspect logs on my jellyfin server I found out that the RemoteClientBitrateLimit is set correctly but the play_method is set to "DirectPlay" even if the bitrate exceeds the RemoteClientBitrateLimit. If I use the same profile but Jellyfin iOS app, the RemoteClientBitrateLimit is set correctly and the play method is "Transcode", which is intended behavior.

All applications are the latest versions and have been downloaded from the AppStore. I use them on my iPhone running iOS 16.2.

Jellyfin server is 10.8.8.

LePips:
I do believe this would be a server issue as clients make a request for the streaming url. If transcoding is required, like in the event of server-set bitrate maximums, clients will (or should at least) use it.

Cloudcatcher888:
exactly same experience with @shimi-ke and @michpas using Infuse Pro and Swiftfin with a frp server (8Mbps)

luyvlei:
In my experience, the jellyfin use client cookies to control the default bitrate limit. When I clean the cookies of my chrome after change the default bitrate limit to 10Mbps, the bitrate limite will be reset to 'auto'. When a new device, e.g. swiftfin, login the server, the default bitrate limit will also be 'auto'. 

[...]

Issue #184, opened 2021-10-16
Quality Menu


maxens-git:
it would be nice if there was a menu to choose the quality directly from the video extent and not from the application settings
Thanks for your reading

LePips:
I'll have this as a part of my video player refactor, but this may come long after that.

gray19950103:
> I'll have this as a part of my video player refactor, but this may come long after that.

is there any updates on this feature? Or is there any other way I can change the quality streaming?
In my case, I need to use Apple TV and Infuse dosen't provide quality menu, with my old non4K Apple TV, some large bitrates movie cause stuttering. (my iPhone & Mac works fine)
So I guess if I could change the transcode option to lower bitrate while playing on Apple TV, maybe will solve this problem.

teddyursa:
> > I'll have this as a part of my video player refactor, but this may come long after that.
> 
> is there any updates on this feature? Or is there any other way I can change the quality streaming with? In my case, I need to use Apple TV and Infuse dosen't provide quality menu, with my old non4K Apple TV, some large bitrates movie cause stuttering. (my iPhone & Mac works fine) So I guess if I could change the transcode option to lower bitrate while playing on Apple TV, maybe will solve this problem.

Did you have any luck with this? I'm struggling with 1080p content and it's frustrating there's not much chatter about the lack of a menu to help with buffering.

gray19950103:
> > > I'll have this as a part of my video player refactor, but this may come long after that.
> > 
> > 
> > is there any updates on this feature? Or is there any other way I can change the quality streaming with? In my case, I need to use Apple TV and Infuse dosen't provide quality menu, with my old non4K Apple TV, some large bitrates movie cause stuttering. (my iPhone & Mac works fine) So I guess if I could change the transcode option to lower bitrate while playing on Apple TV, maybe will solve this problem.
> 
> Did you have any luck with this? I'm struggling with 1080p content and it's frustrating there's not much chatter about the lack of a menu to help with buffering.

No, it seems you can only wait for quality select menu provided by Infuse or Swiftfin someday. Now I'm temporarily using this way: play with infuse on ios and airplay to Apple TV, but sometimes you can't lock the screen of ios device.

isvein:
Pops in to say I miss this too.

matthias-hoste:
Is it possible to give bounties for issues? Because I'd really like to see this feature realized

thornbill:
No bounties.

100tomer:
[...]

PR #1081, opened 2024-05-31
Video Player Cleanup

LePips:

PR #1147, opened 2024-07-07
App-Wide Bitrate Limit

JPKribs:
I was just on a trip and ran into an issue with the hotel WiFi. My Phone could DirectPlay using Swiftfin and my Server had the Bandwidth to DirectPlay but the WiFi I was on couldn't handle that level of bandwidth. The result was a stuttering playback until I remoted into the Server and lowered the server max bitrate.

This PR creates an App-Wide Bitrate Limit for all Playback sessions. I believe this solution should be used in addition to a setting found on the player overlay itself. The reason being there will be some clients that perpetually have poor bandwidth (Mobile Hotspots, Hotel Wifi, Remote Locations, etc.). This allows a single setting to create that limit instead of having to set it per playback session. Additionally, this creates the enum that can be used for the Player overlay down the road. I can look at that if this gets merged in.

Users will have the option of all of the same bitrate options found on Jellyfin-Web with an additional option for Maximum. Jellyfin-Web uses a bitrate option for 'Auto' that I am using as the default option. For Swiftfin, Auto will use the GET /Playback/BitrateTest API to get the duration that it takes to interact with the Server. From the duration and the test package size, it will automatically determine the available bitrate. Since the bitrate calculation is more accurate with larger test packages _(which may result in an increased delay prior to playback)_, I have included an option to manually set the test duration on a scale from 1mb to 10mb.

The Maximum setting is unique to Swiftfin compared to Jellyfin-Web but this will perform the historical behavior where it allows for a 360mbps bandwidth. This was the value formerly used as the tempOverkillBitrate so it should mirror the current state of Swiftfin.

All other bitrate options exactly mirror Jellyfin-Web.

By default, the configuration is **Auto** with a Test Duration of **Standard** (5mb). This configuration means that at runtime, a 5mb download test from the server will occur and the duration will be used to determine the bitrate.

If a user selects a setting that is NOT 'Auto', the Test Duration setting is hidden. Test Duration is only used if 'Auto' is selected. Any other bitrate selected is a static value that can be seen in the PlaybackBitrate enum.

<details>

<summary>How Auto Works</summary>

Screenshot 2024-07-12 at 20 51 08

</details>

### New Menus

**New Settings Menu:**

<details>

<summary>iOS</summary>

Simulator Screenshot - iPhone 15 Pro - 2024-07-12 at 20 40 35

</details>

<details>

<summary>TVOS</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-07-12 at 20 38 30

</details>

**Auto Configuration:**

<details>

<summary>iOS</summary>

Simulator Screenshot - iPhone 15 Pro - 2024-07-12 at 20 40 41

</details>

<details>

<summary>TVOS</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-07-12 at 20 39 49

</details>

**Non-Auto Configuration:**

<details>

<summary>iOS</summary>

Simulator Screenshot - iPhone 15 Pro - 2024-07-12 at 20 40 50

</details>

<details>

<summary>TVOS</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-07-12 at 20 40 13

</details>

**I believe this somewhat resolves: https://github.com/jellyfin/Swiftfin/issues/678 and https://github.com/jellyfin/Swiftfin/issues/184**

_Tested on iPhone 10S, AppleTV 3rd Gen via hardware, and the iPhone 15 Pro via emulator._

### Outstanding Items

- Is my method for Auto okay? Is there a better way to get the bandwidth to the server? Do we _WANT_ an Auto and a Maximum?


PJungkamp:
- It's kind of weird that the `maxBitrate` variable has the same name as the function `maxBitrate`. I'd suggest to rename the function to something like `testMaxBitrate` since it actually runs a test.
- You specify a default for the `testSize` parameter but explicitly override it to `5_000_000` at both call sites. I'd suggest to either make the `5_000_000` the default value and omit it at both call sites. Is there a reason behind that defaul value of `102_400`?

https://github.com/jellyfin/Swiftfin/blob/9fd521aa3b9f5673f19e411c45ca72f672910908/Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto%2BVideoPlayerViewModel.swift#L21

https://github.com/jellyfin/Swiftfin/blob/9fd521aa3b9f5673f19e411c45ca72f672910908/Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto%2BVideoPlayerViewModel.swift#L54

https://github.com/jellyfin/Swiftfin/blob/9fd521aa3b9f5673f19e411c45ca72f672910908/Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto%2BVideoPlayerViewModel.swift#L106-L127

---

I'm not affiliated with Swiftfin and have not contributed in any way to it, but I'd like this feature to be merged. Take my recommendations with a grain of salt!

JPKribs:
> - You specify a default for the `testSize` parameter but explicitly override it to `5_000_000` at both call sites. I'd suggest to either make the `5_000_000` the default value and omit it at both call sites. Is there a reason behind that defaul value of `102_400`?

102,400 is the default size for the GET /Playback/BitrateTest endpoint so removing this default would result in the same API call. I default it in the function to get rid of the build error/remove the need to use a Guard since I'm using that parameter as a denominator for the bitrate calculation. Can't divide by 0.

When I was testing, I found a quick 102kb test gave me very small bitrates. The longer the test (and the bigger the test size) the more accurate the bitrate calculation is. The downsize is this test is kind of just... 'dead bandwidth' so the greater the size, the more that adds up. The other downside is that a 100mb test is extremely accurate but creates a noticeable delay to playback starting.

The reason I made this a parameter then set it to 5,000,000B was from an idea I was throwing around to make the test size a setting as a well. Something like, when Auto is selected, a setting appears under for 'Test Duration' with Slow/Normal/Long... etc. at 1mb, 5mb, 10mb, 25mb, 50mb. So each of those would pull from that test setting instead of that static value.

In the end, I agree I should just change this to whatever the base test is/change the default but I'd be curious what other people's experience is. Does a ~5mb test work for everyone else or are we seeing that cause playback issues on slower networks?

fkribs:
Great work on implementing the app-wide bitrate limit! This feature will be incredibly useful for users dealing with inconsistent internet connections, like in hotels or remote areas. The addition of an 'Auto' setting for adaptive bitrate is a smart move and ensures smoother playback experiences. The consideration for future integration with player overlay settings is also appreciated. Well done!

JPKribs:
> > * You specify a default for the `testSize` parameter but explicitly override it to `5_000_000` at both call sites. I'd suggest to either make the `5_000_000` the default value and omit it at both call sites. Is there a reason behind that defaul value of `102_400`?
> 
> 102,400 is the default size for the GET /Playback/BitrateTest endpoint so removing this default would result in the same API call. I default it in the function to get rid of the build error/remove the need to use a Guard since I'm using that parameter as a denominator for the bitrate calculation. Can't divide by 0.
> 
> When I was testing, I found a quick 102kb test gave me very small bitrates. The longer the test (and the bigger the test size) the more accurate the bitrate calculation is. The downsize is this test is kind of just... 'dead bandwidth' so the greater the size, the more that adds up. The other downside is that a 100mb test is extremely accurate but creates a noticeable delay to playback starting.
> 
> The reason I made this a parameter then set it to 5,000,000B was from an idea I was throwing around to make the test size a setting as a well. Something like, when Auto is selected, a setting appears under for 'Test Duration' with Slow/Normal/Long... etc. at 1mb, 5mb, 10mb, 25mb, 50mb. So each of those would pull from that test setting instead of that static value.
> 
> In the end, I agree I should just change this to whatever the base test is/change the default but I'd be curious what other people's experience is. Does a ~5mb test work for everyone else or are we seeing that cause playback issues on slower networks?

I've updated this PR to include the variable test size. I have removed the default value and instead added a catch to ensure that the test size is greater than 0 to avoid the division errors. I've updated the screenshots to reflect the current state and move the bitrate options to their own screen to avoid cluttering the main settings section.

Any feedback is appreciated!

JPKribs:
### Update

I took a crack at adding this on the player overlay. That part was straightforward enough. Additionally, this value is filtered based on the App-Wide Setting created in this PR so far. At this time, these overlay options just set a published variable on the VideoPlayerManager but that variable doesn't do anything. I think I need some guidance on how to use this value.

<details>

<summary>App-Wide: Maximum</summary>

839F89BC-B759-4FB4-A867-144154977A2D_1_102_o

</details>

<details>

<summary>App-Wide: 10 mbps</summary>

77033BD3-FC05-4389-A2CF-681BC06B3AA1_1_102_o

</details>

The issue I've ran into is that all of the overlay items are for VLC specific functionality using the VideoManager.Proxy. Playback Speed, Subtitles, etc are handled by VLC from the file(s) from Jellyfin have already been provided. This setting would need to alter the connection to Jellyfin. By the time these menus exist, the DeviceProfile has already been created so there isn't an option to change it. At least not in the same manner that I have handled this PR so far.

I will not include these commits for the time being as changing the bitrate from the Player Overlay does not appear to be possible using the same method that I used for the App-Wide setting.

Shared/Objects/PlaybackBitrate/PlaybackBitrate.swift
```diff
@@ -0,0 +1,82 @@
...
+        case .kbps1500:
+            return L10n.bitrateKbps1500
+        case .kbps720:
+            return L10n.bitrateKbps720
+        case .kbps420:
+            return L10n.bitrateKbps420
+        }
+    }
+
+    static func filterValues(
```
LePips, line 68: What's the purpose of this?

LePips (changes requested):
Apologies for the long wait. This is going in a good direction and I'm good with where it is within settings for now.

> I took a crack at adding this on the player overlay.

I honestly wouldn't recommend doing that right now. Currently, this would work by creating a new `VideoPlayerViewModel` object and then telling our proxy to play from that, kind of like next/previous episodes. I have been meaning to refactor the video player a bit to take a queue, allow same-media restarts (this feature), combine the Live TV stuff, and other things in #1081 but just haven't gotten around to it.

Shared/Objects/PlaybackBitrate/PlaybackBitrate.swift
JPKribs: This Filter Values allows a built in function to filter the a picker to only include certain values. The purpose of this is related to future Video Player Overlay work. The reason for this would be to only show valid bitrates on the Player Overlay that do not conflict with App-Wide Setting.

For example, if the App-Wide setting had a cap of 80mbps, having an option for 120mbps or maximum would conflict with App-Wide setting. To prevent confusion, you could set your picker to pull from the value below to only include bitrates that were valid for the App-Wide setting:

`let filteredValues = PlaybackBitrate.filterValues(lessThanOrEqual: .mbps80)`

I have provided an example of this filter in the screenshots found here:

https://github.com/jellyfin/Swiftfin/pull/1147#issuecomment-2230003733

Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto+VideoPlayerViewModel.swift
```diff
@@ -100,4 +97,22 @@ extension BaseItemDto {
...
     }
+
+    private func getMaxBitrate() async throws -> Int {
+        let settingBitrate = Defaults[.VideoPlayer.appMaximumBitrate]
+
+        if settingBitrate == PlaybackBitrate.auto {
+            let settingBitrateTestSize = Defaults[.VideoPlayer.appMaximumBitrateTest]
+            let userSession = Container.shared.currentUserSession()!
+
+            let testStartTime = Date()
```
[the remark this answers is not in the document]
JPKribs: This is something I threw together myself. I like the idea of a default setting just finding your bandwidth to the server and using that which is why I included it. In my mind, this is the most user friendly option but I also don't see it on other clients so if we don't want to stray away I can remove it.

From what I understand, this endpoint is a test based on a package size. I could just be doing something wrong but I don't believe this actually returns anything. The details for this are here:

> https://api.jellyfin.org/#tag/MediaInfo/operation/GetBitrateTestBytes

Timing this appears to work pretty well for automatically finding your available bitrate. Obviously, a longer test works a lot better but in my testing a longer test caused bigger delay in playback so I made it a variable and put a disclaimer under the setting.

For me, this comes down to whether or not we want an auto feature. In Jellyfin-Web, Auto sets the MaxStreamingBitrate to 0 which _I think_ just means 'Anything Goes.' For my implementation, I set Auto to test the speed and find a bitrate that worked on this network. I then added Maximum to mirror the previous overkill bitrate or Swiftfin's 'Anything Goes' setting.

If we want to mirror Web, I can remove Maximum and change Auto to 360mbps.

Shared/Objects/PlaybackBitrate/PlaybackBitrate.swift
JPKribs: As a note, that comment image used a revision to include all bitrates when Auto was selected. If we don't want it, I won't include it but this was the revised version of this:

`    static func filterValues(
        lessThan max: PlaybackBitrate? = nil,
        lessThanOrEqual maxOrEqual: PlaybackBitrate? = nil
    ) -> [PlaybackBitrate] {
        self.allCases.filter {
            $0 == .auto || (
                ((max == .auto || maxOrEqual == .auto) && $0.rawValue >= 0) ||
                    (max != nil && $0.rawValue < max!.rawValue) ||
                    (maxOrEqual != nil && $0.rawValue <= maxOrEqual!.rawValue)
            )
        }
    }`

LePips: I'm gonna have to think about how this is actually an "app-wide maximum" vs just acting as a default. This would also probably be different for music when we get to it. Lastly, I would just do a `PlaybackBitrate.allCases.filter { $0 </<= maxBitrate }` instead.

JPKribs: If we want to allow the player overlay version of this (TBD) to override the App-Wide Setting this filter is not necessary. Alternatively, if we would rather just filter this ad-hoc, this filter is also not necessary.

This filter is only valuable if we want to filter bitrates so only valid ones appear on a picker AND we don't need to filter it ad-hoc per location it's used.

Completely your call on this! I just felt it would take me 5 minutes to add this now to save the trouble later.

Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto+VideoPlayerViewModel.swift
```diff
@@ -100,4 +97,22 @@ extension BaseItemDto {
             playSessionID: response.value.playSessionID!
         )
     }
+
+    private func getMaxBitrate() async throws -> Int {
+        let settingBitrate = Defaults[.VideoPlayer.appMaximumBitrate]
+
+        if settingBitrate == PlaybackBitrate.auto {
+            let settingBitrateTestSize = Defaults[.VideoPlayer.appMaximumBitrateTest]
```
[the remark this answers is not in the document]
JPKribs: I believe I did this? Since this was a defaults, I just moved this a level up? Let me know if that's not what you meant!

Shared/Objects/PlaybackBitrate/PlaybackBitrate.swift
JPKribs: > Lastly, I would just do a `PlaybackBitrate.allCases.filter { $0 </<= maxBitrate }` instead.

Gotcha! So regardless of if this is the Default Bitrate or the Maximum Bitrate, we do not need this filter?

JPKribs:
@LePips In the Shared/Resources there is a bitrate.json that captures the same thing as my enum. I'm not sure how to use a JSON file for this which is part of why I used an enum instead.

Is this Resources > Bitrate.json no longer needed? Should I remove it as part of this project? Alternatively, should that be the enum instead? I would need some guidance on using a json file in that way but I don't want to recreate this if it's not needed.

Shared/Objects/PlaybackBitrate/PlaybackBitrate.swift
JPKribs: This has been removed. For the time being, the maximum vs default bitrate conversion is not required since this is the only location that bitrate can be managed. When the player overlay bitrate is set, this can be revisited based on it's implementation there.

That is, if we want it to be maximum, we can filter the picker and if we want it to be default we just set the default. That decision can be made when 2 bitrate locations exist.

I will mark this as resolved as the code in question is no longer in this PR.

LePips: A default bitrate would be just that: the default bitrate selected when we create the playback URLs. Users can still select an alternate bitrate, even a higher one, if desired.

Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto+VideoPlayerViewModel.swift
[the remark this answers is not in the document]
LePips: At the moment, this all sounds good to me as I don't immediately have a better idea/algorithm for another automatic bitrate detection.

JPKribs:
> I'm good for approval after that has been done.

The Resource/bitrate.json file has been removed.

LePips (approved):

Issue #963, opened 2024-02-04
Screen reader compatibility 


AndrewH2004:
### Describe the bug

Hi,

im a visually impaired user of jellyfin who uses a screen reader (voiceover) on iOS and Apple TV.

the initial web wrapper app was accessible and the only reason I’m considering Swiftfin is that somehow on the jellyfin app it doesn’t work with h.265 playback.

however the problem I’m having with jellyfin is that my screen reader doesn’t announce what movie I’m focused on in the library just button, could a tag be added to these buttons with the films title. On iOS I can read the title underneath the button but on Apple TV it’s completely inaccessible along with any details of a title such as synopsis, ratings etc when I select the title.

### Application version

Current 

### Where did you install the app from?

App Store

### Device information

iPad 8th gen, iPhone SE 2, TV 4k 2nd gen

### OS version

iOS 17

### Jellyfin server version

10.8.11

AndrewH2004:
Any update?

Pa1yn-dev:
Pull request submitted to hopefully fix this issue.

Pa1yn-dev:
Pull request has been merged, please close this issue.

LePips:
I consider this a broader effort and would first like to establish some sort of guidelines for us before this is closed

Pa1yn-dev:
> I consider this a broader effort and would first like to establish some sort of guidelines for us before this is closed.

Sounds like a good idea, in order to keep the implementation standard across the package.

PR #1153, opened 2024-07-14
Update PosterButton.swift | Adding accessibility support to media posters in media section of TVOS app

Pa1yn-dev:
Added accessibility labels to enable voiceover on TVOS to read film titles on posters in media. I am not overly familiar with Swift as I do not develop for the Apple ecosystem, so apologies if this fix does not work as intended as I have no means to test the fix on hardware. 

This is in response to issue #963.

I would appreciate this fix (given it works) being pushed to the TVOS App as soon as possible.

### DIFF
--- a/Swiftfin tvOS/Components/PosterButton.swift
+++ b/Swiftfin tvOS/Components/PosterButton.swift
@@ -77,7 +77,7 @@ struct PosterButton<Item: Poster>: View {
                     }
             }
             .accessibilityLabel(item.displayTitle)
-            
+
             content()
                 .eraseToAnyView()
                 .zIndex(-1)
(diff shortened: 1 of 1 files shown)


LePips:
Apologies as I won't be able to test this for a while since VoiceOver isn't available in the tvOS simulator for testing. Additionally, tvOS is in a bad state and won't be shipped for a while.

Pa1yn-dev:
> Apologies as I won't be able to test this for a while since VoiceOver isn't available in the tvOS simulator for testing. Additionally, tvOS is in a bad state and won't be shipped for a while. However, I suspect everything looks good but I need to test accessibility first.

No worries, appreciate the response.

LePips (approved):
I'll approve and merge now. I'll be looking at tvOS more soon and if this doesn't work, I'll remove it.

LePips (approved):

PR #1071, opened 2024-05-26
[WIP] Clean up Localizations

JPKribs:
Discussed here: https://github.com/jellyfin/Swiftfin/issues/518

The goal of this PR is to:

1. Group labels into categories
2. Add descriptions of labels to the best of my ability (Some may need revisions later based on usage)
3. Remove unused labels
4. Ensure that the project still builds for both tvOS and iOS after labels are removed (see 3)

At this time, this is Draft PR since I am about 30% done. I am creating this now to keep others in the loop. Hoping to clean this up and get this done soon but I am putting this done for the day.

Issue #1034, opened 2024-04-19
Replace or work on Stinsen


LePips:
[Stinsen](https://github.com/rundfunk47/stinsen) is the great coordinator package we have used for view coordination for a while. However, it hasn't seen active development for a long time and I will soon be incorporating some design elements that the package can't handle (eg: dynamic tabs). I have personally contributed to the project and even have merge permissions, but it's a somewhat large project to refactor. Also, it doesn't support the split column view for iPadOS which is why iPadOS has a tab bar instead.

We need to replace or work heavily on Stinsen so that we can support more design features. I would still heavily, heavily want to be using the coordinator behavior. Looking up `swiftui coordinator` on GitHub there are a few other packages to investigate.

bennyyy999:
Was there a reason for the coordinator pattern originally? Or is there a statistic of how many people use the app with iOS 15? We could modernize with a more native approach ([NavigationStack iOS 16+](https://developer.apple.com/videos/play/wwdc2022/10054/)) and like Azam Sharp explained [here](https://azamsharp.com/2023/02/28/building-large-scale-apps-swiftui.html#navigation). It kinda follows the coordinator syntax in my opinion and already had great success in my own apps (smaller apps) and at work (bigger app).

Just a suggestion and would love to hear your opinion on that!

LePips:
I just like the coordinator pattern and it makes navigation really easy, nothing more. On an opinionated note, I never really liked `NavigationLink` and personally find it baffling that there isn't a coordinator equivalent by Apple in SwiftUI.

There wasn't an iOS version restriction per say and whatever solution we use could utilize some back-ported `NavigationStack` APIs for version compatibility. Thank you for sharing that article as I haven't seen it before and am taking it into consideration.

I may take a look at making my own solution, especially since I want to see if I can utilize [Transmission](https://github.com/nathantannar4/Transmission) for some custom transitions.

chickdan:
As far as I can tell, [PresentationDetents](https://developer.apple.com/documentation/swiftui/view/presentationdetents(_:)) and Apple's new [zoom navigation](https://developer.apple.com/documentation/swiftui/zoomnavigationtransition) in iOS 18 can accomplish much of what Transmission offers in terms of navigation handling/animations.

[...]

PR #1158, opened 2024-07-26
[tvOS] Customize Home Tabs / Hide Tab Labels

JPKribs:
_**Relies on: https://github.com/jellyfin/Swiftfin/pull/1161**_

This PR sets out to add some customization to the Navigation Bar for tvOS. I have not included the same customization for iOS as well but I don't see that being valuable since Home, Search, and Media are all pretty much requirements for iOS. tvOS, this feature is supposed to help with scenarios like 'I only have movies, why do I need a TV Shows section?' or 'I don't want the Home screen first.' 

Additionally, this PR adds the ability to turn on/off the section labels, mirroring the setting in Jellyfin-Expo. This feature exists in both tvOS and iOS

### tvOS

<details>

<summary>Setting Location</summary>

Screenshot 2024-08-01 at 23 36 32

</details>

<details>

<summary>Settings</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-08-01 at 23 19 37

</details>

<details>

<summary>Settings  Re-Arrange Menu</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-08-01 at 23 19 56

</details>

<details>

<summary>Re-Arranged With Labels</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-07-25 at 22 24 32

The tabs are re-arranged. Search has been removed. 6 Tabs is the maximum number of tabs that can exist on the screen without overflowing.

</details>

<details>

<summary>Re-Arranged Without Labels</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-07-25 at 22 25 21

The tabs are re-arranged and the labels have been turned off. Although Settings was turned off, it is always added back onto the end of the list if it is not included. This is to prevent users from turning off Settings then not having the Settings tab so they can't add it back. If the Settings exists in the list, it will be placed where it was listed. Otherwise, it will be placed at the end.

</details>

### iOS

<details>

<summary>Settings</summary>

Simulator Screenshot - iPhone 15 Pro - 2024-08-01 at 23 24 24

</details>

<details>

<summary>Labels On vs Off</summary>

Screenshot 2024-08-01 at 23 35 01

</details>

### Outstanding Items

 - I mirrored the OrderedSectionSelectorView into tvOS and it works almost perfectly. The only issue is the whole view freezes up once you move a single item in edit mode. To get around this issue, I just have it end edit mode after moving 1 item. It's not terrible but it does mean you can only move a single item at a time. The ideal solution would be figure out why it locks up after moving a single item but I've stared at it for too long and I'm out of ideas. 

 - I've added a check to refresh the MainTabCoordinator on tvOS when the HomeLabels or HomeSections are changed. This seems to work well but I don't know if I did it the right way... I did this by making the MainCoordinator observable and making it refresh the MainTabCoordinator underneath it. This seems to work but to make it refresh the home view, I added a `self.objectWillChange.send() `as part of the refresh. Let me know if there is a cleaner way to do this!

 - **Localizations are not done since I wanted to make sure that a) I was doing this right and b) this wasn't going to cause some issues. If we decide to move forward with this but I would ask that https://github.com/jellyfin/Swiftfin/pull/1071 gets merged first so I can build on that layout instead of the current version on Main.**
 
 ### Outside this PR
 
~~Any nested Settings views cannot be reached on tvOS right now. This existed previously with the Indicators menu prior to this PR but this PR is impacted. This appears to be related to the SettingsCoordinator not working from anywhere but the SettingsView page?~~

This issue has been resolved by https://github.com/jellyfin/Swiftfin/pull/1161! All Customize settings that were nested are functional.


Shared/ViewModels/LibraryViewModel/ItemLibraryViewModel.swift
JPKribs, line 1: This was a fat finger. This will not be included on a final version.

LePips (changes requested):
I'm glad this is taking shape and thank you for starting this.

1 - It's good that this could initially be done without #1034 but I would prefer we do that first so that tabs can be updated live instead of requiring a restart of the app. The attempt at adding `Observable` (just adding the protocol doesn't do anything) with `Published` is incorrect and shouldn't be/can't be mixed together.

2 - I would actually like this on iOS as well. Users would be able to put Live TV options as tabs, specific libraries, or other views I'm sure we'll make in the future. This specifically would be more complex since we would have to check Live TV and other access permissions if they are no longer valid. I will be implementing something similar for the SuperUser and local user customization, so I'll be hand-wavy and say let's wait to see what I do there.

3 - I had the thought to make the first tab, regardless of what it could ever be, to have the settings button. This would remove its strong relationship with `Home`.

4 - I'll have to think about how we implement the re-arranging menu.

JPKribs:
> 1 - It's good that this could initially be done without #1034 but I would prefer we do that first so that tabs can be updated live instead of requiring a restart of the app.

I will go ahead and leave this as a draft until #1034 is addressed! This PR will be more of a reference since I'm sure there will be some other changes required to make this work when that is changed.

> 2 - I would actually like this on iOS as well. Users would be able to put Live TV options as tabs, specific libraries, or other views I'm sure we'll make in the future. This specifically would be more complex since we would have to check Live TV and other access permissions if they are no longer valid. I will be implementing something similar for the SuperUser and local user customization, so I'll be hand-wavy and say let's wait to see what I do there.

I had this completed and in up until https://github.com/jellyfin/Swiftfin/pull/1158/commits/55a3d253bc1af9fb1ea381d9fef6c797be46f04c. I'll go back and re-add this to this PR so it's there when I pick this back up again.

> 4 - I'll have to think about how we implement the re-arranging menu.

Outside of the one bug/workaround, I like the re-purposed implementation of `OrderedSectionSelectorView` for this purpose. I'd like to get that view in tvOS regardless of this PR since I think it's going to be very helpful for things like Enabling/Disabling filters on tvOS.

JPKribs:
Alright! This *should* be ready. Localizations and live updating the main tabs need to be done still but I will wait on https://github.com/jellyfin/Swiftfin/issues/1034 before I pick this up again.

chickdan:
IMO this is a small enough change to not need to wait on #1034, having to restart the app is a little annoying but theoretically a user will only be making changes once or twice. Including Settings as a tab option and adding an alert to inform users the app will need restarted would make for a solid MVP. Though I'm someone who prefers to get things in and make iterative improvements.

JPKribs:
> IMO this is a small enough change to not need to wait on #1034, having to restart the app is a little annoying but theoretically a user will only be making changes once or twice. Including Settings as a tab option and adding an alert to inform users the app will need restarted would make for a solid MVP. Though I'm someone who prefers to get things in and make iterative improvements.

@chickdan 

I wouldn't personally be opposed to that. That being said, I think this PR is pretty out of date now so resolving conflicts on in it's current PR is gonna be a chore. In addition to this, I think one of the direction @LePips was referring to was the ability to pin something like a specific library. Which, I didn't account for in this version.

I'm game to pick this back up but I'm going to be throwing a lot of my time of Admin Dashboard items (https://github.com/jellyfin/Swiftfin/pull/1271) for a bit so if you wanted to take anything valuable from this PR and work on this I would not be offended at all! Just lemme know and I can close this one out to avoid confusion. Otherwise, once I get all of https://github.com/jellyfin/Swiftfin/pull/1271 completed, split out, and done I can take another look at this.

PR #1161, opened 2024-07-29
[tvOS] Fix Indicators Setting Unreachable

JPKribs:
### Issue Resolved by PR

In the Main branch, the Indicator button exists nested in the Settings > Customize. Because it is nested, the Settings.Router didn't seem to know what to do with it. On the current version of Main, selecting Indicators doesn't do anything. If you move the Indicators button to the main SettingsView, it's usable but it's not usable from Settings > Customize.

To resolve this, I created a CustomizeSettingsCoordinator that is activated in the same way that the existing SettingsCoordinator is when you call on SettingsViews. In this case, CustomizeSettingsCoordinator only contains the Indicators view at this time. But, in https://github.com/jellyfin/Swiftfin/pull/1158 I am hoping to add a Home Sections menu to select which Tabs you want for tvOS. This CustomizeSettingsCoordinator works in this case as well.

### Testing

Pull the current Main Branch and try selecting this button and notice that it does not route you to the Indicators selection view:

<details>

<summary>Settings > Customize > Indicators</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-07-28 at 22 32 13

</details>

Using this PR, you will be able to select this and it will bring you to this view:

<details>

<summary>Indicators View</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-07-28 at 22 29 39

</details>

### Cause

I believe back in an earlier version, all of the settings that routed to other views existed on the main Settings Page. My guess is this was moved to the nested Settings > Customize Page. Since this version isn't on the App Store, the only people who would have noticed / tested probably aren't worried about setting Indicators so it went unnoticed.


LePips (approved):
Looks good! 👍

PR #1163, opened 2024-07-31
[tvOS] Settings Cleanup

JPKribs:
### Goal

This PR originally started with the goal of fixing a slight peeve I had with the settings. All buttons in Forms would clip into their boundaries when focused. See:

Button Clipping

Following the changes here: https://github.com/jellyfin/Swiftfin/pull/1163#pullrequestreview-2217218184, I've removed the clipping. Additionally to the clipping, I have added labels to the settings items and updated it to ensure that all sections have headers.

### Settings 

I created a custom SettingsViewForm and SettingsViewFormSection to handle the Form and Section actions from a single location. The goal was to centralize the button/form configuration in single struct so they could all be updated if there are any customization that we want to apply to all sections. These settings can be seen reflected here:

<details>

<summary>Settings Main</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-08-03 at 17 14 57

1. Updated the Switch User button to indicate that it will take you to another page using the Chevron button.
2. Added the Jellyfin Section category for Jellyfin Server items (First Section)

</details>

<details>

<summary>Server Details</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-08-03 at 17 12 38

1. Created a Checkbox Button that mirrors the Chevron Button but has a variable for true/false. This removes the need to declare the button twice as checked/unchecked while also move the checkmark to more closely resemble the Chevron buttons used elsewhere in settings.

</details>

<details>

<summary>Video Player</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-08-03 at 17 12 54

1. Change Resume Offset button Offset in the Resume category. 
2. Create Section Headers to better split out content since the Header/Footer can start looking a little weird when there are footers at the end of a section but no headers for that section.

</details>

<details>

<summary>Maximum Bitrate</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-08-03 at 17 13 01

</details>

<details>

<summary>Customize</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-08-03 at 17 13 09

</details>

<details>

<summary>Experimental</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-08-03 at 17 13 16

</details>

This last section was not part of the Setting Form Changes but instead just a padding(.horizontal) to resolve this clipping.

<details>

<summary>Server Selection</summary>

Settings

</details>


LePips (changes requested):
Thank you for taking a look at this and the localization but I apologize to say that the clipping issue could just be fixed [1] with the following in `SplitFormWindowView` and attached to other `Form`s [2] if necessary:

```swift
Form {
    contentView()
        .eraseToAnyView()
}
.scrollClipDisabled()
```

Additionally, this keeps the `Form` styling instead of moving to `Button`s in a `VStack`.

---

[1] That modifier came in tvOS 17 which I just now set as the minimum but I would have pursued a way to prevent the clipping somehow with `overlay` or whatever and still use `Form`.

[2] The entire server selection will be changed to use a `Menu` which was introduced in tvOS 17, so I wouldn't worry about my ugly replication.

JPKribs:
```
Form {
    contentView()
        .eraseToAnyView()
}
.scrollClipDisabled()
```
That is so much cleaner of a solution than this whole build out! I can't believe I didn't find this when I researching this. Let me re-submit this with just that single line change.

JPKribs:
> Thank you for taking a look at this and the localization but I apologize to say that the clipping issue could just be fixed [1] with the following in `SplitFormWindowView` and attached to other `Form`s [2] if necessary:
> 
> ```swift
> Form {
>     contentView()
>         .eraseToAnyView()
> }
> .scrollClipDisabled()
> ```
> 
> Additionally, this keeps the `Form` styling instead of moving to `Button`s in a `VStack`.
> 
> [1] That modifier came in tvOS 17 which I just now set as the minimum but I would have pursued a way to prevent the clipping somehow with `overlay` or whatever and still use `Form`.
> 
> [2] The entire server selection will be changed to use a `Menu` which was introduced in tvOS 17, so I wouldn't worry about my ugly replication.

@LePips I've moved back to just using the standard form items again with that one `.scrollClipDisabled()` change. This resolves the exact peeve I had with this so I am happy with the change. I've left in my other items of adding Labels and Section headers but let me know if you want me to drop those as well. 

Lastly, I added that padding to the Server Selection Form. Let me know if you want me to keep the padding or if that should also just be `.scrollClipDisabled()`.

I've updated the original post with screenshots to reflect the current state of this PR!

Swiftfin tvOS/Views/ServerDetailView.swift
```diff
@@ -35,36 +35,49 @@ struct EditServerView: View {
                     .frame(maxWidth: 400)
             }
             .contentView {
-                Section(L10n.server) {
-                    TextPairView(
-                        leading: L10n.name,
-                        trailing: viewModel.server.name
-                    )
+
+                Section {
```
LePips, line 39: Please use the other init instead. I know there are many places throughout that use the other one but I've been moving to this init instead.

```swift
Section(L10n.server) {
    // ...
}
```

Swiftfin tvOS/Views/ServerDetailView.swift
```diff
@@ -35,36 +35,49 @@ struct EditServerView: View {
...
+                    Button {} label: {
```
LePips, line 40: We don't need just the server name to be focusable, but the user on the settings view did in order to have better scrolling to/from the tab bar.

Swiftfin tvOS/Views/SettingsView/IndicatorSettingsView.swift
```diff
@@ -34,13 +34,15 @@ struct IndicatorSettingsView: View {
 
                 Section {
```
LePips, line 35: Since you're here, please use the other init.

Swiftfin tvOS/Views/SettingsView/MaximumBitrateSettingsView.swift
```diff
@@ -24,21 +24,19 @@ struct MaximumBitrateSettingsView: View {
                     .frame(maxWidth: 400)
             }
             .contentView {
+
                 Section {
```
LePips, line 28: Since you're here, please use the other init for the header. Since the footer has a conditional, it does not need to be changed,

Swiftfin tvOS/Views/SettingsView/SettingsView.swift
```diff
@@ -51,14 +51,26 @@ struct SettingsView: View {
...
+
+                            Spacer()
+
+                            Image(systemName: "chevron.right")
+                                .font(.body.weight(.regular))
+                                .foregroundColor(.secondary)
+                        }
                     }
+
+                } header: {
```
LePips, line 67: Since you're here, please use the other init and you can do more throughout this file if you would like to.

Swiftfin tvOS/Views/SettingsView/VideoPlayerSettingsView.swift
```diff
@@ -43,46 +43,54 @@ struct VideoPlayerSettingsView: View {
             .contentView {
 
                 Section {
```
LePips, line 45: Since you're here, please use the other init for both header and footer.

Swiftfin tvOS/Views/SettingsView/VideoPlayerSettingsView.swift
```diff
@@ -43,46 +43,54 @@ struct VideoPlayerSettingsView: View {
...
                         isPresentingResumeOffsetStepper = true
                     }
+                } header: {
+                    L10n.resume.text
                 } footer: {
-                    Text("Resume content seconds before the recorded resume time")
+                    L10n.resumeOffsetDescription.text
                 }
 
                 Section {
```
LePips, line 60: Since you're here, please use the other init for both header and footer.

Swiftfin tvOS/Components/SplitFormWindowView.swift
```diff
@@ -31,6 +31,7 @@ struct SplitFormWindowView: View {
             .if(descriptionTopPadding) { view in
                 view.padding(.top)
             }
+            .scrollClipDisabled()
```
LePips, line 34: This does fix the issue like I said earlier but now introduces a different clipping behavior for long lists when cells clip out. The documentation comment for this modifier has a solution for custom clipping behavior so we'll have to take a look at something like that later with a desirable gradient mask.

Swiftfin tvOS/Views/SettingsView/VideoPlayerSettingsView.swift
```diff
@@ -43,46 +43,54 @@ struct VideoPlayerSettingsView: View {
...
-                    description: "Resume content seconds before the recorded resume time",
-                    value: $resumeOffset,
-                    range: 0 ... 30,
-                    step: 1
-                )
-                .valueFormatter {
-                    $0.secondLabel
+                    Toggle(L10n.pauseOnBackground, isOn: $pauseOnBackground)
+                    Toggle(L10n.playOnActive, isOn: $playOnActive)
+                } header: {
```
LePips, line 75: Please use other init.

Swiftfin tvOS/Components/CheckButton.swift
```diff
@@ -0,0 +1,62 @@
...
+// Swiftfin is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2024 Jellyfin & Jellyfin Contributors
+//
+
+import SwiftUI
+
+struct CheckButton: View {
```
LePips, line 11: I don't feel like this serves much purpose in the area that it's currently being used. I would also prefer we move to pickers for select-able items, especially now that we can use `Menu` or `pickerStyle(.menu)`.

LePips (changes requested):

Swiftfin tvOS/Views/ServerDetailView.swift
JPKribs: Leaving this as button still allows the cursor to go over it but it doesn't focus correctly. I changed this from a button to just the view and I believe this works as intended now.

Swiftfin tvOS/Views/SettingsView/MaximumBitrateSettingsView.swift
JPKribs: For the Footers I am a little lost. Here is what I am seeing as options for Footer usage:

Screenshot 2024-08-06 at 13 43 51

The version I think you are looking for looks like this? But I am getting some syntax errors:

Screenshot 2024-08-06 at 13 44 22

The original version looks like this if we want to retain this. Otherwise, I am more than happy to change this to keep everything consistent!

Screenshot 2024-08-06 at 13 44 50

Swiftfin tvOS/Views/SettingsView/VideoPlayerSettingsView.swift
JPKribs: I believe this is the same item as https://github.com/jellyfin/Swiftfin/pull/1163#discussion_r1706031863 so I will base this change on that discussion!

Swiftfin tvOS/Components/SplitFormWindowView.swift
JPKribs: Sorry for my confusion. Do you want me to leave this in, take this out, or would you look me to look into custom clipping behavior? Happy to help out wherever I can!

Swiftfin tvOS/Views/SettingsView/VideoPlayerSettingsView.swift
JPKribs: I believe this is the same item as https://github.com/jellyfin/Swiftfin/pull/1163#discussion_r1706031863 so I will base this change on that discussion!

Swiftfin tvOS/Views/SettingsView/MaximumBitrateSettingsView.swift
LePips: Ah I apologize as I must have been mixed up with the inits with footers. We can leave this as is.

Swiftfin tvOS/Components/SplitFormWindowView.swift
LePips: Leave this in to fix the original clipping issue. The new clipping issue was something I just noticed while testing this and we can fix later.

Swiftfin tvOS/Views/SettingsView/VideoPlayerSettingsView.swift
LePips: https://github.com/jellyfin/Swiftfin/pull/1163#discussion_r1706053161

Swiftfin tvOS/Views/SettingsView/VideoPlayerSettingsView.swift
LePips: https://github.com/jellyfin/Swiftfin/pull/1163#discussion_r1706053161

Swiftfin tvOS/Components/CheckButton.swift
JPKribs: Oh nice the Menu items look a lot better! I tried to take a crack at this. Is this kind of what you were looking for:
Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-08-06 at 15 01 03

JPKribs: I've included this working version on my latest commit. It's rough around the edges with now and I don't like how the box has a larger button box around it but for the menu specific drop down I think that shows the basic of what I think you're suggesting.

JPKribs:
@LePips It looks like I can for the SettingView > Username that I did here: https://github.com/jellyfin/Swiftfin/pull/1163#discussion_r1705929277

The focus appears to work as we expect it to but I would agree it feels a little weird. Let me know if you want me to include this here:

https://github.com/user-attachments/assets/a823f92d-2201-4c9e-b1a3-4affc1ace8eb

LePips (approved):
Everything looks good as is. The user button should be changed to match iOS soon and I'll probably clean up some other things here as well. I'm experiencing some bugs in the iOS 18 simulator around losing focus when popping a settings view but that can be fixed later.

Thanks for doing this work and going through my requests!

PR #1165, opened 2024-08-02
Media Item HD Attribute Source

chickdan:
Clearing the TODO that suggests using the stream to determine if content supports HD. As a fallback I made the conditional an OR statement just to be safe.

### DIFF
--- a/Swiftfin/Views/ItemView/Components/AttributeHStack.swift
+++ b/Swiftfin/Views/ItemView/Components/AttributeHStack.swift
@@ -23,7 +23,8 @@ extension ItemView {
                 }
 
                 if let mediaStreams = viewModel.selectedMediaSource?.mediaStreams {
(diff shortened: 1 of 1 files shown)


LePips (changes requested):
I was going to ask "please do tvOS as well" but I wonder if past me looked at the [server](https://github.com/jellyfin/jellyfin/blob/a82527dcdff2082131904c41ce0cda66ed8a69d2/MediaBrowser.Controller/Entities/BaseItem.cs#L378), using only used the streams for tvOS and didn't do the same in iOS. `isHDVideo` differs from the server since I used `1920x1080` but I don't particularly care what definition we use.

So, let's just do what I did for tvOS and just do the streams instead. I'm too lazy to see what `BaseItem.Height` reflects on the server but my assumption is that it's from the streams.

LePips (approved):

PR #1173, opened 2024-08-08
Fix iOS 15 Rotation

### LINKED ISSUE #1115: Rotation not working (not switching to landscape mode iOS)
### Describe the bug

On the Swiftfin iOS app it doesn’t go into landscape mode when you launch a video and rotate your phone it. It works when you launch the app while rotated into landscape mode but when you rotate again it goes back to portrait mode and stays there and doesn’t switch back.

### Application version

1.1(78)

App Store

### Device information

iPhone 7

### OS version

iOS 15.8

### Jellyfin server version

10.9.4

LePips:
SwiftUI on iOS 15 has a different view hierarchy and doesn't work with `PreferencesView` nicely. I tried a few things and this was the least intrusive.

- Closes #1115


LePips:
The changes to `AppDelegate` actually breaks for iOS 16+. Found a different way to using `PreferencesView` in #1174

PR #1177, opened 2024-08-09
[WIP] [tvOS] Letter Picker

JPKribs:
### Summary

Takes the same logic from the LetterPickerBar on iOS and implements it to tvOS.

<details>

<summary>Image</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-08-09 at 14 04 11

</details>

<details>

<summary>Video</summary>

https://github.com/user-attachments/assets/a52ef248-4c10-4180-b5ea-17a3bfa0aa23

</details>

### Outstanding Issues

1. On Main, tvOS PagingScrollViews jump their focus to the top of the page whenever a new page of data is loaded. Since the Letter Picker is the tallest item, this is now capturing the focus. I think this is more of an issue with the PagingLibraryView and knocking out that TODO. _(I tried but I couldn't get figure it out)_
2. If you are hovering a letter that is above, below, or between posters, it cannot jump to them. Instead, you have to go back up to a letter that in even to a poster to move the cursor back to the posters.
3. This only works where there are filterViewModels. This means the MainTab Movies/TV Shows cannot have a Letter Picker on them. Only items in Media section can have it. I didn't want to dig too far into adding FilterViewModels to the ItemTypeLibraries. Partially because I'm not sure how but also do we want the Item Types to have filters on them? 

### Outstanding Discussions

- I made the letters white with a shadow for visibility. Let me know if there are any issues with this.
- Since I am able to re-use the LetterPickerBar, I moved this to Shared but the LetterPickerButtons are OS specific. Let me know if it would be better to have a separate bar per OS.


LePips:
I won't be looking at tvOS for a bit and will have to think about the design and interaction of this.

JPKribs:
There isn't much actual work here so this is more of a 'discussion about tvOS filters' which will likely end up in the broader tvOS discussion. So, I'm going to close this out to help avoid PR clutter!

PR #1184, opened 2024-08-14
[tvOS] Fix Episode Description Size & Provide Episode Thumbnail Clarity

JPKribs:
### Ensures that the episode description matches the size of the thumbnail:
Fix Description Spacing

### Provides some additional clarity that selecting the episode thumbnail _plays_ the episode.
Episode Play Clarity
_This comes from an issue where I select the thumbnail thinking it will take me to the episode details. Focusing the thumbnail will now overlay the play icon to let the user know this will play the episode._

_Any extra formatting comes from the "Format File" option from SwiftFormat_


LePips (approved):
Looks great to me, good idea for the play button on focus.

PR #1185, opened 2024-08-14
[tvOS] Episode/Season Focus gets Stuck

JPKribs:
### Issue
When entering a Show, focus can get trapped in the 'Seasons' HStack along the top. This can be reliably recreated by entering a Show, and changing the season prior to focusing on an episode. Because there is no 'LastFocusedEpisodeId', you are stuck until you are able to focus an Episode.

### Resolution
This PR resolves this by adding a check to see if the LastFocusedEpisodeId exists in the current ViewModel. If the LastFocusedEpisodeId does not exist in the current ViewModel, it will always default to the first element in the current ViewModel.

If the LastFocusedEpisodeId does exists in the current ViewModel, it is used. This keeps the same functionality as before where moving between the episodes and seasons will always return you to the same episode you were on prior.

### Focus Cursor Getting Stuck
https://github.com/user-attachments/assets/9eb7508e-13b9-4a04-976d-4a590e4901cc

### Resolved
https://github.com/user-attachments/assets/89f49cdd-de17-4ab0-ac65-b1ba99e2142e


LePips (changes requested):
Hm so this issue is a bit more complex because this view isn't complete yet (by yours truly). For example, this view doesn't work correctly in the loading or error state.

We would have to play around with observing the `viewModel.state` and changing what should be focused when that changes. For example, the following fixes this immediate issue if you put this in the end of `EpisodeHStack.body`:

```swift
.onChange(of: viewModel.state) { _, newValue in
    if newValue == .content {
        lastFocusedEpisodeID = viewModel.elements.first?.id
    }
}
```

JPKribs:
Good call! That resolves the issue I was seeing. I've changed this PR to only that change!

LePips (approved):

Issue #1166, opened 2024-08-03
Play/Pause


bitzerjd99:
### Describe the bug

I'm using Allonis as a Remote control system and it uses the standard Play/Pause key.  The key is mapped to Apple's Play/Pause commands and works throughout the other ATV apps (Infuse, Plex, for example).  While using Swiftfin they don't work at all.  Is there some other command I should be using?  

### Application version

Just installed, so the latest version on the App Store.

### Where did you install the app from?

App Store

### Device information

Apple TV 4K (Original Version)

### OS version

Latest (ATV says it is up to date)

### Jellyfin server version

10.9.8

vywei:
I'm also experiencing the same on ATV remote Version 1.1.1 (1.1.1) play pause. 
https://github.com/bsharper/atv-desktop-remote/releases

play pause works fine on other apps so don't think it's the above app.

winterpetrichor:
Having this problem on an LG tv as well with An MR23GA magic remote. Button works in all other apps but this one. On this remote it's the button in the middle of the d-pad.

reconbot:
I'm experiencing this with AirPods when I squeeze the airpod to pause/play the media. Guessing it's related.

PR #1186, opened 2024-08-15
Listen for AirPods connection changes

### LINKED ISSUE #1160: When AirPods are removed, media continues playing out loud
### Describe the bug

1. Starting playing media with AirPods in
2. Remove airpods while media is playing
3. Witness media continuing to play out loud, and not automatically pause. (This is a break in behavior with every other popular app such as YouTube, Netflix, Hulu, etc.)

### Application version

1.1(78)

App Store

### Device information

iPhone 15 Pro

### OS version

iOS 18 dev beta 4, (this also happened on iOS 17)

### Jellyfin server version

10.8.13

chickdan:
This is by no means the cleanest solution so input is most welcome (primarily on where to put this). Most of this was taken right from Apple's [docs](https://developer.apple.com/documentation/avfaudio/responding_to_audio_route_changes) and from what I've tested it works nicely. 

With this being shared code I do have some concern as I cannot test against an AppleTV. Additionally, I am unsure what else may trigger the `AVAudioSession.routeChangeNotification`.

Closes #1160

### DIFF
--- a/Shared/ViewModels/VideoPlayerManager/VideoPlayerManager.swift
+++ b/Shared/ViewModels/VideoPlayerManager/VideoPlayerManager.swift
@@ -11,6 +11,7 @@ import Combine
 import Defaults
 import Foundation
 import JellyfinAPI
+import MediaPlayer
 import UIKit
 import VLCUI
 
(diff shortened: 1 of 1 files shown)


LePips (approved):
I've thought about making nice wrapper framework for all of the now playing/headphones stuff since they're older APIs and could have some SwiftUI bridges. I'm fine with this currently since I'll be refactoring the entire `VideoPlayerManager` layer.

I'll be able to test and approve later.

chickdan:
Sounds good! Thanks for taking a look!

chickdan:
@LePips I reworked this to use the `MPRemoteCommandCenter` instead which greatly simplifies the code, [Apple docs](https://developer.apple.com/documentation/mediaplayer/mpremotecommand). Hopefully this change will also close #1166.

LePips (approved):
This works and I've been working in the `MPRemoteCommandCenter` anyways for all the now playing stuff 👍

PR #1203, opened 2024-08-19
Video Player Manager Refactor

LePips:
Was originally just implementing the now playable solution but I was discontent with `VideoPlayerManager` and had the intention to refactor it anyways, so I'm just doing that now.

Very much copying solution from [Apple sample app](https://developer.apple.com/documentation/mediaplayer/becoming_a_now_playable_app). These APIs are archaic and this is just the easiest.


JPKribs:
Out of curiosity, will part of this refactor include the TODO to move the playback reporting logic? I only ask because Native Player playback reporting seems to be hit or miss. Particularly when the app is closed instead of closing the player more gracefully in-app.

I was going to look at it but I can hold off if that's an area you are going to be interacting with.

LePips:
Yes, my intention is to fix the playback reporting and some other discrepancies between the players with this as well.

JPKribs:
I'm probably behind the curve here but I was looking at this today:

https://developer.apple.com/documentation/avkit/customizing-the-tvos-playback-experience

Do you think we have any appetite for this for the `Native Player` on tvOS or is the goal to make the `AVKit` implementation as minimal as possible for maintainability?

Not implying this should be part of this PR! Just thought I would ask before I started looking at this more seriously.

LePips:
At least here, I only care about implementing the same capabilities on iOS and tvOS which mostly means only providing the additional metadata. However since tvOS does allow that additional customization with buttons I'm not necessarily against their implementation but I will not personally actively seek to implement them. If someday Apple allows the same customization on iOS by publicizing their private API, I wouldn't necessarily be against it there either.

- https://github.com/MTACS/iOS-17-Runtime-Headers/blob/d1d960dfaa4107765dd7fcf891e4967c0930d5fd/Frameworks/AVKit.framework/AVPlayerViewController.h#L49

LePips:
I had really anticipated for this to make the 1.3 release but I had unexpected exciting changes come up, which has taken much of my time. With other big changes expected like dropping iOS 15 and updating the very outdated SDK, I have made the decision that this will not make the 1.3 release or make it to iOS 15 devices.

LePips:
Replaced by #1581.

Issue #221, opened 2021-11-10
Implement mpv Player


PangMo5:
I think VLCkit is really good, too.
But it'd be better to change to a better performance and more scalable [mpv](https://mpv.io), so I'm going to leave an issue.

As far as I know, plex is also using mpv.
And https://github.com/jellyfin/jellyfin-media-player too.

Of course, I'm currently completely ignorant of mpv. lol

LePips:
I'll consider this in my video player refactor.

LePips:
Actually, building mob for iOS isn't officially supported and looks complicated. We shouldn't even consider it.

https://github.com/mpv-player/mpv/issues/5902

PangMo5:
> Actually, building mpv for iOS isn't officially supported and looks complicated. We shouldn't even consider it.

That's too bad.

LePips:
What performance issues do you have with VLCKit?

PangMo5:
> What performance issues do you have with VLCKit?

When I changed the time scroll bar(?) offset when Playing a specific video(MPEG-4, AVI), I felt that it was much slow than other video players.
And there is a crash when playing a video including a specific subtitle(.kor.srt).

But seeing how it works normally through the VLCPlayer in the App Store,
There seems to be a problem with the implementation of Swiftfin.
In the future, I'll create issues related to this problems again if the same occurs after VideoPlayer refactoring. (Not about "using mpv")

PangMo5:
VLC4 will be released soon, so we can expect better performance 😃

LePips:
I may want to investigate some time as I'm interested in attempting to build for iOS/tvOS with some sources:
- https://github.com/ldwardx/mpv-build-mac-iOS
- https://github.com/mpv-player/mpv/issues/8362
- https://github.com/mpv-player/mpv/issues/5331

At first I wouldn't look at entirely replacing VLCKit but instead creating another player based on mpv and then have the option to render with either of them.

anthonylavado:
Here's also something for the future: https://github.com/manhpham90vn/MPVPN

LePips:
Oh ha that isn't an MPV implementation, but some VPN implementation.

LePips:
Gonna reopen as now this seems like something we need to do and eventually replace (or demote) VLCKit

LePips:
[...]

PR #1206, opened 2024-08-24
MPVKit implementation

stevebyatt10:
(https://github.com/jellyfin/Swiftfin/issues/221)

Quick initial investigation into using MPV.

Clone https://github.com/cxfksword/MPVKit and build locally, add as local swift package dependency

Replacing swift player type with example MPV player for Apple TV. Tested and confirm playback from start working so far.

Also this PR changes `lastSignedInUserID` key to be non optional as it was causing a crash for me for some reason inserting null object into UserDefaults

### DIFF
--- a/Shared/Coordinators/VideoPlayerCoordinator.swift
+++ b/Shared/Coordinators/VideoPlayerCoordinator.swift
@@ -73,9 +73,10 @@ final class VideoPlayerCoordinator: NavigationCoordinatable {
 
         #else
         if Defaults[.VideoPlayer.videoPlayerType] == .swiftfin {
-            PreferencesView {
-                VideoPlayer(manager: self.videoPlayerManager)
-            }
+            MPVVideoPlayer(manager: self.videoPlayerManager)
+//            PreferencesView {
+//                VideoPlayer(manager: self.videoPlayerManager)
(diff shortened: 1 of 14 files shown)


LePips:
Thanks, I also know this works because I've done it before haha. I started work on an mpv wrapper like my VLCUI project before the summer but I got busy and want to work on other things before the iOS 15 cutoff.

My (not finished) project is below and also currently depends on `MPVKit` since it's easy to get started but has the goal of providing a better API for playback and will abstract away the player layers from the app. I just made it public and is very much not finished. I would prefer not to use the views as made in `MPVKit` and make better ones influenced by the views in that project, but also other mature mpv players like iina or Yattee.

- https://github.com/LePips/MPVUI

---

For the `lastSignedInUserID`, I also _sometimes_ ran into that crash but I think that's a bug within the OSs, however we should still change it to be non-`nil`. Do that in another PR where we also make sure to use the `default` suite if it is empty, or we should make some kind of `UserSignInState` enum with `signedOut` and `signedIn(userID: String)` (not final, just a suggestion) and migrate it over.

stevebyatt10:
Yeah cool the project looks good, will follow up with another PR

PR #1209, opened 2024-08-25
Update lastSignedInUserID defaults to use new UserSignInState enum 

stevebyatt10:
Encountered a crash with `lastSignedInUserID` and UserDefaults sometimes, reason was due to inserting a null object. 

Replace optional string with new enum for better handling.


stevebyatt10:
@LePips moved those changes from the other PR to here 🙏

Swiftfin.xcodeproj/project.pbxproj
```diff
@@ -1838,6 +1841,7 @@
 				E18ACA8A2A14301800BB4F35 /* ScalingButtonStyle.swift */,
 				E164A7F52BE4814700A54B18 /* SelectUserServerSelection.swift */,
 				E129429228F2845000796AC6 /* SliderType.swift */,
+				DFB7C3DE2C7AA42700CE7CDC /* UserSignInState.swift */,
```
LePips, line 1844: In Xcode, select all files in `Shared/Objects` and `Sort By Name` please.

LePips (changes requested):
Looks great and I don't know why I didn't think of this at the beginning. We will probably be forcing people to be signed out after this update, and that sucks, but not the end of the world.

Issue #1211, opened 2024-08-25
Want option to limit next up in number of days like the web client


sjorge:
**Describe the feature you'd like**
The web client has an option to limit `Next Up` to a certain number of days:
<img width="1032" alt="Screenshot 2024-08-25 at 18 39 22" src="https://github.com/user-attachments/assets/0267f0f6-95b2-4829-bc78-b434d71c19e9">

This is nice as I have a bunch of shows I've stopped watching a long time ago that show up.

Note: this started with 10.9.10 as the endpoint got tweaked to return all items now.

With it set to 90 days the web client shows these:
<img width="1346" alt="Screenshot 2024-08-25 at 18 37 44" src="https://github.com/user-attachments/assets/4de26e2b-06ad-4fff-bf94-2494c72c43a8">

On 10.9.10 swiftfin shows a very very very long scroll list (not found the end yet)
![Screenshot 2024-08-25 at 18 38 02](https://github.com/user-attachments/assets/b00cc604-d786-4431-b0f2-be574ee1776f)

![Screenshot 2024-08-25 at 18 38 04-1](https://github.com/user-attachments/assets/e23e013f-f7a7-446e-9919-85ea2ae1208a)

![Screenshot 2024-08-25 at 18 38 10-2](https://github.com/user-attachments/assets/578ee9b3-b974-4066-867a-541e058f1f34)

**Additional context**
n/a

JPKribs:
Hey @sjorge, I thought this was a good addition. I was taking a look at how Jellyfin-Web was handling this and realized this was a really straight forward ask. Please see my PR below and let me know if you have any questions/changes:

https://github.com/jellyfin/Swiftfin/pull/1212

sjorge:
Wow that was quick! Not that I have a way to build the PR to test, but I had a glance the code and it looks fine.

A nice solution to setting the number of days that also works well for tvOS!

PR #1212, opened 2024-08-26
Add 'Enable Rewatching' and ' Max days' to Next Up

JPKribs:
Resolves: https://github.com/jellyfin/Swiftfin/issues/1211

### Summary

This PR enables a feature found on Jellyfin-Web that limits the number of items in the 'Next Up' section. To further mirror Jellyfin-Web, this feature uses 0 as disabled _(All items are shown in Next Up)_ and only includes the parameter for NextUpDateCutoff when this is greater than 0.

While I was in there, realized that Jellyfin-Web had an option for EnableRewatching so included that in this PR. I was basing my Parameters off of this section of Jellyfin-Web: https://github.com/jellyfin/jellyfin-web/blob/master/src/components/homesections/sections/nextUp.ts

### Screenshots

**iOS**

<details>
<summary>New Settings (Disabled)</summary>
New Settings - Disabled
</details>

<details>
<summary>New Settings (1 Day)</summary>
New Settings - 1 Day
</details>

<details>
<summary>New Settings (Multiple Days)</summary>
New Settings - Multiple Days
</details>

**tvOS**

<details>
<summary>New Settings (Location)</summary>
New Settings - Location
</details>

<details>
<summary>New Settings (Disabled)</summary>
New Settings - Disabled
</details>

<details>
<summary>New Settings (1 Day)</summary>
New Settings - 1 Day
</details>

<details>
<summary>New Settings (Multiple Days)</summary>
New Settings - Multiple Days
</details>

<details>
<summary>New Settings (Subtitle Update)</summary>
New Settings - Subtitle Update
</details>


sjorge:
Anything still needed for this? Would love for this to help in the next TestFlight build.

Swiftfin/Views/SettingsView/CustomizeViewsSettings/Components/Sections/CustomizeHomeSection.swift
```diff
@@ -0,0 +1,49 @@
...
+                    value: $maxNextUp,
+                    range: 0 ... 999,
+                    step: 1
+                )
+                .valueFormatter { days in
+                    switch days {
+                    case 0:
+                        return L10n.disabled
+                    case 1:
+                        return "1 " + L10n.day
```
LePips, line 37: These `day/days` strings should instead take the integer day as a parameter instead of manual concatenation.

Swiftfin/Views/SettingsView/CustomizeViewsSettings/Components/Sections/CustomizeHomeSection.swift
```diff
@@ -0,0 +1,49 @@
...
+        @Default(.Customization.Home.maxNextUp)
+        private var maxNextUp
+        @Default(.Customization.Home.enableRewatching)
+        private var enableRewatching
+
+        var body: some View {
+            Section {
+                Toggle(L10n.showRecentlyAdded, isOn: $showRecentlyAdded)
+                Toggle(L10n.nextUpRewatch, isOn: $enableRewatching)
+                BasicStepper(
```
LePips, line 26: I'm fine with this for now, but we will have to investigate another solution so that people aren't tapping/holding possibly hundreds of times. I would probably want to mirror the wheeled time interval picker, but there isn't one for months/weeks/days. We will probably have to take a look at my project below to make that for iOS but we'll have to think more for tvOS.

- https://github.com/LePips/MultiComponentPicker

That would change this implementation but we'll deal with that later.

Shared/Services/SwiftfinDefaults.swift
```diff
@@ -167,6 +166,24 @@ extension Defaults.Keys {
...
+        enum Home {
+
+            static let showRecentlyAdded: Key<Bool> = UserKey(
+                "showRecentlyAdded",
+                default: true
+            )
+
+            static let maxNextUp: Key<Int> = UserKey(
+                "homeMaxNextUp",
+                default: 0
```
LePips, line 179: I think instead we should mirror the web and default to 365 days.

LePips (changes requested):
Sorry for the long wait, I just get PR blindness.

Swiftfin/Views/SettingsView/CustomizeViewsSettings/Components/Sections/CustomizeHomeSection.swift
JPKribs: I've updated this. I added a 'dayLabel' as an extension on Int. Let me know if you want me to do that another way. Saw that's how seconds were being handled and I liked centralizing this but I'm also not sure how much additional usage we'll get from this. I can go either way.

LePips (changes requested):
Actually, let's get the underlying data type correct now so we don't have to possibly migrate in the future and not introduce new strings for units.

I've done the necessary work on a branch from this branch here:

- https://github.com/jellyfin/Swiftfin/commit/a8f88ee9272cfc7c133f269ac4cb8214cc3a7f37#diff-1a61dfe3e6a14ee09f839dea3bef02c7cb04876cb911ace5e280d95a1225d29fR36

Notable changes:

- Changes the data type to `TimeInterval`
	- While we are representing days and `TimeInterval` represents seconds, we should be consistent when it comes to time and date representation. I am trying to migrate the many inconsistencies we have, most notably with item runtimes, to `TimeInterval`.
- Uses a provided formatter for formatting the interval, where we get years, months, weeks and days for free!
	- Since Apple is more correct with their calendars, the default has changed to `366` days to get a nicer `1y` default label for English.
	- I would recommend keeping something like https://fuckingformatstyle.com as a reference for provided formatters in Foundation/SwiftUI.
- As a better middle ground for the day selection, I instead present an alert with a numerical textfield that converts between the necessary formats and clamps the valid date ranges.
	- The more correct solution is to use `ParseableFormatStyle`/a formatter but I didn't investigate that and should probably use that instead.

---

I didn't look at tvOS, however a similar alert can be made for the day modification.

Shared/Extensions/Int.swift
```diff
@@ -53,6 +53,24 @@ extension Int {
             .prepending("-", if: isNegative)
     }
 
+    /// Label if the current value represents days
+    var dayLabel: String {
```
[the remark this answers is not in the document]
JPKribs: This is actually perfect. I was struggling with this formatting on https://github.com/jellyfin/Swiftfin/pull/1255 as well so this gives me a good direction there too!

JPKribs:
I am putting this together on this branch and so far it looks so good. There is only one weird interaction that I am seeing. If you click into the alert 3 times in a row, the text becomes all caps like Swiftfin is annoyed with you:

Screenshot 2024-10-06 at 11 46 23

The very next time you click in, it's fine but I just thought that was weird.

LePips:
This solution is probably relevant: https://stackoverflow.com/questions/73469154/in-alert-text-are-coming-as-uppercase-though-no-property-is-added-in-swiftui

JPKribs:
That resolved this! 

iOS should be good to go. tvOS is killing me today. I can get this to mirror on tvOS but the keyboard gets dismissed whenever I add a number. It seems to mirror this but on tvOS:

https://forums.developer.apple.com/forums/thread/716377

I'm going to keep working at it but I might save up my iOS working / tvOS not-working version just so there is some visibility on the iOS side of this

JPKribs:
This branch got really messed up when I tried to rebase on Main. As a result, I've moved all of my work over to a new branch that _**STARTED**_ on Main as of Oct-6, 2024. So this should be easier to merge in but I've linked back to this one so we have some visibility to the original changes and post.

PR #1219, opened 2024-08-31
Fix the subtitle stream index calculation for multiple subtitles

### LINKED ISSUE #1207: Subtitles missing and duplicating in subtitle selection
### Describe the bug

As shown in the screenshots below, the same video in the web client correctly shows 2 different external subtitle tracks, one zh-hans and one zh-hant.
However it shows 2 zh-hans subtitle tracks in Swiftfin. When I enable one of them, both show as enabled. 
IMG_1613
IMG_1612

I don't really have concrete steps to reproduce so the steps below may not reproduce it properly 
1. Prepare media with multiple external .ass subtitles 
2. Play said media in Swiftfin 
3. Open subtitle selection

### Application version

1.1(78)

App Store

### Device information

iPhone 13 Mini

### OS version

iOS 17.5.1

### Jellyfin server version

10.9.9

cebrusfs:
Fix #1207 

This patch make the subtitles map to a different indexes rather than the same index while doing the index recalculation in `adjustExternalSubtitleIndexes`.The original code will let us have only one subtile available in menu.

Minor, rename variable to `copy` to alight the code in the below (in `adjustAudioForExternalSubtitles`)

### DIFF
--- a/Shared/Extensions/JellyfinAPI/MediaStream.swift
+++ b/Shared/Extensions/JellyfinAPI/MediaStream.swift
@@ -225,10 +225,10 @@ extension [MediaStream] {
 
         for (i, mediaStream) in mediaStreams.enumerated() {
             guard mediaStream.isExternal ?? false else { continue }
-            var _mediaStream = mediaStream
-            _mediaStream.index = 1 + embeddedSubtitleCount + audioStreamCount
+            var copy = mediaStream
+            copy.index = (copy.index ?? 0) + 1 + embeddedSubtitleCount + audioStreamCount
 
-            mediaStreams[i] = _mediaStream
+            mediaStreams[i] = copy
         }
 
         return mediaStreams


cebrusfs:
It works on my local device. But would be nice to have some reviewer that understand what those indexes calculation is doing.

LePips:
Thanks for taking a look at this, I can't remember exactly _why_ we had to perform the adjustments (hence the TODO comments) and where I referenced the original implementation. Would probably be best to cross reference with other clients.

cebrusfs:
I see. I am new to here. Do you have any recommendation about who to contact? or where to find them?

BTW, if you don't recall that, should we consider to just somehow test it and then merge it to Testflight to see if it cause any regression?

cebrusfs:
comment from discord. 

> to support STRM we started listing external streams before internal (because external are known at the time of scanning, internal are not). https://github.com/jellyfin/jellyfin/pull/7529
> this is because strm are only probed on playback start to not pull the file on each scan

And some type reference from SDK.

> Stream [types](https://github.com/jellyfin/jellyfin-sdk-swift/blob/a0e848a7aaec55991610818de6128b15cfcec725/Sources/Entities/MediaStream.swift#L117): https://github.com/jellyfin/jellyfin-sdk-swift/blob/a0e848a7aaec55991610818de6128b15cfcec725/Sources/Entities/MediaStreamType.swift

I might have no time to have closer look during weekday. It seems I will need to cross-check all current implementation of indexes. @LePips feel free to take over if that make you recall what you were doing here.

LePips (approved):
I have now started to take a look at this in my work for #1203 and, at least statically without testing here, looks good to me and will moreso keep as reference and fix if needed in my work.

Thank you for reaching out and getting an explanation. After seeing the server PR, I think I remember better now.

Issue #1170, opened 2024-08-06
Native Player Fails to Transcode Opus


JPKribs:
### Describe the bug

Since there are so many variables with Video Encoding, I don't know if this is necessarily AV1 or something in my audio but I have a few files encoded like this that always fail:

### Video
Title: Movie Name (YEAR) - 1080p - AV1 - SDR
Codec: AV1
AVC: No
Profile: Main
Level: 8
Resolution: 1920x960
Aspect ratio: 2:1
Anamorphic: No
Interlaced: No
Framerate: 24
Bitrate: 1298 kbps
Bit depth: 10 bit
Video range: SDR
Video range type: SDR
Color space: bt709
Color transfer: bt709
Color primaries: bt709
Pixel format: yuv420p10le
Ref frames: 1

### Audio
Title: French - OPUS - 5.1 - Default
Language: fre
Codec: OPUS
AVC: No
Layout: 5.1
Channels: 6 ch
Bitrate: 214 kbps
Sample rate: 48000 Hz
Default: Yes
Forced: No
External: No

### Subtitles - Not Selected
Title: English - SUBRIP
Language: eng
Codec: SUBRIP
AVC: No
Default: No
Forced: No
External: No

### Summary
Swiftfin does not attempt to transcode this AV1 file down and I get the following screen:

![Simulator Screenshot - iPhone 15 - 2024-08-06 at 15 50 59](https://github.com/user-attachments/assets/ac3acdf6-84ef-46b1-a8cf-46dc3e30e650)

### Application version

1.1.1 & Main

### Where did you install the app from?

App Store

### Device information

iPhone 13, iPhone 15 Pro, & iPhone XS

### OS version

iOS 17.6

### Jellyfin server version

10.9.9

JPKribs:
I realized using the https://github.com/jellyfin/Swiftfin/pull/1169 custom profile, I can force it to transcode but it tries to retain Opus and fails to play.

When I had this set up with the decoder check, Opus is removed from both the DirectPlay and Transcode profiles, resulting in this working:

<img width="283" alt="Screenshot%202024-08-05%20at%2011 56 40" src="https://github.com/user-attachments/assets/e85664ae-8dbf-4c41-a90b-2f1cde8642b1">

holow29:
Opus 2 channel works for me on Native player, even with AV1 files (which transcode video but leave audio). I need to test 5.1 opus I guess.

JPKribs:
[...]

Issue #1150, opened 2024-07-11
OPUS audio not playing on Apple TV


anyoneyun:
### Describe the bug

1. Click on a movie/show with an OPUS audio track
2. No audio is played, jellyfin doesn't transcode

### Application version

1.0.1 (70)

### Where did you install the app from?

App Store

### Device information

Apple TV 4K 2nd Gen

### OS version

17.5.1

### Jellyfin server version

10.9.7

UnidentifiedX:
Also experiencing the same on Swiftfin iOS

### Application version
1.1

### Where did you install the app from?
App Store

### Device information
iPhone 11

### OS version
17.5.1

### Jellyfin server version
10..9.7

holow29:
@UnidentifiedX Can you please give the mediainfo of the file?

UnidentifiedX:
> @UnidentifiedX Can you please give the mediainfo of the file?

![image](https://github.com/user-attachments/assets/1376affc-33ea-4763-8658-40e1caf3acc8)
![image](https://github.com/user-attachments/assets/12f3d845-8d56-4254-85f9-e769f9f1f55e)

Maybe it's unrelated, but my ASS subtitle also doesn't load on Swiftfin either. I can open another issue if this is the case.

holow29:
@anyoneyun Until the tvOS app is updated past 1.0.1 so that the device profile changes take effect, this will remain the case.
@UnidentifiedX I assume you are trying to use the Native player. I have identified that the native player only supports 2 channel Opus. You can see more discussion and a possible patch which I have yet to test here: https://github.com/jellyfin/Swiftfin/issues/1170

UnidentifiedX:
@holow29 Thanks for the reply! Looked through the thread and I've decided to transcode my files to 5.1 AC3. Works fine now, thanks!

Also on a side note, even running stereo opus doesn't produce any audio. Tried with the LegacyStereo audio track as well - doesn't seem to work either. Not sure what is going on here.

UnidentifiedX:
Actually, upon further investigation, I realised that if you add a .sup file to your subtitles the audio track from the video will be gone if you use the Swiftfin player. Extremely weird, gonna open an issue when I have time. This was probably why my OPUS file couldn't play audio and has nothing to do with the codec.

smebberson:
I've also experience issues with two audio tracks; it will just play nothing. I removed one and it worked fine.

nds2204:
[...]

PR #1226, opened 2024-09-02
Fix Opus Playback Issues on Native Player

JPKribs:
### Resolves:
https://github.com/jellyfin/Swiftfin/issues/1170 & https://github.com/jellyfin/Swiftfin/issues/1150

### Issue:
Officially, Opus audio is not supported for Apple's AVKit. Technically, Opus **IS** supported on AVKit but only up to 2 channels. This means, people with more than 2 channels (5.1 being the most common) who use the Native Player are unable to play that content. Even if it's forced to transcode for other reasons, since Opus is in the Transcode profile, the problem audio gets carried forward.

The issue we run into is that there doesn't appear to be a good way to tell Jellyfin to DirectPlay Opus 2.0 and Transcode Opus 2.0+. @holow29 had a good idea with adding:

```
        let opusCodecConditions: [ProfileCondition] = [
            ProfileCondition(
                condition: .lessThanEqual,
                isRequired: true,
                property: .audioChannels,
                value: "2"
            ),
        ]

        profile.codecProfiles = [
            CodecProfile(
                applyConditions: opusCodecConditions,
                codec: "opus",
                type: .videoAudio
            ),
        ]
```
For the CodecProfiles but in testing it is either bugged or we aren't using it correctly.

For now, I am I pitching that we just force transcode Opus for Native Player and, if the user wants to Direct Play 2.0, they can enable Opus using the changes in: https://github.com/jellyfin/Swiftfin/pull/1169

Alternatively, if we can find a way to differentiate between 2.0 and other channels I think that would be ideal but I think #1169 gives us some additional leniency to be a little over-zealous in favor of compatibility.

### DIFF
--- a/Shared/Objects/VideoPlayerType/VideoPlayerType+Native.swift
+++ b/Shared/Objects/VideoPlayerType/VideoPlayerType+Native.swift
@@ -21,7 +21,6 @@ extension VideoPlayerType {
             AudioCodec.alac
             AudioCodec.eac3
             AudioCodec.flac
-            AudioCodec.opus
         } videoCodecs: {
             VideoCodec.h261
             VideoCodec.hevc
@@ -110,7 +109,6 @@ extension VideoPlayerType {
             AudioCodec.alac
             AudioCodec.eac3
             AudioCodec.flac
-            AudioCodec.opus
         } videoCodecs: {
             VideoCodec.h264
             VideoCodec.hevc


LePips (approved):

holow29:
Thanks for moving forward on this! Haven't had a ton of time to look into it more, but ideally will try to figure out what is going on with the codecConditions (for all I know, it could be a server bug with them applying or only applying to directPlay and not to transcode, etc. There have been some PRs to Jellyfin server recently to address similar things.)

Issue #627, opened 2022-11-11
Hide Chanel & Folders


madswaord:
Hi, I was wonder Swiftfin can have the feature just like web version, I can choose which Chanel and folder can shows on the startup page and new added, sometimes there are something needs to hide from kids. And for the tv Swiftfin, it’s all films and tvshows listed combined in one Chanel, it’s better we can choose just like infuse, once we uncheck this folder it wont show on the media library.

PR #1227, opened 2024-09-03
Hide Media Library

JPKribs:
### Issue
https://github.com/jellyfin/Swiftfin/issues/627

### Summary
The goal of this PR is to allow users to hide libraries at a Client level. This should also remove items from those Libraries from the tvOS MediaType views like 'TV Shows' and 'Movies' on the Main screen.

### Remaining Work
- [ ] Filter out the Excluded Libraries from Home > Next Up
- [ ] Filter out the Excluded Libraries from Home > Resume
- [ ] Find a way to Exclude libraries from the Excluded Libraries View opposed to only from Media
- [ ] Mirror the Excluded Libraries view in tvOS
- [ ] Filter out the Excluded Libraries from tvOS MediaType views 
- [ ] Refresh Media when an Excluded Library is re-included

### Outstanding Issues
1. I would rather use the OrderedSectionSelectorView for the Exclude Library selection screen but I'm not positive there is a good way to populate this? Since the Name/Id combo for Media Libraries is unknown and varies from Server to Server. I could just store the Library names but I'm not sure if Jellyfin allows for multiple libraries with the same name or not?
2. I am filtering the Latest/Next Up views AFTER the API call is completed to pull all of the data. I'd like to have this trim the API call. I need to research this more to see if that's a filter I can do on the API call opposed to getting all data and filtering.

### Screenshots
<details>
<summary>Exclusion Context Menu</summary>
Simulator Screenshot - iPhone 15 Pro - 2024-09-03 at 11 47 26
</details>

<details>
<summary>All Libraries Excluded</summary>
Simulator Screenshot - iPhone 15 Pro - 2024-09-03 at 11 47 33
</details>

<details>
<summary>Excluded Libraries Setting Location</summary>
Simulator Screenshot - iPhone 15 Pro - 2024-09-03 at 11 47 48
</details>

<details>
<summary>Excluded Libraries Setting</summary>
Simulator Screenshot - iPhone 15 Pro - 2024-09-03 at 11 47 50
Simulator Screenshot - iPhone 15 Pro - 2024-09-03 at 11 47 57
</details>


JPKribs:
Unfortunately, without a change to the API, I do not believe this PR is possible. The reason is that the API for ItemType Library (tvOS), Recently Added, Next Up, and Resume Watching do not allow for filtering based on Library Id. To make matters harder, the LibraryId is not available for these. Instead, the ParentID value on these are the ItemType or the Show/Season that holds the episode.

One route that could be taken to get here would be to pull X API calls for each of the libraries that are excluded then, using the item Ids found in the excluded API calls, use those in the Excluded ItemIds option on a final API call to get only the included Libraries for that section. The issue becomes that we will have to read the entire library of excluded libraries. This costly both in memory but also for server bandwidth.

An easier route would be a server-side change for excludedParentIDs which would accept an array of LibraryId to exclude OR change the current parentID parameter to be parentIDs and allow for an array to be provided. The latter change might break some other usages of this API. 

Screenshot 2024-09-04 at 14 19 44

I'll leave this branch open if anyone has any ideas but for now I am going to throw in the towel and say this isn't a realistic change at this time.

LePips:
My understanding of the first part of the request was to implement the corresponding web customization of `User/Home/` for home screen customization (which I believe you are tackling here).

When it comes to content exclusion, that should be entirely controlled by user permissions as set by the server admin. _However_, I am unsure if we have to add parental rating filters to API calls if the user has any.

thornbill:
> However, I am unsure if we have to add parental rating filters to API calls if the user has any.

No the server should handle that automatically.

JPKribs:
I might have gone the wrong way with this PR. I thought the goal was to hide/show library content. The roadblock I hit was for the non-library specific sections. Where there was an option to select a parentID to pull from but not a way to filter out a parentID without pulling in the full call and filtering from there. If I'm wrong, I'd be happy to pick this back up again!

Issue #608, opened 2022-10-18
Admin Dashboard


bwlok:
Add the admin Dashboard that's already available on the website and AppStore app

![image](https://user-images.githubusercontent.com/65543013/196376821-ef43e274-2aa3-482e-9c99-38a12e06d145.jpeg)

JPKribs:
So this is very very early but I was taking a crack at this My big question is what are people looking from for the 'admin dashboard' in Swiftfin? Is it primarily just the Active Devices and Scan Library? Or do people want to use Swiftfin to change settings / update users / update libraries? Here is what my screen look likes right now: 

![Simulator Screenshot - iPhone 15 Pro - 2024-09-04 at 22 03 54](https://github.com/user-attachments/assets/cfa21266-58cf-4153-bd2a-6185df9c92e3)

The red bars are covering the usernames but those are there.

sjorge:
Active Sessions, Scan Library and start a scheduled task is what I sometimes use on my phone. In order of most -> least used.

bwlok:
Same here mostly Scan Library for me

JPKribs:
Please see https://github.com/jellyfin/Swiftfin/pull/1230

This PR includes:

1. Active Sessions
*LOCKED TO ADMINS ONLY*
2. Scan Library
3. Restart Server
4. Shutdown Server

I see a route for adding scheduled tasks that shouldn't be too difficult but I have a feeling this current PR is going to have some revisions. I'm going to focus on getting this cleaned up and merged then I will look at Scheduled Tasks at a later date.

Issue #508, opened 2022-08-02
Home network detection and remote URL switching


ReubenBTalbott:
Could you add a feature where when SwiftFin detects the local network it'll use an internal address to access JellyFin, and when that local network isn't present it'll switch to a second external URL/address?   Home assistant has a similar feature which I've posted screenshots of below.

Thanks!
![IMG_3346](https://user-images.githubusercontent.com/19376692/182471365-97a1e6a2-2787-46f1-ad9a-6741ee3949e2.PNG)
![IMG_3347](https://user-images.githubusercontent.com/19376692/182471627-44ac142c-1d98-4d24-b5c6-df8647afb7b2.PNG)

michaelkrieger:
This would be amazing for iPads and iPhones. Less necessary for Apple TV.

maherbeg:
This would be amazing especially if you use Tailscale VPN to get into your local network. They currently don't support mDNS so you have to modify the address when switching networks.

michaelkrieger:
Detecting via SSID (or network name for wired) of course would be the ideal feature.
That said, I'd also be good with a (try: 192.168.123.123:1111 and if it doesn't find Jellyfin use the external address).

JBHemloque:
I'm a new Jellyfin convert, and I use swiftfin on my phone. This feature would be absolutely amazing.

I'd be happy simply being able to specify multiple addresses for the same server, which would let me manually switch between the two.

PR #1230, opened 2024-09-05
[iOS] Admin Dashboard

JPKribs:
### Issue
Resolves: https://github.com/jellyfin/Swiftfin/issues/608

### Summary
This PR Creates a User/Admin Dashboard. This view replaces the current EditServerView that is accessible from the SettingsView. For now, this feature will only exist on iOS and iPadOS. If there is an appetite for adding this to tvOS I can look into it but, on a personal level, I don't see a lot of value in adding this to tvOS.

When clicking into this view, the user will always be provided with name of the server and the ability to switch to another URL for the same server. For Non-Admin Users, this PR adds an additional section that shows all Active Devices.

Active Devices are handled by the Server. With this, the Server only returns Sessions that the user is allowed to see. For Non-Admins, this means they can only see their own Sessions. For Admins, they are able to see all Sessions. All filtering/security for this section is handled by the Server API.

Clicking an Active Device will show the user Active Device Details which includes:

- Content currently being played
- Current progress of the Content
- Stream details including transcoding
- Transcoding reasons (if transcoding is occurring)

Both the Active Devices List and Active Device Details update every second with the status and progress of the Active Device. If the device is active but not streaming, both view will show the last time that the device interacted with the Jellyfin Server.

If the User accessing the Dashboard is an Admin they will see 2 additional Administration buttons. These buttons are:

- Scan All Libraries
- Scheduled Tasks

Scan All Libraries is the same function as Jellyfin-Web. When triggered, it returns the current progress of the process. There is a Red X on the left side while this task is running which can be selected to cancel the Libraries Scan. The progress percentage pings every second, even if not triggered from the Server, so this shows Library scans that are triggered from other devices as well with the ability to cancel them too.

Scheduled Tasks takes the Admin User to another View that contains all of the available Tasks on the Server. If selected, this runs these tasks. The same as the Scan All Libraries, these tasks can be cancelled and progress is showed. In addition to the Scheduled Tasks, the Shutdown/Restart Server tasks are included on this page at the top. These prompt the user with a confirmation prior to running.

The only difference between iOS and iPadOS is the Active Devices section contains 1 column on iOS and 2 columns on iPadOS.

Please find all Screenshots in the Screenshots Section below.

### Difficult to Recreate Issues
- [ ] The container in the Active Device Details > Streams Section comes from SessionInfo.NowPlayingItem.Container. On my server, it's always 100% correct but for some reason demo.jellyfin.org/stable is giving me 7 containers for every item.

### Screenshots
<details>
<summary>User Dashboard</summary>
User Dashboard
</details>

<details>
<summary>User Dashboard - URL Switching</summary>
User Dashboard - URL Switching
</details>

<details>
<summary>Admin Dashboard</summary>
Admin Dashboard
</details>

<details>
<summary>Admin Dashboard - URL Switching</summary>
Admin Dashboard - URL Switching
</details>

<details>
<summary>Admin Dashboard - Scan in Progress</summary>
Admin Dashboard - Scan in Progress
</details>

<details>
<summary>Admin Dashboard - Scan Cancelling</summary>
Admin Dashboard - Scan Cancelled
</details>

<details>
<summary>Admin Dashboard - iPad</summary>
Admin Dashboard - iPad
</details>

<details>
<summary>Shutdown/Restart Warnings</summary>
Restart Warning
Shutdown Warning
</details>

<details>
<summary>Active Device Details - DirectPlay</summary>
Active Device Details - DirectPlay
</details>

<details>
<summary>Active Device Details - Transcode</summary>
Active Device Details - Transcode
</details>

<details>
<summary>Active Device Details - Idle Sessions</summary>
This ticks up every second. If the user starts playback, the view populates with the new session details.
Active Device Details - Idle
</details>

<details>
<summary>Active Device Details - Offline Session</summary>
This will only happen when you are viewing a User Session Detail and the user goes offline while you are looking at it.
Active Device Details - Offline
</details>

<details>
<summary>Scheduled Tasks</summary>
Scheduled Tasks
</details>

<details>
<summary>Scheduled Tasks - Many Running</summary>
Scheduled Tasks - Many Running
</details>


JPKribs:
I can't seem to find a good way right now to make the ViewModel for Active Sessions update at an interval. For now, I have a refresh button. I guess I don't know what people's usage for this will be. Is a refresh button adequate? 

Otherwise, I am doing some research to hopefully figure out how to refresh a ViewModel without redrawing the view on top of it.

### Any feedback is welcome to how I am handling the UI for this!

JPKribs:
@LePips **Absolutely no rush on this** but I wanted to get your thought on this PR and make sure I was going in the right direction for what the Admin Dashboard would include.

sjorge:
When it's transcoding, does it also show the reason/format ? That generally the most useful thing I sometimes check on the go.

I don't think we should fully duplicate all functionality of the web interface, even vue is struggling with that.

JPKribs:
I have the type broken out like Jellyfin-Web:

'''
static func getDisplayPlayMethod(_ session: SessionInfo) -> SessionPlaybackMethod {
        if let transcodingInfo = session.transcodingInfo {
            if (
                transcodingInfo.isVideoDirect ?? false ||
                    transcodingInfo.videoCodec == nil
            ) &&
                transcodingInfo.isAudioDirect ?? false
            {
                return .remux
            } else if transcodingInfo.isVideoDirect ?? false {
                return .directStream
            } else if session.playState?.playMethod == .transcode {
                return .transcode
            }
        }
        if session.playState?.playMethod == .directStream {
            return .directStream
        } else if session.playState?.playMethod == .directPlay {
            return .directPlay
        }

        return .unknown
    }
'''

<detail>
<summary>But I'm guessing you mean:</summary>
image
</detail>

I could look at adding this in but I originally skipped it since phone screen real estate is a lot smaller and I was worried it was starting to get cluttered.

sjorge:
Correct, it seems those activity tiles already can already be expanded based on the little V arrow.

At least the transcode reason would be nice to have as an extra item there.

Target audio and video code would be useful too, but IMHO not as important.

JPKribs:
I still need to reorganize since this has gotten pretty messy behind the scenes but is this the time of information we're looking for or is the reason string important too?
Screenshot 2024-09-06 at 10 31 36

sjorge:
> I still need to reorganize since this has gotten pretty messy behind the scenes but is this the type of information we're looking for or is the reason string important too? Screenshot 2024-09-06 at 10 31 36

What is show here would generally be sufficient. (For me at least)

With the info in the screenshot we see what the client is and what is being transcoded vs copied, in this case AC3 -> AAC for an iOS client makes sense as they generally do not play dolby surround.

So we know it's a lightweight transcode in this case.

The reason string would take the guess work out of it, but it does tend to be rather 'lengthy'. But it would also tell the transcode is because of audio stream incompatibly. It might be less clear for a video example though, a HVEC to h264 transcode could be because: a) the device does not support the HVEC profile b) subtitles are being burned-in because they are not supported c) original bitrate is too high. (I think b, can be covered by showing if the subtitle stream is copy or not though)

I'm not sure what the right thing to do is, we need to weigh the clutter vs providing information vs code complexity.
e.g. we could make it so tapping the 3 lines showing the stream detail would swap between this display and just the reason string so the user could easily switch without it taking up a lot of extra space but that would make things more complex code wise (I assume, swift/object C is confusing for me) or be too intuitive. 

Also adding the reason below, would probably make it take up to much space. 

Maybe we can have Erik our ui-ux guy chime in?

JPKribs:
I'd be game for any insight someone may have related to what gets included here.

This is my current version:
Screenshot 2024-09-06 at 11 56 39

sjorge:
That looks fine to me, but I'll link Erik to this PR to see if he wants to chime in.

LePips:
Since these can be very detailed views, I think a better design would be to have the list of sessions with bare details and we push to a session with more details instead of expanding.

JPKribs:
I can do that. What should be included in the high level view? My thought is:
- Username
- Content Name
- Playback State ⏯️
- Playback Progress

Then I can add everything else to the details page.

sjorge:
> I can do that. What should be included in the high level view? My thought is:
> 
>     * Username
>     * Content Name
>     * Playback State ⏯️
I think maybe we can use play and pause icon instead of the text? That should save some space
>     * Playback Progress
2 Colored progress bar like in web? It shows playback and transcoding progress in 1 bar. 
> 
> Then I can add everything else to the details page.
> 
> Also, for the progress bar, do we want the color to indicate Play/Pause or should it use the accent color?

I think accent color, using color to indicate playback is going to be less obvious than using just a string or icon.

JPKribs:
What do people think of this version:
Reduced Version

Swiftfin/Views/UserDashboardView/ActiveSessionsView/ActiveSessionRowView.swift
```diff
@@ -0,0 +1,246 @@
...
+        case .landscape:
+            contentWidth = 133
+            imageSource = item.landscapeImageSources().randomElement()
+        }
+
+        if let cachedPoster = posterCache[item.id] {
+            output = cachedPoster
+        } else {
+            output = AnyView(
+                AsyncImage(url: imageSource?.url) { phase in
```
LePips, line 213: I know this is a draft, but why remake `ImageView`?

Shared/ViewModels/ActiveSessionsViewModel.swift
```diff
@@ -0,0 +1,69 @@
...
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2024 Jellyfin & Jellyfin Contributors
+//
+
+import Combine
+import Foundation
+import JellyfinAPI
+
+class ActiveSessionsViewModel: ViewModel {
```
LePips, line 13: I would heavily recommend making this `Stateful`. If you are unsure of how to do so, I can help.

Shared/ViewModels/ActiveSessionsViewModel.swift
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    func loadSessions() {
+        isLoading = true
+        error = nil
+
+        Task {
+            do {
+                let fetchedSessions = try await get()
+                DispatchQueue.main.async {
```
LePips, line 43: As a tip when inside an `async` context, use the following instead to switch to the main actor:

```swift
await MainActor.run {
// stuff on main thread
}
```

Swiftfin/Views/UserDashboardView/ActiveSessionsView/ActiveSessionRowView.swift
JPKribs: To be honest, I didn't see that exists! I'll move the images over to that.

Shared/ViewModels/ActiveSessionsViewModel.swift
JPKribs: I'm not positive how to do that but I can take a crack at it. Does that eliminate the need to do the timer() functions in there to refresh?

JPKribs:
I'm going to call it for now. I've pushed my latest changes so anyone who wants to look it over / recommend anything this should be the latest version. There is still a **TON** of reorganization that I need to do. I've also overloaded the Details Page with just about every detail I could find. If there is anything else needed, I can add it but I think I've hit the point where it's more likely that we start removing things from it. Screenshots can be found below:

<details>

<summary>Active Devices</summary>

Simulator Screenshot - iPhone 15 Pro - 2024-09-07 at 01 35 23

</details>

<details>
<summary>Active Device Details 1</summary>

Simulator Screenshot - iPhone 15 Pro - 2024-09-07 at 01 36 11

</details>

<details>
<summary>Active Device Details 2</summary>

Simulator Screenshot - iPhone 15 Pro - 2024-09-07 at 01 36 15

</details>

<details>
<summary>Video In Action</summary>

https://github.com/user-attachments/assets/fe9216a3-485b-4aa7-aedd-a4153fa534e6

</details>

I have some outstanding issues I want to cleanup but it may be a bit:

- [ ] ViewModel / Stateful Changes - I might need help with this
- [ ] Live Updating ViewModel in the Details Tab
- [ ] Ensure the Container/Codecs are accurate since some look weird in testing
- [ ] Reorganize Files
- [ ] Localization

sjorge:
The video shows it without the poster/thumbnail, I think that looks much cleaner than with. Also what happens when there is only a poster (Primary) available?

JPKribs:
I think I'm calling images a wrong way? I need to look at that. I get images from my server but demo.jellyfin.org doesn't seem to pull images. I'm leaning towards no images on the main screen with posters on the details? I'm only able to get 4 sessions on a single screen with these posters.

Right now, I am looking at the Library Customization for Poster Type to determine with poster is used for this. I think I need to look at where I get images from since there should be some sort of placeholder if the image isn't there. I had one prior but I moved to use the ImageView and I haven't taken the time to dig into it.

sjorge:
> I think I'm calling images a wrong way? I need to look at that. I get images from my server but demo.jellyfin.org doesn't seem to pull images. I'm leaning towards no images on the main screen with posters on the details? I'm only able to get 4 sessions on a single screen with these posters.

No images on the main screen would be good to fit more, and also no need to download a bunch of extra images. 👍 

> 
> Right now, I am looking at the Library Customization for Poster Type to determine with poster is used for this. I think I need to look at where I get images from since there should be some sort of placeholder if the image isn't there. I had one prior but I moved to use the ImageView and I haven't taken the time to dig into it.

I think the image type you want is `Primary` it's either the Show Poster, Movie Poster, or album cover for music.
The compiled URI should be `/Items/<itemId>/Images/Primary`, optionally you can also get them resize by adding `?fillHeight=400&fillWidth=240&quality=70`

erikbucik:
Why are there two sections (STREAMS & ONLINE)? What's the difference?

sjorge:
> Why are there two sections (STREAMS & ONLINE)? What's the difference?

Streams are those clients actively streaming something. Online is clients marked online in jellyfin, IIRC recent activity in the last X minutes. 

Also welcome 😃

JPKribs:
I split this into people are are connected to Jellyfin and people who are connected and streaming something. Personally, on Jellyfin-Web, I'm always more interested in what's being streamed opposed to whose only. So splitting these out is partially a preference from me. See an example below. In this case, the iPhone is ONLINE and the Android TV would be STREAMING.

I'm not sold on that verbiage so if there is a better way to name that I'm all ears!

image

sjorge:
I'd even be inclined to only show those that are actively streaming.

JPKribs:
This is an alternative that is closer to how Jellyfin-Web handles it. Just throwing another option out there while I'm screwing around to try and figure out getting the ViewModel to work better.
Screenshot 2024-09-07 at 18 51 46

Shared/ViewModels/ActiveSessionsViewModel.swift
JPKribs: This is what I have now. It builds and it works and it conforms for Stateful but I don't think I did this right:

This is the warning I've been getting on var sessions:

`Publishing changes from background threads is not allowed; make sure to publish values from the main thread (via operators like receive(on:)) on model updates.`

```
import Combine
import Foundation
import JellyfinAPI
import OrderedCollections
import SwiftUI

final class ActiveSessionsViewModel: ViewModel, Stateful {

    // MARK: Action

    enum Action: Equatable {
        case error(JellyfinAPIError)
        case loadSessions
        case refreshSessions
    }

    // MARK: State

    enum State: Hashable {
        case initial
        case loading
        case loaded
        case refreshing
        case error(JellyfinAPIError)
    }

    @Published
    var sessions: [SessionInfo] = []

    @Published
    final var state: State = .initial

    @Published
    final var lastAction: Action? = nil

    private var timer: Timer?

    // MARK: Initialization

    override init() {
        super.init()
        Task { @MainActor in
            await self.send(.loadSessions)
        }
        Task { @MainActor in
            startRefreshing()
        }
    }

    deinit {
        timer?.invalidate()
    }

    // MARK: Stateful Conformance

    func respond(to action: Action) -> State {
        switch action {
        case .loadSessions:
            loadSessions()
            return .loading

        case .refreshSessions:
            refreshSessions()
            return .refreshing

        case let .error(error):
            return .error(error)
        }
    }

    func send(_ action: Action) async {
        state = await respond(to: action)
        lastAction = action
    }

    // MARK: Session Management

    private func loadSessions() {
        Task { @MainActor in
            do {
                let fetchedSessions = try await fetchSessions()
                self.sessions = fetchedSessions
                self.state = .loaded
            } catch {
                self.state = .error(JellyfinAPIError(error.localizedDescription))
            }
        }
    }

    private func refreshSessions() {
        Task { @MainActor in
            do {
                let fetchedSessions = try await fetchSessions()
                self.sessions = fetchedSessions
                self.state = .loaded
            } catch {
                self.state = .error(JellyfinAPIError(error.localizedDescription))
            }
        }
    }

    private func fetchSessions() async throws -> [SessionInfo] {
        let request = Paths.getSessions(parameters: createParameters())
        return response.value
    }

    private func createParameters() -> Paths.GetSessionsParameters {
        var parameters = Paths.GetSessionsParameters()
        parameters.activeWithinSeconds = 960
        return parameters
    }

    private func startRefreshing() {
        timer = Timer.scheduledTimer(withTimeInterval: 2.0, repeats: true) { [weak self] _ in
            DispatchQueue.main.async {
                self?.handleRefreshSessions()
            }
        }
    }

    // MARK: Helper Methods

    @MainActor
    private func handleRefreshSessions() {
        Task {
            await self.send(.refreshSessions)
        }
    }
}

```

sjorge:
Overall I personally still like the one without any images on the main view the most.

Shared/ViewModels/ActiveSessionsViewModel.swift
```diff
@@ -0,0 +1,138 @@
...
+        case .refreshSessions:
+            refreshSessions()
+            return .refreshing
+
+        case let .error(error):
+            return .error(error)
+        }
+    }
+
+    func send(_ action: Action) async {
```
LePips, line 79: This function doesn't need to be implemented explicitly, it has a default implementation via `Stateful`.

Shared/ViewModels/ActiveSessionsViewModel.swift
```diff
@@ -0,0 +1,138 @@
...
+    @Published
+    final var lastAction: Action?
+
+    private var timer: Timer?
+
+    // MARK: Initialization
+
+    override init() {
+        super.init()
+        Task { @MainActor in
```
LePips, line 50: Don't start loading in an `init`. Instead, have the view send the `loadSessions` in a `.onFirstAppear`.

Shared/ViewModels/ActiveSessionsViewModel.swift
```diff
@@ -0,0 +1,138 @@
...
+    enum State: Hashable {
+        case initial
+        case loading
+        case loaded
+        case refreshing
+        case error(JellyfinAPIError)
+    }
+
+    @Published
+    var sessions: [SessionInfo] = []
```
LePips, line 36: Should probably be an `OrderedSet<SessionInfo>`

Shared/ViewModels/ActiveSessionsViewModel.swift
```diff
@@ -0,0 +1,138 @@
...
+        case refreshSessions
+    }
+
+    // MARK: State
+
+    enum State: Hashable {
+        case initial
+        case loading
+        case loaded
+        case refreshing
```
LePips, line 31: I'm going to have to think about having `loading` and `refreshing`. Instead, we should probably have the passive-timer loading sessions state be through a `BackgroundState` as well as a hard refresh state.

See `HomeViewModel` for a similar implementation.

Shared/ViewModels/ActiveSessionsViewModel.swift
```diff
@@ -0,0 +1,138 @@
...
+
+    deinit {
+        timer?.invalidate()
+    }
+
+    // MARK: Stateful Conformance
+
+    func respond(to action: Action) -> State {
+        switch action {
+        case .loadSessions:
```
LePips, line 66: We will need `Task` cancellation or possibly ignoring timer fires if a `Task` takes too long.

Shared/ViewModels/ActiveSessionsViewModel.swift
```diff
@@ -0,0 +1,138 @@
...
+        return response.value
+    }
+
+    private func createParameters() -> Paths.GetSessionsParameters {
+        var parameters = Paths.GetSessionsParameters()
+        parameters.activeWithinSeconds = 960
+        return parameters
+    }
+
+    private func startRefreshing() {
```
LePips, line 122: I'm leaning more towards the `Timer` being the `View` instead.

- https://www.hackingwithswift.com/quick-start/swiftui/how-to-use-a-timer-with-swiftui

Shared/ViewModels/ActiveSessionsViewModel.swift
```diff
@@ -0,0 +1,138 @@
...
+                self.sessions = fetchedSessions
+                self.state = .loaded
+            } catch {
+                self.state = .error(JellyfinAPIError(error.localizedDescription))
+            }
+        }
+    }
+
+    private func fetchSessions() async throws -> [SessionInfo] {
+        let request = Paths.getSessions(parameters: createParameters())
```
LePips, line 111: I would recommend looking at how I create requests that have parameters, where I don't make an explicit function for it. I used to, but find it too verbose in most cases.

Reference: https://github.com/jellyfin/Swiftfin/blob/0a997aeb5ec71eead36f7b062897cff1eaab99f9/Shared/ViewModels/HomeViewModel.swift#L177

JPKribs:
@LePips I've committed my current version of this PR. The new Active Sessions are closer to Jellyfin-Web now and, in my opinion, look at home on Swiftfin. Let me know if you would prefer I revert to an earlier version of this. Please find a screenshot of this current version below:

<details>
<summary>Current Version (Light Mode / iPad)</summary>
Screenshot 2024-09-09 at 10 30 37
<details>

I ended up making a new file for almost every section used on the button since a lot of the work was being re-used by the details page. Let me know if there are any of these that you think I should just combine out or if there is a better way to do this. I had the idea of using a ActiveSession Struct that had all of these elements as a var but I was trying to mirror the VideoPlayerView sections. I'm game for whatever direction you prefer.

At this current point, I feel like this is very close to being done but I think there are 3 issues that are outstanding. 2 of them I think are ViewModel issues and 1 of them is more of a missing feature but would rely on the ViewModel working right:

- [ ] I am using the CollectionVGrid for the buttons now but I get this weird flickering every so often. Is this me populating the ViewModel wrong or is the CollectionVGrid not built for live updating? If the latter, should I use SwiftUI's built in Grids instead? This all kind of depends on whether we like the tiles/buttons opposed to the previous List that I made.
- [ ] The LastSeen in the bottom left of online but not streaming sessions doesn't update unless the ViewModel has new changes. So, if no one is streaming then the times don't update. I am able to fix this by adding a second timer to the ConnectionSection but I was hoping there was a more elegant solution to this.
- [ ] The ActiveSessionsDetails doesn't live update. I think I can do that by binding sessions to the input or I can have this page call its own instance of the ViewModel. In this case, I would add an optional input of Session.DeviceID where, if provided, it would limit the Sessions to only that DeviceID and allow live updating there. I'm leaning towards two ViewModel instances because binding a session for the button from a loop was giving me a lot of trouble but I could be doing this wrong.

As a note, something I found is that Jellyfin handles Sessions by DeviceID. So, if you have 2 browser windows open it will not create 2 sessions but instead will try to share that session. This is not Swiftfin specific but it freaks out Jellyfin-Web as well so for testing you'll need to either use 2 browser sessions or multiple devices.

JPKribs:
I've updated all the screenshots. This is working how I would expect it to now. I've updated to include the 3 remaining potential issues. I don't believe any of these are show stoppers but they would be nice to resolve on this PR. I've hit a bit of a wall working on them but I am open to any feedback/ideas on how to resolve them!

JPKribs:
I'm going to call this finally ready for review! The button flickering has been resolved and I have added the admin buttons:

Screenshot 2024-09-12 at 23 31 12

These buttons are only accessible if the user has the isAdministrator policy. Otherwise, the buttons are not available. The Restart and Shutdown buttons have warnings prior to triggering.

The scan button tracks progress back from the server as seen below:

https://github.com/user-attachments/assets/dfc23a64-9656-428b-a760-b84cbde90d78

At this point in time, all of the screenshots on the post should be valid.

JPKribs:
Okay, _NOW_ I'm ready for review. This should encompass almost all of the Admin/User Dashboard items. This is getting up there in size so if there are any additional features, I can tackle those in another PR. For now, this covers the bulk of Dashboard Features with:

1. Active Devices
2. Scan Libraries
3. Shutdown/Restart Server
4. Run Scheduled Tasks

Screenshots will be updated as I flip this to 'Ready for review'

LePips (changes requested):
I have many comments regarding some basic SwiftUI/architecture stuff, but I won't really have time to cover it all soon. Most things look good right now but just needs some polish to take into account more-extreme device counts and performance.

---

**Active Devices**

This should be its own tab. This would save us an issue in the long run if people have many, many devices active within 16 minutes (there are people who do) and makes design easier imo. That page should use a `CollectionVStack` instead where we can have a card and a list view.

I apologize I didn't respond to earlier comments on design but I don't necessarily fancy the backdrop with all text overlaid. I would prefer a basic design closer to just having the portrait poster, title, subtitle (if applicable), and basic stream information. I also apologize I don't have a concept here and due to that, you can forgo doing this redesign and I can do that instead.

**Active Session**

I would also use a design similar to what I would like above for the item info. Additionally, I don't really care to show the description. That should have also been in a `TruncatedText` to defend against long descriptions.

**Tasks**

Good job on the task progress tracking. I would also want to show the last time run but I don't necessarily care about the "took X time". We will have to update the tasks to get the new `TaskResult`s. I would also like to account for the states within `TaskCompletionStatus` and the error message, if applicable.

We also have to think about our interactions as a client when performing server administration. From our client, people with large libraries could easily slow down their server with a large library scan or retrieving images. I am unsure if I want a confirmation alert before running each task, but I'll have to think about it more.

---

Lastly, all of the above does require active updates but polling every second can be very heavy, _especially_ for active sessions (those are some big structures and the endpoint doesn't support pagination). Even for Quick Connect, I poll 5 seconds for essentially a small string. With short polling intervals, we could be cancelling successful API calls that take > 1 second, so we enter a loop where we don't get any updates at all.

The web doesn't face this issue because it uses sockets to respond to session update events instead of polling. Sockets would be nice in the long run but I won't block this work because we don't have that.

We should increase the interval to at least 3-5 seconds. That would lose the "real time" seconds on sessions that are playing but I'm fine with that. We could implement some mechanism with the 1 second timer to give the illusion of real-time playing in the future (but then if the session pauses within the interval it would "jump back").

JPKribs:
I can start on that! I might ping you with the new grid button vs list button just to make sure I'm on the right page.

My only concern is:
> That page should use a CollectionVStack instead where we can have a card and a list view.

I was finding that items in the CollectionVStack were flickering when the button was updated. There is a good chance that removing the layered poster eliminates this as well but that's the only reason I moved to a VGrid was to eliminate flickering each time it got a response from the server.

Lastly, I know you were asking me to make these ViewModels stateful. I *think* I did that but please feel free to let me know if these are not correct and I can take another look!

JPKribs:
> ### Tasks
> 
> Good job on the task progress tracking. I would also want to show the last time run but I don't necessarily care about the "took X time". We will have to update the tasks to get the new TaskResults. I would also like to account for the states within TaskCompletionStatus and the error message, if applicable.
> 
> We also have to think about our interactions as a client when performing server administration. From our client, people with large libraries could easily slow down their server with a large library scan or retrieving images. I am unsure if I want a confirmation alert before running each task, but I'll have to think about it more.

Do the screenshots below look correct:

<details>
<summary>Filled Out Tasks:<summary>
Simulator Screenshot - iPhone 16 Pro - 2024-09-17 at 19 19 32
</details>

Which type of warning would you prefer I use? Or, should I be splitting this how I have this how where Server Shutdown/Restart are the Type 1 and Type 2 is the alert style:

<details>
<summary>Warning (Type 1)<summary>
Simulator Screenshot - iPhone 16 Pro - 2024-09-17 at 19 21 50
</details>

<details>
<summary>Warning (Type 2)<summary>
Simulator Screenshot - iPhone 16 Pro - 2024-09-17 at 19 22 41

</details>

Swiftfin/Views/UserDashboardView/ScheduledTasksView/ScheduledTasksView.swift
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    @ViewBuilder
+    private func sectionForCategory(_ category: String, tasks: [TaskInfo]) -> some View {
+        Section(header: Text(category)) {
+            ForEach(tasks, id: \.id) { task in
+                if let taskName = task.name,
+                   let taskID = task.id
+                {
+                    ScheduledTaskButton(
```
LePips, line 118: To make this simpler, just pass the task itself instead of extracting all of the values.

Swiftfin/Views/UserDashboardView/ScheduledTasksView/ScheduledTasksView.swift
```diff
@@ -0,0 +1,135 @@
...
+
+        ForEach(groupedTasks.keys.sorted(), id: \.self) { category in
+            sectionForCategory(category, tasks: groupedTasks[category] ?? [])
+        }
+    }
+
+    @ViewBuilder
+    private func sectionForCategory(_ category: String, tasks: [TaskInfo]) -> some View {
+        Section(header: Text(category)) {
+            ForEach(tasks, id: \.id) { task in
```
LePips, line 114: Make a `TaskInfo` file in `Shared/Extensions/JellyfinAPI` and have `TaskInfo` conform to `Identifiable`.

LePips:
The new tasks looks better but I personally don't really care for the "ran for X minutes" bit or showing that it was completed outright, only that it failed or cancelled/aborted as I see that more pressing. I would also want to use the label I have been using to indicate warnings elsewhere:

```swift
Label(/* message */, systemImage: "exclamationmark.circle.fill")
```

I think I like the `confirmationDialog` more than an `alert` for these (so same as shutdown server) and think we should transition some of my pre-existing alerts to that instead (where applicable later). We can do another layer of polish by checking whether the task had failed and show `TaskResult.longErrorMessage` within the dialog if cancelled/aborted and exists and have the button title of `Rerun task`.

To make this easier in code, we don't need a bunch of booleans for showing which dialog for each task. There are overloads that take a `presenting: T?` parameter that refer to an object that will be used when constructing the `actions` and `message`.

---

The session design was a bit more what I was looking for however I would like to use the portrait images instead. I will have to think more about the overall design.

sjorge:
> The session design was a bit more what I was looking for however I would like to use the portrait images instead. I will have to think more about the overall design.

At least on my server, Portrait (for movies/shows) is (almost) always available, backdrops, thumbs, episode thumbs are not. So I think it would indeed make sense to use portraits here, I think for most servers they will have a higher chance of being available.

JPKribs:
> The new tasks looks better but I personally don't really care for the "ran for X minutes" bit or showing that it was completed outright, only that it failed or cancelled/aborted as I see that more pressing. I would also want to use the label I have been using to indicate warnings elsewhere:
> 
> ```swift
> Label(/* message */, systemImage: "exclamationmark.circle.fill")
> ```
> 

Is this what you were thinking? There are 4 potential statuses (Completed, Cancelled, Aborted, Failed) do we just want Yellow for anything not Completed or do differing colors based on result:

Screenshot 2024-09-18 at 08 44 13

> I think I like the `confirmationDialog` more than an `alert` for these (so same as shutdown server) and think we should transition some of my pre-existing alerts to that instead (where applicable later). We can do another layer of polish by checking whether the task had failed and show `TaskResult.longErrorMessage` within the dialog if cancelled/aborted and exists and have the button title of `Rerun task`.

I've updated the ScheduledTaskButton to use confirmationDialog for everything. Do we always want the run button to be destructive or should I only use that for the Server Shutdown/Restart?

Screenshot 2024-09-18 at 08 36 41

I have the TaskResult.longErrorMessage in there as well but I can't seem to figure out how to manually get that to trigger on server side. Even aborted doesn't have one.

LePips:
- Yes, the labels look good like that. I will probably go in and adjust the font sizes.
- Let's only have the destructive role for the shutdown/restart. Also, it's `Rerun`.
- It is `portraitImageSources`, however you don't need to be storing that anywhere on the view. You can just pass that directly into the `ImageView`.

JPKribs:
I've made those updates. I still have some cleanup I want to do with the pivots but I wanted to get the listView version of this out there so I could make sure this was going the right direction:

details>
<summary>Grid View:</summary>
Simulator Screenshot - iPhone 16 Pro - 2024-09-17 at 19 19 32
</details>

<details>
<summary>List View</summary>
Simulator Screenshot - iPhone 16 Pro - 2024-09-17 at 19 21 50
</details>

Let me know if we want anything added or removed. The list view I think looks a little weird with the text alignment when there is no NowPlayingItem but it'd look weirder to align it with no image.

I wanted to pull the same device icons that server is using but it looks like those are just hardcoded paths in the web folder. I could still get those by doing like... `{Server URL}/path/to/image.svg` or we could provide our own images but I thought I would throw this out there before I started looking at this.

JPKribs:
I think I'm happy with the current state of this. I'm sure there is some room for improvement in terms of the Swift / Code Layout side of this. I have a lot of stuff broken out into files for repeated sections but I've pulled back a bit for smaller sections. I'm also having a tough time determining when to use a Func vs a Var so I'm sure some of that can be cleaned up as well.

The big highlights since the last review are:
- Added a List Mode
- Added a List/Grid Toggle that saves to User data
- Truncated the Session Details Overview Text with a See More view
- Re-organized the User Dashboard folders
- Moved the Grid Button posters from a background to a Portrait Poster
- Lowered the Timers for refresh from 1 sec to 5 sec
- Moved all the Active Device / Scheduled tasks off the User Dashboard into their own Sections
- Added an Active Device Count to the Chevron Button on the User Dashboard
- I'm sure some other cleanup I forgot about as well...

Please see our screenshots below! For feedback, I'm primarily concerned about format / UI just to make sure that's all up to snuff. Outside of that, I am not sure if my stateful view models are correct since I kind of based them off HomeView.

<details>
<summary>Dashboard</summary>
Dashboard
</details>

<details>
<summary>List View</summary>
ListView
</details>

<details>
<summary>Grid View</summary>
GridView
</details>

<details>
<summary>Grid View</summary>
TaskView
</details>

LePips:
I've had time to take a look at this and have started on my changes. Some main changes:

1 - I have split the "server connection" and "Dashboard" settings views as they serve different purposes. The former being how the client is connected to the server.
2 - I have removed the `CurrentUserViewModel` as we already retrieve the latest user data when we do the "server check". This is done [here](https://github.com/jellyfin/Swiftfin/blob/9ea989c5131acd305344f0bd859c652f5f674eea/Shared/ViewModels/ServerCheckViewModel.swift#L43). While we only get this data when we perform the check, a better way in the future would be implement a background task system that runs periodically.
3 - Some changes to the task progress tracking and most views for tasks.

I have really only taken a look at the tasks and haven't gotten to the active devices yet.

JPKribs:
I'm doing a bit of travel for work but I'd love to assist when I'm back. Is there anything you'd like me to work on for this or is this mostly cleanup/design to make sure it mirrors the rest of Swiftfin?

> 1 - I have split the "server connection" and "Dashboard" settings views as they serve different purposes. The former being how the client is connected to the server.

Since this view only has one function (Switch URL) would it make sense to just make the existing Chevron button for this view into the Menu to select the URL instead? Flatten out the hierarchy a little? Then again, I know that view is re-used so the benefit might not be there.

LePips:
No, as that view is also used for the deletion and could gain more functionality with stuff like #508.

You've made a great foundation and I'm almost done with my changes. After I am done I'll make comments on the major things I changed as well as some things I may have misguided on.

JPKribs:
Sounds great! Thank you (once again) for helping me get this over the finish line!

Feel free to let me know if there's anything that I can do to help out. If you'd like, I'd be more than happy to go through and localize all of the strings when you're done with it. I was looking through your commit, and I saw that you TODO'd localizing the transcoding reasons. I'd be more than happy to do that part before we have our final merge.

LePips:
Still have some work to do but most visual things should be done or has relevant TODOs.

Major changes:
- Copying the Settings app headers, have descriptive headers in some areas where I feel is applicable. This header should not be used necessarily everywhere, but we should have in more places.
- Instead of having the single tasks view model track the tasks, each task tracks its own progress since it can query itself.
- Tasks can be pushed to a detail view to show more information about the task and where to edit the triggers. Most of this is still left as a TODO.
- I misled on having a grid and row on the active devices, thinking to match the libraries for items. However the amount of information doesn't/shouldn't differ in each view type unlike item libraries, so it seemed redundant to have both. However, this can change with the issue I describe below when more space is allowed.
- For the session rows, I switched to using `ListRow` since that view is made for these view contexts.
- For sessions, we don't need to see a lot of detail on the item itself, even though the information is available. It may be desirable in the future to have the ability to deep link to that item.
- I was under a previous assumption that devices could have multiple sessions, so then that would cause the previous reused session observer to point to wrong sessions. However that assumption was incorrect and I didn't check before I over-engineered a session updating/removing. I'll be taking a look at that again to reuse the session observer like you had it.
- For polling, I've now learned that we want to instead delay calls after successful calls instead of view polling. Polling causes issues for cancelling previous tasks and we can live with tasks taking longer than the polling interval. The QuickConnect object also does a delay instead of polling and I haven't really had to think about these constant updates, so now in the future we have more guidance.
- It was quick for me to add the ability to see server logs.
- Handle music art for sessions. However I haven't been able to test with other media like books.

For many of the changes I would recommend learning from in future work:
- Instead of making strings that are `Title:`, use a `TextPairView`. These should be in their own rows in `List/Form` contexts instead of in a `VStack`.
- For formatted text/dates, use [`Text(_:format:)`](https://developer.apple.com/documentation/swiftui/text/init(_:format:)-9d2x4) instead. I have been moving to this primarily in #1203 and was able to bring over some of my work done there. I would recommend learning more about `FormatStyle` and the many (admittedly confusing) provided formats in SwiftUI.
- There are many uses of `foregroundColor`. Please use [`foregroundStyle`](https://developer.apple.com/documentation/swiftui/view/foregroundstyle(_:)) instead. There are still uses of it throughout the project, however they should all be replaced.
- Most other changes are small that would match other app design. Like proper usage of `secondary` styles, fonts, spacing, and animations. A lot of this stuff I've developed myself over my years of SwiftUI.

---

When testing on the iPad, I realized using the dashboard can be rather small. The user is performing administrative work and we are restricting their viewport. This is a side effect iOS and iPadOS having the same presentations but we may need to change that once we take a look at replacing our navigation stack. Meaning, iPadOS would have some entirely different views to allow them to be full screen. However, that is something I will reevaluate after we drop iOS 15.

JPKribs:
I am still traveling but I had some time today at the airport to pick this back up. All I did was fix the CodeFactor issue and localize all of the Strings. I can take a look at the Edit Tasks screens this weekend. Just to confirm, is our goal there to allow updating the crons for scheduled tasks?

Other than that, I will follow the TODOs!

I'm loving the changes you made for this! It looks great!

JPKribs:
@LePips Do we want to mirror Jellyfin-Web like this (see below) or should an editView? The TaskTriggers don't have an ID so you can't really edit them. I assume that's why you can't do that on Jellyfin-Web. That being said, I am able to find and remove a trigger then replace it with a new one from the array as a psuedo edit. Do we want to have edit functionality or do we just want to mirror what's available on web?

Simulator Screenshot - iPhone 16 Pro - 2024-10-04 at 17 08 37

LePips:
Thank you for doing that work. I have completed the work that I feel needs to be done except for a few things I am willing to kick down the road:

- I had a hard time thinking about the timer/delay change for the active sessions/tasks and it works now, so I wasn't going to worry about it. Currently, if you are trying to view the active sessions and the server is shut down/response takes a really long time the view will go to a "cancelled error" because on the poll we cancel the previous request. I'm fine with that for now and I'm sure we'll come up with a universal solution for these kind of active updates/polling in the future.
- I thought about reverting back the active sessions observation like I said in my previous comment, but actually think my solution works better since the list constantly updates. I frankly didn't know what to do in this situation but will have to think about it.
- Added a few `TODO`s for polish investigations.

---

However, I ran into issues of testing the tasks since I wasn't able to run a task for a long amount of time. I'm somewhat confident in the observation of the tasks so I think it's okay.

---

For updating a `Task`, we should have the `Save` button in the navigation bar like in `Security` and `Custom Profile`. That then posts to `Paths.UpdateTask` with the list of triggers that we have.

However, this PR has laid the groundwork for the dashboard view and think is done. `TODO`s like editing tasks and having the device logo for inactive sessions (which I don't think is an explicit `TODO`, but wouldn't mind having) can be saved for later PRs.

LePips (approved):

JPKribs:
Sounds good. I'll hold onto what I'm doing for editing the task trigger and throw that into a new PR. I agree what exists now exceeds a lot of the original issue request so I'm happy.

I'll work on the Edit Task Triggers and look at some of the TODOs to see if I can wrap those up over the coming weeks.

PR #1243, opened 2024-09-24
Fix App Icon picker Xcode16/iOS 18

chickdan:
Something changed with iOS 18/Xcode 16 and the app icons are no longer showing up in the picker.

Apparently, this change was in [Xcode 16 beta 5](https://forums.developer.apple.com/forums/thread/757162) but I'm not finding the entry in the final release notes. So all I did was add them to the assets catalog as image sets.


chickdan:
I think the app has to be built with Xcode 16, I'm also getting a build failure in 15.4

LePips:
Thought this was just a development quirk, thanks for finding the depreciation in the release notes. We'll have to change app icons anyways for iOS 18.

LePips (approved):

PR #1247, opened 2024-09-28
Fixed Reset button for Selector view

### LINKED ISSUE #1246: The "Reset" button in Filters in search doesn't work
### Describe the bug

Go to the Search tab
click on any filter below the search bar
choose any filter
click on the "Reset" button
The previous choice will not be reverted

### Application version

1.2 (4)

TestFlight

### Device information

iPhone 11 Pro

### OS version

iOS 18.0

### Jellyfin server version

10.9.11

FrankreedX:
Closes #1246
Fixes the "Reset" button in the filter selector not resetting the selector itself, only the data outside the selector.
Screenshot 2024-09-28 at 12 46 21 PM


LePips (changes requested):
Sadly, this isn't the correct solution. That `StateObject` is there as an intermediate "binding" to update the view if it couldn't be from a parent view. For example with this change, the font selection breaks. See `BindingBox` for its justification.

- https://github.com/jellyfin/Swiftfin/blob/4cba7622263e1aad9c3777acb8ca3eb9b1392511/Shared/Objects/BindingBox.swift#L15

The difference in behavior comes in `FilterView` and how we perform some indirection wackiness to update the correct filters. Filters are a bit confusing.

- https://github.com/jellyfin/Swiftfin/blob/4cba7622263e1aad9c3777acb8ca3eb9b1392511/Swiftfin/Views/FilterView.swift#L44

---

I would be able to take a look at this a bit more later but I may just revert the `BindingBox`'s usage from #1216. I only implemented it since I thought I could save some code, but will revert if it's causing more issues than solving them.

FrankreedX:
If we want to support both use cases, maybe BindingBox can enable or disable its 2 way binding. From what i can see, BindingBox is a binding but only 1 way right? The other way can be a boolean that is passed in with a default of 2 way, and in special cases use 1 way instead.

LePips:
Bindings should ideally/really always be both ways. `BindingBox` essentially making a one way binding is more of an unwanted side effect and I'll have to take a look at it again, but below is quick fix.

- https://github.com/jellyfin/Swiftfin/compare/main...LePips:SwiftFin:fix-filter-reset

PR #1254, opened 2024-10-05
Expanded subtitle font size and change range

orenodinner:
Expanded subtitle font size and change range from min 8 to min 2


LePips (approved):
Hadn't approved earlier because I'm currently working on the video player.

PR #1258, opened 2024-10-06
 Add 'Enable Rewatching' and ' Max days' to Next Up

JPKribs:
### Context
Resolves: https://github.com/jellyfin/Swiftfin/issues/1211

This is a rework https://github.com/jellyfin/Swiftfin/pull/1212 where I DIDN'T try to rebase on Main, resulting in a git nightmare.

### Summary

This PR enables a feature found on Jellyfin-Web that limits the number of items in the 'Next Up' section. To further mirror Jellyfin-Web, this feature uses 0 as disabled _(All items are shown in Next Up)_ and only includes the parameter for NextUpDateCutoff when this is greater than 0.

While I was in there, realized that Jellyfin-Web had an option for EnableRewatching so included that in this PR. I was basing my Parameters off of this section of Jellyfin-Web: https://github.com/jellyfin/jellyfin-web/blob/master/src/components/homesections/sections/nextUp.ts

### Screenshots

**iOS**

<details>
<summary>New Settings (Disabled)</summary>
New Settings - Disabled
</details>

<details>
<summary>New Settings (1 Day)</summary>
New Settings - 1 Day
</details>

<details>
<summary>New Settings (Multiple Days)</summary>
New Settings - Multiple Days
</details>

**tvOS**

<details>
<summary>New Settings (Location)</summary>
New Settings - Location
</details>

<details>
<summary>New Settings (Disabled)</summary>
New Settings - Disabled
</details>

<details>
<summary>New Settings (1 Day)</summary>
New Settings - 1 Day
</details>

<details>
<summary>New Settings (Multiple Days)</summary>
New Settings - Multiple Days
</details>

<details>
<summary>New Settings (Subtitle Update)</summary>
New Settings - Subtitle Update
</details>


JPKribs:
This is up to date, merge-able on Main now, and cleaned up with only the new PR items.

I have a single outstanding issue:

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-10-06 at 17 37 06

This works, but the Ok/Done button is always the number??? I don't set that anywhere and I've been screwing around with this trying to fix it to no avail. I'm going to try this again with some fresh eyes later in the week. Unless @LePips, are you aware of what is going on here? It's only impacting tvOS.

Other than the tvOS Alert Button, I think all other changes are ready.

Shared/ViewModels/LibraryViewModel/NextUpLibraryViewModel.swift
```diff
@@ -7,11 +7,17 @@
 //
 
 import Combine
+import Defaults
 import Foundation
 import JellyfinAPI
 
 final class NextUpLibraryViewModel: PagingLibraryViewModel<BaseItemDto> {
 
+    @Default(.Customization.Home.maxNextUp)
```
LePips, line 16: Don't use the property wrappers in non-view contexts. While they technically can still work, use the subscript instead at the usage sites.

Documentation:
https://github.com/sindresorhus/Defaults?tab=readme-ov-file#swiftui-support

LePips, line 16: Don't use the property wrappers in non-view contexts. While they technically can still work, use the subscript instead at the usage sites.

Documentation:
[sindresorhus/Defaults#swiftui-support](https://github.com/sindresorhus/Defaults?tab=readme-ov-file#swiftui-support)

Swiftfin tvOS/Views/SettingsView/CustomizeViewsSettings/Components/Sections/HomeSection.swift
```diff
@@ -0,0 +1,74 @@
...
+                )
+                .onSelect {
+                    isPresentingNextUpDays = true
+                }
+                .alert(L10n.nextUpDays, isPresented: $isPresentingNextUpDays) {
+
+                    TextField(
+                        L10n.nextUpDays,
+                        value: $maxNextUpDays,
+                        format: .number
```
LePips, line 59: Use the `dayInterval(0 ... 1000)` format instead, then we don't need `maxNextUpDays`.

Swiftfin tvOS/Views/SettingsView/CustomizeViewsSettings/Components/Sections/HomeSection.swift
```diff
@@ -0,0 +1,74 @@
...
+                            )
+                        } else {
+                            return Text(L10n.disabled)
+                        }
+                    }()
+                )
+                .onSelect {
+                    isPresentingNextUpDays = true
+                }
+                .alert(L10n.nextUpDays, isPresented: $isPresentingNextUpDays) {
```
LePips, line 54: To at least help with the tvOS button oddity, we can add a `Button(L10n.done) { }` button to at least help with navigation. Match in iOS as well.

LePips (changes requested):
Per the tvOS issue of dismissal in https://github.com/jellyfin/Swiftfin/pull/1212#issuecomment-2395583707, that happens in the simulator for me when I use `return` on the keyboard and works as expected when I use the on-screen remote.

Per the additional button on tvOS with the alert, that looks like a SwiftUI bug and probably nothing we can do about it.

Swiftfin tvOS/Views/SettingsView/CustomizeViewsSettings/Components/Sections/HomeSection.swift
JPKribs: I added this at one point but ended up with 2 buttons. One with the number and one with Done.

IMO, the number button **appears** a little more intentional than a number next to a done. But also, I do NOT have a knack for UX. Let me know if you'd prefer 1 number button or a number next to a done button and I'll take that route!

Shared/Strings/Strings.swift
```diff
@@ -456,6 +464,10 @@ internal enum L10n {
   }
   /// No title
   internal static let noTitle = L10n.tr("Localizable", "noTitle", fallback: "No title")
+  /// %@ Days
+  internal static func numberOfDays(_ p1: Any) -> String {
```
LePips, line 468: No longer need this string.

Shared/Strings/Strings.swift
```diff
@@ -726,6 +738,8 @@ internal enum L10n {
   internal static func signInToServer(_ p1: UnsafePointer<CChar>) -> String {
     return L10n.tr("Localizable", "signInToServer", p1, fallback: "Sign In to %s")
   }
+  /// 1 Day
```
LePips, line 741: No longer need this string.

Swiftfin tvOS/Views/SettingsView/CustomizeViewsSettings/Components/Sections/HomeSection.swift
```diff
@@ -0,0 +1,69 @@
...
+        private var showRecentlyAdded
+        @Default(.Customization.Home.maxNextUp)
+        private var maxNextUp
+        @Default(.Customization.Home.resumeNextUp)
+        private var resumeNextUp
+
+        @State
+        private var isPresentingNextUpDays = false
+        @State
+        private var maxNextUpDays: Int
```
LePips, line 25: This is no longer needed.

Swiftfin/Views/SettingsView/CustomizeViewsSettings/Components/Sections/HomeSection.swift
```diff
@@ -0,0 +1,69 @@
...
+        private var showRecentlyAdded
+        @Default(.Customization.Home.maxNextUp)
+        private var maxNextUp
+        @Default(.Customization.Home.resumeNextUp)
+        private var resumeNextUp
+
+        @State
+        private var isPresentingNextUpDays = false
+        @State
+        private var maxNextUpDays: Int
```
LePips, line 25: This is no longer needed.

Swiftfin tvOS/Views/SettingsView/CustomizeViewsSettings/Components/Sections/HomeSection.swift
LePips: I'm a bit perplexed on why that button shows up but I also haven't tested on an actual device. There have been differences and quirks between the tvOS simulator and device in particular. Sure, let's just not add the `Done` button and maybe in the future when this issue is fixed we will get the system `OK` button.

LePips (changes requested):

Swiftfin tvOS/Views/SettingsView/CustomizeViewsSettings/Components/Sections/HomeSection.swift
JPKribs: Both HomeSections should mirror each other now. I tacked on a TODO to look at this later. I know the next tvOS build is a ways away but I'll double check this when there is the next Testflight.

JPKribs:
I've updated this with the last outstanding changes from the review. There is a TODO for the weird tvOS Alert Button bug(?) and I've merge in missing changes on Main. Let me know if there is anything else needed! Otherwise, I think this is ready.

sjorge:
I'm not a developer so this is probably a dumb question, but why not use a simple text field that only accepts numbers for this?

That should work on iOS and tvOS, there is even a simple number one I think? (Although that might be 18.x only, I've only recently seen that for the first time)

JPKribs:
> That should work on iOS and tvOS, there is even a simple number one I think? (Although that might be 18.x only, I've only recently seen that for the first time)

I might be mistaken but this **IS** using the default TextField:

```
TextField(
    L10n.nextUpDays,
    value: $maxNextUp,
    format: .dayInterval(range: 0 ... 1000)
)
.keyboardType(.numberPad)
```

The primary item we are discussing is the delivery method for this. The popup is an 'alert' which is working fine on both iOS and tvOS but for whatever reason, the done button that all alerts just have by default is showing the wrong text on tvOS. The alert presents the TextField where edits are made without taking the user to a new screen or covering up the existing screen (see below). This is either something not working correctly with the tvOS emulator for XCode or SwiftUI itself.

iOS:
Screenshot 2024-10-07 at 14 55 55

Since we're a ways away from a tvOS build on TestFlight **(NOT a statement just _my_ assumption)**, I'd argue this is worth revisiting at a later point instead of engineering 2 separate solutions to account for functionality that should just work in SwiftUI.

I intend on keeping an eye on this and, as a build is available for tvOS, I can confirm if this issue is present on a non-emulator and make changes accordingly!

LePips (approved):

sjorge:
According to Testflight this should be in 1.3 ... but It's not showing up for me. (The active devices dashboard is, so I'm definitely on the correct build)


IMG_2766

LePips:
Oops, I got the notes wrong. A new build should be up shortly.

Issue #1239, opened 2024-09-20
Double Touch not recognized


mdebbar:
### Describe the bug

The "Double Touch" gesture isn't being recognized during playback. I tried setting it to "Pause/Play" and "Aspect fill" but it doesn't work. Other gestures are working fine for me.

<img width="784" alt="image" src="https://github.com/user-attachments/assets/1afb02b1-dba8-4c3f-b033-e521e5fc05be">

### Application version

1.1 (78)

### Where did you install the app from?

App Store

### Device information

iPhone 15 Pro & iPad Pro 12.9" (4th gen)

### OS version

iOS 18.0 & iPadOS 17.6.1

### Jellyfin server version

10.9.10

lfsty:
yeah, got the same problem

PR #1260, opened 2024-10-08
Make `GestureView` respond to double touch gesture

StevenWong12:
This PR resolves https://github.com/jellyfin/Swiftfin/issues/1239

I find it a low-hanging fruit. If there is any other problem I didn't consider, please let me know.


LePips (approved):
The `didPerformTap` is a silly oversight from me. This will make it into TestFlight, but will be reworked by next release by #1203, which will/should clean up a lot of these files.

PR #1272, opened 2024-10-14
[iOS] Admin Dashboard - Active Sessions Play Method Fix

JPKribs:
Previously, I was just using the PlayState.PlayMethod and directly using that value. It looks like Jellyfin-Web is using a helper method for this:

https://github.com/jellyfin/jellyfin-web/blob/50dd1fb855a00633659a92430b8e97e397f06222/src/components/playback/playmethodhelper.js

On Jellyfin-Web, some items were PlayState.DirectStream were showing as DirectPlay and I tracked it down to this helper. This should just make sure they are consistent.


JPKribs:
So, Remux isn't a PlayMethod so I make this output a String from the SessionInfo. The reason I chose to do this from SessionInfo is in order to mirror Jellyfin-Web because it's the location where the TranscodingInfo and PlayMethod live. I have it output a string instead of a PlayMethod? because Remux isn't included in the PlayMethod enum.

```
    var playMethodDisplayTitle: String? {
        guard self.nowPlayingItem != nil,
              let playState = self.playState,
              let playMethod = playState.playMethod
        else {
            return nil
        }

        if (self.transcodingInfo?.isVideoDirect ?? false || self.transcodingInfo?.videoCodec == nil) &&
            self.transcodingInfo?.isAudioDirect ?? false
        {
            return L10n.remux
        } else if self.transcodingInfo?.isVideoDirect ?? false {
            return PlayMethod.directStream.displayTitle
        } else if playMethod == .transcode {
            return PlayMethod.transcode.displayTitle
        } else if playMethod == .directStream || playMethod == .directPlay {
            return PlayMethod.directPlay.displayTitle
        }

        return nil
    }
```

Shared/Extensions/JellyfinAPI/SessionInfo.swift
```diff
@@ -0,0 +1,36 @@
...
+// Copyright (c) 2024 Jellyfin & Jellyfin Contributors
+//
+
+import Foundation
+import JellyfinAPI
+
+extension SessionInfo {
+
+    var playMethodDisplayTitle: String? {
+        guard self.nowPlayingItem != nil,
```
LePips, line 15: Here is a refactor to be a little more Swift-y with optional handling and to be exhaustive.

```swift
guard nowPlayingItem != nil, let playState, let playMethod = playState.playMethod else { return nil }

if let transcodingInfo {

    let isVideoDirect = transcodingInfo.isVideoDirect ?? false
    let hasVideoCodec = transcodingInfo.videoCodec != nil
    let isAudioDirect = transcodingInfo.isAudioDirect ?? false

    if isVideoDirect || hasVideoCodec, isAudioDirect {
        return L10n.remux
    } else if isVideoDirect {
        return PlayMethod.directStream.displayTitle
    }
}

switch playMethod {
case .transcode:
    return PlayMethod.transcode.displayTitle
case .directPlay, .directStream:
    return PlayMethod.directPlay.displayTitle
}
```

I would have also made the following comments:
- use `if let` shorthand instead, where applicable
	- https://www.hackingwithswift.com/swift/5.7/if-let-shorthand
- use implicit `self` instead

LePips (changes requested):

Shared/Extensions/JellyfinAPI/SessionInfo.swift
JPKribs: I can swap that out! Just to confirm, outputting a String (for this instance) and its location on SessionInfo is okay? 

I wasn't in love with the positioning but that logic is wonky IMO vs just using .PlayMethod literally.

LePips: This is where it belongs, since it uses other data from the session itself.

JPKribs:
Validated this should now mirror other Active Devices sections for other Jellyfin Clients.

JPKribs:
I've fixed the merge issues on this. This should be ready unless there are any additional updates that we want to make.

LePips (approved):

PR #1271, opened 2024-10-14
[WIP] [iOS] Admin Dashboard Enhancements

JPKribs:
### Formerly
https://github.com/jellyfin/Swiftfin/pull/1255

### Issue
This is an extension of https://github.com/jellyfin/Swiftfin/pull/1230

### Summary
This PR is where I have a couple different branches living together. I need to split them out into more manageable PRs but I wanted to make this visible so it wasn't just me working in the shadows. All of the Goals below are in this branch with varying levels of polish. As items are done, I'll check them off below.

### Goals
- [x] Fix PlayMethod DirectPlay vs DirectStream not mirroring Web | https://github.com/jellyfin/Swiftfin/pull/1272
- [x] Add device icons to Active Devices without NowPlayingItems | https://github.com/jellyfin/Swiftfin/pull/1275
- [x] Finish EditScheduledTaskView _(Task Triggers, Errors, Last Run)_ | https://github.com/jellyfin/Swiftfin/pull/1276
- [x] Add Devices view for managing Devices _(Think 'All Devices' in Web)_ | https://github.com/jellyfin/Swiftfin/pull/1277
- [x] Add Users view for managing Users _(All Users + Password Setting / Resetting)_ | https://github.com/jellyfin/Swiftfin/pull/1278
- [ ] Expand User management _(Permissions, Access, Parental Controls)_
- [ ] Migrate all AdminDashboard items into their own Coordinator
- [x] Create HomeView ActiveSessionsIndicator that takes the user into the AdminDashboard | https://github.com/jellyfin/Swiftfin/pull/1282

### ActiveSessionsIndicator
4BF72DC7-FD87-4CF7-B04E-67CD8386840E_4_5005_c

PR #1273, opened 2024-10-15
Generic Button to Input from an Alert

JPKribs:
I think this is my most re-used component that I was using in https://github.com/jellyfin/Swiftfin/pull/1271 so I thought it made sense to make sure this was done right before I threw this into 20 places.

The goal is to create a generic 'Chevron Button' that displays a value but, when selected, presents an input in an alert. Instead of mandating an input type, I have a trailing 'menu' input option that allows any view to be used as the input type.

Finally, I added some optional logic so this can be Init with strings or text, then the same options with onSave and onCancel logic so this can be used to input values and have those values not take effect until saved.

I wanted to split this out from any real logic changes since I intend on inputting values like this in a few different places so I wanted to be sure this was solid before ended up in a few different locations. For this PR, this is only replacing the maxNextUp input logic in SettingsView > CustomizeViewSettings > Components > HomeSection.

This logic is identical in iOS and tvOS.


chickdan (commented):
Given the ChevronInputButton.swift is identical for iOS and tvOS it could go in Shared/Components to reduce duplication.

JPKribs:
> Given the ChevronInputButton.swift is identical for iOS and tvOS it could go in Shared/Components to reduce duplication.

Good call! Changed.

LePips:
Since you were essentially wrapping `ChevronButton`, I took the opportunity to perform some refactors I had wanted.

Main changes:
- For cleanup, I was able to bring over my `TimeIntervalFormatStyle` from my current video player work
- Rename to `ChevronAlertButton`, since this at the very least presents an `alert` and one doesn't even need to implement "input" functionality. However, there is a better name or implementation of this functionality since "Chevron" is a design detail. Perhaps a `ViewModifier` that creates the `TextField` and a few other properties like keyboard style and format, since I anticipate this will mostly be for single-form input. At the same time, I anticipate us moving away from `alert` in many input-contexts.
- Move `ChevronButton` to `Shared` since it's the same view both for iOS/tvOS, probably didn't notice that on initial implementation.
- Changed the next up day formatting to days for now, as it actually seemed weird to me that you input days and get different date fields back. This may change when we implement field interval input.

Code review:
- When making views that essentially wrap other views don't copy their implementation as that reduces reusability.
- When making a few customized `init`s, please put them below the initial implementation in an extension and try to use the default generated init inside that. You will see that same pattern within most views.

LePips (approved):

PR #1275, opened 2024-10-15
[iOS] Admin Dashboard - Active Devices Icons

JPKribs:
This adds client icons when there they are active devices but have no NowPlayingItem. Also maps colors to the client to use as the background for the icons. Lastly, this forces the spacing & poster sizes. This prevents square posters from overflowing into the text when there are no other posters. This also makes sure the spacing between rows doesn't get weird when a square poster is present and all of the other posters shift up. See screenshots for details:

<details>
<summary>New Icon Examples</summary>
Screenshot 2024-10-15 at 09 12 00
</details>

<details>
<summary>Spacing Changes</summary>
Screenshot 2024-10-15 at 09 12 00
</details>

<details>
<summary>Force Albums into Size</summary>
Screenshot 2024-10-15 at 09 14 24
</details>


JPKribs:
I think the only weird part from this is that other clients are just parsing the DeviceName and ClientName to get the DeviceType and, if needed, browser type. I've recreated this logic in my DeviceType enum but I don't love how complicated it is. I'd much prefer fuzzier match where instead of "Swiftfin iOS, Swiftfin iPadOS, Swiftfin tvOS, Jellyfin iOS..." We could get apple with just "\bi(?:OS|PadOS)\b|\btvOS\b" + "Infuse" to capture all of these. I wasn't sure what the best route was so I went for the static route but I'd love to change this to be less maintenance.

Shared/Extensions/JellyfinAPI/DeviceType.swift
```diff
@@ -0,0 +1,176 @@
...
+        case .opera:
+            return Color(red: 1.0, green: 0.0, blue: 0.0) // Opera Red
+        default:
+            return Color.black
+        }
+    }
+
+    // MARK: - Client Title
+
+    var displayTitle: String {
```
LePips, line 134: I disagree that these should be localized as these are proper nouns. I don't have the time to see if every single brand here has opted into localized branding, but at least Android, Apple, and Firefox don't.

Shared/Extensions/JellyfinAPI/DeviceType.swift
```diff
@@ -0,0 +1,176 @@
...
+// Swiftfin is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2024 Jellyfin & Jellyfin Contributors
+//
+
+import SwiftUI
+
+enum DeviceType: String, Displayable, SystemImageable, Codable, CaseIterable {
```
LePips, line 11: I'm thinking this may also be a good addition to the SDK 🤔. However since this is now getting large enough I also think we should make this a `struct` with a private init with static instances.

Swiftfin/Resources/Assets.xcassets/DeviceIcons/Clients/Device-android.imageset/android.svg
LePips, line 1: This is the old Android logo, let's use the new one with the full android instead.

Swiftfin/Views/SettingsView/UserDashboardView/ActiveSessionsView/Components/ActiveSessionRow.swift
```diff
@@ -38,29 +35,47 @@ extension ActiveSessionsView {
...
-                    .squarePosterStyle()
-                } else {
-                    ZStack {
-                        Color.clear
-
-                        ImageView(session.nowPlayingItem?.portraitImageSources(maxWidth: 60) ?? [])
-                            .failure {
-                                SystemImageContentView(systemName: session.nowPlayingItem?.systemImage)
-                            }
+                        DeviceType(
```
LePips, line 41: Instead, we should have a new `device` property on `SessionInfo` that returns the device for that session.

```swift
extension SessionInfo {
  var device: DeviceType {
    // ...
  }
}
```

Shared/Extensions/JellyfinAPI/DeviceType.swift
LePips: I've also only meant to have `SystemImageable` to be used for `SFSymbol` providers. ie: when used in `Image(systemName:)` or `Label(_:systemName:)`.

Instead, make the property return an `ImageResource`.

```swift
var image: ImageResource {
    switch self {
    case .android:
        .deviceAndroid
    // ...
    }
}
```

Swiftfin/Views/SettingsView/UserDashboardView/ActiveSessionsView/Components/ActiveSessionRow.swift
```diff
@@ -38,29 +35,47 @@ extension ActiveSessionsView {
...
+                            client: session.client,
+                            deviceName: session.deviceName
+                        ).clientColor
+
+                        Image(DeviceType(client: session.client, deviceName: session.deviceName).systemImage)
+                            .resizable()
+                            .aspectRatio(contentMode: .fit)
+                            .frame(width: 40)
                     }
                     .posterStyle(.portrait)
```
LePips, line 51: I may want these as square instead.

LePips (changes requested):
All SVGs need to be changed to `Single Scale`.

- https://www.avanderlee.com/xcode/svg-image-assets/#how-can-i-use-an-svg-in-xcode

Swiftfin/Views/SettingsView/UserDashboardView/ActiveSessionsView/Components/ActiveSessionRow.swift
JPKribs: This is what that would look like:
Screenshot 2024-10-15 at 13 19 37

I originally had it as a square so I can go either way. Lemme know if you like this route and I'll leave this in the next commit!

LePips: I think I like the square better. Just apply the `squarePosterStyle` to it so that we can get matched styling.

Swiftfin/Resources/Assets.xcassets/DeviceIcons/Clients/Device-android.imageset/android.svg
JPKribs: This logo?
Device-android

LePips: Apologies, I was incorrect. Just the head is the official branding, so we can leave it as-is.

https://developer.android.com/distribute/marketing-tools/brand-guidelines

Shared/Extensions/JellyfinAPI/DeviceType.swift
JPKribs: I'm sorry if I'm not understanding this correctly. I have something that works for ImageResource but is this what you meant?
```
    var image: ImageResource {
        ImageResource(
            name: self.rawValue,
            bundle: Bundle.main
        )
    }
```

LePips: I'd like to use the generated asset enum instead so that we don't need strings for the assets as the raw value of this type.

- https://sarunw.com/posts/swift-symbols-for-asset-catalog/

LePips:
I did see your comment on the fuzzy matching and that sounds like a good general idea. Since that's just a nice-to-have I would want to keep it simple with natural regex matching, however a consensus wasn't made on its design.

https://forums.swift.org/t/se-0357-regex-string-processing-algorithms/57225/2

Shared/Extensions/JellyfinAPI/DeviceType.swift
JPKribs: I *think* I did this right. Part of what makes me think I didn't is I couldn't find an extension on Image Resources anywhere? I thought that would be used for Jellyfin-Blue-Blob since I see that used in .jellyfinBlueBlob. I created my own extension and there are no conflicts / duplicates. Let me know if I'm off there!

JPKribs:
> I did see your comment on the fuzzy matching and that sounds like a good general idea. Since that's just a nice-to-have I would want to keep it simple with natural regex matching, however a consensus wasn't made on its design.
> 
> https://forums.swift.org/t/se-0357-regex-string-processing-algorithms/57225/2

I tried to keep simple. Nothing crazy. I've tested on the following devices with no missed matches:

- Internet Explorer
- Google Chrome
- MSFT Edge
- FireFox
- Swiftfin
- Infuse
- Samsung TV _(Dev-Build of the Tizen)_
- Jellyfin Vue **_[This doesn't return a DeviceName so the result is the generic other but working as expected]_**
- Safari
- Android Phone

Let me know if you want to make any changes. My REGEX is functional but I'm not offended at all if there is any critique since I'm sure there are more efficient uses

JPKribs:
Also, I added checks to parse out webOS (LG TVs) but for the life of me I couldn't find a webOS logo svg. I left a TODO for that. I don't feel a ton of pressure since other clients groups webOS as part of "Other" so we'd be the first if we had this parsed in Swiftfin.

LePips (approved):
I haven't tested with all of the types but looks good to me. Thank your wife for making the webos logo!

PR #1280, opened 2024-10-17
[Fix] NextUp Updates Require Restart

### LINKED ISSUE #1279: NextUpViewModel Not Updating After Changing ‘Days in Next Up’
### Description:
When adjusting the “Days in Next Up” setting, the NextUpViewModel does not immediately reflect the updated value. The only way for the change to take effect is by restarting Swiftfin after modifying the setting.

### Steps to Reproduce:

	1.	Open Swiftfin.
	2.	Navigate to the “Settings > Customize” and change the “Days in Next Up” value.
	3.	Return to the Home Screen section.
	4.	Observe that the items in the Next Up item range are not updated according to the new “Days in Next Up” setting.
	5.	Restart Swiftfin and notice that the Next Up view now reflects the updated value.

### Expected Behavior:
The Next Up list should update immediately when the “Days in Next Up” value is changed, without requiring a restart of the app.

Actual Behavior:
The Next Up view only reflects changes to the “Days in Next Up” setting after restarting Swiftfin.

Additional Notes:
The issue seems related to the NextUpViewModel not reloading data when the “Days in Next Up” value changes. The current workaround is to restart the app, but this should be addressed so the view updates in real-time.

### Application version

1.3(1)

TestFlight

### Device information

All Devices

### OS version

iOS 17.6

### Jellyfin server version

10.9.11

JPKribs:
### Issue

Fixes #1279.

When I made this change: https://github.com/jellyfin/Swiftfin/pull/1258#discussion_r1789501879, I updated the maxNextUp and resumeNextUp settings to use let constants. However, this introduced an issue where you cannot update the values for maxNextUp or resumeNextUp without restarting Swiftfin. Changing them to var doesn’t fully resolve the problem either, as the values still don’t update dynamically.

### PR Fix

If I revert the change from #1258 (comment), the issue is resolved.

### Alternative Fix

I also explored another approach that allows these values to update dynamically without requiring a restart. The fix uses `Published` properties in combination with Defaults.publisher to observe changes and update the values in real-time.

```
@Published 
private(set) var maxNextUp: TimeInterval = Defaults[.Customization.Home.maxNextUp]
@Published 
private(set) var resumeNextUp: Bool = Defaults[.Customization.Home.resumeNextUp]

init() {
    Defaults.publisher(.Customization.Home.maxNextUp)
        .map { $0.newValue }
        .assign(to: &$maxNextUp)

    Defaults.publisher(.Customization.Home.resumeNextUp)
        .map { $0.newValue }
        .assign(to: &$resumeNextUp)

    // Other init code...
}
```

Let me know if we want to go the alternative route or if there is a better solution I haven't explored yet! For this PR, I chose to start with the 4 line change vs the 20 line change.

### DIFF
--- a/Shared/ViewModels/LibraryViewModel/NextUpLibraryViewModel.swift
+++ b/Shared/ViewModels/LibraryViewModel/NextUpLibraryViewModel.swift
@@ -13,11 +13,6 @@ import JellyfinAPI
 
 final class NextUpLibraryViewModel: PagingLibraryViewModel<BaseItemDto> {
 
-    @Default(.Customization.Home.maxNextUp)
-    var maxNextUp
-    @Default(.Customization.Home.resumeNextUp)
-    var resumeNextUp
-
     init() {
         super.init(parent: TitledLibraryParent(displayTitle: L10n.nextUp, id: "nextUp"))
     }
@@ -37,10 +32,10 @@ final class NextUpLibraryViewModel: PagingLibraryViewModel<BaseItemDto> {
         parameters.enableUserData = true
         parameters.fields = .MinimumFields
         parameters.limit = pageSize
-        if maxNextUp > 0 {
-            parameters.nextUpDateCutoff = Date.now.addingTimeInterval(-maxNextUp)
+        if Defaults[.Customization.Home.maxNextUp] > 0 {
+            parameters.nextUpDateCutoff = Date.now.addingTimeInterval(-Defaults[.Customization.Home.maxNextUp])
         }
-        parameters.enableRewatching = resumeNextUp
+        parameters.enableRewatching = Defaults[.Customization.Home.resumeNextUp]
         parameters.startIndex = page
         parameters.userID = userSession.user.id
 
--- a/Shared/ViewModels/LibraryViewModel/NextUpLibraryViewModel.swift
+++ b/Shared/ViewModels/LibraryViewModel/NextUpLibraryViewModel.swift
@@ -13,8 +13,10 @@ import JellyfinAPI
 
 final class NextUpLibraryViewModel: PagingLibraryViewModel<BaseItemDto> {
 
-    let maxNextUp = Defaults[.Customization.Home.maxNextUp]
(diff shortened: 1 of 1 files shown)


LePips (changes requested):
Per my comment, the wrappers shouldn't be used in non-`View` contexts.

- https://github.com/jellyfin/Swiftfin/pull/1258#discussion_r1789501879

> use the subscript instead at the usage sites

I must have missed making sure these were used correctly, but instead of making properties of the class use the `Defaults[.Customization.Home.maxNextUp]` _in_ the function.

JPKribs:
Apologies on this! I was copying a usage I saw elsewhere but using them in-function resolves this issue too. 

While I'm looking at this, is this the best way to do this? Using `nextUpDateCutoff = 0` creates different results than not including the parameter. 

This is what I currently have, which starts feeling a little worse now we have to call `Defaults[.Customization.Home.maxNextUp]` twice. I could create a `Defaults[.Customization.Home.limitMaxNextUp]` toggle if we think that is better. I was just trying to mirror other clients with the single input field but while I'm changing this, I thought I would check this:

```
if Defaults[.Customization.Home.maxNextUp] > 0 {
    parameters.nextUpDateCutoff = Date.now.addingTimeInterval(-Defaults[.Customization.Home.maxNextUp])
}
```

LePips:
You can just have a local variable:

```swift
let maxNextUp = Defaults[.Customization.Home.maxNextUp]
if maxNextUp > 0 {
    parameters.nextUpDateCutoff = Date.now.addingTimeInterval(-maxNextUp)
}
```

Yes, we need to account for `0` because irr the server doesn't.

JPKribs:
> ```
> let maxNextUp = Defaults[.Customization.Home.maxNextUp]
> if maxNextUp > 0 {
>     parameters.nextUpDateCutoff = Date.now.addingTimeInterval(-maxNextUp)
> }
> ```

Sounds good. Committed this change. I did a quick test with 0, 7, 31, & 1000. All look good and refreshing correctly.

LePips (approved):

PR #1278, opened 2024-10-17
[iOS] Admin Dashboard - Users Management (Passwords)

JPKribs:
I wanted to try and make this one a smaller one but it's still coming in at like 800 lines. Luckily, the majority of this work is the ViewModel, Observer, and I almost 100% copy/pasted the ResetUserPasswordView so that one should be familiar with my only addition being a Reset Password button.

This creates a view with all of the users on the Jellyfin Server. It says their name, shows their icon, and has the last time they were seen. This is ordered alphabetically.

Selecting a user takes you to a UserDetailsView. This view *WILL* contain more items:
1. Profile
2. Permissions _(Done, but this is a bigger PR so I wanted to split it up)_
3. Access
4. Parental Controls
5. Devices _(This is honestly a single line of code once https://github.com/jellyfin/Swiftfin/pull/1277 is merged)_

Currently, only the PasswordView is done. This is a direct steal of ResetUserPasswordView but uses my observer's actions instead. This Observer is Eventful so I am capturing a success/failure message depending on the response. The reset password is the only addition to ResetUserPasswordView.

Localizations are missing from this PR. Primarily because, if I localize everything then https://github.com/jellyfin/Swiftfin/pull/1277 gets merged, I have to do a whole thing to update it to this branch. Instead, I am just going to localize everything either once https://github.com/jellyfin/Swiftfin/pull/1277 is done OR, if this branch does in first, I have a branch where I have the User Permissions almost done that I can add the localizations to.

### Views

It's out of order... But UsersView | Password | UserDetailsView
Screenshot 2024-10-16 at 22 38 17

PR #1277, opened 2024-10-16
[iOS] Admin Dashboard - Device Management

JPKribs:
### Summary
Creates a DeviceView that allows for viewing all hardware that has connected to the Jellyfin Server and allows for assigning a custom name to them or deleting them.

On both the ViewModel and on the View, I do not allow the user to delete their own session. I think if were allowed to happen, then you would get 401'd and need to remove and re-add the user to get back in. 'Delete All' skips your device session. Not only does this mirror other clients with this functionality but it also avoids some weird interactions.

This should bring Swiftfin up to feature parity with Jellyfin-Web for the All Devices page features.

### Ideas
I think this is where I want this. The only feature I was realizing would be valuable is instead of a Delete All function, making the list so you can multselect, with a select all function on the list. Then, use that selection as the delete population. API-wise, there is no "Delete All" endpoint. It's always just been a forLoop to delete all Devices. There is no reason we couldn't make this a list pick instead. I stuck with Delete All to keep things matching but I'd be happy to look at the list idea too.

### Weird Behavior
Custom names on this SDK version are a little weird since we can **update** the customName but we cannot **see** the custom name. I have a TODO to update the deviceName to use CustomName when available. For now, you can update the custom name but it doesn't really take affect on the Swiftfin side so visually it's a little weird but it's all working fine behind the scene.

### Screenshots

<details>
  <summary>DevicesView</summary>

  Simulator Screenshot - iPhone 16 Pro - 2024-10-15 at 23:22:58
</details>

<details>
  <summary>Attempt to Delete Own Session</summary>

  Simulator Screenshot - iPhone 16 Pro - 2024-10-15 at 23:23:15
</details>

<details>
  <summary>Delete 1 Session</summary>

  Simulator Screenshot - iPhone 16 Pro - 2024-10-15 at 23:23:19
</details>

<details>
  <summary>Delete All Sessions</summary>

  Simulator Screenshot - iPhone 16 Pro - 2024-10-15 at 23:23:29
</details>

<details>
  <summary>Showing off the Device Icons in Action</summary>

  Simulator Screenshot - iPhone 16 Pro - 2024-10-15 at 23:24:51
  
</details>

PR #1287, opened 2024-10-23
[iOS] Admin Dashboard - Users

JPKribs:
### Reference
This is a rewrite of https://github.com/jellyfin/Swiftfin/pull/1278 with the goal of staying more focused on Users. I intend on adding password reset/changes later but for the sake of making this easier to review this is going to only tie back to users. Even that wasn't totally accomplished by this PR but it's much more focused than https://github.com/jellyfin/Swiftfin/pull/1278.

### Summary
Creates a list of all Users from the Jellyfin Server _(Mirroring DevicesView)_ and a DetailView. For now the DetailView is pretty empty but I linked it to the previous work done here: https://github.com/jellyfin/Swiftfin/pull/1277. This allows you to see devices owned only by a user. Additionally, this allows for deleting only the devices for a specific user when doing select all.

I have linked this UserView so it can be reached from the DeviceDetailsView by selecting the User or from the ActiveSessionsDetailView by selecting the User.

### Issues/Quirks
- My UserAdminViewModel has a loadDetails action. This is because it's possible to get to the user page with an incomplete UserDto. From DeviceDetailsView & ActiveSessionsDetailView you only have UserID and UserName so any future password resets or permission changes would not be possible. Loading Details ensures all information for that user exists in the Observer.
- There is a parameter in getDevices to filter by UserID. I don't know if this is an SDK issue or an API issue but this filter does not work. To get around this, I create var deviceBox which filters by the UserId. I want to eventually resolve the underlying issue but if it comes down to C# that's gonna bum me out...
- I use grayscale on user icons for users who are IsDisabled == True. Not a quirk but it could not the not-right way to go about it.

### Screenshots

<details>
    <summary>Updated Dashboard</summary>
    Simulator Screenshot - iPhone 16 Pro - 2024-10-23 at 15:51:20
</details>

<details>
    <summary>User List</summary>
    Simulator Screenshot - iPhone 16 Pro - 2024-10-23 at 15:51:28
</details>

<details>
    <summary>User Details</summary>
    Simulator Screenshot - iPhone 16 Pro - 2024-10-23 at 15:51:33
</details>

<details>
    <summary>Filtered Devices</summary>
    Simulator Screenshot - iPhone 16 Pro - 2024-10-23 at 15:51:55
</details>

### Future Steps
**UsersView**
- Create New User (AddUserView)
- Delete User (**Discussion Required**)

**UserDetailsView**
- Password (Reset/Set New)
- Profile (UserDto.Configuration)
- Permissions (UserDto.Policy)
- Access (UserDto.Policy)
- Parental Controls (UserDto.Policy)


Shared/ViewModels/UserAdminViewModel.swift
```diff
@@ -0,0 +1,232 @@
...
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2024 Jellyfin & Jellyfin Contributors
+//
+
+import Combine
+import Foundation
+import JellyfinAPI
+
+final class UserAdminViewModel: ViewModel, Eventful, Stateful, Identifiable {
```
LePips, line 13:
```suggestion
final class ServerUserAdminViewModel: ViewModel, Eventful, Stateful, Identifiable {
```

Swiftfin/Views/SettingsView/UserDashboardView/UsersView/Components/UsersRow.swift
```diff
@@ -0,0 +1,162 @@
...
+
+        // MARK: - Environment Variables
+
+        @Environment(\.colorScheme)
+        private var colorScheme
+
+        @CurrentDate
+        private var currentDate: Date
+
+        private var user: UserDto
```
LePips, line 32:
```suggestion
        private let user: UserDto
```

Swiftfin/Views/SettingsView/UserDashboardView/UsersView/Components/UsersRow.swift
```diff
@@ -0,0 +1,162 @@
...
+        private var rowContent: some View {
+            HStack {
+                VStack(alignment: .leading) {
+
+                    Text(user.name ?? L10n.unknown)
+                        .font(.headline)
+                        .lineLimit(2)
+                        .multilineTextAlignment(.leading)
+
+                    TextPairView(
```
LePips, line 99: I don't think this is that essential of information that it should be on every row. I do think it would be good to show something if the user is disabled though, but I have a different design in mind with a badge.

Swiftfin/Views/SettingsView/UserDashboardView/UsersView/UsersView.swift
```diff
@@ -0,0 +1,90 @@
...
+            }
+        }
+        .navigationTitle(L10n.users)
+        .onFirstAppear {
+            viewModel.send(.getUsers)
+        }
+        .refreshable {
+            viewModel.send(.getUsers)
+        }
+        .topBarTrailing {
```
LePips, line 46: We may want to retrieve users based on the hidden/disabled filters and have a filter for that, if we want to do that now.

LePips (changes requested):
Initial pass, looks great!

Swiftfin/Views/SettingsView/UserDashboardView/UsersView/Components/UsersRow.swift
```diff
@@ -0,0 +1,162 @@
...
+
+        var body: some View {
+            ListRow(insets: .init(horizontal: EdgeInsets.edgePadding)) {
+                userImage
+            } content: {
+                rowContent
+                    .padding(.vertical, 8)
+            }
+            .onSelect(perform: onSelect)
+            // TODO: Do we even want to allow User Deletion?
```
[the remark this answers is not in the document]
JPKribs: I'm happy to add this but I'm kind of stuck again for the UI. I could do a delete & a multi-select for deletion but I also have a branch where I am working on allow user creation via an add button in the top left.

Would we want to have edit and add?

Swiftfin/Views/SettingsView/UserDashboardView/DevicesView/DevicesView.swift
```diff
@@ -8,9 +8,12 @@
 
 import Defaults
 import JellyfinAPI
+import OrderedCollections
 import SwiftUI
 
 // TODO: Replace with CustomName when Available
+// TODO: Why doesn't this filter on the user from the Init?
```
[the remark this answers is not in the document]
JPKribs: The getUsers() call from the SDK is filtering or the filtering I added manually? That is, was this resulting in the filtered data Devices on only the selected user:

```
    init(userID: String? = nil) {
        _viewModel = StateObject(wrappedValue: DevicesViewModel(userID))
    }
```

I can give it another go, but I was getting all devices regardless of my userID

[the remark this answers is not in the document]
LePips: Ah my bad. Looking at the server code, it looks like the intended use of the endpoint is not to filter on the user id, but to return all devices on the server that the user _could_ connect to, regardless if they ever have. It looks like the server doesn't do what we want it to do, so I may be less inclined to link to devices from users this way.

- https://github.com/jellyfin/jellyfin/blob/a0c634a6edcebdf0ee1ed28fc51dfefaab8b9392/Jellyfin.Server.Implementations/Devices/DeviceManager.cs#L193
- https://github.com/jellyfin/jellyfin/blob/a0c634a6edcebdf0ee1ed28fc51dfefaab8b9392/Jellyfin.Server.Implementations/Devices/DeviceManager.cs#L233

I think overall the device/user system on the server doesn't do what we want it to do since the device doesn't store all past users that have connected through it, but only the last one.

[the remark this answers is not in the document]
JPKribs: That makes way more sense! I mean, we can still link through with 'lastUsername'. Primarily, what I see as valuable is the ability to nuke out all devices for a specific user. I wanted to add that as a toggle for a password reset function as well since this would help with a compromised password. What I have now gets us _pretty close_ to that functionality but it's definitely not 1 to 1 like I thought it was...

Does this still sound valuable or should I remove the User/Devices link?

Swiftfin/Views/SettingsView/UserDashboardView/UsersView/Components/UsersRow.swift
```diff
@@ -0,0 +1,162 @@
...
+                        }()
+                    )
+
+                    if let lastActivityDate = user.lastActivityDate {
+                        let timeInterval = currentDate.timeIntervalSince(lastActivityDate)
+                        let twentyFourHours: TimeInterval = 24 * 60 * 60
+
+                        TextPairView(
+                            L10n.lastSeen,
+                            value: timeInterval <= twentyFourHours ?
```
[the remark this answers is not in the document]
JPKribs: Does this look like how you're expecting it to or should it _just_ be the date and not the time?
Screenshot 2024-10-26 at 18 51 07

Swiftfin/Views/SettingsView/UserDashboardView/UsersView/UsersView.swift
JPKribs: Sorry, I don't think I fully understand what you mean. Do you mean add an action for getDisabledUsers / getHiddenUsers? Or add an extra parameter to the getUsers action like getUsers(active: true, inactive: false)?

Swiftfin/Views/SettingsView/UserDashboardView/UsersView/Components/UsersRow.swift
JPKribs: I've removed this TextPair. Do you want me to take a shot at adding an icon for this or do you want me to leave this as is?

Swiftfin/Views/SettingsView/UserDashboardView/UsersView/Components/UsersRow.swift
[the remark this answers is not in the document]
LePips: Just the date is good here. I can't remember what's done on the detail view, but we can present the time there since it's "more detail".

Swiftfin/Views/SettingsView/UserDashboardView/UsersView/Components/UsersRow.swift
LePips: I just tried my design and didn't entirely like it, I think I'm fine with just not showing that row.

JPKribs: It looks like web just makes the disabled users grayscale so we're mirroring that functionality. An icon would be a nice to have but I think just the grayscale should be enough. I have that all on my latest commit.

Swiftfin/Views/SettingsView/UserDashboardView/UsersView/UsersView.swift
LePips: The latter. We can just have 2 `State` variables for the filter states and are set through a `Menu`.

JPKribs: This is giving me some trouble since we have a state enum in the same struct. Is there a way to force this so it knows I mean the properly NOT the enum?
Screenshot 2024-10-26 at 22 44 24

LePips: I am unsure of what exactly that is coming from. I mean having two `@State` variables within the view like below:

```swift
struct ... : View {

	@State
	private var isDisabledFilterActive = false
	@State
	private var isHiddenFilterActive = false
}
```

When these are updated, we update the device list.

JPKribs:
I've got most of these changes in! I have a couple more design changes I'd love input on but I will pick this back up tomorrow/later in the week. I'm going to call it for now since I'm mostly happy with the current working version!

Swiftfin/Views/SettingsView/UserDashboardView/UsersView/UsersView.swift
JPKribs: Ohhhh..... I misunderstood. I was trying to put states on the viewModel... Sorry about that! I know what direction to go now

JPKribs: All right! I have this configured. This works how I would expect it and I have this defaulted to include all users regardless of hidden/disabled status. I think the only outlier now is making sure the filter is placed in a location that makes sense. I currently have it where it just sits on top of the user list.

Swiftfin/Views/SettingsView/UserDashboardView/UsersView/Components/UsersRow.swift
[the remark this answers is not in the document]
JPKribs: I think I've got this! Let me know if this UI is bad but I think I've got a good starting point for this:
Screenshot 2024-10-27 at 14 07 22
Screenshot 2024-10-27 at 14 07 31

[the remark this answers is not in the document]
JPKribs: I stole the Filters & Add button formatting from the SeeAllButton. I tried to get fancy with the icon on the outside but that might not be desirable.

As for the other select all/multi-select this is a pure drop in steal from how the DeviceView did this.

If we like this layout, after this PR I can add the filter option to the Devices as well. I also need to move Devices to be an array instead of using a box so I can do both of those at the same time. I don't even know what an add button would be on Devices so I think I would just add the filter and nothing else. For devices, I was thinking filter on Client?

JPKribs:
I think I have all of the changes accounted for now. This PR is getting a tad bigger because I decided to move my addUser logic into this PR. I'm happy with how the UI looks but this ServerUsersView now allows:

- Creation
- Deletion (Bulk/Single)
- Filtering

So if this feels too clutter I won't be offended. I've left the 2 outstanding review changes that I'm unsure about open. One for design and one for the User/Device link.

Swiftfin/Views/SettingsView/UserDashboardView/DevicesView/DevicesView.swift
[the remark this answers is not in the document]
LePips: Let's just remove the User -> Devices link. When a user password is reset all old tokens are deleted, so we wouldn't need to nuke the devices to get that functionality.

Swiftfin/Views/SettingsView/UserDashboardView/UsersView/Components/UsersRow.swift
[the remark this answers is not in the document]
LePips: See: https://github.com/jellyfin/Swiftfin/pull/1287#issuecomment-2445014476

JPKribs:
> I change the buttons to be in a `Menu`, as I don't really want to overload with buttons, but I may be too minimalistic. I've also made the delete button a pill button to live on the bottom toolbar and will want the other delete buttons to be migrated to it.
> 
> For user creation, that should live in its own view model.

I can take a look at this tomorrow! Thanks for the second set of eyes!

For the new viewModel, do you want a viewModel for the user list, user creation, user deletion or are we fine with a viewModel for the user list, & user deletion with the user creation on it's own viewModel only because it's on it's own View?

The reason I added user creation to the existing viewModel was to push the new user onto the .users set afterwards instead of reloading the whole list.

JPKribs:
> I change the buttons to be in a Menu, as I don't really want to overload with buttons, but I may be too minimalistic.

I had a chance to look this over. I'm actually a big fan! I think that might be a better look that some of the existing Edit/Add/Delete buttons I'm made for other parts of the Admin Dashboard. Do you want me to use this style for the other Edit/Add/Delete buttons in place of, for example, the add button so the ... is used there instead? Or should I leave views with only a single button as those single buttons?

I updated the Devices page to have the same Delete button.

I've updated everything related to users and I've tested that it's still all working as expected. I did see you commented out some of the backgroundStates.remove() calls in the ViewModels. I was guessing that's related to the warning they give since I'm 'not using' the results of the call. I'm not sure if there is a better way to do it but for now I have that re-enabled. Let me know if you have want to change this back.

Otherwise, I think this is ready to go from my end! Let me know if you want anything else adjusted!

LePips:
I did my cleanup and some refactoring for the device details since I noticed that was still in a box design and made its own view model.

For the add user view, that is now a modal since information creation/editing should now be in modals. Other design cleanup as well.

JPKribs:
> I did my cleanup and some refactoring for the device details since I noticed that was still in a box design and made its own view model.
> 
> For the add user view, that is now a modal since information creation/editing should now be in modals. Other design cleanup as well.

Thank you for your help cleaning this up! I do see 2 things I can address from my end but wanted your opinion:

Filtering users looks like this if there are no users (see below). I don't think this is bad but would we prefer this none is centered? Does the None need to have the separator removed? I'd guess this is the same style for the Devices as well but I don't think that matters since there will never be a case where there are no devices. Since, the device you are on will have to exist in the devices list.
Simulator Screenshot - iPhone 16 Pro - 2024-10-31 at 13 40 01

Secondly, when creating a user, I was having the viewModel for ServerUsersView always .getUsers onAppear to account for user creation. I don't think that was the *best* way to do this but the current version doesn't update the user list when a user is created. I see you have a comment for 

`TODO: somehow route to new user details`

on the addUser that kind of resolves this since the need to update the user list isn't there. That being said, for the time being do we feel fine using the onAppear() call for the ServerUserList for this purpose or is there a better route to add the new user to the ViewModel from another ViewModel?

LePips:
You can create a new notification in `SwiftfinNotifications` called `didAddServerUser` and use `onNotification` to listen when that is fired, passing in the new user object in the notification. From that, I would want us to push from the user list view to the user detail view so that the user could be further customized.

JPKribs:
> You can create a new notification in `SwiftfinNotifications` called `didAddServerUser` and use `onNotification` to listen when that is fired, passing in the new user object in the notification. From that, I would want us to push from the user list view to the user detail view so that the user could be further customized.

This was way easier than I thought it was going to be! I have something that is working how we expect it. All the existing Notification logic was really easy to follow and implement. I only have one quirk that works but might not be the best route. I put in a 0.1 second delay. Otherwise, the addServerUser `router.dismissCoordinator()` dismisses the new serverUserDetailsView that pops up. So, without the delay it all runs then immediately kicks me out of details. With this 0.1 second delay it triggers appropriately and at the end of the AddUser process, it leaves the user on the ServerUserDetailsView for the newly created UserDto.

Sorry to bother you further on this one. Let me know if there is anything I am missing or can improve on for this!

LePips:
At least for now with Stinsen, we can have an action fire after the coordinator is dismissed.

LePips (approved):

JPKribs:
> At least for now with Stinsen, we can have an action fire after the coordinator is dismissed.

Very cool! Thank you for your help on this!

PR #1296, opened 2024-10-29
[tvOS] ItemView Button Cleanup

JPKribs:
I originally made a branch with some lofty tvOS goal about making completing views and fixing the episode selector. Instead I'm going to settle for a very small change as I dip my toes back into tvOS.

### Summary
This PR 'cleans' up the action buttons & the episode cards on ItemViews. More accurately, it's just a bit of style changing for the action buttons but it makes it easier to create new action buttons in the future / change existing action buttons.

My main issue was the focus highlight around the action buttons was both way too big for the button and also a rectangle for a square/circle button selection. So I updated the ActionButton logic into it's own struct and performed all the formatting once for re-use with any additional buttons we may need there (Shuffle, Version Selection?, etc.). Additionally, it used more of the available room for these icons since they were 100x100 but confined to a much smaller area in reality.

I also increased the episode card spacing so focusing wont cause them to overlap anymore. Finally, I changed the play icon I added in a previous PR to be less jarring ( White -> Gray / .Secondary ). I added some spacing to the episode text as well to prevent that being so close to the edges. Finally, some spacing between the rows as well to give them more separation.

### Action Button Screenshots
<details>
<summary>This PR</summary>
Screenshot 2024-11-08 at 11 26 46
</details>

<details>
<summary>Current</summary>
Screenshot 2024-11-08 at 11 59 36
</details>

https://github.com/user-attachments/assets/52952828-de41-4913-bf54-546b561aadcc

### Episode Card Screenshots
<details>
<summary>This PR</summary>
Screenshot 2024-10-30 at 12 26 12
</details>

<details>
<summary>Current</summary>
PNG image
</details>


JPKribs:
Looking at this a bit later I think I can do better... I'm putting this on draft and coming back to this later with a fresh set of eyes.

Swiftfin tvOS/Extensions/ButtonStyles/BorderlessFocus.swift
```diff
@@ -0,0 +1,36 @@
...
+// Swiftfin is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2024 Jellyfin & Jellyfin Contributors
+//
+
+import SwiftUI
+
+struct BorderlessFocus: ButtonStyle {
```
LePips, line 11: We should actually try to "avoid" making `ButtonStyle`s for tvOS. I would like us to stick to the default focus and scaling that the system provides unless we absolutely want a different design.

Making buttons is tricky in tvOS and actually quite annoying back when I was first working on the platform.

Swiftfin tvOS/Views/ItemView/Components/EpisodeSelector/Components/EpisodeCard.swift
```diff
@@ -37,7 +37,7 @@ extension SeriesEpisodeSelector {
                     Image(systemName: "play.fill")
                         .resizable()
                         .frame(width: 50, height: 50)
-                        .foregroundColor(.white)
```
LePips, line 40: We will actually always be in dark mode, so passing in a hard color isn't bad. Additionally, if changing instances of `foregroundColor` we should switch to `foregroundStyle`.

Swiftfin tvOS/Views/ItemView/Components/ActionButtons/ActionButtonHStack.swift
```diff
@@ -0,0 +1,55 @@
...
+                // MARK: - Toggle Played
+
+                ActionButton(
+                    icon: "checkmark.circle",
+                    selectedIcon: "checkmark.circle.fill",
+                    color: Color.jellyfinPurple
+                ) {
+                    viewModel.send(.toggleIsPlayed)
+                }
+                .environment(\.isSelected, viewModel.item.userData?.isPlayed == true)
```
LePips, line 32:
```suggestion
                .environment(\.isSelected, viewModel.item.userData?.isPlayed ?? false)
```

Swiftfin tvOS/Views/ItemView/Components/ActionButtons/ActionButtonHStack.swift
```diff
@@ -0,0 +1,55 @@
...
+                // MARK: - Toggle Favorite
+
+                ActionButton(
+                    icon: "heart.circle",
+                    selectedIcon: "heart.circle.fill",
+                    color: .pink
+                ) {
+                    viewModel.send(.toggleIsFavorite)
+                }
+                .environment(\.isSelected, viewModel.item.userData?.isFavorite == true)
```
LePips, line 47:
```suggestion
                .environment(\.isSelected, viewModel.item.userData?.isFavorite ?? false)
```

Swiftfin tvOS/Extensions/ButtonStyles/BorderlessFocus.swift
JPKribs: No worries! That's kind of what I was wondering because this was a crazy amount of effort to get something that wasn't crazy looking. Doing this for every view would be... unpleasant... 

For now, do you want me to rollback this PR to just be the episodeCard spacing?

JPKribs:
I think that looks good! I think my only concern is, as we get more buttons, that might start to get a little overpopulated? I think the current size is good for having the ability to scale, but I agree right now with the two it looks a little bare.

On the other side, do we want to have the action buttons mirror each other between iOS and tvOS? Not necessarily in the focus at all of that but more in the shape and style.

JPKribs:
One revision - Flipped the ellipsis since the that gives us more horizontal spacing:

Screenshot 2024-11-07 at 15 51 43

I'm personally pretty happy with how this looks! Let me know if you have any feedback!

LePips (changes requested):
> as we get more buttons, that might start to get a little overpopulated?

That is something I've had to think about and will have some trouble with. Overall, I think a maximum of 3 actions there is the max but as you have listed there are many more actions that might be appropriate there, with my top 2 being the version selection and trailers.

I don't think that a menu is the best ux choice where we place everything else, as trailers and version selection should be more easily selectable. I also don't want more layers of buttons as that looks busy. So, idk what to entirely do with this area. We could play with widening that area to allow 4 buttons on the second row based on the available content (so that if there's only 2 available actions they won't be so wide).

Swiftfin tvOS/Views/ItemView/Components/ActionButtons/ActionMenu.swift
```diff
@@ -0,0 +1,57 @@
...
+        @Environment(\.isSelected)
+        private var isSelected
+        @FocusState
+        private var isFocused: Bool
+
+        @ViewBuilder
+        let menuItems: Content
+
+        @State
+        private var isShowingMenu = false
```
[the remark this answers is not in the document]
JPKribs: Shoot! That one was left over from an earlier version where I was putting the additional menu items in a sheet instead of using a Menu.

JPKribs:
I've added these changes to this PR. There is a way to get maybe 1 more on a single row with the ellipsis but it looks like this which might feel a little cramped:

Screenshot 2024-11-09 at 13 19 32

Outside of the menu button, I was thinking something like:

- Put the buttons in a CollectionHStack that can be scrolled through
or
- 2 rows getting us up to 4 to 6 buttons with everything else in a menu button

For scrolling, this is my **_rough_** mock up:

https://github.com/user-attachments/assets/461662ff-cd0a-4325-862b-13199dd74cdf

If we do the latter, it might make sense to have a setting for which buttons get included on the screen and which ones are in the menu. But I'm not sure what that looks like on tvOS... Style-wise, I like this PR is but yeah, the scaling side of this is hard to wrap my head around what would look good / make sense.

LePips:
Hm, 4 buttons doesn't look the worst and maybe the version selection we can make the small menu you made be on the play button row.

Anyways, this PR good to be merged if the current empty menu is commented out and iOS builds. 👍

JPKribs:
> Anyways, this PR good to be merged if the current empty menu is commented out and iOS builds. 👍

I think I have both of these. Please let me know if there are any additional changes I can make!

LePips (approved):

PR #1297, opened 2024-10-30
Project Settings Changes

chickdan:
- Replaced a deprecated option for SwiftFormat
- Added a compiler flag to warn against long expression type checking
    - Fixed an instance of long type checking (was around 800ms and is now silenced under 200ms)
    - There are still some warnings for expressions taking 5000-8000ms, but those stem from keypath resolving so it's a little more difficult to improve.
- Removed duplicate file references
- Bumped versions of SwiftFormat and the Swift version it works against

P.S. I will create incremental PRs as I work through increasing the Swift version that SwiftFormat targets as moving from 5.6 -> 5.10 creates many changes. So I'll move one at a time and open PRs for each to keep them small.

### DIFF
--- a/.swiftformat
+++ b/.swiftformat
@@ -1,4 +1,4 @@
-# version: 0.49.11
+# version: 0.54.6
 
 --swiftversion 5.5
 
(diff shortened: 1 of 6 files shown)


LePips (approved):
Thanks for doing this background work!

PR #1282, opened 2024-10-20
[WIP] [iOS] Home - Activity Indicator

JPKribs:
### Originally Discussed
https://github.com/jellyfin/Swiftfin/discussions/1269#discussioncomment-10926588

### Summary
Creates an optional indicator to the HomeView that shows how many active sessions are currently active on the Server. Active sessions are online sessions where there is a nowPlayingItem. 

This indicator has an action tied to it. Selecting this indicator takes the user to the AdminDashboard.

### Considerations
- I baked the viewModel into the ActiveSessionsIndicator. The reason I did this was because I didn't want to add an extra viewModel to the HomeView if the setting was disabled or the user was a non-admin. This way, only when the indicator is present is the viewModel attached to the HomeView.
- This viewModel loads onAppear and every 60 seconds. The video below I changed this to every 4 seconds to capture what that looks like better.

### Screenshot
<details>
<summary>Inactive Indicator</summary>
Screenshot 2024-10-19 at 22 00 10
</details>

<details>
<summary>Active Indicator</summary>
Screenshot 2024-10-19 at 22 01 57
</details>

<details>
<summary>New Setting (Only Visible when User.isAdministrator)</summary>
I had to inject currentUserSession to get this. I hope that was right?

Screenshot 2024-10-19 at 21 18 43
</details>

PR #1300, opened 2024-10-31
[iOS] Admin Dashboard - Migrate to Own Coordinator/Folder

JPKribs:
This looks a lot bigger than it is...

### Summary
This does the follow tasks:
1. Rename all instances of UserDashboard to AdminDashboard
2. Move all AdminDashboard items into their own folder (No longer inside the SettingsView folder)
3. Create an AdminDashboardCoordinator and move all coordinator AdminDashboard items from the SettingsCoordinator
4. Ensure that all AdminDashboard ViewModels are in the iOS Target Build only
5. Add the following to the AdminDashboardView to better mirror the SettingsView:

```
        .navigationBarTitleDisplayMode(.inline)
        .navigationBarCloseButton {
            router.dismissCoordinator()
        }
```

I also added all the AdminDashboard ViewModels to a single folder to clean up the ViewModels folder a little bit. Everything is tested and working as expected. Functionally, the only difference is the AdminDashboard is a Modal instead of a Push.


LePips (approved):
Thank you for doing this! I only changed it back to a push instead of a modal.

JPKribs:
> Thank you for doing this! I only changed it back to a push instead of a modal.

My only reason for that was so I could init from [here](https://github.com/jellyfin/Swiftfin/pull/1282). But, I realized that would be called from a totally different coordinator (HomeCoordinator) so push completely works here!

Issue #420, opened 2022-04-27
Noticible audio delay compared to video when using airplay audio output


jmesmon:
### Describe the bug

1. Set the audio output for the apple tv to airplay speakers (I do this by changing the default output in the apple tv settings, and am using a pair of homepods in a stereo group)
2. Play a video from swiftfin
3. note that the audio is somewhat delayed compared to the video. After pressing back or otherwise pausing, the audio continues playing for a few seconds. Other video streaming apps (Netflix, for example) have video and audio syncronized in this same configuration, and respond immediately to pausing the audio (they don't continue playing it for a few seconds)

### Application version

1.0.0(74)

### Where did you install the app from?

TestFlight

### Device information

Apple TV 4K, 32GB, A2169

### OS version

tvOS 15.4.1 (19L452)

### Jellyfin server version

10.7.7

LePips:
Thank you for the issue, however this is an issue of VLCKit and there is nothing that we can do about it.

jmesmon:
For reference, this appears to also affect bluetooth (though to a lesser degree because the bluetooth latency is lower).

Is there an open issue for this in VLCKit? If it is fixed in VLCKit, would any changes need to be made to Swiftfin to resolve this, or is this completely handled within VLCKit?

LePips:
Over the few years it must have come back up, with a more recent issue:

https://code.videolan.org/videolan/VLCKit/-/issues/544

Sapd:
It is an issue for me, too. Makes it completely unusable.
 I guess a lot of people use the Apple TV together with Homepods, probably affecting not few users.

Sapd:
Seems to be fixed with the latest release. Tested it with multiple movies.

edeustace:
Note that the native player does not have this issue, so you can switch to that if you're stuck.

JPKribs:
There are a few duplicates of this issue that have been made but please find my response here (on one of the closed issues 🤦‍♂️)

https://github.com/jellyfin/Swiftfin/issues/734#issuecomment-2723290284

The TL;DR is this issue is related to VLCKit and the issue is tracked on their repo. If there is any movement/update to VLCKit we will update it to incorporate the fix. That URL is correct. That issue on the VLCKit side is a bit broad but specifically references "Delays".

On our end, I've left #937 open to track this as to try and consolidate communication around this. That being said, it's still a bit more scattered than I'd like but that's unfortunately the nature of the beast wrangling a project like this.

[...]

PR #1302, opened 2024-11-02
[Meta] Player Comparison Page

JPKribs:
### Summary
Since there is a strange mix between AVKit and VLCKit features, I thought it would be nice to document the differences between the two of these. I was motivated by [this discussion](https://github.com/jellyfin/Swiftfin/discussions/953) and expanded on it by mirroring supported codecs directly from the device profile code that was written.

The goal for this PR is to have a reference to point to for questions about the differences between players.

[New Players Page](https://github.com/JPKribs/Swiftfin/blob/playerDocumentation/players.md)


RC2225:
Maybe add to the VLCKit audio section that audio over airplay has around 2 second delay while nativ is fine. See #420

Citrus0299:
This is a great addition, what about the TRUE HD, TRUE HD Atmos and EAC3 Atmos audio codecs?
It should also be noted that just because the player does support the codec, the hardware might not be able to pass it through, and it will be converted on device instead.

players.md
```diff
@@ -0,0 +1,85 @@
+# Player Differences
+
+Swiftfin offers two player options: the default Swiftfin player and Native (AVKit). The Swiftfin team recommends using the Swiftfin player for optimal compatibility and features, though the Native player is also available for certain cases that benefit from Apple's native capabilities. All video, audio, and subtitle formats listed are supported for direct playback. If transcoding is enabled on your server, any unsupported formats will be converted automatically.
```
holow29, line 3: Language here is a bit imprecise because most things on Native are going to be repackaged in HLS even if video/audio codec are supported for DirectPlay, given container limitations (e.g. anything in mkv). This means that most things, disregarding subtitles, not in MP4 container will be remuxed or DirectStreamed if video codec is supported and transcoded otherwise.

players.md
```diff
@@ -0,0 +1,85 @@
...
+
+---
+
+## Feature Support
+
+| Feature                 | Swiftfin (VLCKit)                                                                                               | Native (AVKit)                                                               |
+|-------------------------|-----------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------|
+| **Framerate Matching**  | ❌                                                                                                             | ✅                                                                           |
+| **HDR Playback**        | ❌                                                                                                             | ✅ **                                                                        |
```
holow29, line 12: Might be good to call out HDR supported formats: HDR10, HDR10+ (on certain devices like Apple TV 4k 3rd gen), Dolby Vision profiles 5, 8.
As a note for future, unsure about HLG and Dolby Vision profile 10 (used in AV1)

players.md
```diff
@@ -0,0 +1,85 @@
...
+| **WavPack**             | ✅                                                                                                             | ❌                                                                           |
+| **WMA (Lossless, Pro)** | ✅                                                                                                             | ❌                                                                           |
+
+---
+
+## Video Support
+
+| Video Codec             | Swiftfin (VLCKit)                                                                                               | Native (AVKit)                                                               |
+|-------------------------|-----------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------|
+| **AV1**                 | ✅                                                                                                             | ❌                                                                           |
```
holow29, line 85: Note for future research: this might have changed for Native since iOS/tvOS 18. There were some announcements/updates to Apple's HLS specs at WWDC 24, and AV1 with Dolby Vision profile 10 seems supported now, but I haven't been able to modify device profile and do any tests yet.

players.md
```diff
@@ -0,0 +1,85 @@
...
+| **H.265 (HEVC)**        | ✅                                                                                                             | ✅                                                                           |
+| **MJPEG**               | ✅                                                                                                             | ❌                                                                           |
+| **MPEG1/2/4**           | ✅                                                                                                             | ✅ (MPEG4 only)                                                              |
+| **VC1**                 | ✅                                                                                                             | ❌                                                                           |
+| **VP9**                 | ✅                                                                                                             | ❌                                                                           |
+
+---
+
+## Subtitle Support
```
holow29, line 111: Since subtitle selection isn't working on Native (AFAIK?), it might be good to call that out specifically in the subtitle support section.

holow29 (commented):
Thanks for working on this. Added a couple of comments, but I think it looks good otherwise.

players.md
JPKribs: I added a container section, do you think that suffices? Here is my up-to-date version:

https://github.com/JPKribs/Swiftfin/blob/playerDocumentation/players.md

holow29:
```suggestion
Swiftfin offers two player options: the default Swiftfin player and Native (AVKit). The Swiftfin team recommends using the Swiftfin player for optimal compatibility and features, though the Native player is also available for certain cases that benefit from Apple's native capabilities. All video, audio, and subtitle formats listed are supported for direct playback but might be repackaged based on container support. If transcoding is enabled on your server, any unsupported formats will be converted automatically.
```
I might consider adding that addendum just to be clear, but ultimately it doesn't make too much of a difference.

JPKribs:
> Thanks for working on this. Added a couple of comments, but I think it looks good otherwise.

Thank you for giving this a second look over! To be frank, I don't really encode HDR or anything too fancy so most of what I have is coming from your initial work on #519, and our finalized Device Profiles we have in code now. Additionally, there are some items in there from VLCKit / AVKit documentation but I'm not going to be the best expert for this.

I'm not at my desktop so I can't update the screenshot but my current version can be found here:

https://github.com/JPKribs/Swiftfin/blob/playerDocumentation/players.md

Let me know if there are any additional revisions!

Citrus0299:
Thanks a lot. I just noticed that DTS-HD MA is also missing from the list.

> I don't really encode HDR or anything too fancy
That's interesting as I only keep the highest quality possible in it's native format.

players.md
```diff
@@ -0,0 +1,152 @@
...
+
+---
+
+## HDR Support
+
+| Subtitle Format         | Swiftfin (VLCKit)                                                                                               | Native (AVKit)                                                               |
+|-------------------------|-----------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------|
+| **Dolby Vision Profile 5**              | ❌                                                                                                             | ✅                                                                           |
+| **Dolby Vision Profile 8**              | ❌                                                                                                             | 🟡 Compatible devices only                                                                         |
+| **Dolby Vision Profile 10**              | ❌                                                                                                             | ✅                                                                           |
```
holow29, line 140: IIRC this is only going to be found with AV1, so support is limited to devices with AV1 hardware decode (iPhone 15 Pro+)

players.md
```diff
@@ -0,0 +1,152 @@
...
+| **HDR10**                 | ❌                                                                                                             | ✅                                                                           |
+| **HDR10+**              | ❌                                                                                                             | 🟡 Limited support                                                                           |
+| **HLG**                 | ❌                                                                                                             | ❌                                                                           |
+
+## Notes
+
+- HDR10+ support in Native (AVKit) is limited to certain devices, such as the Apple TV 4K (3rd Generation) and recent iPhones and iPads with compatible hardware.
+- HLG (Hybrid Log-Gamma) support in Native (AVKit) is limited and may vary by device and OS version.
+- Dolby Vision Profile 10 (used in AV1) is currently not supported by Native (AVKit).
```
holow29, line 149: I'm not sure this is technically accurate. I haven't tested it, but it wouldn't surprise me if it worked in Native for the very limited amount of devices that support AV1 hardware decode (iPhone 15 Pro+)

holow29:
Made a couple more comments on HDR section, but otherwise looking good!

players.md
JPKribs: Specifically just HDR10 or all HDR profiles?

JPKribs:
> Made a couple more comments on HDR section, but otherwise looking good!

I had one last question on HDR, but then I feel ready to go with this. I appreciate your help looking this over!

players.md
holow29: Just Dolby Vision Profile 10 (which is different from HDR10)

LePips (approved):
Thank you for doing this and I'm glad that others were able to contribute!

~~I~~ Copilot just condensed the tables and I mostly made grammatical changes, as well as moving into `Documentation` since I feel like we may have more in the future.

PR #1305, opened 2024-11-03
[WIP] [iOS] Refresh, Edit, or Delete Media Items

JPKribs:
### Summary
The goal is to allow refreshing metadata for a select media item. Additionally, allow for manually editing metadata from iOS. I want to get the refresh working from tvOS but have no intention on allowing tvOS to change metadata text.

Eventually, I want to add the ability to add/edit metadata images, and edit subtitles but that's going to be in a later PR. Images and Subtitles require files to POST API so those are going to be a slower process.

### To Do:

- [x] Allow Refreshing an Item/Show/Season Metadata 
- [x] Finish the MetadataEditorView for ALL BaseItemTypes
- [x] Fix Community Rating Rounding
- [x] Localize
- [x] Validate Studio / Person / Genre / Tags Input Views

### Screenshots

<details>
<summary>New Button (Admin Only)</summary>
New Button (Admin Only)
</details>

<details>
<summary>Genres</summary>
Genres
</details>

<details>
<summary>Main</summary>
Main
</details>

<details>
<summary>Metadata</summary>
Metadata
</details>

<details>
<summary>People</summary>
People
</details>

<details>
<summary>Studios</summary>
Studios
</details>

<details>
<summary>Tags</summary>
Tags
</details>


JPKribs:
Hey @LePips no rush at all, but part of the metadata I am trying to wrap my head around allowing edits are Studios, Genres, People, and Tags. These are lists of values and I've gotten a system around editing them but I'm getting stumped on UI stuff again. Since these are edit screens, I made them modal but then I have trouble working around the X button in the top left. This is my current version below:

<details>
  <summary>Editing</summary>
  Simulator Screenshot - Editing view on iPhone 16 Pro, showing editable fields.
</details>

<details>
  <summary>Not Editing</summary>
  Simulator Screenshot - Non-editing view on iPhone 16 Pro, showing uneditable fields.
</details>

<details>
  <summary>Deleting</summary>
  Simulator Screenshot - Delete option on iPhone 16 Pro, showing swipe to delete action.
</details>

LePips:
I would actually like the `Metadata` view to be a modal and then the other edit views are pushed. That should probably help with the close button. I have other comments on the data input of some things, but I'll save those for a bit later.

JPKribs:
> I would actually like the `Metadata` view to be a modal and then the other edit views are pushed. That should probably help with the close button. I have other comments on the data input of some things, but I'll save those for a bit later.

I have a newer version of this I'm cleaning up. I was going to more the list items to another PR since this one is steadily getting out of hand. This version is a much most basic:

Pencil Icon
- Item Details

Item Details
- Refresh Metadata
- Delete Item
- Edit Metadata (But none of the list items)

Then a later PR I was going to add Studios, Cast, Genres, & Tags but 2 of those are just [String] and the other two are weird datatypes so I want to make sure I'm approaching a more custom solution there. I had something that was working but definitely NOT matching the current Swiftin style...

I think my views are good now but the navigation is giving me a hard time. Specifically, making the Items Details View to Modal loses the navigation title and the buttons on the page don't route either. I'm sure it's something stupid I'll figure out so for now just ignore that the main page is Push instead of Modal.

Simulator Screenshot - iPhone 16 Pro - 2024-11-05 at 15 39 23

JPKribs:
@LePips Let me know what you think of that Design! Frankly, this PR was way way harder than I thought it was going to be so there is some stuff I'm getting stuck on:

1. Making the ItemDetailsView modal is giving me some weird interactions
2. I'm trying to figure out how to track progress on a metadata refresh. I found [dateLastRefreshed](https://github.com/jellyfin/jellyfin-sdk-swift/blob/main/Sources/Entities/ItemFields.swift#L62) here but I'm also not sure where 'here' is. For now, I am just waiting 2 seconds which works for my system but slower networks this is probably not going to work.
3. My delete button looks jarring. I think I styled it wrong. 
4. This version of the PR still has some groundwork for doing the list metadata like people, tags, studios, etc. I can split that out if we want to make that it's own PR but that might take me some time haha.
5. I have a couple refresh types in there but let me know if we want less options.

Some stuff I think is going really well though:

1. I have notifications set up so metadata changes are flowing back appropriately!
2. The delete and refresh buttons work exactly how I would expect them to. These are the only buttons I see as valuable on tvOS.

Things I still want to do:

1. I want to make all the itemDetailsView to be a setting. Frankly, I don't want these settings available to me even as an admin on Swiftfin so I want the ability to turn this off this pencil as a setting in the settingsView.
2. Navigation cleanup (Modal + close EditMetadataView on save)

3. Studios add/remove (New View)
4. People add/remove (New View)
5. Genre add/remove (New View)
6. Tags add/remove (New View)
7. Update item images (Different PR Later)
8. Update subtitles (Different PR Later)

JPKribs:
Current state of this: This PR contains the ability to edit all item metadata or delete an item. All items are functional and work in editing Jellyfin items on the server. I've updated all my screenshots on the main post to reflect this state. However there are some issues I need to work out:

- [x] Metadata list should update with changes
- [x] Genre list should update with changes
- [x] Tags list should update with changes
- [ ] People list should update with changes
- [ ] Studio list should update with changes
- [ ] The People portraits are stretched in a weird way

This PR is frankly massive but there is some room to trim down:

- [ ] Genre & Tags are both [String] so I should have both re-use the same view instead of unique views
- [ ] The ViewModels are functional but messy
- [ ] There is some room to trim down what exists between views

Things that would be cool but I'm not sure what/how to do it:

- [ ] Make the Genre/People/Studio/Tags validate against existing items on the server. So, select an existing lookup item OR create a new item.
- [ ] On 'Refresh Metadata' check the progress and report that back.

LePips:
Per my comments in #1296 around the button styling, I would ideally want something similar for that row of buttons on iOS. The same issue with button crowding would still exist here, but luckily this and many other actions might be better on iOS in a `Menu` on the navigation bar

I've looked at adding more navigation bar items before but I can't seem to find the code for the specific style that I want these buttons to be and how they would change in style how the Apple TV app does it when you scroll... so I'll have to remake that to provide some guidance.

JPKribs:
I can add that to the final version of this! I think edit/delete is better there as well since it's more out of the way. I don't imagine these are functions that people will be using that often so toolbar makes a lot of sense. 

One question, this PR is very close to being ready. From there, I am going to break it up into more digestible parts for approval. This is more just a place for me to get my head around everything before I start splitting this up. That all being said, I wanted to check with you if this is the correct way I should be doing my ViewModels for this:

https://github.com/jellyfin/Swiftfin/pull/1305/commits/f83220325de4b50a9025e37ca945688d6f955c86

I tried to model this after ItemViewModel. I'd appreciate some feedback!

I have one outstanding item where I have a dismissCoordinator that is crashing Swiftfin when it's triggered but only for 2 out of the 6 views. Once I have that, I'll start splitting this out for Review!

Shared/ViewModels/ItemDetailsViewModel/ItemDetailsViewModel.swift
```diff
@@ -0,0 +1,185 @@
...
+    private func handleTaskError(_ error: Error) async {
+        guard !Task.isCancelled else { return }
+        let apiError = JellyfinAPIError(error.localizedDescription)
+        await setState(.error(apiError))
+        eventSubject.send(.error(apiError))
+    }
+
+    // MARK: - Update Item on Server
+
+    func updateItem(_ newItem: BaseItemDto, refresh: Bool = false) -> State {
```
LePips, line 129: What would be the reason we would want to refresh locally vs on the server? I see that we do so differently based on the genres, people, etc.

LePips (commented):
The `ItemDetailsViewModel` works looks good to me!

Shared/ViewModels/ItemDetailsViewModel/ItemDetailsViewModel.swift
JPKribs: Since the Genre & Tag are just [String] updating `item = newItem` just works. However, the person/studio is being input as a name. Jellyfin turns these into full BaseItemPerson/NameGuidPair. I've tried updating it as `item = newItem` but the list view isn't able to get the image for the person and the Studio doesn't actually link to a Studio object since it's just `guid=nil` and `name=name`. 

The only reason I'm not updating the Genre and Tag is it felt unnecessary. 

The other route for this is to make the addPeople/addStudio view look up the item instead of just inputting from a name. I'd be game to make those lookups on the server but I might need some guidance on what we'd what that view to look like.

JPKribs:
I spun off the first batch of this to https://github.com/jellyfin/Swiftfin/pull/1310.

All of the next steps are done on this PR here but there is still 1 bug that I need to work out so the impact views for that bug are going to be last:

1. EditMetadataView _(I might split the lock metadata options to their own view as well)_
2. EditGenreView
3. EditTagsView
4. EditPeopleView _[Impacted by routing issue]_
5. EditStudioView _[Impacted by routing issue]_

JPKribs:
After https://github.com/jellyfin/Swiftfin/pull/1336 this should all be done now.

PR #1310, opened 2024-11-10
[iOS] Media Item Menu | Refresh Metadata & Delete Item

JPKribs:
### Optional Item Editing / Deletion
Creates an optional setting in CustomizeSettings to allow editing items. When editing is enabled, this creates a setting that takes the user from the itemView to an itemEditorView. This view, will eventually allow metadata editing, deletion, image changes, etc. On this PR, this only allows for refreshing an item's metadata or deleting the item. To prevent issues, deletion is checked 2 ways. The first is if the item.canDelete is enabled from the server. The second is a CustomizeSetting for 'Allow media item deletion'. This is to help ensure that the user is aware this is enabled.

This is currently iOS only. The only two options that I would want on tvOS are the refresh and deletion settings that are created by this PR. They will likely be a menu option on tvOS with similar constraints. Since editing metadata won't be enabled on tvOS, there is no new for a itemEditorView, only the two button options. I will add these to tvOS once there is a plan for the button formatting: https://github.com/jellyfin/Swiftfin/pull/1296

### Screenshots
<details>
<summary>Editor</summary>
Editor
</details>

<details>
<summary>MenuLocation</summary>
MenuLocation
</details>

<details>
<summary>ItemViewMenu</summary>
ItemViewMenu
</details>

<details>
<summary>Settings</summary>
Simulator Screenshot - iPhone 16 Pro - 2024-11-09 at 18:54:08
Simulator Screenshot - iPhone 16 Pro - 2024-11-09 at 18:54:13
</details>


JPKribs:
Quick pivot on this. It turns out you can be given permissions to delete items from a library without being given permissions to edit metadata. I don't know what circumstance someone would want that interaction where you trust a user to delete but not edit but I want to make sure this PR can account for this:

Simulator Screenshot - iPhone 16 Pro - 2024-11-10 at 11 29 17

I've moved the delete option from inside the ItemEditorView and instead it's just triggerable from the ItemView. Additionally, the setting for allow media item deletion no longer relies on the allow media item editing setting. This way we can lock the edit page to Tested and working. I just need to remove the item from the PagingLibraryViewModel / ItemViewModels on deletion. I have the notification set up but I've been having some trouble figuring out the removal from the set?

Swiftfin/Views/ItemEditorView/Components/RefreshMetadataButton.swift
```diff
@@ -0,0 +1,175 @@
...
+
+        @StateObject
+        private var viewModel: RefreshMetadataViewModel
+
+        @State
+        private var isPresentingEventAlert = false
+        @State
+        private var error: JellyfinAPIError?
+        @State
+        private var isFinishing = false
```
LePips, line 24: There isn't a way to query the task that's spun up for refreshing the metadata. It's good to see some attempt at UX for it but I don't think we could provide a good enough abstraction for refreshing the metadata. I would consider this an uncommon enough task and that admins are somewhat understanding that the item must be popped and reselected in order to see the new metadata.

Swiftfin/Views/ItemEditorView/Components/RefreshMetadataButton.swift
```diff
@@ -0,0 +1,175 @@
...
+        @State
+        private var isLoading = false
```
LePips, line 26: We shouldn't have internal states that are different from the view model.

LePips (changes requested):

LePips:
> I have the notification set up but I've been having some trouble figuring out the removal from the set?

In the `PagingLibaryViewModel.init` at least you should have something like the following.

```swift
Notifications[.didDeleteItem].publisher
    .sink(receiveCompletion: { _ in }) { [weak self] notification in
        guard let item = notification.object as? Element else { return }
        self?.elements.remove(item)
    }
    .store(in: &cancellables)
```

Swiftfin/Views/ItemEditorView/Components/RefreshMetadataButton.swift
JPKribs: Sounds good. Are we fine with this just disabling the button for 2 seconds and showing the checkmark? Just some sort of feedback to prevent this being spammed on accident thinking nothing happened.

Swiftfin/Views/ItemEditorView/Components/RefreshMetadataButton.swift
```diff
@@ -0,0 +1,175 @@
...
+
+        init(item: BaseItemDto) {
+            _viewModel = StateObject(wrappedValue: RefreshMetadataViewModel(item: item))
+        }
+
+        // MARK: - Body
+
+        var body: some View {
+            Menu {
+                Button {
```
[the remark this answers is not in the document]
JPKribs: I've made this change, but I'm not sold on these icons. These were kind of my 'eh, it works' pick so let me know if you'd prefer something else!

Screenshot 2024-11-11 at 16 54 07

Swiftfin/Views/ItemEditorView/Components/RefreshMetadataButton.swift
JPKribs: Here's the version I have now that waits 5 seconds before pulling the new item:

https://github.com/user-attachments/assets/7e980ea3-7d84-4611-bf3d-9a65b8cba70b

JPKribs:
> > I have the notification set up but I've been having some trouble figuring out the removal from the set?
> 
> In the `PagingLibaryViewModel.init` at least you should have something like the following.
> 
> ```swift
> Notifications[.didDeleteItem].publisher
>     .sink(receiveCompletion: { _ in }) { [weak self] notification in
>         guard let item = notification.object as? Element else { return }
>         self?.elements.remove(item)
>     }
>     .store(in: &cancellables)
> ```

I had a version of this working but your's is a lot more concise so I'm using it instead of what I made lol. I think I have all of the changes in this next commit. I've tested and this works how I would expect it to. My only open items are UI related.

Just checking if those Icons look good to you:

https://github.com/jellyfin/Swiftfin/pull/1310#discussion_r1837281944

and whether we like this way to show refresh "progress" & if ticking 100 times over 5 seconds is excessive:

https://github.com/jellyfin/Swiftfin/pull/1310#discussion_r1837306362

Swiftfin/Views/ItemEditorView/Components/RefreshMetadataButton.swift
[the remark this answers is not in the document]
LePips: I don't really have any other icons in mind, so these seem okay.

However, it is easy to be confused per what each option does, especially between `Refresh` and `Replace All`. It would be handy if we had a small description in the footer of this button which then has a `Learn More...`, similar to many rows in iOS Settings. I would then want us to pop up a modal that describes each option. I had thought about that as well for the `Playback Quality/Device Profile/Compatibility` options.

Shared/ViewModels/ItemEditorViewModel/RefreshMetadataViewModel.swift
```diff
@@ -0,0 +1,171 @@
...
+import Foundation
+import JellyfinAPI
+
+class RefreshMetadataViewModel: ViewModel, Stateful, Eventful {
+
+    // MARK: Events
+
+    enum Event: Equatable {
+        case error(JellyfinAPIError)
+        case refreshTriggered(Date)
```
LePips, line 19: Don't need the date if we aren't doing anything with it.

Shared/Coordinators/ItemEditorCoordinator.swift
```diff
@@ -0,0 +1,30 @@
...
+import SwiftUI
+
+final class ItemEditorCoordinator: ObservableObject, NavigationCoordinatable {
+
+    let stack = NavigationStack(initial: \ItemEditorCoordinator.start)
+
+    @Root
+    var start = makeStart
+
+    private var item: BaseItemDto
```
LePips, line 20:
```suggestion
    private let item: BaseItemDto
```

Shared/ViewModels/ItemEditorViewModel/RefreshMetadataViewModel.swift
```diff
@@ -0,0 +1,171 @@
...
+
+            let currentProgress = Double(i) / Double(steps)
+            await MainActor.run {
+                self.progress = currentProgress
+            }
+        }
+
+        // After waiting for 5 seconds, fetch the updated item
+        let request = Paths.getItem(userID: userSession.user.id, itemID: itemId)
+        let response = try await userSession.client.send(request)
```
LePips, line 164: We should send a `itemMetadataDidChange` notification which the item views listen for and redraw their views by manipulating an `.id(...)`.

We would also have to replace the item in libraries. We will also have to test when you have a series item open in one tab, and update an episode in another tab, and how that updates correctly.

Shared/ViewModels/ItemEditorViewModel/RefreshMetadataViewModel.swift
```diff
@@ -0,0 +1,171 @@
...
+
+    enum State: Hashable {
+        case content
+        case error(JellyfinAPIError)
+        case initial
+        case refreshing
+    }
+
+    @Published
+    var progress: Double = 0.0
```
LePips, line 44:
```suggestion
    private(set) var progress: Double = 0.0
```

Swiftfin/Views/ItemView/ItemView.swift
```diff
@@ -15,8 +15,34 @@ import SwiftUI
...
+    @State
+    private var showConfirmationDialog = false
+    @State
+    private var isPresentingEventAlert = false
+    @State
+    private var isPresentingFailedAlert = false
+    @State
+    private var error: JellyfinAPIError?
+
+    private let permDelete = StoredValues[.User.enableItemDeletion]
```
LePips, line 35: Used the `StoredValue` wrappers instead:

```swift
@StoredValue(.User.enableItemDeletion)
private var enableItemDeletion: Bool
```

Swiftfin/Views/ItemView/ItemView.swift
```diff
@@ -15,8 +15,34 @@ import SwiftUI
...
+    private let permEdit = StoredValues[.User.enableItemEditor]
```
LePips, line 36: Same here.

Swiftfin/Views/ItemView/ItemView.swift
```diff
@@ -100,6 +127,68 @@ struct ItemView: View {
...
+            }
+            Button(L10n.cancel, role: .cancel) {}
+        }
+        .onReceive(deleteViewModel.events) { event in
+            switch event {
+            case let .error(eventError):
+                error = eventError
+                isPresentingEventAlert = true
+            case .deleted:
+                if deleteViewModel.item == nil {
```
LePips, line 150: There shouldn't be a situation where we succeed on deleting the item and it wasn't set to `nil`. If it fails, we should have gotten an `error` event. So, we don't need this kind of check - if we get the `deleted` event then it succeeded.

LePips (changes requested):

Swiftfin/Views/ItemView/ItemView.swift
JPKribs: Potentially dumb question: Why is this the format for `StoredValue` but not for `Defaults' when the value is being used as a constant? Should `StoredValue` just always be used with the wrapper?

LePips: Using them as constants isn't necessarily incorrect, however they just wouldn't respond to the updates if an item view was open in another tab and the setting was changed. It is also just consistent in their usage and not tied to the lifetime of the view.

In the places that we use the `Defaults[...]/StoredValues[...]` subscripts, those are in functions that will lookup the value every time they are run and will effectively respond to new changes.

Swiftfin/Views/ItemEditorView/Components/RefreshMetadataButton.swift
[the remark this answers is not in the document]
JPKribs: Do you have a design in mind for that? That feels like something we should create a learnMoreView to keep it consistent across uses. 

I can do that either on this PR or I can add this separately to keep the learnMoreView _(Or a better named view)_ separate and apply it to these two spots (Refresh metadata / customDeviceProfile types)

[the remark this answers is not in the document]
JPKribs: Very rough but this commit in on this PR now if you want to see what I have as an early version. Lemme know if you want me to move this to another PR instead!

Simulator Screenshot - iPhone 16 Pro - 2024-11-12 at 16 35 08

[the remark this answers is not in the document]
JPKribs: Sorry, and here is what the footer looks like:

Screenshot 2024-11-12 at 16 36 39

Right now this is basically just a button. I'm sure there's some kind of view extension / section extension that might be better for future usage?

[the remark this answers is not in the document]
JPKribs: I'm happy with the text and the modal but I don't have a strong idea of what the Item | Description should look like. This is my working version but 🤷‍♂️

Simulator Screenshot - iPhone 16 Pro - 2024-11-13 at 12 12 01

JPKribs:
Hm... same issue here as I'm having [here](https://github.com/jellyfin/Swiftfin/pull/1312#issuecomment-2476969257)

Swiftfin/Components/LearnMore/Components/ItemDescription.swift
```diff
@@ -0,0 +1,22 @@
...
+// Swiftfin is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2024 Jellyfin & Jellyfin Contributors
+//
+
+import Foundation
+
+struct ItemDescription: Identifiable {
```
LePips, line 11: Instead of making another type, just use a `TextPair`. In the future, we may remove this and instead move to `LabeledContent` from SwiftUI.

Swiftfin/Components/LearnMore/LearnMore.swift
```diff
@@ -0,0 +1,65 @@
...
+struct LearnMore: View {
+    @State
+    private var isPresented: Bool = false
+
+    let title: String
+    let itemDescriptions: [ItemDescription]
+
+    // MARK: - Initializer
+
+    init(_ title: String, items: [ItemDescription]) {
```
LePips, line 20: Instead of passing an array, we can use a `ArrayBuilder<TextPair>` instead.

LePips (changes requested):

Swiftfin/Components/LearnMore/Components/ItemDescription.swift
JPKribs: Sounds good. I've make this change but I just wanted to confirm UI wise we liked how this looked:

Simulator Screenshot - iPhone 16 Pro - 2024-11-14 at 21 41 50

LePips: I don't have a different design in mind so that's good 👍

Swiftfin/Views/ItemEditorView/ItemEditorView.swift
```diff
@@ -0,0 +1,77 @@
...
+            )
+
+            Section {
+                RefreshMetadataButton(item: item)
+                    .environment(\.isEnabled, userSession?.user.isAdministrator ?? false)
+            } footer: {
+                LearnMoreButton(L10n.metadata) {
+                    TextPair(
+                        title: L10n.refresh,
+                        subtitle: "Default refresh with the ability to override metadata without affecting existing images."
```
LePips, line 55: I changed the strings a little bit and I was a bit confused for what this option actually does. So, I took a look at the flags being sent to the server and tried to track down what each flag does, mainly `.default` but was unable to really understand what that flag specifically does. `fullRefresh` and `none` are more straightforward.

On the server, this will eventually be called to refresh the metadata for an item:

- https://github.com/jellyfin/jellyfin/blob/5e8c0fe40c0eb4c29b0f11cdf397c9880708e74b/MediaBrowser.Providers/Manager/MetadataService.cs#L74

Do you know what `.default` actually does?

LePips (changes requested):

Swiftfin/Views/ItemEditorView/ItemEditorView.swift
JPKribs: Here is where this is used on Web: https://github.com/jellyfin/jellyfin-web/blob/2b86995fc77d52b8682ed87fb0cf3b0f88d15553/src/components/refreshdialog/refreshdialog.js#L81

My understanding is that:

*Full:* Replace Metadata
*Default:* Get Missing / Update new or updated 
*None:* Do not change Metadata

Web has a checkmark for "Replace Images" but I went the route of having:

Refresh: Default, just get metadate for new/update files
Refresh Metadata: Replace Metadata but not images
Replace Images: Replace Images but not Metadata
Replace All: Get new items for all sources

From the web link I sent, it looks like 10.10 has some additional items we may want to account for like TrickPlay. Finally, when I first set this up I did recursive unset which should default to false. Do we want to enable that? Or make that a toggle?

This is making me wonder if I should have broken this out into a refreshMetadataView to accommodate checkmarks or other variations?

JPKribs: I've removed "Replace" since I wasn't finding the difference I was expecting. Default seems to have the same impact as Full when I tested it one by one. I also localized the surviving strings.

LePips (approved):
> From the web link I sent, it looks like 10.10 has some additional items we may want to account for like TrickPlay.

You are correct and we should think about how to account for these additional settings when we update the API.

PR #1312, opened 2024-11-12
[iOS] Admin Dashboard - User Passwords

JPKribs:
### Summary
Instead of creating a new view, I added an optional init to the resetUserPasswordViewModel and resetUserPasswordView to allow it to be init using a user that was NOT the current device user. Then, I moved both of these to be modal since they're password reset forms (forms being modal?). Not that I think that's a requirement but more in line with other edit screens.

I've tested that this works for both the current user and another server user. The benefit to this approach is a single ViewModel and also allowing changes to this view in a single location for a more cohesive feel.

The only liberty I took with this (besides changing to this to modal) is I hide the current password field for Server Users since it's likely a password reset on their behalf. That is, you're approaching as an admin so you don't have their current password to reset so there is no reason for the field IMO.

### Screenshot
<details>
<summary>UserView</summary>
UserView Screenshot
</details>

<details>
<summary>Server User Password</summary>
Server User Password Screenshot
</details>

<details>
<summary>Device User Password</summary>
Device User Password Screenshot
</details>


JPKribs:
Any idea what I am doing wrong? I haven't changed anything on my end, it builds locally, I'm on xcode 16.1 but I've been getting this issue about 70% of the time:

Screenshot 2024-11-14 at 09 48 58

If I select 'Retry' there is a 50/50 chance it fixes itself. I don't know if this is a github issue or a me issue.

LePips:
Oh, that's due to Xcode 16 being removed from the macos-14 image.

https://github.com/actions/runner-images/issues/10703

JPKribs:
That looks like it did it! Thank you @LePips!

LePips (approved):
I just cleaned things up to be more declarative in when the current password is required and how the `ResetUserPasswordViewModel` gets supplied the `userID`, since it's now a bit more generic.

PR #1313, opened 2024-11-12
[iOS] Admin Dashboard - User Permissions

JPKribs:
### Summary
Creates a Permissions view in the userDetailsView to allow for permission editing.

At this time, I am still working out the organization for the buttons I have. The working views are:

**- Permissions**
 > - Disabled Toggle
 > - Administrator Toggle
 > - External Access / Bitrate
 > - SyncPlay Access
 > - Remote Control of Other Devices
 > - Max Sessions / Failed Logins
 > - Other Permissions

**- Profile**
 > - Username
 > - Profile Picture Editing

**- Parental Controls**
 > - Max Ratings
 > - Access Schedule
 > - Blocked Items
 
 **- Media Access**
 > - Accessible Folders
 
 **- Media Deletion**
> - Deletable Folders

 **- Live TV**
> - Accessible Channels
> - Live TV Recording/Permissions

This PR will only cover the items that end up in the more generic Permissions view as a catch all for items that don't have a better home.

### Screenshots

Screenshot 2024-11-13 at 16 18 01


LePips (changes requested):
Having just done a quick skim, let's remove the `Allow` prefix from each string, as the toggle should be enough to understand the permission setting.

JPKribs:
> Having just done a quick skim, let's remove the `Allow` prefix from each string, as the toggle should be enough to understand the permission setting.

Sounds good! These have been updated.

Shared/ViewModels/AdminDashboard/ServerUserAdminViewModel.swift
```diff
@@ -72,137 +66,67 @@ final class ServerUserAdminViewModel: ViewModel, Eventful, Stateful, Identifiabl
...
+    // MARK: - Perform Action
 
-    private func resetPassword() async throws {
-        guard let userId = user.id else { return }
-        let parameters = UpdateUserPassword(isResetPassword: true)
-        let request = Paths.updateUserPassword(userID: userId, parameters)
-        try await userSession.client.send(request)
+    private func performAction(action: @escaping () async throws -> Void) -> State {
+        userTask?.cancel()
+        state = .updating
```
LePips, line 98: You don't need to set the state, this flow will set the state that is returned.

Shared/ViewModels/AdminDashboard/ServerUserAdminViewModel.swift
```diff
@@ -16,49 +16,43 @@ final class ServerUserAdminViewModel: ViewModel, Eventful, Stateful, Identifiabl
     // MARK: Event
 
     enum Event {
-        case success
-    }
-
-    // MARK: BackgroundState
-
-    enum BackgroundState {
-        case updating
```
LePips, line 25: We should still use `BackgroundState.updating`, as this indicates that the policy update is being made. It shouldn't be a primary `State` as we don't have any larger changes in design during that state and shouldn't be mixed up with `State.content`.

LePips (changes requested):

JPKribs:
I had 2 changes.

1. Using the .custom.rawValue to show the button meant the button would hide as soon as you updated the button to anything besides the custom default value. I changed the logic to be 'custom' by process of elimination. I don't know if swift has an easier way to do that but I'm just saying 'show the button when it's not unlimited' etc.

2. Similar issue to 1 where if I entered a value that was another enum it would immediately hide the input and change the value to that input. So I updated it to have a save/cancel where it still makes -1 unlimited (like it should) but the user needs to save first.

Original for 2:
https://github.com/user-attachments/assets/28f4bc69-e008-4fb1-b666-279d54f66b99

New for 2:
https://github.com/user-attachments/assets/ecefb3c1-3979-447a-979a-13e70ca470a5

LePips:
Ah, I see the issue. I tested with your changes but was still able to get into a weird state:

1 - Set `Maximum failed login policy` to `Custom`
2 - Set value to -1
3 - The value is set to `Unlimited` like you said
4 - Set it back to `Custom`
5 - Value is still -1
6 - Edit value, just hit `Cancel`
7 - Value is set to `0`

Instead of having custom values default to something else, let's just have it so it will perform validation on the input value. Let's just have custom values have a minimum of 1.

JPKribs:
I've added the learnMore side of this:

Screenshot 2024-11-26 at 15 09 09

I'll get the validation here in the next commit. Let me know if any of those strings are weird.

JPKribs:
I've updated this. No value should be able to be 0 now which fixes the unlimited issues. I did the same for the bitrate as well as the sessions details. 

The only remaining interaction that's a little jarring is the 3 / 5 changing the enum to Default based on User / Admin. I'm a bit confused on what this validation would look like? I think I might be a bit off on this implementation.

Swiftfin/Views/AdminDashboardView/ServerUserPermissionsView/Components/Sections/SessionsSection.swift
```diff
@@ -0,0 +1,193 @@
...
+
+                if policy.maxActiveSessions != ActiveSessionsPolicy.unlimited.rawValue {
+                    MaxSessionsButton()
+                }
+
+            } footer: {
+                VStack(alignment: .leading) {
+                    Text(L10n.maximumConnectionsDescription)
+                    LearnMoreButton(L10n.maximumSessionsPolicy) {
+                        TextPair(
```
LePips, line 155: I believe this isn't relevant to this section?

LePips (changes requested):
I just implemented more `Binding` extensions which is perfect for these kinds of settings.

Swiftfin/Views/AdminDashboardView/ServerUserPermissionsView/Components/Sections/SessionsSection.swift
JPKribs: You are absolutely right. I mixed that up. I moved that to the FailedLoginsView where I meant to put that.

JPKribs:
I've made these changes (see latest commit). I think the only outstanding issue is this:

https://github.com/user-attachments/assets/78d4378a-a3d7-4679-9b5c-18d482e07fc6

I know you said change it from using the enum rawValue to instead use it as validation but I wasn't quite sure what you meant. Could you give me an example of what that would look like?

LePips:
I took a look at web and it says that `A value of zero means inheriting the default ...`, so I think that we're incorrectly having `userDefault` and `adminDefault` that have separate values since the default seems to be set by the server. Instead, we should have a `default` value with 0 and the custom will have a minimum of 1, so we will no longer run into the issue where a custom value will cause those changes.

LePips:
Oh, I see the issue again. Apologies for not testing anything before talking haha. I also thought about just dropping `default`, so let's do that. Since we are providing a different UX for these settings, it's okay that there's that difference.

LePips (approved):
The behavior sounds good to me. Thanks for working on this!

LePips (approved):

PR #1316, opened 2024-11-17
[iOS] Playback Quality - Learn More

JPKribs:
### Summary
This is a pretty small change. It's just adding Learn More descriptions to Playback quality since we had a TODO for it and it's one of our more confusing sections.

I created a styleFormat for bitrate and I am passing in the Maximum bitrate into the description for it so we don't have to worry about updating that later if we ever increase above 360mbps.

I also updated the LearnMoreButton to have an optional footer input. This way, any Footer+LearnMore inputs are formatted the same instead of recreating the spacing each time.

<details>
<summary>Main</summary>
Main Screenshot
</details>

<details>
<summary>Bitrate</summary>
Bitrate Screenshot
</details>

<details>
<summary>Test</summary>
Test Screenshot
</details>

<details>
<summary>Profiles</summary>
Screenshot 4
</details>


JPKribs:
I think this is almost ready. My only issue is the alignment for the text on row 2 or greater. It wants to center it and I can't seem to figure out how to get it to both have AccentColor for LearnMore AND be Leading Alignment.

JPKribs:
Here is what tvOS looks like. I'm not sure if there is a better way to handle focus but I think this *looks* good. Focus, I'm open to feedback for how this should work:

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-11-17 at 18 13 02
Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-11-17 at 18 13 06
Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-11-17 at 18 13 59

LePips (changes requested):
Don't make any changes to `LearnMoreButton` since that should just hold a button. While in iOS Settings it sometimes does have the `Learn More...` as a suffix to the description, I would want it to just always be leading aligned below the original description.

So the footers would look something like the following to get the desired layout:
```swift
VStack(alignment: .leading) {
    L10n.bitrateTestDisclaimer.text

    LearnMoreButton(
        L10n.bitrateTest
    ) {
        TextPair(
            title: L10n.testSize,
            subtitle: L10n.bitrateTestDescription
        )
    }
}
```

In order to get the desired affect you want you would have to make something like an `AttributedString/NSAttributedString` but that is a bit more work.

Edit: I started writing this before you fixed it with `buttonStyle(.plain)`, but I would still personally want the layout I describe.

LePips:
I am unsure of what to do on tvOS, as I don't know how other apps have a `Learn More` UX. We shouldn't necessarily just copy from iOS.

JPKribs:
In an earlier version I had of this, I had the regular footer and the button in a Vstack. I don't think it was bad so if we like that layout I can move back to that. My primary reason for moving away from it was that is wasn't the same as the iOS settings.

For tvOS, maybe the solution is a single button at the end of the list for 'details' or something like that with all of the buttons that need context in a single view. Let me roll this back to an earlier version so we can work from a closer-to-main place!

JPKribs:
> I am unsure of what to do on tvOS, as I don't know how other apps have a `Learn More` UX. We shouldn't necessarily just copy from iOS.

I think "Learn More..." isn't the best text to go there but we're really only going to need additional information in settings for tvOS. In this case, we could have it popup in a sheet with sections for the items that might need some additional detail?

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-11-17 at 20 13 47

Swiftfin/Views/SettingsView/PlaybackQualitySettingsView.swift
```diff
@@ -44,18 +57,19 @@ struct PlaybackQualitySettingsView: View {
                 } header: {
                     L10n.bitrateTest.text
                 } footer: {
-                    VStack(alignment: .leading, spacing: 8) {
-                        L10n.bitrateTestDescription.text
-                        L10n.bitrateTestDisclaimer.text
+                    VStack(alignment: .leading) {
+                        Text(L10n.bitrateTestDisclaimer)
+                        LearnMoreButton(L10n.bitrateTest) {
```
LePips, line 62: `Learn More` should be for more detail on multiple items or for a single section that requires a lot more description. This specific section can be described concisely, so we don't need a `Learn More` for it.

LePips (changes requested):
I think the idea for tvOS works now, but I can see issues if it is too large and requires scrolling. We can see if other solutions work better in the future.

JPKribs:
> I think the idea for tvOS works now, but I can see issues if it is too large and requires scrolling. We can see if other solutions work better in the future.

I have your changes in. One idea I was thinking of was expanding out ScrollIfLargerThanContainerModifier to have a toggle for 'autoScroll' then have it just scroll through a view with this enabled. This way, if an item on the left was too large, this would treat it as a scrollview and automatically scroll through the content. This is my working version which 'works' but also scrolls way to quickly then never resets back to the top:

```
struct ScrollIfLargerThanContainerModifier: ViewModifier {

    @State
    private var contentSize: CGSize = .zero
    @State
    private var layoutSize: CGSize = .zero

    let padding: CGFloat
    let autoScroll: Bool

        AlternateLayoutView {
            Color.clear
                .trackingSize($layoutSize)
        } content: {
            ScrollViewReader { proxy in
                ScrollView {
                    VStack {
                        content
                            .trackingSize($contentSize)
                        Color.clear
                            .frame(height: 0)
                            .id("BottomAnchor")
                    }
                }
                .frame(maxHeight: contentSize.height >= layoutSize.height ? .infinity : contentSize.height)
                .scrollDisabled(contentSize.height < layoutSize.height)
                .scrollIndicators(.never)
                .onChange(of: contentSize) { _, newSize in
                    if autoScroll && newSize.height >= layoutSize.height {
                        withAnimation {
                            proxy.scrollTo("BottomAnchor", anchor: .bottom)
                        }
                    }
                }
            }
        }
    }
}
```

If you like this idea, I can work on it more or make it it's own ViewModifier since the overlap might not make sense to include on this existing ViewModifier.

JPKribs:
If we aren't worried about the tvOS size quirks, I think this should be ready to go from my end!

LePips (approved):
Looks great!

PR #1326, opened 2024-11-24
[iOS] LetterPicker - Dynamic Scaling

JPKribs:
### Summary
I noticed some LetterPickers looked wrong. I realized it was because he was using the Accessibility > Font Size to increase his font to the maximum allowed font. Since the LetterPicker was build using static padding and frame numbers, this resulted in a weird look:

Screenshot 2024-11-23 at 21 45 27

Coming back at this with some additional experience, I just changed each letter to accept it's size as an input. Then, from the LetterPickerBar, I use the `String().height()` to create a static `size` for the Headline font that we are using for the LetterPicker. So, all letters get that size passed in for it's from the LetterPickerBar and the LetterPickerBar frame is always appropriately sized to ensure no letters get cut off.

Finally, I am calling `LetterPickerBar.size` on the `PagingLibraryView` and using that to get the padding when the LetterPickerBar is present. I am adding 10 to this value since the LetterPickerBar is padded 10 away from the edge, centering it between the edge of the screen and the content.

I've tested this from iOS and iPad from the smallest to the largest letters and this looks to be working a lot better. An unintended side effect is this is turning into a scroll more often but that seems ideal over cramming everything in.

Screenshot 2024-11-23 at 21 57 56

### DIFF
--- a/Swiftfin/Components/LetterPickerBar/Components/LetterPickerButton.swift
+++ b/Swiftfin/Components/LetterPickerBar/Components/LetterPickerButton.swift
@@ -20,10 +20,12 @@ extension LetterPickerBar {
         private var isSelected
 
         private let letter: ItemLetter
+        private let size: CGFloat
         private let viewModel: FilterViewModel
 
-        init(letter: ItemLetter, viewModel: FilterViewModel) {
+        init(letter: ItemLetter, size: CGFloat, viewModel: FilterViewModel) {
             self.letter = letter
+            self.size = size
             self.viewModel = viewModel
         }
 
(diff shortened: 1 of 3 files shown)


JPKribs:
Same comparison but on iPad:
Screenshot 2024-11-23 at 22 08 52

LePips (approved):
Good idea considering accessibility. In the future, it would be better to somehow let the letter bar lay itself out and get the width that way, without having to recalculate `size` every time it's called. However, that would require some refactoring of `.scrollIfLargerThanContainer()`.

PR #1328, opened 2024-11-26
[iOS] Admin Dashboard - User Profiles

JPKribs:
### Summary
Allows for editing/deleting a user image. This is just repurposing the existing UserProfileImagePicker to be callable from both current userSession or another user. Also allows for editing a user's account name. 

This edits the layout of the ServerUserDetailView to be more in line with the UserProfileView

### Screenshots
<details>
    <summary>Updated View</summary>
    Updated View Screenshot
</details>

<details>
    <summary>Update Username</summary>
    Updated View Screenshot
    
</details>

<details>
    <summary>Image Options</summary>
    Username Editor Screenshot
</details>


JPKribs:
This is currently 'ready' but the notification to update the image on update isn't firing 100% of the time. So sometimes, when I change a user image it just never changes even though it's been updated on the server. Happens about 50% of the time so I think I have an issue on my end to work out.

LePips:
I've played around with this and I don't think that the issue is due to the notification being sent but a caching issue. If I edit the profile image for a different user then nothing will be properly updated for that user, even on the user sign in view. If I exit the admin user and then sign into the user I edited, then all of the images will appear properly.

JPKribs:
Interesting. What is being done on the local user level to clear the cache on pfp update? I tried to mirror/reuse what I could but I could have missed something. Plus, I need to clean up this PR a bit it's a little hectic...

JPKribs:
Daw, I merged it wrong...

JPKribs:
This *should* be good but I'm back to having the same issue as before where I'm unable to get images to refresh appropriately throughout. Going from no image -> image works appropriately but image -> new image doesn't always refresh. Is that image caching or did I screw up the notifications?

There is also a weird one where if you change your username from `Server Users` -> `Server User Details` it doesn't update for the settings. The reason for this is I'm not sure how to update the userSession.user.name.

Last item, I am using the `UserID` as the payload for `didChangeUserProfile`. If I change that to `UserDto`, that would be a bigger payload but it would also mean I can refresh the user once at update then use that across multiple `ViewModels` because I think right now there is a chance that like 2 refreshes for the same data can occur?

Shared/ViewModels/AdminDashboard/ServerUsersViewModel.swift
```diff
@@ -144,6 +186,21 @@ final class ServerUsersViewModel: ViewModel, Eventful, Stateful, Identifiable {
...
+    // MARK: - Refresh User
+
+    private func refreshUser(_ userID: String) async throws {
+        let request = Paths.getUserByID(userID: userID)
+        let response = try await userSession.client.send(request)
+
+        let newUser = response.value
+
+        await MainActor.run {
+            if let index = self.users.firstIndex(where: { $0.id == userID }) {
```
LePips, line 198: Looks like we should migrate `users: [UserDto]` to `users: IdentifiedArrayOf<UserDto>`. I believe the impact should be minimal.

Swiftfin/Views/UserProfileImagePicker/UserProfileImage.swift
```diff
@@ -0,0 +1,102 @@
...
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2024 Jellyfin & Jellyfin Contributors
+//
+
+import Defaults
+import JellyfinAPI
+import SwiftUI
+
+struct UserProfileImage: View {
```
LePips, line 13: If we're going to make some kind of central view to handle showing a user profile image we should make it applicable to the following areas: `UserProfileRow`, `UserRow`, `PublicUserRow`. (Although I think some reflection on the names of those views is warranted).

All of these show a user profile image but some are currently incomplete due to design (like the background colors and placeholder/failure views) having been done after their implementations. We should probably make a component view for showing a user profile image in a circle. That would change _this_ specific view to something like `UserProfileHeroImage`, since this is a hero view of editing or presenting a user.

I think the caching issues come down to which pipeline is used. The `branding` pipeline is a separate "cache" used for images that should be "offline", like local user profile images and the server splash screen (at least by choice). Not all user images, especially when managing them as a server administrator and they haven't signed in locally, should be considered "something to be seen offline" and use the default pipeline, which is like a normal memory cache.

You can see the branding pipeline results in `Files/On My iPhone/Swiftfin iOS/Cache/org.jellyfin.swiftfin.branding`.

Swiftfin/Views/AdminDashboardView/ServerUserDetailsView/ServerUserDetailsView.swift
```diff
@@ -43,13 +71,26 @@ struct ServerUserDetailsView: View {
...
                         }
                 }
-            }
-
-            Section(L10n.advanced) {
                 ChevronButton(L10n.permissions)
                     .onSelect {
                         router.route(to: \.userPermissions, viewModel)
                     }
+                ChevronAlertButton(
```
LePips, line 78: Username should probably be at the top of this section.

LePips (changes requested):

Swiftfin/Views/AdminDashboardView/ServerUserDetailsView/ServerUserDetailsView.swift
JPKribs: Do we want Username as a button or should I add it to the `.confirmationDialog` when selecting the profile picture? I used to have it there but I removed it. I'm not sure the button right below the Picture & Username VStack is good. For whatever reason UX/UI is always a struggle for me haha.

Swiftfin/Views/UserProfileImagePicker/UserProfileImage.swift
JPKribs: I can work on this! 

On a similar note, is there a way to use: `RedrawOnNotificationView(.didChangeUserProfile)` where I can take the `.didChangeUserProfile` payload into account? I pass the `userID` as the payload and it would be nice to reserve redrawing for matching `userIDs`

LePips: I've added the capability for `RedrawOnNotificationView` to filter based on the payload if necessary.

Swiftfin/Views/AdminDashboardView/ServerUserDetailsView/ServerUserDetailsView.swift
LePips: The confirmation dialog is for whether to delete or set the image. Adding modification of the username there doesn't fit what it's being used for.

Swiftfin/Views/UserProfileImagePicker/UserProfileImage.swift
JPKribs: I centralized all user icons to `UserProfileHeroImage`. I made a `UserEditableHeroImage` for this `pencil` + `UserProfileHeroImage` where the `confirmationDialog` pops up to replace/delete:

Screenshot 2024-12-19 at 10 20 17

I updated the verbiage on this to better reflect caching:

simulator_screenshot_C9E6F7F8-F0FE-4B96-AACA-FED83995629C

I hope this was what you meant! Let me know if there are changes/adjustments that are needed for this. This is also used for tvOS so merging with some other branches might start getting a bit weird.

LePips (commented):
As some clarification, the current `UserProfileHeroImage` should be `UserProfileImage`. The `Hero` attribute should be used for the views that have the profile image as the large image, which then can use `UserProfileImage`.

- https://www.optimizely.com/optimization-glossary/hero-image/

JPKribs:
Ah! I probably should have googled that first haha... I've updated this to reflect what the names *should* be!

JPKribs:
I was just briefly looking at Nuke. When we update a user profile image, should we be doing something like:

```
        let pipeline = ImagePipeline.Swiftfin.default (or branding)
        let request = ImageRequest(url: imageURL)

        pipeline.cache[request] = nil
        pipeline.configuration.dataCache?.removeData(for: imageURL.absoluteString)
```

To remove the cached image? This and https://github.com/jellyfin/Swiftfin/pull/1345 are going well in terms of functionality but the image caching is making it look a like janky. Adding the info above *appears to work* in my testing but I don't know if that's a fluke. I also don't know if this is just the wrong way to do this.

I was looking at a Notification for .didImageGetReplaced that takes a URL payload and we can nuke it out of Nuke from there? Just wanted to double check before I started down this road!

LePips:
Debugging this a little bit further I see a larger issue: most images aren't cached at all in the expected directories. The only images that are cached as expected are the splash screens. For the rest of the images, the name generator that is used makes invalid filenames.

`maxWidth` is an issue since it would cause different keys for the same images just at different sizes. I'm pursuing a new caching system to help with that as well as hopefully some performance improvements, but I won't let that stop this.

JPKribs:
> Debugging this a little bit further I see a larger issue: most images aren't cached at all in the expected directories. The only images that are cached as expected are the splash screens. For the rest of the images, the name generator that is used makes invalid filenames.

Is that a Nuke issue or have I been using Nuke incorrectly? Is the .onNotification I did [here](https://github.com/jellyfin/Swiftfin/pull/1328#issuecomment-2558056739) the correct usage or will that need to change?

> `maxWidth` is an issue since it would cause different keys for the same images just at different sizes. I'm pursuing a new caching system to help with that as well as hopefully some performance improvements, but I won't let that stop this.

I didn't even think about that for the URL maxWidth. Is there anything on my end that I can help with? Sorry, I feel like I've been creating a lot of additional work for you so please let me know if there is anything I can do to help out!

LePips:
Nah, I've just been holding Nuke incompletely. The widths-causing-different keys has long been on my mind since I first implemented the caching but the lack-of-caching is new to me.

It's just that since now we need to manipulate the caches, the need for a better implementation is more pressing.

LePips:
I have largely refactored our caches to, well, work as expected. I've taken a lot of time to really understand our use case of images since on the server images are keyed by the item id _and_ a tag, which causes issues on our side for cache invalidation. The total implementation is not complete and I will be asking the creator of Nuke some clarifying questions to fit our use case a little bit more.

You should run through this really quick to check that the user images update as expected.

JPKribs:
> You should run through this really quick to check that the user images update as expected.

This works great! This is going to help a ton with https://github.com/jellyfin/Swiftfin/pull/1345 as well! I only found 2 things. One was a comment that said 500mb when it was 1000mb and one was updating tvOS to use the new cache.

Tested on both iOS and tvOS and everything is working as expected for the areas that we touched with this PR. The only thing that's a little weird (and this was weird before your changes) is that the caching behaves weirdly on the `UserSignInView`. If I switch to a new server, it uses the images from the previously selected server:

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-12-28 at 18 30 15
Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-12-28 at 18 30 25

This is behaving in the same way as before this PR so I don't think we need to fix it for this PR. Just something where, if you knew what was causing this on the caching side, might be nice to make a quick swap on. Otherwise, if we need to do. some digging, I can do that later on another PR!

LePips (approved):

PR #1333, opened 2024-11-27
[iOS] Admin Dashboard - Media Access / Deletion Settings

JPKribs:
### Summary
Creates a menu to enable/disable access to a media library. Also allows for enable/disable access to deleting from a media library.

Getting very close to the end!

### Screenshots

<details>
<summary>New Button</summary>
New Button Screenshot
</details>
    
<details>
<summary>View 1</summary>
View 1 Screenshot
</details>
    
<details>
<summary>View 2</summary>
View 2 Screenshot
</details>

### DIFF
--- a/Shared/Coordinators/AdminDashboardCoordinator.swift
+++ b/Shared/Coordinators/AdminDashboardCoordinator.swift
@@ -52,6 +52,8 @@ final class AdminDashboardCoordinator: NavigationCoordinatable {
     @Route(.push)
     var userDetails = makeUserDetails
     @Route(.modal)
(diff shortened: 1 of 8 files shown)


JPKribs:
I think this is good to go unless there is a good binding extension for this scenario. https://github.com/jellyfin/Swiftfin/pull/1323 has this scenario a few different places as well:

There is a list of potential options that should all be their own toggle that come from a datasource. When selected, they should be added to a [String] but when deselected, they should be removed from the [String].

LePips (approved):
I just made a new `Binding` function for a `contains`, which I've seen other people use too.

You're really burning through these!

PR #1338, opened 2024-12-05
Resolve Prolonged Type Checking

chickdan:
The use of keypaths in the where clauses were causing type check warnings due to being in excess of 200ms after the compiler flag I turned on. Prior to these changes the type checking took roughly 2000-3000ms but with these changes the warnings are resolved, meaning it now takes <200ms.


LePips (approved):
Thanks for doing this!

PR #1342, opened 2024-12-07
[iOS] Admin Dashboard - User Device & TV Access

JPKribs:
### Summary
Adds a new section to the ServerUserDetailsView to limit device access. On web, this is in the Access section. This should roughly mirror our media access section but I've added some additional device details like the latest user on that device, when was it last seen, and the device icon to spruce it up a little.

I also re-odered the ServerUserDetailsView so the Media Access and Device Access options are both just in their own 'Access' section. Permissions are now in the 'Advanced' section where the last two sections will be added as well. So this will be: Permissions, Parental controls, & Live TV.

I've added LiveTV as well. The reason I did this was because it's just 2 toggles. The enabledChannels is NOT LiveTV Channels. So, this is a lot smaller and an easy implementation. This can probably just get rolled up into the Permissions section but it also makes sense for it to be in its own Access section as well 🤷‍♂️

### Screenshots

<details>
  <summary>User Details</summary>
  
</details>

<details>
  <summary>Enable All Devices</summary>
  Enable All
</details>

<details>
  <summary>Enable Individual Devices</summary>
  Enable Devices
</details>

<details>
  <summary>Live TV Access</summary>
  Enable Devices
</details>


Translations/en.lproj/Localizable.strings
```diff
@@ -1933,9 +1933,13 @@
...
-// Media Access - Section Title
-// Section Title for Server User Media Access Editing
-"mediaAccess" = "Media Access";
+// Enable All Devices - Toggle
+// Toggle to enable a setting for all devices
+"enableAllDevices" = "Enable all devices";
+
+// Item Access - Section Title
+// Section Title for Server User Media / Device Access Editing
+"stringWithAccess" = "%@ Access";
```
LePips, line 1942: We shouldn't have compound string like this for localization, as inflection may be different depending on the parameter.

For example: in English to Spanish:

`User Access` -> `acceso de usuario`, while `Device Access` -> `Acceso al dispositivo`. The `User` and `Device` strings don't have the inflective `de` and `al`.

LePips (changes requested):

Translations/en.lproj/Localizable.strings
JPKribs: Didn't even think about that! I've changed this out.

JPKribs:
I set up Live TV on Jellyfin only to find that none of the 'Channel' settings are related. I'm not able to find a web equivalent so I think I'm safe to skip those. This is for items like:

enabledChannels
blockedChannels

So, I am treating this PR as wrapping up the User Details 'Access' section. I'm very close to the end for this. I left TODOs for my last items in the Admin Dashboard which are:

**PR 1**
- Username Editing
- Profile Pictures

_I just need to figure out the PFP caching/refreshing the cache on change_

**PR 2**
- Allow Tags (10.10+)
- Block Tags

_I am going to make 1 view to be shared between both. Will likely reuse the tag searching from https://github.com/jellyfin/Swiftfin/pull/1336 for the add screen._

**PR 3**
- Access Schedules

_I need the listView and an add screen. Should roughly mirror the EditServerTasksView & AddServerTaskView._

**PR 4**
- Parental Rating (maxParentalRating & blockUnratedItems)

_I already have this done but I don't like how it works. I don't like how it looks on web either but I'm not sure what is a good route to do this cleaner. See below._

<details>
<summary>Parental Rating</summary>
https://github.com/user-attachments/assets/4048ce64-65ce-4c76-affb-7a918792532e
</details>

Swiftfin/Views/AdminDashboardView/ServerUserDeviceAccessView/ServerUserDeviceAccessView.swift
```diff
@@ -0,0 +1,177 @@
...
+                Toggle(
+                    L10n.enableAllDevices,
+                    isOn: $tempPolicy.enableAllDevices.coalesce(false)
+                )
+            }
+
+            if tempPolicy.enableAllDevices == false {
+                Section {
+                    ForEach(devicesViewModel.devices, id: \.self) { device in
+                        Toggle(
```
LePips, line 109: For these rows, I would rather reuse the `DevicesView.DevicesRow` and refactor it a bit so that the `onDelete` action is optional and if not defined, then it isn't able to be deleted.

Swiftfin/Views/AdminDashboardView/ServerUserDeviceAccessView/ServerUserDeviceAccessView.swift
```diff
@@ -0,0 +1,177 @@
...
+    // MARK: - Environment
+
+    @EnvironmentObject
+    private var router: BasicNavigationViewCoordinator.Router
+
+    // MARK: - ViewModel
+
+    @ObservedObject
+    private var viewModel: ServerUserAdminViewModel
+    @ObservedObject
```
LePips, line 24:
```suggestion
    @StateObject
```

Swiftfin/Views/AdminDashboardView/ServerUserDetailsView/ServerUserDetailsView.swift
```diff
@@ -47,25 +46,55 @@ struct ServerUserDetailsView: View {
...
+                    }
 
-                // TODO: Parental: accessSchedules, maxParentalRating, blockUnratedItems, blockedTags, blockUnratedItems & blockedMediaFolders
+                ChevronButton(L10n.media)
+                    .onSelect {
+                        router.route(to: \.userMediaAccess, viewModel)
+                    }
 
-                // TODO: Live TV: enabledChannels & enableAllChannels
+                ChevronButton(L10n.liveTV)
```
LePips, line 67: Put this above media, so that it is in english alphabetical order.

LePips (changes requested):

JPKribs:
I've made the changes. The padding was getting weird when I tried using the DevicesView.DeviceRow and, in fixing that, I made the same changes to ServerUsersView.ServerUserRow since I made those at the same time and they mirrored eachother.

I had the RowInsets in the row instead of from the view. So, it looked good in a plain list and cramped in a non-plain list. I've moved to just using those from the Views above them so it would be reusable for both Device Access and Device List. Users mirrors that just for parity sake + if we ever need a user row somewhere in the future.

JPKribs:
For consistency's sake, EditItemElementRow needed the same thing... The downside of doing this all in such quick succession is, any weird decision I made for 1 listRow, I made for all of them. Just checked though all the ItemEditing & AdminDashboard and we should be consistent now!

LePips:
That sounds like a good addition.

LePips (changes requested):

Swiftfin/Views/AdminDashboardView/ServerUserDeviceAccessView/ServerUserDeviceAccessView.swift
```diff
@@ -0,0 +1,132 @@
...
+                Toggle(
+                    L10n.enableAllDevices,
+                    isOn: $tempPolicy.enableAllDevices.coalesce(false)
+                )
+            }
+
+            if tempPolicy.enableAllDevices == false {
+                Section {
+                    ForEach(devicesViewModel.devices, id: \.self) { device in
+                        Toggle(
```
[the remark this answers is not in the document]
JPKribs: Just to confirm, this is the Before and After of this change:

Screenshot 2024-12-09 at 12 42 39

Just want to make sure we're okay with the custom toggle underneath the default toggle.

Swiftfin/Views/SelectUserView/Components/UserRow.swift
```diff
@@ -117,25 +117,9 @@ extension SelectUserView {
...
-                    Image(systemName: "circle")
-                        .resizable()
-                        .backport
-                        .fontWeight(.bold)
-                        .aspectRatio(1, contentMode: .fit)
-                        .frame(width: 24, height: 24)
-                        .foregroundStyle(.secondary)
-                }
+                ListRowCheckbox()
+                    .environment(\.isEditing, isEditing)
```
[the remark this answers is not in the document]
JPKribs: Whoops yes, consider it gone!

LePips:
I made the toggle mimic the inset grouped list style like we do for the header, which allows the list to be plain and give more room for the device rows like we have elsewhere. How do you think that looks?

JPKribs:
> I made the toggle mimic the inset grouped list style like we do for the header, which allows the list to be plain and give more room for the device rows like we have elsewhere. How do you think that looks?

Oh that looks way better! I was hesitant to try and figure out mixing standard & plain lists but I'm very happy with how that looks! Only change I made was resetting the development team back to None & org.jellyfin.swiftfin

LePips (approved):

PR #1345, opened 2024-12-08
[iOS] Media Item Menu - Edit Item Images

JPKribs:
### Summary

Adds an image management view for editing Movies, Shows, Episodes, Season*, or People*.

* *Seasons & People don't have an ItemView so there isn't a place where you can select them to edit but they should work if there is every a good view to edit them from.*

### Features
•	View all images associated with a media item.
•	Add, delete, and edit images for media items.
•	Delete existing images.
•	Upload a remote image by searching and selecting from Jellyfin Server's configured sources.
•	Upload new images from device files.
•	Upload and crop new images from device photos.

### Demo

https://github.com/user-attachments/assets/dcdb8a16-0d51-4f17-9acd-e28107f628a2


JPKribs:
@LePips Absolutely no rush on this at all. I'm still a ways away from being ready but let me know if you have any feedback on the UI! Not in love with anything I've done so far but I don't have a great sense of what this *should* look like. The edit button on the poster is because you can have more than one banner, etc. so I make it so you edited at that level opposed to from the section as a whole.

The remote image section, with the option from remote sources from the server, is paginated so it only loads in 50 at a time. Web is only 25 so I can limit that further if we want. 

Functionally, I think my only issue I'm not sure of is I am using `item.imageSource(imageType)` to get the image based on the type but that only every gives me index 0 so items like banners only let you edit/delete/upload to the first item which, if possible, I'd like to make this usable for any index.

LePips:
For each grid, we can make assumptions on what the shape of the images are going to be and have the layouts match accordingly. ie: primary is portrait, backdrop is landscape, etc. 

For the image contents, we can instead show nothing. When an image is selected, lets push to a confirmation view that shows the larger image with the size and metadata source details.

JPKribs:
This is will In-Progress but this is my current working version:

https://github.com/user-attachments/assets/a0d38adb-051e-40f0-bef5-dde7727a765a

I still need figure out the following:

1. When deleting an Image, if there are multiple Images for that `ImageType`, the order of the items shuffles around in a jarring way.
2. The `UploadImage` + icon doesn't actually work. It always returns a 500 error even though I mirrored `UserProfileImagePicker` so I need to understand that all better:

Screenshot 2024-12-22 at 18 41 36

3. At the end of the `AddItemImageView` scroll, the last image always explodes like this:

Simulator Screenshot - iPhone 16 Pro - 2024-12-22 at 18 46 54

Other than that, I think this is almost ready!

Shared/ViewModels/ItemAdministration/RemoteImageInfoViewModel.swift
```diff
@@ -0,0 +1,253 @@
...
+import Get
+import IdentifiedCollections
+import JellyfinAPI
+import OrderedCollections
+import UIKit
+import URLQueryEncoder
+
+private let DefaultPageSize = 50
+
+class RemoteImageInfoViewModel: ViewModel, Stateful {
```
LePips, line 20: Why not conform to `PagingViewModel`?

Shared/Extensions/JellyfinAPI/RemoteImageInfo.swift
```diff
@@ -0,0 +1,18 @@
...
+//
+
+import Foundation
+import JellyfinAPI
+import SwiftUI
+
+extension RemoteImageInfo: @retroactive Identifiable {
+
+    public var id: String {
+        UUID().uuidString
```
LePips, line 16: I see this being done often but is an error, since a new `UUID` will be created every time this property is accessed on the same object. Using `hashValue` here is fine.

LePips (changes requested):

Swiftfin/Views/ItemEditorView/ItemImages/EditItemImagesView/EditItemImagesView.swift
```diff
@@ -0,0 +1,207 @@
...
+        contentView
+            .navigationBarTitle(L10n.images)
+            .navigationBarTitleDisplayMode(.inline)
+            .onFirstAppear {
+                viewModel.send(.refresh)
+            }
+            .sheet(item: $selectedImage) { imageInfo in
+                deletionSheet(imageInfo)
+            }
+            .fileImporter(
```
LePips, line 65: Hm, this makes me think that we should want to allow both image and/or file imports. This does differ from the user profile image selection which seems okay since this is more administrative. It would be cool in the future to allow the option of selection between files and photos.

Swiftfin/Views/ItemEditorView/ItemImages/EditItemImagesView/EditItemImagesView.swift
```diff
@@ -0,0 +1,207 @@
...
+            .frame(maxHeight: 150)
+            .shadow(radius: 4)
+            .padding(16)
+        }
+    }
+
+    // MARK: - Delete Image Confirmation Sheet
+
+    @ViewBuilder
+    private func deletionSheet(_ imageInfo: ImageInfo) -> some View {
```
LePips, line 178: Instead of this being tailored for deletion, we can just have this as a more general `ImageInfoView`. I would like to to mirror the changes I made to the `RemoteSearchResultsView`, where the details are instead rows in a `List` vs in a `VStack`. The image can be in a larger section as the first item in the list.

But with all that, the `Delete` button can still be present. I would probably want it in the bottom toolbar to be a bit more consistent, but we can play with it.

Swiftfin/Views/ItemEditorView/ItemImages/EditItemImagesView/EditItemImagesView.swift
```diff
@@ -0,0 +1,207 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    // MARK: - Section Header
+
+    @ViewBuilder
+    private func sectionHeader(for imageType: ImageType) -> some View {
```
LePips, line 135: Have a single add button that is a `Menu`, where the buttons for `Search` and `Upload file` are menu items.

LePips (changes requested):

LePips:
Per the exploding images, images that don't have size data (height, width) will break because the images use `fill` for aspect mode. We can try to debug that for a bit and I do have somewhat of a solution, but I'd almost rather filter out images that don't have that information.

Swiftfin/Views/ItemEditorView/ItemImages/EditItemImagesView/EditItemImagesView.swift
JPKribs: I originally went for File since, in my own usage, I'm just downloading a file from a theposterdb or some other source. I do want to include the Photo Picker as well but I wanted to start with .fileImporter. I'm pretty stuck on this upload API call though. I've tried formatting the image in a dozen different ways and I always get 400/500 errors... Nothing in the Server logs either so I'm having a hard time making progress on the upload side of this.

Is there anything obvious I'm doing wrong? Otherwise, I can keep chipping away at it! No need to sidetrack you troubleshooting IMO. 

When we move to iOS 16, are we planning on using [PhotoPicker](https://developer.apple.com/documentation/photosui/photospicker) for the Profile Image? I was looking at using that for this but it looks like iOS 16 is the minimum.

Swiftfin/Views/ItemEditorView/ItemImages/EditItemImagesView/EditItemImagesView.swift
JPKribs: I had this idea that we could somewhat mirror web:

Screenshot 2025-01-03 at 14 21 39

Is this what you mean? 1 add/upload then have the type be a `menu` on the search (add) / confirmation (upload) view?

Shared/ViewModels/ItemAdministration/RemoteImageInfoViewModel.swift
JPKribs: Because I don't think ahead lol. I can make that swap!

Swiftfin/Views/ItemEditorView/ItemImages/EditItemImagesView/EditItemImagesView.swift
LePips: Not a `Menu` for the image type, as we have the sections for that. I think we're on the same page that it's a `Menu` for `Search` and `Upload file`.

Swiftfin/Views/ItemEditorView/ItemImages/EditItemImagesView/EditItemImagesView.swift
LePips: You aren't converting to the image data into base 64. Just adding `.base64EncodedData()` will make file upload pass.

Yes, we may look into using that when we drop iOS 15. For now, there is the `UserProfileImagePicker.PhotoPicker`.

Swiftfin/Views/ItemEditorView/ItemImages/EditItemImagesView/EditItemImagesView.swift
JPKribs: Just confirming:

Option 1:
Screenshot 2025-01-03 at 16 14 41

Option 2:
Screenshot 2025-01-03 at 16 17 47

Option 2, I would add a dropdown on the upload confirmation sheet. Something like this:

Screenshot 2025-01-03 at 16 27 55

Swiftfin/Views/ItemEditorView/ItemImages/EditItemImagesView/EditItemImagesView.swift
JPKribs: Oh my god that was it... I have a new error but it's more understandable since there is a server log:

`Jellyfin.Api.Middleware.ExceptionMiddleware: Error processing request: "Request body too large. The max request body size is 30000000 bytes". URL "POST" "/Items/bb1fd1f76a66e5a5a77e29d283a9ba4f/Images/Primary"`

Should I be doing something in code to compress the image down or should I throw if the image is too large? It looks like 30mb is the default but I'm sure people have configs with larger images so it might not make sense to cap this.

Swiftfin/Views/ItemEditorView/ItemImages/EditItemImagesView/EditItemImagesView.swift
LePips: I would prefer the first option, and instead of the ellipsis circle icon we can still use the plus.

Swiftfin/Views/ItemEditorView/ItemImages/EditItemImagesView/EditItemImagesView.swift
JPKribs: This was my solution for it. It's a tiny bit brute force capping at 3000 pixel for the longest side. I'm open to other ideas but I'm just glad I have something working for this:

```
    // MARK: - Upload Image

    private func uploadImage(_ url: URL, type: ImageType) async throws {
        guard let itemID = item.id else { return }

        guard url.startAccessingSecurityScopedResource() else {
            throw JellyfinAPIError("Unable to access file at \(url)")
        }
        defer { url.stopAccessingSecurityScopedResource() }

        let data = try Data(contentsOf: url)
        guard let image = UIImage(data: data) else {
            throw JellyfinAPIError("Unable to load image from file")
        }

        let maxDimension: CGFloat = 3000
        let resizedImage = image.size.width > maxDimension || image.size.height > maxDimension
            ? resizeImage(image, maxDimension: maxDimension)
            : image

        let contentType: String
        let imageData: Data

        if let pngData = resizedImage.pngData() {
            contentType = "image/png"
            imageData = pngData.base64EncodedData()
        } else if let jpgData = resizedImage.jpegData(compressionQuality: 0.8) {
            contentType = "image/jpeg"
            imageData = jpgData.base64EncodedData()
        } else {
            throw JellyfinAPIError("Failed to convert image to png/jpg")
        }

        var request = Paths.setItemImage(
            itemID: itemID,
            imageType: type.rawValue,
            imageData
        )
        request.headers = ["Content-Type": contentType]

        _ = try await userSession.client.send(request)
    }

    // MARK: - Resize Large Image

    private func resizeImage(_ image: UIImage, maxDimension: CGFloat) -> UIImage {
        let size = image.size
        let aspectRatio = size.width / size.height

        var newSize: CGSize
        if size.width > size.height {
            newSize = CGSize(width: maxDimension, height: maxDimension / aspectRatio)
        } else {
            newSize = CGSize(width: maxDimension * aspectRatio, height: maxDimension)
        }

        UIGraphicsBeginImageContextWithOptions(newSize, false, 1.0)
        image.draw(in: CGRect(origin: .zero, size: newSize))
        let resizedImage = UIGraphicsGetImageFromCurrentImageContext()
        UIGraphicsEndImageContext()

        return resizedImage ?? image
    }
```

LePips: Restricting size isn't necessarily valid,  as thumbs or logos are easily larger than that by default from remote sources. I'm really not worried about people uploading anything larger than 30MB, a near 16k image, that causes issues to all clients.

For file selection, we can just check the size of the file data and throw if larger than 30MB.

In the future if we need to manipulate downloaded images, we should use `Nuke.ImageRequest` and stuff like the `resize` processors instead.

JPKribs: I'm figuring it out. It looks like the `UIImage.pngData()` and `UIImage.jpegData()` can result in some inflated sizes. I'm going to add a check to the image at the start to just not convert it if it's already a PNG/JPEG. 

Otherwise, what do you think of this:

Simulator Screenshot - iPhone 16 Pro - 2025-01-04 at 11 46 26

JPKribs: This might be too verbose now:
Screenshot 2025-01-04 at 12 09 01

LePips: Oh, this issue is because the size of the compressed image inevitably will differ from the image laid out into memory, which happens when converting it into `UIImage`. Reading online, it may sometimes take into account the screen scaling which isn't what we necessarily want.

So, actually, the existing implementation for uploading user profile images is also incorrect. We are given the data for the image already and just need to try and determine what image it is by the extension.

The stubbed implementation:

```
private func uploadImage(_ url: URL, type: ImageType) async throws {
    guard let itemID = item.id else { return }

    let data = try Data(contentsOf: url)
	let imageType: String

    switch url.pathExtension {
    case "png":
		imageType = "image/png"
    case "jpg", "jpeg":
		imageType = "image/jpeg"
    default:
		// if heic photo, we could just take the jpeg data from it instead.
    }

    var request = Paths.setItemImage(
        itemID: itemID,
        imageType: type.rawValue,
        data.base64EncodedData()
    )

    request.headers = ["Content-Type": imageType]

    _ = try await userSession.client.send(request)
}
```

JPKribs:
I'm getting a lot closer but there are a lot of moving parts. This all works now but I still need to resolve some issues:

1. For some reason, when selecting an image from the search, there is a 50% chance it exits both views so I think there is a doubled up dismissCoordinator() in there.
2. The Search is empty if there are no images in a group the search is there are no existing images with that imageType. That one I have no idea is going on.
3. I need to implement photo uploading

Swiftfin/Views/ItemEditorView/ItemImages/EditItemImagesView/EditItemImagesView.swift
JPKribs: Here is the layout for selected images. We have `RemoteImages`, which are returned from a search with the `Save` in the top right. Then, we have `LocalImages` which is just `ImageInfo` but accounts for images that already exist for an item. These have the option to `Delete` at the bottom.

Screenshot 2025-01-05 at 16 44 07

JPKribs:
This is now fully functional! No rush for reviews. I've been staring at this all weekend so I'm going to revisit this later in the week and make sure there isn't anything I am missing.

As far as quirks, I thinks this is my first PR where I can come out of the gates and say everything is working exactly how I am expecting it. I think the only thing that might need a second look *(Later this week for me)* is the PhotoPicker logic for Items vs User Profile Images. It's essentially the same logic with the major difference being:

1. User Profiles lock to a square while items are a freeform rectangle
2. Router/ViewModel differences.

I'm sure there is a way to make that more generic so I can re-use this better but for now this is functional so I'm going to leave it for a second and revisit it later on with a clear head.

This should be the last PR for https://github.com/jellyfin/Swiftfin/discussions/1294 until the SDK update to 10.10.

Shared/ViewModels/ItemAdministration/ItemImagesViewModel.swift
```diff
@@ -0,0 +1,492 @@
...
+                            imageIndex: imageInfo.imageIndex
+                        )
+                        let request = Paths.getItemImage(
+                            itemID: itemID,
+                            imageType: imageInfo.imageType?.rawValue ?? "",
+                            parameters: parameters
+                        )
+                        let response = try await self.userSession.client.send(request)
+
+                        // TODO: Is there a way for me to just get the Image URL so I can use
```
LePips, line 299: Yes, and I would prefer that we don't just pass around the images like this:

```swift
userSession.client.fullURL(with: request)
```

LePips (changes requested):

Shared/ViewModels/ItemAdministration/ItemImagesViewModel.swift
JPKribs: Sounds good. I will make that change to use the URL and ImageView instead. That's cool we can get the full URL back out. I wasn't sure how that would work with the GET Body.

As a note, I committed a new version where I re-use the Image Crop/Picker as much as possible between Items and UserProfile. Differences in the `ViewModel`, `Router`, and `ViewModel.Actions` made it a little messier than I would like but it centralizes it fairly well. Let me know if you have any suggestions on how to do it better!

JPKribs: @LePips I've changed this to use the URL instead of the UIImage. I was googling this but is there a way to just extend `ImageInfo` to add another var for `imageURL`? I would prefer this over the truple(`ImageInfo`, `URL`. The other route I was considering was creating a new `LocalImageInfo` that was just `imageInfo: ImageInfo` &  `imageURL: URL` but I didn't want to add clutter with a struct that was just barely not `ImageInfo`.

`LocalImageInfo` since the other struct is `RemoteImageInfo` for remote sources and this would be local, saved to the server Image Info

JPKribs: Sorry about the burst of changes there. I should be set now. I just saw some room to trim down / reuse code so I took it while I was getting the UIImage changed for URL.

LePips: Instead of keeping a map of the URLs, you can instead create a similar function like `UserState.profileImageSource` that takes the client to be used and returns the constructed `URL`.

JPKribs: Gah, that's a really good idea. I am working on that now!

Something I'm stuck on... Should this be a `PosterHStack`?

```
    // MARK: - Image Scroll View

    @ViewBuilder
    private func imageScrollView(for imageType: ImageType) -> some View {
        if viewModel.images.isNotEmpty {
            ScrollView(.horizontal, showsIndicators: false) {
                HStack {
                    ForEach(viewModel.images, id: \.self) { imageInfo in
                        imageButton(imageInfo: imageInfo) {
                            router.route(to: \.deleteImage, imageInfo)
                        }
                    }
                }
            }
        }
    }
```

I tried making `ImageInfo` conform to `Poster` and everything built but I couldn't get a `PosterHStack` to populate so I think I'm doing something wrong. Thought I would ask before I spent too much time troubleshooting.

Shared/ViewModels/ItemAdministration/ItemImagesViewModel.swift
```diff
@@ -0,0 +1,452 @@
...
+        case initial
+        case content
+        case updating
+        case deleting
+        case error(JellyfinAPIError)
+    }
+
+    // MARK: - Image Variables
+
+    private let includeAllLanguages: Bool
```
JPKribs, line 48: No longer used. Delete.

Shared/ViewModels/UserProfileImageViewModel.swift
```diff
@@ -151,6 +151,12 @@ class UserProfileImageViewModel: ViewModel, Eventful, Stateful {
         )
         request.headers = ["Content-Type": contentType]
 
+        guard imageData.count < 30_000_000 else {
```
JPKribs, line 154: Greater than or equal to.

Swiftfin/Views/ItemEditorView/ItemImages/AddItemImageView/AddItemImageView.swift
```diff
@@ -0,0 +1,172 @@
...
+                .placeholder { source in
+                    if let blurHash = source.blurHash {
+                        BlurHashView(blurHash: blurHash, size: .Square(length: 8))
+                            .scaledToFit()
+                    } else {
+                        Image(systemName: "circle")
+                    }
+                }
+                .failure {
+                    VStack(spacing: 8) {
```
JPKribs, line 162: Just do a question mark.

Swiftfin/Views/ItemEditorView/ItemImages/ItemImageDetailsView/Components/ItemImageDetailsDetailsSection.swift
```diff
@@ -0,0 +1,109 @@
...
+//
+
+import JellyfinAPI
+import SwiftUI
+
+extension ItemImageDetailsView {
+
+    struct DetailsSection: View {
+
+        let imageURL: URL?
```
JPKribs, line 16: Remove "image" prefix.

Swiftfin/Views/PhotoPickerView/Components/PhotoCropView.swift
```diff
@@ -0,0 +1,183 @@
...
+    private var accentColor
+
+    // MARK: - State, Observed, & Environment Objects
+
+    @StateObject
+    private var proxy: _PhotoCropView.Proxy = .init()
+
+    // MARK: - Image Variable
+
+    var isReady: Bool
```
JPKribs, line 27: Let

Shared/ViewModels/ItemAdministration/ItemImagesViewModel.swift
JPKribs: I think, aside from potentially using `PosterHStack` this should be set? Let me know if there is anything that stands out to you! 

I did have 1 quirk I have a workaround for:

Swiftfin/Views/ItemEditorView/ItemImages/AddItemImageView/AddItemImageView.swift:

```
            .onReceive(viewModel.events) { event in
                switch event {
                case .deleted:
                    break
                case .updated:
                    UIDevice.feedback(.success)
                    *DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {*
                        *router.pop()*
                    *}*
                case let .error(eventError):
                    UIDevice.feedback(.error)
                    error = eventError
                }
            }
```

On Selection > Add, I `DismissCoordinator()` on the selection View. This brings the user back to `AddItemImageView` From there, I want to immediately `pop()` the `AddItemImageView` which would take the user back to the main `ItemImagesView`. If I have this `pop()` without a delay, it closes out the whole Coordinator and looks like this:

https://github.com/user-attachments/assets/f5d65470-5aed-4e16-a3f4-25bc14a3326d

With the delay, it works correctly:

https://github.com/user-attachments/assets/8d05235a-9fd1-4b09-9b52-974114affd37

LePips: This is because you have 2 areas to listen to viewmodel events that will interact with the routers and do some other stuff. The other area is `ItemImageDetailsView` that will call `router.dismissCoordinator`, and both of these areas also perform a success feedback.

We should not have multiple areas responding to events like this. Also, on an error from uploading an image (500), the image selection will be in an error state which isn't necessarily correct. We should find another way to handle that.

JPKribs:
I feel good about this current state. There's nothing that stands out to me. My only question is should we make a generic `ListRowDeleteButton`? I think we have delete buttons that look just like this in some other places as well?

LePips:
> generic ListRowDeleteButton?

No, as I've left a comment on `ListRowButtonStyle` that it should take into consideration the `configuration.role`. When `ButtonStyle.destructive` is used we would style it like we currently do the delete buttons.

LePips (approved):
I've completed my changes that should finish this out. I've left a couple `TODO`s that should be polishes in the future. The only real issue that we've observed is images with no sizing data which cause those layout issues but that's something we can solve in `ImageView` which then gets applied to all images throughout the app.

Issue #1339, opened 2024-12-05
Move from `OrderedSet` to `IdentifiedArrayOf`


LePips:
I've long thought about this but think we're going to run into this issue soon, especially with item deletion and updating throughout the app. I've run into some implementation issues while working on static notification payloads, and this will help with that.

We are very susceptible to the following structure, where our items all have 

```swift
struct MyItem: Hashable {
    
    let id: String?
    let name: String?
    let otherValue: Int?
}

let foo = MyItem(id: "123", name: "foo", otherValue: 123)
let bar = MyItem(id: "123", name: nil, otherValue: nil)

// false, but we actually care about the IDs
print(foo.hashValue == bar.hashValue)

```

We should migrate to [swift-identified-collections](https://github.com/pointfreeco/swift-identified-collections) instead for things like `PagingLibraryViewModel` or other contexts where we currently use `OrderedSet` to represent `BaseItemDto`, `BaseItemPerson`, and other items that conform to `Identifiable`.

LePips:
Oof, this requires a bit of thinking for my `CollectionHStack` and `CollectionVGrid` packages to take `Hashable` or `Identifiable` elements...

JPKribs:
@LePips I found a new one. Selecting `See All` for Cast & Crew results in:

![Image](https://github.com/user-attachments/assets/f2f8d479-280e-4bc7-b9ec-0245458b8cd9)

I tried some other similar locations like Similar Items / Recommendations, Search > `See All`, & the Home View `See All`. All of these work but give me:

`Accessing StateObject's object without being installed on a View. This will create a new instance each time.`

I did some digging and it looks like every on of these appropriately uses `viewModel.elements` which is formatted as `IdentifiedArray<Int, Element>`. Only Cast & Crew is just a pure `[BaseItemPerson]`.

![Image](https://github.com/user-attachments/assets/c26da9c9-561e-4099-8e03-5943ab47debe)

LePips:
Ah, good catch. Fixed in https://github.com/jellyfin/Swiftfin/pull/1354.

For the runtime warning, that's always kind of been an issue when working with Stinsen. However since we'll be moving away from it I haven't thought about what it's warning about much but we shouldn't have to worry about it for much longer.

JPKribs:
@LePips Do you think this is related: https://github.com/jellyfin/Swiftfin/issues/1368? Otherwise, I can start looking at this on my end!

LePips:
[...]

PR #1346, opened 2024-12-08
Move to `IdentifiedArray` in `PagingLibraryViewModel`

LePips:
As laid out in #1339, we need to move to `IdentifiedArray` where applicable. This starts with `PagingLibraryViewModel`, probably the most important target of the issue.

I largely refactored my `CollectionHStack` and `CollectionVGrid` packages to no longer depend on `OrderedSet`, mirroring `List`, and fixed all instances accordingly.


JPKribs (approved):
I took a look through and it looks good. Tested it locally and everything seems to be working as expected.

My only note is on the runtime labels which isn't really related to this PR but just makes the commit history a little weird moving runtime up and down 3 lines.

PR #1348, opened 2024-12-08
[tvOS] Media Item Menu - Refresh / Delete Items

JPKribs:
### Summary
This should be the full extent of tvOS item management. I can't imagine that anyone want to manually write metadata or replace images from their TV. Instead, this should allow refreshing metadata if data is missing or some information is wrong. Additionally, if enabled, this should allow for deleting items which was a popular ask on the Android TV side: https://github.com/jellyfin/jellyfin-androidtv/discussions/2368

This should almost 1 to 1 mirror iOS with some small differences:

1. Since this is on the ActionHStack, I don't have `router` access to dismiss the coordinator on item Deletion. The result is the user needs to manually go back.
2. iOS has a loading `ProgressView()` on the `Refresh Metadata` button. Since tvOS's button is hidden behind the menu, I removed the loading circle since it was giving some weird flashing (at least on the emulator)

### Screenshots

<details>
<summary>Buttons - No Access vs Access</summary>
Screenshot 2024-12-08 at 15 31 37
</details>

<details>
<summary>Menu Buttons</summary>
Screenshot 2024-12-08 at 15 32 46
</details>

<details>
<summary>Refresh Options</summary>
Screenshot 2024-12-08 at 15 33 15
</details>

<details>
<summary>Delete Warning</summary>
Screenshot 2024-12-08 at 15 33 04
</details>


LePips:
> Since this is on the ActionHStack, I don't have router access to dismiss the coordinator on item Deletion. The result is the user needs to manually go back.

That isn't necessarily true as routers are retrieved via `EnvironmentObject`. For example the following works, as all items are presented with an `ItemCoordinator`. I did test this, however I ran into the same issue as https://github.com/jellyfin/Swiftfin/pull/1090#issuecomment-2392425336, where I had to delete the `.receive(on: RunLoop.main)` in `DeleteItemViewModel.events`. I'm still unsure why this is an issue. I also did not test iOS.

```swift
@EnvironmentObject
private var router: ItemCoordinator.Router

//...

.onReceive(deleteViewModel.events) { event in
    switch event {
    case let .error(eventError):
        error = eventError
        isPresentingEventAlert = true
    case .deleted:
        router.dismissCoordinator()
    }
}
```

Swiftfin/Views/ItemEditorView/AddItemElementView/Components/SearchResultsSection.swift
```diff
@@ -13,13 +13,19 @@ extension AddItemElementView {
...
         var name: String
+        @Binding
+        var id: String?
+
+        // MARK: - Element Search Variables
 
         let type: ItemArrayElements
         let population: [Element]
+
+        // TODO: Why doesn't environment(\.isSearching) work?
```
LePips, line 28: We can't always re-use SwiftUI provided environment values as they can be controlled internally and not be set/change unexpectedly. The documentation for `EnvironmentValues.isSearching` states that this is used when utilizing the SwiftUI `searchable` stuff, which we don't use.

- https://developer.apple.com/documentation/swiftui/environmentvalues/issearching

LePips (approved):
Looks great!

PR #1349, opened 2024-12-08
Static Notification Payloads, Move more to `IdentifiedArray`

LePips:
- Overhauls current notifications to allow static payloads, so no more casting the user info is necessary.
- Some refactoring in `SeriesItemViewModel` and `SeasonItemViewModel` to use `IdentifiedArray`. Instead of passing around objects we look at IDs.


Shared/ViewModels/ItemViewModel/ItemViewModel.swift
```diff
@@ -90,22 +90,44 @@ class ItemViewModel: ViewModel, Stateful {
...
-                    } else if let seriesID = userInfo["seriesID"], seriesID == item.id {
-                        Task { [weak self] in
-                            await self?.send(.backgroundRefresh)
-                        }
-                    }
-                } else if let newItem = notification.object as? BaseItemDto, newItem.id == self?.item.id {
-                    Task { [weak self] in
-                        await self?.send(.replace(newItem))
-                    }
+//        Notifications[.itemMetadataDidChange].publisher
```
JPKribs, line 93: Do we need this commented code or can we delete this?

Swiftfin tvOS/Views/SelectUserView/SelectUserView.swift
```diff
@@ -310,33 +310,27 @@ struct SelectUserView: View {
...
+                } else {
+                    serverSelection = .all
                 }
-
-                // change splash screen selection if necessary
-//                selectUserAllServersSplashscreen = serverSelection
             }
+
+            // change splash screen selection if necessary
+//            selectUserAllServersSplashscreen = serverSelection
```
JPKribs, line 333: Same as above.

JPKribs (dismissed):
@LePips Looks good minus the 2 comments I tagged. If we don't think we're going to use this later, I say we remove it and we can revisit version history if we ever need to reference it.

You know better than I do if it makes sense to keep those as comments. I pulled this and I didn't notice anything functionally different which is great!

Shared/ViewModels/ItemViewModel/ItemViewModel.swift
LePips: Good catch on this one, forgot to clean up.

Swiftfin tvOS/Views/SelectUserView/SelectUserView.swift
LePips: This was previously commented because this isn't complete on tvOS.

JPKribs (approved):

PR #1352, opened 2024-12-09
Fix optional ID case for libraries

LePips:
I noticed that items weren't being deleted from libraries. I tracked this down due to the following behavior:

```swift
let fooA = "Foo".hashValue // String.hashValue
let fooB = Optional("Foo").hashValue // Optional.hashValue
let fooC = Optional("Foo")?.hashValue

print(fooA)
print(fooB)
print(fooC)

// 7631599267132626175
// -2986088522783645984
// Optional(7631599267132626175) 
```

Many of the objects like `BaseItemDto` have `id: String?`, which would mean that when we receive the item ID `String` it will be comparing something like `fooA == fooB` from above, which will never be true. I have had to make a new protocol for items that may be in libraries and must return the intended `.hashValue` target or 0 otherwise. When using `Optional`, this would be `Wrapped.hashValue`. I don't see us running into an issue with the 0 case, but something to be mindful of.

This would be better fixed if there were parameterized generics so I could extend `Identifiable` or provide the default implementation. Or, I could've compared `Optional(id).hashValue` if the first failed, but I would rather be statically safe than try to remember typing issues.

### DIFF
--- a/Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto.swift
+++ b/Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto.swift
@@ -27,6 +27,13 @@ extension BaseItemDto: LibraryParent {
     }
 }
 
+extension BaseItemDto: LibraryIdentifiable {
+
+    var unwrappedIDHashOrZero: Int {
+        id?.hashValue ?? 0
+    }
+}
+
 extension BaseItemDto {
 
     var episodeLocator: String? {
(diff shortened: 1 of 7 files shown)


JPKribs (approved):
This makes sense to me! That's annoying that hashes vary based on the `Optional` but that does make sense if I stop to think about it.

Apologies for not testing the deletion portion of this on the first round. Despite creating the delete functionality, I don't really like to delete items so that slipped my testing 😆.

LePips:
Nah it broke from my change to `IdentifiedArray` and worked with your implementation.

PR #1353, opened 2024-12-10
[iOS] Admin Dashboard - Parental Ratings

JPKribs:
### Summary
Allows settings the maximum Parental Rating for a User. Also allows changing how unrated or unrecognized ratings are handled based on Type.

The way this works on Server is ratings have a name and a number associated to them. This way, TV-14 and PG-13 are roughly the same value. This is why they are grouped together like this:

Screenshot 2024-12-09 at 19 02 14

I've done the same thing for us. It's a bit cluttered looking but I'm not sure how else to show this.

### Screenshots
<details>
<summary>Dropdown</summary>
Screenshot 2024-12-09 at 18 57 09
</details>

<details>
<summary>Other Options</summary>
Screenshot 2024-12-09 at 18 57 23
</details>


LePips:
Hm, I think that there is something better we can do than just a `Menu` and these offerings will only ever get larger. I'll have to think about it.

LePips:
I had thought about something like that as well: to just show the shared ages of the parental ratings. I think that's the way go since we've both come across it.

I think we can make it better with a `Learn More` where each age lists what ratings are a part of that age cap.

JPKribs:
> I think we can make it better with a `Learn More` where each age lists what ratings are a part of that age cap. That will be dynamic if it ever gets more granular.

Here is my working version. Let me know what you think!
Simulator Screenshot - iPhone 16 Pro - 2024-12-11 at 10 08 18
Simulator Screenshot - iPhone 16 Pro - 2024-12-11 at 10 08 22

LePips (changes requested):
`Learn More` without a bulleted list looks great. However, I think the way the ages are presented is backwards. The ages are actually for the maximum age for content, whereas we are presenting them like "this age and up" are allowed. We should indicate this is the `Max parental rating` and don't need the `_ +` strings.

JPKribs:
> We can simplify that this is the `Max parental rating` and don't need the `+` in the strings.

Here is my latest commit. Made the StateObject change + the String change:
Screenshot 2024-12-11 at 11 40 08

LePips (dismissed):
I just did some cleanup and switched `LearnMoreButton` to use `SeparatorVStack`.

JPKribs:
> I just did some cleanup and switched `LearnMoreButton` to use `SeparatorVStack`.

That's a good change! Only note, without the foregroundStyle, the LearnMore inherits the style from the footer and looks like this:

simulator_screenshot_6ABDAF78-7CED-4D88-B394-BE02F95F72F5

Not an issue just so you're aware before we merge if we want to change the section headers to Color.primary instead.

LePips:
Thanks for pointing that out. I've fixed it on the `LearnMoreButton` level so that passing in the `foregroundStyle` on the callers isn't necessary.

LePips (approved):

PR #1354, opened 2024-12-10
Fix possible duplicate ids

LePips:
Since we don't check the uniqueness of ids with `uniqueElements` we need to at least employ the uniqueness consolidation in `IdentifiedArray`. Just keep the first item in the event of collisions.

### DIFF
--- a/Shared/ViewModels/LibraryViewModel/PagingLibraryViewModel.swift
+++ b/Shared/ViewModels/LibraryViewModel/PagingLibraryViewModel.swift
@@ -57,7 +57,7 @@ protocol LibraryIdentifiable: Identifiable {
        on remembering other filters.
  */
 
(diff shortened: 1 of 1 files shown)


JPKribs (approved):
Verified all the See Alls and this resolved what I was seeing. Tested deleting which worked. Editing metadata did not work for pagingLibraryViewModel but I realized that isn't actually implemented.

Looks good to me!

JPKribs:
@LePips for `PagingLibraryViewModel`, updating a single item on metadata change, using the new Notification payloads would that just look like this:

```
        Notifications[.itemMetadataDidChange]
            .publisher
            .sink { item in
                guard let newItem = item as? Element else {
                    return
                }
                self.elements[id: newItem.unwrappedIDHashOrZero] = newItem
            }
            .store(in: &cancellables)
```

The id.hashvalue is what I'm having trouble wrapping my head around. Like, a pure `self.elements[index] = newItem` wouldn't work, right?

PR #1356, opened 2024-12-10
Update README

LePips:
- Direct to tvOS update discussion for TestFlight
- Put loc at the bottom
- Convert the TestFlight icon svg text to outlines instead of raw text


JPKribs (approved):
Looks good!

PR #1357, opened 2024-12-11
[iOS & tvOS] Error Cleanup

JPKribs:
### Summary
This PR finds all instances where `isPresentingError` is used as a `State`. I'm using this to to find where errors are presented via an alert and I am replacing them with the new `errorMessage(error)`. In addition to this, I updated a couple typos like `Uplicates` vs `Duplicates`. Finally, I tried to `Mark` the classes and break up the variables to try and make everything more readable.

The primary goal of this is the errors so if the `MARK`s are unwanted I can remove those.

Here is an example of why I wanted to get all of these on the same page. Left is current. Right is updated. Should make this easier to maintain as well!
Screenshot 2024-12-10 at 18 00 06


JPKribs:
Drafting real quick. Going to localize everything in those views while I'm touching them.

PR #1358, opened 2024-12-11
[iOS] Admin Dashboard - User Access Schedules

JPKribs:
### Summary
This creates a view to edit existing Access Schedules for a User and Add a new schedule to a user. Schedules are periods of time where a user is allowed to access Jellyfin and consists of a Start Time, End Time, and a Day/DayOfWeek.

🎉 This is the last **new** PR needed for the Admin Dashboard. 🎉

Context below:

_I still need to work out the User Profile Picture & Username on https://github.com/jellyfin/Swiftfin/pull/1328 but change reworks some existing code so I'd like to focus on that and knock it all out in one go. This PR can sit for a bit since it's new functionality I can weave into other PRs without too many merge items._

_Access Tags are still on the list but I'm considering moving that to a later date since BlockedTags exists in 10.8 but AllowedTags need 10.10 (Maybe 10.9?) SDK to work. This means I would need to come back later to finish this regardless so I would feel better just doing it all at once when we're on the 10.10 SDK._

### Screenshots

<details>
  <summary>View</summary>
  View
</details>

<details>
  <summary>Single Delete</summary>
  Single Delete
</details>

<details>
  <summary>Multi-Select</summary>
  Multi-Select
</details>

<details>
  <summary>Add View</summary>
  Add View
</details>

<details>
  <summary>Dropdown</summary>
  Dropdown
</details>

<details>
  <summary>Duplicate Schedule</summary>
  Duplicate Schedule
</details>


JPKribs:
This should be ready to go. The only thing missing I would like to do is that the current version of this spits out an error if you save a duplicate. I would rather this worked like how the AddServerTaskTrigger works:

Screenshot 2024-12-11 at 19 08 27

The only reason I wasn't able to do this is because the `AccessSchedule`'s `StartHour` and `EndHour` formats are just bad. Instead of a regular time, or dateTime, or even ticks, it's represented as ##.#. So, 10:30pm is 22.5. As a result of this, I have so add in some extra processing on `Save` to make get this into format. So getting it from `Date Components` to `String` or from `Date` to `Double` I couldn't find a clean way to appropriately check if an Access Schedule existed.

I settled on the error on creation which is more than we're currently doing for web now but if there is a smarter way to handle this where it's more preemptive that would be ideal.

There is a tiny bit of cleanup where I'm trying to get the Admin Dashboard to be a little less chaotic. So, I've taken this for both the Server Tasks and the Server User Access Schedules since they have Edit & Add Views:
Screenshot 2024-12-11 at 19 12 16

Shared/Extensions/JellyfinAPI/DynamicDayOfWeek.swift
```diff
@@ -0,0 +1,38 @@
...
+
+import Foundation
+import JellyfinAPI
+
+extension DynamicDayOfWeek {
+
+    var displayTitle: String {
+        switch self {
+        case .sunday:
+            DayOfWeek.sunday.displayTitle ?? self.rawValue
```
JPKribs, line 17: On DayOfWeek, we are using this to get the day of the week instead of us localizing it: https://github.com/jellyfin/Swiftfin/pull/1255#discussion_r1797778181

Because of this, I am just using this to steal the localization from there. I think I could get away with `!`  instead of ?? rawValue but in case of failure I thought this would be less crash-y.

LePips (approved):
Implemented being able to check for duplicate schedules and some cleanup.

This list should also have been `insetGrouped` (automatic) because it isn't showing much complex information like users, devices, or media elements do. This more closely matches the task triggers and custom device profiles. Also, having the `Add` button in the list for empty data cases helps so that the user doesn't have to go through the menu and matches other design like for the task triggers and custom device profiles. I missed checking this stuff on  `EditItemElementView`, but that needs to be migrated to this as well.

Also, I reverted the `Add trigger` -> `Add Trigger` casing. I haven't been strong on casing before, but I think we should strongly move to `Sentence case` over `Title Case` where appropriate. I may have missed other casing changes before but don't want to move pre-existing strings to `Title Case` for now.

PR #1360, opened 2024-12-12
[WIP] [tvOS] Standardized Overlay Menu

JPKribs:
### Summary
Creates a new overlay type similar to a sheet in tvOS but to appear like a popup menu. I'm not sold on this and if there is a more standard design I'm all for using this instead. Primarily, I think this would be good for for a tvOS version of the `ChevronAlertButton` since (at least in emulator) that alert gives us issues. Additionally, I think this would be a good delivery method for things like the `StepperView`, and eventually tvOS filters to allow interacting with a menu without `taking` the user away from the page.

Finally, instances where a `Menu` is not possible, IMO, this could be a viable route.

I'm fine using `Alerts`, & `Sheets` as well but tvOS seems to have some weird limitations on where those can be called from.

Completely open to feedback or not proceeding with this! This was just an idea I wanted to build on top of from the [Naviation Bar Menu Issues](https://github.com/jellyfin/Swiftfin/pull/1359) that tvOS seems to have.

### Example Screenshot
Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-12-12 at 10 14 16

### Example Usage
https://github.com/user-attachments/assets/06f989c2-97b3-4b6b-9234-6587c952201e

https://github.com/user-attachments/assets/056bcb9d-0a5b-44b9-bd81-a3b47330a086


JPKribs:
Just as note, this design is just kind of ripped from the `ErrorMessageModal` but with a white border. I don't have an strong opinions on it personally but I think it would be nice to get something set once that we can use to ensure we're not doing anything that's going to look out of place.

chickdan:
Thanks for seeing my struggle and helping work around the limitations!

The direction I'd take is to have the title left aligned for a "menu" style usage like the screenshot you provided in #1359, and center aligned (as you have here) for "alert" style usage.

I don't, however, care for the white border and maybe that's what is bothering you as well? The divider also feels out of place to me, it's pretty rare to see that in Apple UI outside of lists.

JPKribs:
> I don't, however, care for the white border and maybe that's what is bothering you as well? The divider also feels out of place to me, it's pretty rare to see that in Apple UI outside of lists.

Here's the revision from this. Is that an improvement:

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-12-12 at 17 39 08

chickdan:
Yeah, that's good with me! Though I am not very familiar with tvOS design as much as mobile 😅

JPKribs:
I screwed around with this a little more. Here is Apple's Default `Alert` (left) next to this new `overlayMenu()` (right). I've updated the `ChevronAlertButton` to use this for tvOS as well. Feeling a lot better about this version! Open to feedback!

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-12-14 at 17 42 37

This part doesn't exist in this commit but I was playing around with putting the `IndicatorSettingsView` into here. I think this looks good and would be usable for menu options:

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-12-14 at 17 42 37

JPKribs:
Changing direction on this PR. The new goal is:

1. Add an `orientation` option to `FullScreenMenu` to allow it to be on either `leading` or `trailing`. 
2. Add a `subtitle` option to `FullScreenMenu` in case we need descriptive text included with the title. 
3. Fix up some of the padding for `FullScreenMenu` so items don't clip out by default.
4. Add a `View Modifier` to to be able to call this similarly to how we call `Alerts`.

JPKribs:
Replacing this PR with https://github.com/jellyfin/Swiftfin/pull/1363

LePips:
Apologies for not commenting earlier. I do agree that something like this may be necessary depending on the context. However I will typically always want to prioritize using provided modifiers or `Menu`.

PR #1363, opened 2024-12-16
[tvOS] SelectServerView Change to Menu

JPKribs:
### Summary
Changes up some of the padding, coloring, and rounding of the `FullScreenMenu`. Enable `FullScreenMenu` as a `ViewModifier` to allow it to be used as `View.fullScreenMenu`. Create a modifier to the `FullScreenMenu` for `orientation` to allow the view to be either `leading` or `trailing` for more dynamic usage.

Clean up the `SelectServerView` to utilize the new `FullScreenMenu` and better mirror Apple's default `Alert`.

### Videos

Old:

https://github.com/user-attachments/assets/cc4bb7e0-8921-4dc4-8800-b675b23e194f

New:

https://github.com/user-attachments/assets/1264242f-e5dc-4134-a10a-bb571d75356a

### Screenshots

<details>
<summary>SelectServerView Buttons (Old vs New)</summary>
Screenshot 2024-12-16 at 11 48 25
</details>

<details>
<summary>SelectServerView Server List (Old vs New)</summary>
Screenshot 2024-12-16 at 11 49 59
</details>

<details>
<summary>FullScreenMenu Color Changes (Alert vs Old | Alert vs New)</summary>
Screenshot 2024-12-16 at 11 57 51
</details>


LePips:
Why not first investigating `Menu`? Moving to tvOS 17 as minimum to use `Menu` for this screen at the very least was a primary motivation.

JPKribs:
I mostly went this direction since the `Menu` doesn't work from the `Toolbar`. I have a branch where I put this into a `Menu` which I think looks like this:

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-12-16 at 14 36 38

The only downside I see is the scroll scrolls with everything so we can't pin anything at the top like I did in my `FullScreenMenu`. Not the end of the world at all but just a note:

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-12-16 at 14 40 40

If you'd prefer, I can use this instead. This does mean there is no other usage of `FullScreenMenu` so would we want to delete that or do we think there is some potential usage elsewhere?

JPKribs:
I've updated this to be the `Menu` and I removed the 2 views that are no longer used. Actually a big fan of the `Menu` and it seems to be a lot sturdier on the emulator as well so I think that's the best way to go! 

Let me know if you want me undo any of the deletions on this PR in case we need those views in the future.

Swiftfin tvOS/Views/SelectUserView/Components/ServerSelectionMenu.swift
```diff
@@ -43,33 +40,97 @@ extension SelectUserView {
...
+                Group {
+                    switch serverSelection {
+                    case .all:
+                        Label(L10n.allServers, systemImage: "person.2.fill")
+                    case let .server(id):
+                        if let server = viewModel.servers.keys.first(where: { $0.id == id }) {
+                            Label(server.name, systemImage: "server.rack")
+                        }
+                    }
+                }
```
LePips, line 126: This should also have the menu indicator like on iOS:

https://github.com/jellyfin/Swiftfin/blob/b0583125f78f791e3a053619abae370d7c989089/Swiftfin/Views/SelectUserView/Components/ServerSelectionMenu.swift#L93

Swiftfin tvOS/Views/SelectUserView/Components/ServerSelectionMenu.swift
```diff
@@ -43,33 +40,97 @@ extension SelectUserView {
         }
 
         var body: some View {
-            Button {
-                let parameters = SelectUserCoordinator.SelectServerParameters(
-                    selection: _serverSelection,
-                    viewModel: viewModel
-                )
+            Menu {
```
LePips, line 43: I somewhat disagree that the menu layout should differ compared to iOS. In fact, just copying over the iOS implementation compiles but the order of the elements differ. At least right now, on iOS I do like that the `Add`, `Edit`, and `All Servers` buttons have bottom layout priority on iOS and the list would grow up with the additional servers.

<details>
<summary>iOS Menu</summary>

IMG_6673
</details>

Now, the fact that happens on iOS is currently just manual manipulation of where the `Menu` is and where the contents will be laid out. [`menuOrder`](https://developer.apple.com/documentation/swiftui/view/menuorder(_:)) has since been implemented and we can currently use that for tvOS to get the order of the menu elements that we would like. We will soon do that for iOS since it was implemented in iOS 16.

LePips (changes requested):

Swiftfin tvOS/Views/SelectUserView/Components/ServerSelectionMenu.swift
JPKribs: Mirroring iOS is a a little off on the centering:

Screenshot 2024-12-17 at 14 32 21

```
                HStack {
                    switch serverSelection {
                    case .all:
                        // Image(systemName: "person.2.fill")
                        Label(L10n.allServers, systemImage: "person.2.fill")
                    case let .server(id):
                        if let server = viewModel.servers.keys.first(where: { $0.id == id }) {
                            // Image(systemName: "server.rack")
                            Label(server.name, systemImage: "server.rack")
                        }
                    }

                    Image(systemName: "chevron.up.chevron.down")
                        .foregroundStyle(.secondary)
                        .font(.subheadline.weight(.semibold))
                }
```

If I change it to this, it centers a bit better:

```
                HStack {
                    switch serverSelection {
                    case .all:
                        Image(systemName: "person.2.fill")
                        Text(L10n.allServers)//, systemImage: "person.2.fill")
                    case let .server(id):
                        if let server = viewModel.servers.keys.first(where: { $0.id == id }) {
                            Image(systemName: "server.rack")
                            Text(server.name)//, systemImage: "server.rack")
                        }
                    }

                    Image(systemName: "chevron.up.chevron.down")
                        .foregroundStyle(.secondary)
                        .font(.subheadline.weight(.semibold))
                }
```

Screenshot 2024-12-17 at 14 31 38

LePips: Ah, the issue is due to the spacing differences between the `Label` contents and between `HStack` contents. However the server and disclosure icons should more closely hug the text, so maybe setting the `HStack(spacing:)` to something sensible would be better.

Swiftfin tvOS/Views/SelectUserView/Components/ServerSelectionMenu.swift
JPKribs: I'm having some trouble with this. I can manually force this order where `AddServer` is at the bottom but the `Menu` defaults to focusing the top item first. So, if I had 50 servers, I would need to scroll through all of them to get to `AllServers`. I was hoping `Priority` would be helpful but I am getting:

Screenshot 2024-12-17 at 15 05 09

JPKribs: This is the current version I have:

image

It's kind of the reverse of iOS's order. The focus in this screenshot matches where focus goes on first Menu selection

LePips: If you use `menuOrder(.fixed)` then it will be laid out in the order as statically defined in the code. That means that it will require switching around the sections and buttons. At least given our design where the `Menu` is always on the bottom, we can guarantee that the menu will be presented upwards and get the ordering that we want. Although, I find it odd `priority` isn't offered on tvOS.

JPKribs: This is where I'm at now. The order is good but I don't love that the focus starts at the top. I've tried a couple different `@FocusState` attempts but it doesn't seem to be working how I expect it to. Maybe it's just getting the `onSelect` for `Menu` isn't working how I think it should. I've tried `onSubmit` and `onTapGesture` to no avail.

Here's what this looks like with a lot of items:

https://github.com/user-attachments/assets/8ea5ec66-1e4c-48c1-a442-a8d5caf858ad

LePips: Hm, you're right that it may be annoying that it starts at the top. However for this I don't anticipate many servers being added, especially on tvOS, and we can figure that out later.

JPKribs: Sounds good! If you're happy with that layout, then this should be good to go!

LePips (approved):

PR #1113, opened 2024-06-15
[WIP][POC] update swiftfin tvos signin

Isaiah-Hamilton:
## Changes made

Updated the swiftfin tvos signin view

## Before

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-06-15 at 02 17 00

## After

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-06-15 at 02 08 41
## Feedback

Let me know which one is best and what you think.

PR #1365, opened 2024-12-17
[tvOS] Update ConnectToServerView & UserSignInView

JPKribs:
### Summary
Continues the work from https://github.com/jellyfin/Swiftfin/pull/1113 with updates from the feedback provided. Increases padding in a lot of locations and swaps out the `Public Users` list for a grid.

Since this view is essentially the same thing between `ConnectToServerView` and `UserSignInView` I created `SplitLoginWindowView` in the same vein as `SplitFormWindowView`. This works as 2 sections, with titles and content, separated by a divider with a Jellyfin logo at the top.

For `ConnectToServerView`, the only real changes made are changes to make the `LocalServerButton` better mirror the `PublicUserButton` and making the connect button match the sign in button.

In my screenshots below, I duplicated by Server/Users to fill out the list more. This is why there are duplicates.

### Connect To Server

---

### Before
Screenshot 2024-12-17 at 12 44 41

---

### After
Screenshot 2024-12-17 at 12 26 26

---

### User Sign In

---

### Before
Screenshot 2024-12-17 at 12 43 15

---

### After

Screenshot 2024-12-17 at 12 27 28

---

### DIFF
--- a/Swiftfin tvOS/Components/SplitLoginWindowView.swift
+++ b/Swiftfin tvOS/Components/SplitLoginWindowView.swift
@@ -51,7 +51,7 @@ struct SplitLoginWindowView: View {
                 }
             }
         }
-        .navigationBarBranding(isLoading: true)
+        .navigationBarBranding(isLoading: isLoading)
     }
 }
 
--- a/Swiftfin tvOS/Components/SplitLoginWindowView.swift
+++ b/Swiftfin tvOS/Components/SplitLoginWindowView.swift
@@ -9,18 +9,18 @@
 import Foundation
 import SwiftUI
 
-struct SplitLoginWindowView: View {
+struct SplitLoginWindowView<Leading: View, Trailing: View>: View {
 
     // MARK: - Loading State
 
-    private var isLoading: Bool
+    private let isLoading: Bool
 
     // MARK: - Content Variables
 
-    private var leadingTitle: String
-    private var leadingContentView: () -> any View
-    private var trailingTitle: String
-    private var trailingContentView: () -> any View
+    private let leadingTitle: String
+    private let leadingContentView: () -> Leading
+    private let trailingTitle: String
+    private let trailingContentView: () -> Trailing
 
     // MARK: - Body
 
@@ -57,19 +57,17 @@ struct SplitLoginWindowView: View {
 
 extension SplitLoginWindowView {
 
-    init(isLoading: Bool = false, leadingTitle: String, trailingTitle: String) {
+    init(
+        isLoading: Bool = false,
+        leadingTitle: String,
+        trailingTitle: String,
+        @ViewBuilder leadingContentView: @escaping () -> Leading,
+        @ViewBuilder trailingContentView: @escaping () -> Trailing
+    ) {
         self.isLoading = isLoading
         self.leadingTitle = leadingTitle
         self.trailingTitle = trailingTitle
-        self.leadingContentView = { EmptyView() }
-        self.trailingContentView = { EmptyView() }
-    }
-
-    func leadingContentView(@ViewBuilder _ content: @escaping () -> any View) -> Self {
-        copy(modifying: \.leadingContentView, with: content)
-    }
-
-    func trailingContentView(@ViewBuilder _ content: @escaping () -> any View) -> Self {
-        copy(modifying: \.trailingContentView, with: content)
+        self.leadingContentView = leadingContentView
+        self.trailingContentView = trailingContentView
     }
 }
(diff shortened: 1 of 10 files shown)


Swiftfin tvOS/Components/SplitLoginWindowView.swift
```diff
@@ -0,0 +1,94 @@
...
+    private var leadingTitle: String
+    private var leadingContentView: () -> any View
+    private var trailingTitle: String
+    private var trailingContentView: () -> any View
+
+    // MARK: - Body
+
+    var body: some View {
+        VStack {
+            HStack {
```
LePips, line 29: This "custom navigation bar" should be used across the user selection, server selection, and user sign in screens. So, it should probably be broken out into its own view.

Swiftfin tvOS/Views/UserSignInView/Components/PublicUserButton.swift
```diff
@@ -41,39 +50,49 @@ extension UserSignInView {
...
+                    .failure {
+                        fallbackPersonView
+                    }
+            }
+        }
+
+        // MARK: - Body
+
         var body: some View {
             Button {
```
LePips, line 77: I've recently checked out the new additions in tvOS since there has been more traction in this area. A lot has been added to make better buttons that we will adopt for the platform.

Changing the button to below will get us a better focus effect and will automatically move the labels on hover:

```swift
Button(action: action) {
    personView
        .aspectRatio(1, contentMode: .fill)
        .posterShadow()
        .clipShape(.circle)
        .frame(width: 150, height: 150)
        .hoverEffect(.highlight)

    Text(user.name ?? .emptyDash)
        .fontWeight(.semibold)
        .foregroundStyle(.primary)
        .lineLimit(1)
        .padding(.bottom)
}
.buttonBorderShape(.circle)
.buttonStyle(.borderless)
.disabled(!isEnabled)
.foregroundStyle(.primary)
```

I would recommend looking at the following sample project to see the new additions that we will be referencing.

- https://developer.apple.com/documentation/swiftui/creating-a-tvos-media-catalog-app-in-swiftui

LePips (changes requested):

Swiftfin tvOS/Components/SplitLoginWindowView.swift
JPKribs: Sounds good. Make this into a `View` or a `View Modifier` like how the iOS filters are handled?

JPKribs: This is what I came up with. Looks the same but can be tacked onto any view we need it to including `SplitLoginWindowView`

```
import Defaults
import SwiftUI

struct NavigationBarBrandingModifier: ViewModifier {

    var isLoading: Bool?

    func body(content: Self.Content) -> some View {
        VStack {
            VStack(alignment: .trailing) {
                if let loading = isLoading, loading == true {
                    ProgressView()
                }
            }
            .frame(maxWidth: .infinity)
            .frame(height: 100)
            .overlay {
                Image(.jellyfinBlobBlue)
                    .resizable()
                    .aspectRatio(contentMode: .fit)
                    .frame(height: 100)
                    .edgePadding()
            }
            .padding(.bottom)

            content

            Spacer()
        }
    }
}
```

Swiftfin tvOS/Views/UserSignInView/Components/PublicUserButton.swift
JPKribs: What!? That looks so much better! Big fan of how this looks!

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-12-17 at 16 26 20

JPKribs:
Here's what this now look like:

Just moved this to the actual toolbar to work better. Here are all of the views that have this:

### User Selection

With Background
Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-12-17 at 22 08 24

Without Background
Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-12-17 at 22 08 30

---

### User Login

With example `ProgressView()`
Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-12-17 at 22 20 22

---

### Server Connection
Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-12-17 at 22 08 48

LePips (approved):
I'm moving away from the pattern of functions on views that supply the content through `() -> any View`, at least where the views aren't actually customizable and should be using generics instead.

These views still have some work that can be done like having the server splash screen on the background of `UserSignInView` as well as some design refinements.

LePips (approved):

PR #1369, opened 2024-12-18
[iOS] Media Item Menu - Identify Media Item

JPKribs:
### Summary
Creates an 'Identify' View to allow users to identify an item against Jellyfin Server's remote sources. This searches and matches for an item based on the Title, Original Title or Year. See the video for more detail!

### Video

Fixing a mismatch where Jellyfin thought `Star Wars` was `Phineas & Ferb: Star Wars`. This video is correcting that match back to `Star Wars'.

https://github.com/user-attachments/assets/990af6dd-a480-482c-9806-cc75aa4c685c


JPKribs:
Okay, I officially have too many open PR's to keep up with all of them so this one is on pause for a bit. That being said, here is the latest version of this:

https://github.com/user-attachments/assets/3e72c2e0-dcfb-4b6d-94e3-92e3db603eba

It's all completely functional but I think there are some Views that look weird. The whole abstraction for allowing multiple search parameters on a single view is kind of janky. I also think how I am handling the media types via coordinator is wrong:

```
    @ViewBuilder
    func makeEditIdentity(item: BaseItemDto)  -> some View {
        switch item.type {
        case .boxSet:
            IdentifyItemView<BoxSetInfo>(viewModel: BoxSetInfoViewModel(item: item))
        case .movie:
            IdentifyItemView<MovieInfo>(viewModel: MovieInfoViewModel(item: item))
        case .person:
            IdentifyItemView<PersonLookupInfo>(viewModel: PersonInfoViewModel(item: item))
        case .series:
            IdentifyItemView<SeriesInfo>(viewModel: SeriesInfoViewModel(item: item))
        default:
            ErrorView(error: JellyfinAPIError("Invalid media type"))
        }
    }
```

Swiftfin/Views/ItemEditorView/IdentifyItemView/IdentifyItemView.swift
```diff
@@ -0,0 +1,340 @@
...
+
+    // MARK: - Error State
+
+    @State
+    private var error: Error?
+
+    // MARK: - Potential States
+
+    @State
+    var boxSetInfo = BoxSetInfo()
```
LePips, line 40: As much as its annoying that there are these objects that are the exact same structure for each item type, we don't need to have variables for each one.

The fields themselves should have their own source and then the view models receive the strings and construct their own info objects.

LePips (changes requested):

Swiftfin/Views/ItemEditorView/IdentifyItemView/IdentifyItemView.swift
JPKribs: Sounds good! That doesn't seem too bad. I was having trouble wrapping my head around all this. Is that right from the coordinator or should I put the switch on the View instead?

LePips: Outside of making the objects, there shouldn't really be any `switch`ing in `IdentifyItemView` as the objects are subclassed. If the base search method instead took as parameters the fields being filled out then they would instead construct the necessary `_Info` objects for their eventual request.

JPKribs: I took a swing at this. Since I have the `BaseItemDto` in the `ViewModel`, I changed this to take the generic title, original title, & year then switch inside of the `ViewModel` based on the type. This lets me use the view and the viewModel without specifying type.

I *think* this is better? Less code and it works as expected. Open to feedback if you think there is a better route!

JPKribs:
I think this is ready to go! There are three items that could be controversial:

1. The Selection Modal that pops up in the Identify View is a `.sheet`. I can change this to a full Stinsen > Router but I thought this was easier solution opposed to passing the `viewModel` between `views`. I'd be happy to move that over but I thought I would submit this with the working version.

2. Do we have a more standardized `Cancel Button` than this:

Simulator Screenshot - iPad Pro 13-inch (M4) - 2024-12-19 at 21 52 23

3. I created a very simple `struct` for storing my search items. This way expanding it in the future is less intrusive and allows me to validate all search terms with a single comparison. I wasn't sure if there was a more localized way to do this:

```
struct ItemIdentifySearch: Equatable {
    var name: String?
    var originalTitle: String?
    var year: Int?
}
```

Other than that, I believe this is ready to go! Please let me know if there are any changes or feedback that should be taken into account!

LePips (approved):
Apologies for not responding earlier for some feedback. Instead of just commenting and waiting I've just updated with some of my changes, most notably by renaming and some redesign using other existing views.

Per the implementation of a unified cancel view like the video player, that view and flow is changed in my work in #1203. Something _like_ that should be in `DelayedProgressView`, but even then that view probably isn't the best here.

There are many changes that this flow should receive though:

- Change poster styling depending on media type. This will be more important when music is implemented.
- `ItemView` and all subsidiary views are way old, before I knew what I know now about about proper ways to update views in SwiftUI. They will need some refactoring to handle item updates, which is why I changed a bit in `ItemView.AboutView` to accommodate item updating but even that should be changed.
- We need to figure out cache invalidation for the previous item images.

What's here is good for now!

JPKribs:
> Apologies for not responding earlier for some feedback. Instead of just commenting and waiting I've just updated with some of my changes, most notably by renaming and some redesign using other existing views.
> 
> Per the implementation of a unified cancel view like the video player, that view and flow is changed in my work in #1203. Something _like_ that should be in `DelayedProgressView`, but even then that view probably isn't the best here.
> 
> There are many changes that this flow should receive though:
> 
>     * Change poster styling depending on media type. This will be more important when music is implemented.
> 
>     * `ItemView` and all subsidiary views are way old, before I knew what I know now about about proper ways to update views in SwiftUI. They will need some refactoring to handle item updates, which is why I changed a bit in `ItemView.AboutView` to accommodate item updating but even that should be changed.
> 
>     * We need to figure out cache invalidation for the previous item images.
> 
> 
> What's here is good for now!

This looks great! Thank you again for helping get this over the finish line. These changes all make sense and I'm very happy with the result!

Issue #518, opened 2022-08-08
Clean up Localizations


LePips:
The base file `Translations/en.lproj/Localizable.strings` needs to be cleaned up as:

1 - there are many unused strings
2 - strings have just always been appended to the end of the file. This has lead to merge conflicts and has no organization.

So we need to find the unused strings, remove them, and then sort the remaining strings by some organization format. I was thinking sections by purpose and strings are alphabetized within each section:

```
/* General */
"close" = "Close";
...

/* QuickConnect */
"quickConnect" = "QuickConnect"
"quickConnectStep1" = "..."
...
```

As far as I know, re-organizing the strings does not cause Weblate to break.

JPKribs:
@LePips I can go through and do this. I was starting to work on this for the PR I was working on but it should be pretty straight forward for me to apply this to the other labels as well. I can make a PR for this, but does this layout make sense:

<details>
<summary>In-Progress Labels</summary>

![Screenshot 2024-05-25 at 3 49 19 PM](https://github.com/jellyfin/Swiftfin/assets/37599297/c5522cd6-9de3-41e0-a6b7-82e6120d1093)

</details>

At this time, my testing is that I do a search for the label before I organize it. If not found, I comment out labels and try a clean build for both tvOS and iOS to see if the label didn't show up in the original search on the project.

JPKribs:
As a note, the indentation does not interfere with the build. Let me know if this looks too busy with the indentation. This is what looked good to me but I'm game for whatever.

LePips:
Thank you for wanting to take a look at this. I like the comments for each string, but let's not indent anything.

The main issue that has prevented me from working on this over the years is how this interacts with Weblate/other languages and whether it will either move all of the strings or just nuke them.

For determining which strings are unused, you can use one of the following which detects unused code and you can just see what it outputs for the `Strings.swift` file:

- https://github.com/PaulTaykalo/swift-scripts
- https://github.com/peripheryapp/periphery

JPKribs:
> Thank you for wanting to take a look at this. I like the comments for each string, but let's not indent anything.
> 
> The main issue that has prevented me from working on this over the years is determining how this interacts with Weblate/other languages and whether it will either move all of the strings or just nuke them. If it does the latter, then we cannot do this. If we have to manually perform the same movements in all files, then we'll probably need a script.
> 
> For determining which strings are unused, you can use one of the following which detects unused code and you can just see what it outputs for the `Strings.swift` file:
> 
[...]

PR #1362, opened 2024-12-13
[iOS & tvOS] Unused Localization Cleanup

JPKribs:
### Summary
Removes Localizations with no usages in code. Alphabetizes localizations and cleans up comments. This is a must more intrusive part of https://github.com/jellyfin/Swiftfin/issues/518. This may change our webplates but Swiftfin itself will be unchanged.

Some localizations may be worth retaining for future usage. An example of this is `WIP` is currently unused but if we ever need something to be tagged as `WIP` in the future it will be good to keep this localized.

I have validated that both tvOS and iOS still appropriately build with these removed. The next step in this process, I have a remove with all `"Strings"` that exist in Swiftin. Most of them are fine like `" / "` but I'll submit another PR with the unlocalized strings localized.

The following localizations will be removed by this PR. Let me know if any of these should be retained and I can place them back in:

```
/// Who's watching?
"WhosWatching" = "Who's watching?";

/// Active Devices
"activeDevices" = "Active Devices";

/// Administration
"administration" = "Administration";

/// All Genres
"allGenres" = "All Genres";

/// Allowed tags
"allowedTags" = "Allowed tags";

/// Only show media to this user with at least one of the specified tags.
"allowedTagsDescription" = "Only show media to this user with at least one of the specified tags.";

/// Apply
"apply" = "Apply";

/// Audio & Captions
"audioAndCaptions" = "Audio & Captions";

/// Audio Track
"audioTrack" = "Audio Track";

/// Blocked tags
"blockedTags" = "Blocked tags";

/// Hide media with at least one of the specified tags.
"blockedTagsDescription" = "Hide media with at least one of the specified tags.";

/// Cancelled
"canceled" = "Cancelled";

/// CAST
"cast" = "CAST";

/// Change Server
"changeServer" = "Change Server";

/// Changes not saved
"changesNotSaved" = "Changes not saved";

/// Cinematic Views
"cinematicViews" = "Cinematic Views";

/// Closed Captions
"closedCaptions" = "Closed Captions";

/// Coming soon
"comingSoon" = "Coming soon";

/// Confirm Close
"confirmClose" = "Confirm Close";

/// Connect Manually
"connectManually" = "Connect Manually";

/// Connect to Jellyfin
"connectToJellyfin" = "Connect to Jellyfin";

/// Connect to a Jellyfin server
"connectToJellyfinServer" = "Connect to a Jellyfin server";

/// Continue Watching
"continueWatching" = "Continue Watching";

/// Current Position
[...]

PR #1372, opened 2024-12-20
[Meta] Automatic String Organization

JPKribs:
### Summary

Adds a step at the beginning of the build process that alphabetizes localizations.

The changes localizations in this commit were re-alphabetized by this new Swift Script. I think Swift and Python have different Alphabetization methods but if we're just using this by default, the Swift method should be used.

Finalizes https://github.com/jellyfin/Swiftfin/pull/1362 and hopefully resolves https://github.com/jellyfin/Swiftfin/issues/518

This script only comments & alphabetizes strings. This step does not validate whether a String is used in the project. I can add that step in as well if we want it but I do think there are some unused strings that are good to keep. A good example is `Experimental` since we're not using it now but may want to use it in the future.

If we want to remove unused strings, I can add it via another script if we like this script process!


Scripts/AlphabetizeStrings.swift
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// This regular expression pattern matches lines of the format:
+// "Key" = "Value";
+let regex = #/^\"(?<key>[^\"]+)\"\s*=\s*\"(?<value>[^\"]+)\";/#
+
+// Process each .strings file found.
+for fileURL in files {
+    // Attempt to read the file content.
+    guard let content = try? String(contentsOf: fileURL, encoding: .utf8) else {
```
LePips, line 43: This can be an issue since it only picks up the English file. The file is encoded differently than all of the other language files and frankly I haven't investigated why it needs to be, which was done in https://github.com/jellyfin/Swiftfin/pull/934. Regardless, we shouldn't do a loop and instead explicitly get the English file if we only need to sort that single file.

LePips (changes requested):

Scripts/AlphabetizeStrings.swift
JPKribs: Gotcha! I just made this change. So the top half is pretty much all replaced by just:

`var fileURL = URL(fileURLWithPath: "./Translations/en.lproj/Localizable.strings")`

Any issue specifying the file like that?

JPKribs: Also, do we want logic to remove unused localizations as part of the build? I'd like to think we can just avoid creating unused strings in the future but it would be nice for it to just be automated in case it does happen. On the other hand, it'd get annoying to lose localizations that are temporarily used that we might want later like `Replace`.

If we do want to purge unused strings, should that be a separate `PurgeStrings.swift` file or make one `OrganizeLocalizations.swift`?

LePips (approved):
The automatic removal during the build phase isn't correct because it removes strings before they can even be used. I think I ideally want a job on GitHub that can detect unused strings and comment on PRs.

However, having the script around is still useful to run.

For the scripts they don't need to run every time and only when `en.lproj/Localizable.strings` changes. However, even though I've followed the instructions at the section below the script still runs on every build. It's fine for now since these take < 0.5s on my machine.

- https://developer.apple.com/documentation/xcode/improving-the-speed-of-incremental-builds#Declare-inputs-and-outputs-for-custom-scripts-and-build-rules

JPKribs:
> The automatic removal during the build phase isn't correct because it removes strings before they can even be used. I think I ideally want a job on GitHub that can detect unused strings and comment on PRs.

Sounds good! I probably wont take a crack at that one (at least for now) since Git is still a pretty weak skill for me. I'm fine with just the Alphabetizing for now since that should handle the primary duplicate entries + organization issues. Unused localized isn't *that bad* IMO since there are some items worth keep localized even if we aren't using them: (Eg, `WIP`, `Experimental`, etc.).

I'll let you make the call for whether or not this closes out https://github.com/jellyfin/Swiftfin/issues/518!

PR #1373, opened 2024-12-21
[iOS] Select all Users When Editing

chickdan:
Simple change, adds a button when editing users to select all users for faster management and moves the buttons to the toolbars.

### DIFF
--- a/Shared/Extensions/Set.swift
+++ b/Shared/Extensions/Set.swift
@@ -17,4 +17,10 @@ extension Set {
             insert(value)
         }
     }
+
+    mutating func insert(contentsOf elements: [Element]) {
+        for element in elements {
+            insert(element)
+        }
+    }
 }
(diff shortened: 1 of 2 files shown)


LePips:
Over the development of the dashboard we've established a pattern on iOS where the `Select/Remove All`, `Cancel`, and even `Delete` are in the toolbars instead. We should be moving to that here too.

chickdan:
> Over the development of the dashboard we've established a pattern on iOS where the `Select/Remove All`, `Cancel`, and even `Delete` are in the toolbars instead. We should be moving to that here too.

Gotcha, I can make that change. Seems like it will be a bit crowded with all three buttons up there, should `Delete` stay at the bottom?

LePips:
I just added `Remove All` as well.

LePips (approved):

chickdan:
Whoops, sorry for leaving the old button behind.

PR #1377, opened 2024-12-28
[iOS] Admin Dashboard - User Access Tags

JPKribs:
### Summary

This PR starts the works for managing Access Tags. On 10.8, we only have the ability to set `BlockedTags` which removes the ability for a user to see media with that tag. In 10.10, there will be `AllowedTags` which block ALL media for that user unless the media has one of the `AllowedTags`.

The view I am using should work for both but I have the `AllowedTags` logic commented out for us to enable when it's available. I reused the Trie from the Tags/Genres/People/Studios editing: https://github.com/jellyfin/Swiftfin/pull/1336. This is almost a copy/paste of `EditItemElementView` but I am turning all of the tags into `[String:Bool]` where I am tracking both the tag and whether it allows or blocks access. This allows us to do all of this on a single view instead of breaking it out. Since the object we are editing are just tags, having a secondary `Bool` doesn't add much clutter.

Since this item is incomplete before 10.10, I'm fine sitting on this one for a bit but the change to get this working for 10.10 should be very easy and this should be feature complete for 10.8.

### Screenshots 

<details>
<summary>New Button</summary>
New Button
</details>

<details>
<summary>Empty View</summary>
Empty View
</details>

<details>
<summary>New Entry</summary>
New Entry
</details>

<details>
<summary>LearnMore</summary>
LearnMore
</details>

<details>
<summary>Tag Searching</summary>
Tag Searching
</details>

<details>
<summary>Tag Validation</summary>
Tag Validation
</details>

<details>
<summary>Populated View</summary>
Populated View
</details>

### DIFF
--- a/PreferencesView/Sources/PreferencesView/Box.swift
+++ b/PreferencesView/Sources/PreferencesView/Box.swift
@@ -3,7 +3,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, you can obtain one at https://mozilla.org/MPL/2.0/.
 //
-// Copyright (c) 2024 Jellyfin & Jellyfin Contributors
+// Copyright (c) 2025 Jellyfin & Jellyfin Contributors
 //
 
 class Box {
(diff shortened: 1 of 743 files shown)


LePips:
The largest design change I made was to remove the label for what kind of tag it is. Instead, once `allowed` tags are implemented, we can then use a `DisclosureGroup` to group the `blocked/allowed` tags together and make the sections collapsable. Then on iOS 17 we can use a plain `Section` for a more desirable implementation.

- https://developer.apple.com/documentation/swiftui/section/init(_:isexpanded:content:)-5hjcw

LePips (approved):

PR #1381, opened 2025-01-02
[Meta] 2025 Disclaimer

JPKribs:
### Summary

I just pulled in Main then built. This should update all of the Disclaimer '2024' to '2025. I'm making this a stand alone PR so we don't have a random PR with 736 file changes. Instead, this one has 736 file changes.

For example, https://github.com/jellyfin/Swiftfin/pull/1377 has "doubled" in size to include the 2025 changes.

### DIFF
--- a/PreferencesView/Sources/PreferencesView/Box.swift
+++ b/PreferencesView/Sources/PreferencesView/Box.swift
@@ -3,7 +3,7 @@
 // License, v2.0. If a copy of the MPL was not distributed with this
 // file, you can obtain one at https://mozilla.org/MPL/2.0/.
 //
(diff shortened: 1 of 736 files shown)


JPKribs:
@LePips No changes made here. Just ran the build on Main to update all of the 2024 to 2025. Goal of the PR is to have the huge 736 file change here instead of cluttering another PR. 

The only other change in here looks like Strings file was changed without a build before the last commit. So the discrepancy is just these strings being removed since there is no longer a localization:

Screenshot 2025-01-02 at 12 00 15

chickdan (approved):

LePips (approved):

PR #1383, opened 2025-01-02
[tvOS] Add pin prompt to sign-in screen

chickdan:
Most of this is brought over from iOS with little changed. I removed `LocalAuthentication` since the framework does not support tvOS so the security options right now are `none` or `pincode`.

One oddity I noticed that seems specific to tvOS is that TextField with placeholder text will add an additional button to alerts that have the same text as the placeholder. Clicking the button opens the keyboard but it's a bit odd so I'm not sure if that is desired to be kept in there.
Example:

Screenshot 2025-01-08 at 12 33 24 AM

I also left in the password reset and reset settings buttons (but commented out) for later use if it's desired to include that functionality.

<details><summary>Set pin</summary>
<p>

https://github.com/user-attachments/assets/0fbb4752-645f-4df5-a128-8c15bd64f767

</p>
</details> 

<details><summary>Attempt Login</summary>
<p>

https://github.com/user-attachments/assets/58c72a40-42cc-4504-b149-2b43e453b969

</p>
</details>


JPKribs:
I am so sorry, @chickdan! I didn't realize those would just direct commit I thought those would send you a PR to your local branch! You are totally welcome to revert those!

Here is what I wanted to send as changes. I'm sure @LePips will have a much better insight but just from my prior tvOS PRs:

- Move UserProfileSettings to their own Coordinator
> I did this just to help clean it out a little. It's not a requirement by any stretch but as we add other UserProfileSettings it would help keep things from getting cluttered.

- Make Views Modal to better reflect existing items
> The other Settings & CustomizeSettings use Modal instead of Push. Just mirroring those.

- Fix CustomizeSettingsCoordinator (This is on me!)
> This was just an issue I accidentally where I was treating this as BasicNavigationViewCoordinator when it should have been CustomizeSettingsViewCoordinator. Changing this resolves a routing issue I created but was not aware of until now.

- Change PINs to use SecureField
> Just a preference. I'd be interested to hear everyone else's take on this!

- Move all Settings View to use SplitFormWindowView to mirror existing Settings
> All Settings are SplitFormWindowView with an icon on the left. This just moves those over to this format to be a little more cohesive.

- Use user profile image for SplitFormWindowView Icon
> iOS we have the profile image at the top. I used SplitFormWindowView to put it on the side. IMO, it looks good but open to feedback!

- Change Profile Security to use LearnMoreModal
> We've moved a few longer descriptions over to this layout. With this change, I'm not sure the text lines up exactly now but it de-clutters the list on the right now.

- Hide Device Authentication option
> Either this setting is not possible or maybe we are able to assign this to a tvOS Profile in the future? For now, I think Pin or None are our only options so this makes sense as an inline toggle instead of another view selection.

- Use suggestion from https://forums.developer.apple.com/forums/thread/739545
> In trying to find a good article on this, I found this! It looks like this resolves the issue.

- Tag Alert > TextFields with TODO so we can check this on tvOS 18
> In the future, placeholder text would be nice so I left TODOs for us to re-enable this on tvOS and hopefully it's resolved!

Sorry again for the commit over! Please let me know if you have any questions!

LePips (approved):
Thank you for implementing this @chickdan and thank you for the changes @JPKribs. tvOS "design" is still pretty dynamic as we figure out better UX interactions on the platform. For example, we may want to change the `Save` button from the navigation bar into a button in the list, similar to `Change User`.

PR #1388, opened 2025-01-09
[WIP] tvOS - Episode Item View Backdrop Substitute

JPKribs:
### Summary

The issue I am experiencing is that Episode images tend to be lower quality than other images in Jellyfin. As a result, episode views get blown out when used as the backdrop:

Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-01-08 at 12 03 31

My WIP solution is to use the Series Backdrop for Episodes but also include the Episode Primary image in a more standard size for it's resolution. Below is my current version of this. I don't love it since I feel like it cramps the description/title. I'd love some feedback on ideas for how we'd want this to look!

Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-01-08 at 16 56 29

<details>
<summary> Other Attempts </summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-01-08 at 11 43 57

Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-01-08 at 12 48 29

</details>

### DIFF
--- a/Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto+Poster.swift
+++ b/Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto+Poster.swift
@@ -20,7 +20,12 @@ extension BaseItemDto: Poster {
         case .episode:
             seasonEpisodeLabel
         case .video:
-            extraType?.displayTitle
+            /// Don't show any subtitle if it's unknown
+            if let extraType, extraType != .unknown {
+                extraType.displayTitle
+            } else {
+                nil
+            }
         default:
             nil
         }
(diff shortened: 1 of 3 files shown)


LePips:
I have seen this long happen with episodes but they also get their images from metadata providers. It kind of makes sense that there probably aren't that many good-quality images for episodes compared to series or movies.

I haven't entirely dismissed the idea of an alternate view for video media that could be used, but that would be larger change than just placing the episode image elsewhere.

JPKribs:
> I haven't entirely dismissed the idea of an alternate view for video media that could be used, but that would be larger change than just placing the episode image elsewhere.

Sounds good! I was really hoping this was going to be a straight forward one but I got caught in design hell. I get what you mean about an alternative view since I kept having the issue that only episodes would look completely different than any other item...

I'm good to either close this out for now, or, if you think it's valuable I can add a toggle for tvOS and iPadOS for 'Use Series Backdrop' but I worry the customization section is getting a bit overwhelming? I dunno, your call! This really isn't my domain so I might be losing the plot a little here haha

LePips:
I'm good to close this out for now. I'll do a write up soon in Discussions about thoughts I have for navigation on tvOS specifically.

PR #1401, opened 2025-01-20
[Meta] Re-Organize Admin Dashboard

JPKribs:
### Summary

Now that we're all wrapped up on Admin Dashboard (until 10.10), I wanted to group this a little better to be less chaotic. Groups the Admin Dashboard items into folders by function. Now looks like this:

New:
Screenshot 2025-01-20 at 14 59 38

Previous:
Screenshot 2025-01-20 at 15 03 06


LePips (approved):
We're fine since we squash PR commits but I would recommend not re-using branches.

PR #1402, opened 2025-01-21
Refresh Server Info

### LINKED ISSUE #1389: Server name change not picked up (Tv, iPad)
### Describe the bug

I did my jellyfin server installation as a podman container.  This resulted in the hostname and jellyfin server name being a hexadecimal string.

I then used swiftfin on both iPad and Apple TV where the hexadecimal string was picked up as the server name.

I later changed the jellyfin server name to what to something human and nice.

Despite Tv and iPad restarts, swiftfin still shows the server name as the hex string.  I believe it should pick up and use the updated server name,

### Application version

1.2(4) (ipad)

App Store

### Device information

iPad Pro (7th gen), Apple TV 4k (2nd gen)

### OS version

iPadOS 18.2, TvOS 18.2

### Jellyfin server version

10.10.3

chickdan:
Working to resolve #1389, looking through where server information is set this is the partial solution I came up with. It's not ideal as the user won't see the name change until after relaunching the app so pointers on where to put this would be appreciated!

There is also a `TODO` in ServerCheckViewModel that says "Also server stuff", what needs to be done to resolve the todo? I can fill it out since I'm already touching that file.

### DIFF
--- a/Shared/SwiftfinStore/SwiftfinStore+ServerState.swift
+++ b/Shared/SwiftfinStore/SwiftfinStore+ServerState.swift
@@ -78,4 +78,21 @@ extension ServerState {
         let request = Paths.getSplashscreen()
         return ImageSource(url: client.fullURL(with: request))
     }
+
+    func updateServerInfo() async throws {
+        guard let server = try? SwiftfinStore.dataStack.fetchOne(
+            From<ServerModel>()
+        ) else { return }
+
+        let publicInfo = try await getPublicSystemInfo()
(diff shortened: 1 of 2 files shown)


LePips (approved):
This seems okay for now. I've only pushed a small workaround to reset the `UserSession` which will make future view models use the updated information so we don't need to wait for the next restart.

But `UserSession` needs more architectural consideration so that it can update its own values and the UI picks up on it.

PR #1403, opened 2025-01-21
[tvOS] Login Flow Cleanup - Second Pass

JPKribs:
# Login Flow UI Updates

## Initial Login Screen
<details>
<summary>Connect to Server - Remove Duplicate Icons</summary>

**New Version:**
New Initial Screen

**Previous Version:**
Old Initial Screen
</details>

## Server Management
<details>
<summary>Server Bar - Ensure Menus are same sizing</summary>

**New Version:**
New Server Bar

**Previous Version:**
Old Server Bar
</details>

## Server Configuration
<details>
<summary>Edit Server View - Clean up menu buttons</summary>

New Edit Server View
</details>

## Additional UI Elements
<details>
<summary>User Sign-In - Add Background Image for Server</summary>

Login View Background
</details>

---

## Outstanding Issue
<details>
<summary>tvOS User Sign-In Crash</summary>

This could be an emulator issue but I see this issue where the first time I login with a new user, Swiftfin crashes. Restarting, the user still exists but it's not great to restart after each login.

Screenshot 2025-01-21 at 07 52 19


Swiftfin tvOS/Views/SelectUserView/Components/UserGridButton.swift
```diff
@@ -71,7 +71,13 @@ extension SelectUserView {
             )
             .aspectRatio(1, contentMode: .fill)
             .overlay {
-                if isEditing {
+                if isEditing && isSelected {
+                    RelativeSystemImageView(systemName: "trash.circle.fill", ratio: 1)
```
LePips, line 75: I would prefer if this was similar to iOS with a smaller checkmark icon in the `bottomTrailing` position.

LePips (changes requested):
As I was typing a comment for [this change](https://github.com/jellyfin/Swiftfin/pull/1403/files#diff-21c9dad0b010a44f596130cb19fceb0151b5b104084312d6d9f38e05fd9e3da3R244) I thought instead to just implement the same `Menu/Button` conditioning for server selection when it is `all`.

Using the user selection a bit more, I think we may want to move the `Add User` button from an inline item with the users and instead have it be a button on the left/leading side of the server selection to make the stack symmetrical. Maybe also for iOS to move it to the top trailing menu. We can take a look at that later.

JPKribs:
> Using the user selection a bit more, I think we may want to move the `Add User` button from an inline item with the users and instead have it be a button on the left/leading side of the server selection to make the stack symmetrical. Maybe also for iOS to move it to the top trailing menu. We can take a look at that later.

I can do that! One item of note, on user sign in on tvOS, disabled buttons square off. IMO, it's a little wonky looking but not an issue per-say. Wanted to run it by you while I was touching up login flow:
Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-01-21 at 12 15 52

Swiftfin tvOS/Views/SelectUserView/Components/UserGridButton.swift
JPKribs: I have a version of this. I am doing the hoverEffect on both the checkbox AND the icon since this is what this looks like without it:

https://github.com/user-attachments/assets/9aa8920e-18a5-4ca9-8d93-6e00295f7d5a

JPKribs:
> I think we may want to move the `Add User` button from an inline item with the users and instead have it be a button on the left/leading side of the server selection

This is my implementation of this. Instead of disabling/hiding when there is no selected server I turned this into a menu. 1) It's hard to mirror Button & Menu styling to this avoids that that headache. 2) you can now always select the server you want to add a user to regardless of if you are in 'All Servers' or have a selected server.

Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-01-21 at 13 08 10

Swiftfin tvOS/Views/SelectUserView/Components/UserGridButton.swift
LePips: Hm, I didn't realize that the `buttonBorderShape(.circle)` would cause issues with this overlay. Instead of `bottomTrailing`, I'm fine with just setting in the `.bottom` for now. It won't need a separate `hoverEffect`.

JPKribs:
Okay! Take 2. I've made some additional changes. 

- I updated the `ListRowButton` to have a `.destructive` that mirrors iOS.
- I updated the `ListRowButton` to retain its rounded corners even when disabled.
- I changed changed the add user button to be a menu to allow selecting the server you want to add a user from.
- There is only an add user, user card if there are no users. This is mostly to be a bigger target to let the user know where to add profiles.
- Delete button now uses the same `ListRowButton` `.destructive` to mirror.

<details>
<summary>Delete User Buttons</summary>

Selected User - Non-Focused Delete:
Selected User   Non-Highlighted Delete

Selected User - Focused Delete:
Selected User   Highlighted Delete

Disabled Delete:
Disabled Delete

</details>

<details>
<summary>Add User Buttons</summary>

This is the new layout. See - no 'Add User' button in the grid:
Add User - Layout

Adding a user. See - Choosing the server for adding a user:
Add User - Some Existing User(s)

Adding a user. See - The Add User Grid button exists
Add User - No Existing Users

</details>


<details>
<summary>New Found Bug</summary>
Old Buttons were cards inside of forms... I actually did this a while ago so that's on me:
Screenshot 2025-01-21 at 15 03 16

The new buttons fill this out appropriately:
Screenshot 2025-01-21 at 15 05 46

<details>

---

<details>
<summary>New Found Bug</summary>

Single User Bug - Only occurs if you only have a single account. I believe this precedes my changes but before there was almost always 2 users since 1 user + the add user "user":
Weird Bug

</details>

LePips (changes requested):
Let's leave out the add user button changes as that change would also consist of more cleanup within `SelectUserView`.

JPKribs:
> Let's leave out the add user button changes as that change would also consist of more cleanup within `SelectUserView`.

Sounds good. Those changes should now be reverted!

Swiftfin tvOS/Views/ServerDetailView.swift
```diff
@@ -75,14 +75,17 @@ struct EditServerView: View {
                                 .foregroundColor(.secondary)
                         }
                     }
+                    .listRowBackground(Color.clear)
+                    .listRowInsets(EdgeInsets(top: 0, leading: 0, bottom: 0, trailing: 0))
```
LePips, line 79:
```suggestion
                    .listRowInsets(.zero)
```

Swiftfin tvOS/Views/ServerDetailView.swift
```diff
@@ -75,14 +75,17 @@ struct EditServerView: View {
...
 
                 if isEditing {
                     Section {
-                        ListRowButton(L10n.delete) {
+                        ListRowButton(L10n.delete, role: .destructive) {
                             isPresentingConfirmDeletion = true
                         }
-                        .foregroundStyle(.primary, .red.opacity(0.5))
+                        .listRowBackground(Color.clear)
+                        .listRowInsets(EdgeInsets(top: 0, leading: 0, bottom: 0, trailing: 0))
```
LePips, line 88:
```suggestion
                        .listRowInsets(.zero)
```

Translations/en.lproj/Localizable.strings
```diff
@@ -1618,6 +1618,9 @@
 /// Select Image
 "selectImage" = "Select Image";
 
+/// Select Server
+"selectServer" = "Select Server";
```
LePips, line 1622:
```suggestion
"selectServer" = "Select server";
```

LePips (changes requested):

LePips (approved):
Looks great!

PR #1404, opened 2025-01-21
[iPadOS] Fixes #1368 Shrunken posters on iPad home screen

### LINKED ISSUE #1368: [iPadOS] Home Screen - Shrunken Posters
### Describe the bug

The posters on the home screen are all very small with the exception of Continue Watching:

Image

It looks like `Continue Watching` uses `CollectionHStack` while the other Home Views use `PosterHStack`. I could be totally wrong but it seems like this could be related to https://github.com/jellyfin/Swiftfin/issues/1339.

### Application version

1.3 (3)

TestFlight

### Device information

iPad, iPad Emulator

### OS version

18.1

### Jellyfin server version

10.10.3

cooksey:
It looks like the iPad code got switched to use the wrong initializer. This commit reverts a change from commit `e856303`

### DIFF
--- a/Swiftfin/Components/PosterHStack.swift
+++ b/Swiftfin/Components/PosterHStack.swift
@@ -26,7 +26,7 @@ struct PosterHStack<Element: Poster & Identifiable, Data: Collection>: View wher
     private var padHStack: some View {
         CollectionHStack(
             uniqueElements: data,
-            columns: type == .portrait ? 140 : 220
+            minWidth: type == .portrait ? 140 : 220
         ) { item in
             PosterButton(
                 item: item,


LePips (approved):
🤦‍♂️ Thank you. I've spent a bit too much time in my package thinking this was most likely a math issue instead.

JPKribs:
This is a selfish ask, but could we get a new TestFlight with this PR included? Technically, it would also have the finalized Admin Dashboard / Item Editing too but you don't realize how many time we use the PosterHStack until they're all shrunken lol.

No pressure/rush at all! Just thought I'd ask!

PR #1407, opened 2025-01-23
[WIP] [tvOS] Letter Picker / Filters

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/333

Takes the same logic from the LetterPickerBar on iOS and implements it to tvOS. Spiritual successor to: https://github.com/jellyfin/Swiftfin/pull/1177

This PR is functional and 'ready' but there is still a lot of discussion that needs to occur before tvOS filters are implemented. This is a precursor to that conversation and, hopefully, somewhat reusable if we end up going another direction.

This PR stands as more of a placeholder for the logic opposed to what the final version of this will look like. I've rebased this on Main again so it should be up-to-date compared to my old version of this PR which was very far behind without a reasonable path to rebase.

I have a current working version of this that can be seen below:

<details>

<summary>Current Version</summary>

https://github.com/user-attachments/assets/0756a364-2d16-4a70-81ba-e82f87cc05d0

</details>

### Outstanding Issues

1. On Main, tvOS `PagingLIbraryView` jump their focus to the top of the page whenever a new page of data is loaded. Since the Letter Picker is the tallest item, this is now capturing the focus. I think this is more of an issue with the `PagingLIbraryView` and knocking out that TODO. I am not sure if this is a package issue where the Focused item just gets lost when new items are added in?

2. If you are hovering a letter that is above, below, or between posters, it cannot jump to them. Instead, you have to go back up to a letter that in even to a poster to move the cursor back to the posters. I will need to figure out the FocusGuide to use this appropriately but this should be doable.

### Outstanding Discussions

There is a whole conversation about tvOS UI I know LePips wants to have. This should provide some clarity on what we want this to look like so we aren't reworking the same view multiple times to no benefit. This PR's work should hopefully be re-usable when that conversation is had but this is likely *not* our ideal version of this. I personally have a poor eye for UX/UI so I want to leave plenty of room for news items/revisions. However, this PR introduces the FULL `FilterViewModel` so any filter logic we need to do should be feasible.

PR #1213, opened 2024-08-27
[tvOS] ListView + Grid Landscape Poster Padding

JPKribs:
This is almost an exact copy/paste from iOS to tvOS. I added the Focus elements from the tvOS PosterButton to the LibraryRow elements in tvOS. Tested to work and the Cinematic Backdrop is unaffected.

While I was in here, I noticed that the landscape grid didn't have any item spacing, resulting in some overlap in grid + landscape. I added in spacing to mirror the grid + portrait.

Resolves: https://github.com/jellyfin/Swiftfin/issues/687

<details>
<summary>tvOS Current Implementation</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2024-08-27 at 12 31 37

</details>

<details>
<summary>tvOS Implementation (1 Column, 2 Columns, 3 Columns)</summary>

321 ListViews

</details>

<details>
<summary>tvOS Settings</summary>

Settings ListViews

</details>

<details>
<summary>Landscape Posters Padding Changes</summary>

Fix Landscape Overlap

</details>

PR #1408, opened 2025-01-23
[tvOS] PagingLibraryView - Mirror iOS "Hot Reload" Functionality

JPKribs:
### Summary

This is almost just a complete copy / paste of the iOS functionality into tvOS. The only changes I made to this on this tvOS side is the `.onChange( before, after...` that tvOS wants. Additionally, tvOS only needs one layout while iOS needs Phone vs Tablet. 

Finally, I just added the settings to the `CustomizeSettingsView`. The Remember Sort / Layout *work* but since we don't have filters/sorting they aren't really usable. I know that is another conversation so I'm fine to disable those as well but I thought I would start with those enabled for now.

I have tested this and confirmed that this works as intended. I left a comment in place of where a LetterPicker *would* go. See the video for an example of this.

This is a continuation / resolution to this:

https://github.com/jellyfin/Swiftfin/pull/1213#pullrequestreview-2353000974

Additionally, this is a good starting point for https://github.com/jellyfin/Swiftfin/pull/1407 but without the UX/UI discussions that are needed to make that work appropriately.

The only other change that shows up in iOS as well is adding Remember Sort/Layout labels instead of plaintext.

*Edit: See additional examples of hot reloading the `PagingLibraryView`*

### Videos

<details>
<summary>Poster Change</summary>

https://github.com/user-attachments/assets/9374b92a-016a-4650-aeb1-8df75ae56132

</details>

<details>
<summary>Grid/List Change</summary>

https://github.com/user-attachments/assets/31b0b526-3d9f-48e0-bd40-5d777414fc8a

</details>

<details>
<summary>Column Change</summary>

https://github.com/user-attachments/assets/36c29fa9-3499-4cb6-be8c-3f401c87f5c9

</details>

### DIFF
--- a/Swiftfin tvOS/Views/PagingLibraryView/PagingLibraryView.swift
+++ b/Swiftfin tvOS/Views/PagingLibraryView/PagingLibraryView.swift
@@ -223,7 +223,7 @@ struct PagingLibraryView<Element: Poster & Identifiable>: View {
                 listItemView(item: item, posterType: posterType)
             }
         }
-        .onReachedBottomEdge(offset: .offset(300)) {
+        .onReachedBottomEdge(offset: .rows(3)) {
             viewModel.send(.getNextPage)
         }
         .proxy(collectionVGridProxy)
--- a/Swiftfin tvOS/Views/PagingLibraryView/PagingLibraryView.swift
+++ b/Swiftfin tvOS/Views/PagingLibraryView/PagingLibraryView.swift
@@ -338,23 +338,11 @@ struct PagingLibraryView<Element: Poster & Identifiable>: View {
          .onChange(of: posterType) {
              setCustomLayout()
          }
-         .onChange(of: defaultPosterType) {
-             guard !Defaults[.Customization.Library.rememberLayout] else { return }
-             setDefaultLayout()
-         }
          .onChange(of: displayType) {
              setCustomLayout()
          }
-         .onChange(of: defaultDisplayType) {
-             guard !Defaults[.Customization.Library.rememberLayout] else { return }
-             setDefaultLayout()
-         }
          .onChange(of: listColumnCount) {
              setCustomLayout()
-         }
-         .onChange(of: defaultListColumnCount) {
-             guard !Defaults[.Customization.Library.rememberLayout] else { return }
-             setDefaultLayout()
          }*/
     }
 
--- a/Swiftfin tvOS/Views/PagingLibraryView/PagingLibraryView.swift
+++ b/Swiftfin tvOS/Views/PagingLibraryView/PagingLibraryView.swift
@@ -12,32 +12,45 @@ import JellyfinAPI
 import SwiftUI
 
 // TODO: Figure out proper tab bar handling with the collection offset
-// TODO: list columns
-// TODO: list row view (LibraryRow)
 // TODO: fix paging for next item focusing the tab
 
 struct PagingLibraryView<Element: Poster & Identifiable>: View {
 
     @Default(.Customization.Library.cinematicBackground)
     private var cinematicBackground
-    @Default(.Customization.Library.posterType)
-    private var posterType
-    @Default(.Customization.Library.displayType)
-    private var viewType
-    @Default(.Customization.showPosterLabels)
-    private var showPosterLabels
+    @Default(.Customization.Library.enabledDrawerFilters)
+    private var enabledDrawerFilters
+    @Default(.Customization.Library.rememberLayout)
+    private var rememberLayout
+
+    @Default
+    private var defaultDisplayType: LibraryDisplayType
+    @Default
+    private var defaultListColumnCount: Int
+    @Default
+    private var defaultPosterType: PosterDisplayType
 
     @EnvironmentObject
     private var router: LibraryCoordinator<Element>.Router
 
     @State
     private var focusedItem: Element?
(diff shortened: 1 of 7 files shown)


Swiftfin tvOS/Views/PagingLibraryView/PagingLibraryView.swift
```diff
@@ -174,40 +226,144 @@ struct PagingLibraryView<Element: Poster & Identifiable>: View {
...
         }
+        .animation(.linear(duration: 0.1), value: viewModel.state)
         .ignoresSafeArea()
         .navigationTitle(viewModel.parent?.displayTitle ?? "")
         .onFirstAppear {
             if viewModel.state == .initial {
                 viewModel.send(.refresh)
             }
         }
+        .onChange(of: defaultPosterType) { _, newValue in
```
LePips, line 297: With these new closures, we can instead use the variant that doesn't pass in the values to the closure. The values that we observe for changes (`defaultPosterType` here), will be updated by the time the closure is called. So, this would mean instead of using `newValue` we would just use the variable throughout instead.

https://developer.apple.com/documentation/swiftui/view/onchange(of:initial:_:)-8wgw9

LePips (changes requested):

Swiftfin tvOS/Views/PagingLibraryView/PagingLibraryView.swift
JPKribs: Working on getting this changed but I have some weird interactions. Since we have a lot of `.onChange`, we're getting the standard "too large" warning. But, for some reason, the order I put the `onChange` matters?

Same code but different order. The closure-less ones at the bottom work but only for some of them. Moving these all to another var, like innerContent also works, in ANY order but I wasn't sure if there was something I just wasn't thinking of? I am going to keep plugging away at this but if you had any insight I'd be curious to know if there was something simple I was missing:

Screenshot 2025-01-23 at 21 43 59

Screenshot 2025-01-23 at 21 44 25

LePips: Ah, I see. Now we're pushing the checker a bit too much. Honestly I'm okay with these `onChange`s being on the `innerContent` along with a small comment it's for type checking reasons. At least in this view, all of those observed changes are meant for the `content` state so it works out.

LePips: With these new changes we hardly need to worry about the `newValue` passing around, so I think that the move to functions is great!

JPKribs: Sounds good! I had that one locked and ready to go so that version is all committed up now. Here are the only ones that still use the value in the closure. Without digging into it, they all look like they need the `newValue` but just wanted to make sure there wasn't a revision here we should be doing while we're looking at this.

If these look good, and unless there is anything else you see we should change, I've tested this on the latest commit and everything is working as expected!

```
        .onChange(of: rememberLayout) { _, newValue in
            let newDisplayType = newValue ? displayType : defaultDisplayType
            let newListColumnCount = newValue ? listColumnCount : defaultListColumnCount
            let newPosterType = newValue ? posterType : defaultPosterType

            layout = Self.makeLayout(
                posterType: newPosterType,
                viewType: newDisplayType,
                listColumnCount: newListColumnCount
            )
        }
        .onChange(of: viewModel.filterViewModel?.currentFilters) { _, newValue in
            guard let newValue, let id = viewModel.parent?.id else { return }

            if Defaults[.Customization.Library.rememberSort] {
                let newStoredFilters = StoredValues[.User.libraryFilters(parentID: id)]
                    .mutating(\.sortBy, with: newValue.sortBy)
                    .mutating(\.sortOrder, with: newValue.sortOrder)

                StoredValues[.User.libraryFilters(parentID: id)] = newStoredFilters
            }
        }
        .onChange(of: focusedItem) { _, newValue in
            guard let newValue else {
                withAnimation {
                    presentBackground = false
                }
                return
            }

            cinematicBackgroundViewModel.select(item: newValue)

            if !presentBackground {
                withAnimation {
                    presentBackground = true
                }
            }
        }
```

LePips: For `rememberLayout` and `focusedItem`, those can still be replaced by just using the variables instead of `newValue`. The filters change looks like a good example in the future where we would probably want to use the 2 parameter closure, or at least manually create the `newValue` based on the value we are observing.

LePips (approved):
Looks great!

Swiftfin tvOS/Views/PagingLibraryView/PagingLibraryView.swift
JPKribs: Sounds good. All of these are now value-less closures with the exception of viewModel.filterViewModel?.currentFilters. 

I moved defaults to under the `body` so the only part that's under innerContent is:

Screenshot 2025-01-24 at 14 10 15

While I'm doing this, should I be mirroring this in iOS or should I keep this PR tvOS Only?

LePips: Let's keep this tvOS for now. I'm confident that we will make these changes sometime in the future however.

PR #1409, opened 2025-01-25
[tvOS] ItemTypeLibraryViewModel - Implement FilterViewModel

JPKribs:
### Summary

I'm trying to get in the habit of doing more manageable / single item PRs. This one is the last step for https://github.com/jellyfin/Swiftfin/pull/1407 before we start needing to have UI discussions.

This PR changes the `ItemTypeLibraryViewModel` and `PagingViewModel`. The primary change is the `init` for `ItemTypeLibraryViewModel` that accepts `ItemFilterCollection` as a parameter. This replaces the existing `init` and results in `ItemTypeLibraryViewModel` having a `FilterViewModel`. This enables the ability to perform filtering on the generic Shows / Movies. 

I've tested this implementation works via https://github.com/jellyfin/Swiftfin/pull/1407. I am splitting this out so we can get all the groundwork for filtering done now so we can hit the ground running when we have a plan for what we want that to look like. Additionally, this lets us confirm this implementation is done in the cleanest fashion possible without the clutter of other filtering components at the same time.

### DIFF
--- a/Shared/Coordinators/MainCoordinator/tvOSMainTabCoordinator.swift
+++ b/Shared/Coordinators/MainCoordinator/tvOSMainTabCoordinator.swift
@@ -48,7 +48,10 @@ final class MainTabCoordinator: TabCoordinatable {
     }
 
     func makeTVShows() -> NavigationViewCoordinator<LibraryCoordinator<BaseItemDto>> {
-        let viewModel = ItemTypeLibraryViewModel(itemTypes: [.series])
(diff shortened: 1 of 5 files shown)


LePips (approved):
It's been a while since I've looked at these and realize there is potential for a good amount of cleanup, like removing `ItemTypeLibraryViewModel` entirely, and some API improvements. I've added those comments.

PR #1411, opened 2025-01-26
[iOS & tvOS] ItemLibraryViewModel Cleanup

JPKribs:
### Summary

The goal of this. PR is to touch up the `ItemLibraryViewModel` to accommodate some of the new items that we've added since it was created. @LePips left some notes in in there and `ItemTypeLIbraryViewModel` and mostly I am going to try and follow them.

Primary change was I removed `ItemTypeLIbraryViewModel` and moved `ItemType` to the the `ItemFilterCollection` instead. I've tested on tvOS and all libraries are correct including the the `ItemType` sections for TV Shows and Movies. Because of this, we are still able use a single ViewModel for both existing Library types we have on iOS and tvOS.

This PR is based on: https://github.com/jellyfin/Swiftfin/pull/1409#pullrequestreview-2574373538


JPKribs:
@LePips This all looks correct and is working as intended. My big question is whether this is the correct way to handle initing from a single itemType on the `ItemFilterCollection`:

```
    /// The default collection of filters by a BaseItemKind
    static func typeDefault(_ type: BaseItemKind) -> ItemFilterCollection {
        var filters: ItemFilterCollection = .init()

        filters.types = [type]

        return filters
    }
```

Last item I'd like to roll into this PR is from `ItemLibraryViewModel`. I was looking at adding this but might need a bit more context. Since `LibraryParent` is a protocol, do we want that there? Or on the enum for libraryType? 

```
    func itemParameters(for page: Int?) -> Paths.GetItemsByUserIDParameters {

        var libraryID: String?
        var personIDs: [String]?
        var studioIDs: [String]?
        var isRecursive: Bool? = true

        // TODO: this logic should be moved to a `LibraryParent` function
        //       that transforms a `GetItemsByUserIDParameters` struct, instead
        //       of having to do this case-by-case.

        if let libraryType = parent?.libraryType, let id = parent?.id {
            switch libraryType {
            case .collectionFolder, .userView:
                libraryID = id
            case .folder:
                libraryID = id
                isRecursive = nil
            case .person:
                personIDs = [id]
            case .studio:
                studioIDs = [id]
            default: ()
            }
        }

        var parameters = Paths.GetItemsByUserIDParameters()
        parameters.enableUserData = true
        parameters.fields = .MinimumFields
        parameters.isRecursive = isRecursive
        parameters.parentID = libraryID
        parameters.personIDs = personIDs
        parameters.studioIDs = studioIDs
        ...
```

Swiftfin tvOS/Views/PagingLibraryView/PagingLibraryView.swift
```diff
@@ -102,7 +102,7 @@ struct PagingLibraryView<Element: Poster & Identifiable>: View {
     private func select(item: BaseItemDto) {
         switch item.type {
         case .collectionFolder, .folder:
-            let viewModel = ItemLibraryViewModel(parent: item)
+            let viewModel = ItemLibraryViewModel(parent: item, filters: .default)
```
JPKribs, line 105: I am adding this to mirror iOS.

Shared/ViewModels/LibraryViewModel/ItemLibraryViewModel.swift
```diff
@@ -117,6 +115,11 @@ final class ItemLibraryViewModel: PagingLibraryViewModel<BaseItemDto> {
             if filters.sortBy.first == ItemSortBy.random {
                 parameters.excludeItemIDs = elements.compactMap(\.id)
             }
+        } else {
+            // Default item types & sort order
```
JPKribs, line 119: This does 2 things:

1. Right now, when filters are not present, there is no sort order. For an example, take the current 1.2 or 1.3(5) version and open select a person. It will show you all of their movies & shows but in no specific order. This is because the order is set from the `FilterViewModel`. This ensure, if the `FilterViewModel` is not present, this will default to ordering the results by name.

2. Since I am using the `includeItemTypes` in the `FilterViewModel`, I don't want to set this value twice, This `[.movie, .series, .boxSet]` is the previously default value for this, so I am ensuring this is set when a `FilterViewModel` is not present (see people again). This is the same default found on the `FilterViewModel` so it will be set there as well if the view does not use these values. The alternative to this is I can reference `FilterViewModel.default.types` so this is only being set from the `FilterViewModel`?

LePips: Good catch that some necessary fields aren't set when we aren't filtering. I just moved this above so then the parent or the filters will override them.

LePips (approved):
Apologies as I haven't had time recently to think deeply about code, but I had other ideas about how this would be implemented. Filtering on the `itemTypes` should have been on the `ItemFilterCollection` instead of the `FilterViewModel`, as the `FilterViewModel` largely serves as an observable source of information for the selected and total collection of filters.

This work is good to clean up the `ItemTypeLibraryViewModel`. I am deferring filtering on `BaseItemKind` due to some issues I observed during testing.

JPKribs:
No worries at all! Your changes make a ton of sense. Thank you for the revisions this looks great!

PR #1412, opened 2025-01-28
[iOS & tvOS] FilterViewModel - Cleanup

JPKribs:
### Summary

**Resolves**: https://github.com/jellyfin/Swiftfin/issues/1246

This PR updates the `FilterViewModel` to be `Stateful`. Additionally, all of the filter logic is moved from the `FilterView` to the `ViewModel`. Finally, while I was working on the `FilterView`, I found that https://github.com/jellyfin/Swiftfin/issues/1246 was still present. I've updated the `SelectorView` and tested this from both the `FontPickerView` and `FilterView` and this is now working as expected.

---

I feel very good about the `FilterViewModel` change to `Stateful`. As I understand it, the data functionality should be tied to the `ViewModel` instead of the view. Please let me know if that is not the case or if there is any reason this view was intentional left as non-`Stateful`.

The `SelectorView` is functional in both locations it's called from but not pretty. To make this functional for both the `String` and `AnyItemFilter` I had to have the selection as both a `State` and a `Binding`. I'm open to feedback if there is a cleaner way to do this but I was trying to accomplish this without having to change the callers too dramatically.

If we decide the `SelectorView` changes are a wash, I'm perfectly fine with this PR focusing on the `FilterViewModel` changes instead! Let me know!

Final item, the `SearchViewModel` didn't have a filter for Letters. I've added that as a drop in copy/paste from the `LibraryItemViewModel`

---

### Before:
The reset appropriately resets the filter but the UI does not update. This is because the selection in a binding box is not updating correctly. My solution was to have a `State` item for editing that updated the `Binding` item on change.

https://github.com/user-attachments/assets/df47cd85-c408-4e93-bb42-9c09a02c6d0d

### After:
https://github.com/user-attachments/assets/8886f0e5-48c2-4091-ae0b-feb38b86c9aa


JPKribs:
This one should be ready to go. The only item that I was considering adding but didn't was filtering the filters. Sometimes, there is a navigationFilterBar when there are no options for that filter.

A good example of this Favorites. You can look at the favorites but there is no option for Genres or Filters in the dropdown. I would argue if the `FilterViewModel` returns nothing in `AllFilters` that dropdown should be hidden.

LePips:
While it is nice that things are moved out of `FilterView` and makes `FilterViewModel` easier to use, `FilterViewModel` doesn't really need to be fully `Stateful` - it's only really meant as an observable source of filter information. The only thing that would be "stateful" would be getting the query filters for the library parent, but that can kind of be hidden away from the user in this case.

I'll have to think about the overall architecture of this.

LePips (approved):
I mostly just cleaned up `FilterViewModel` away from manual tracking of the set filters as checking the filter state isn't that expensive.

PR #1414, opened 2025-01-29
[tvOS] ErrorViews - Creation

JPKribs:
### Summary

This is a simplified rework of https://github.com/jellyfin/Swiftfin/pull/1382 that exclusively focuses on the `ErrorView`. `HomeView` does behave strangely when it needs to be reloaded but that will need to be handled in a later PR.

This PR cleans up some of the buttons in tvOS to make then better reflect native `.card` functionality. Additionally, the `PrimaryButton` is added to tvOS and is used in the new `ErrorView`. I then mirrored all of the locations in iOS that use the `ErrorView` and added them to tvOS. 

The only unique thing that isn't a direct mirror of iOS is that `HomeView` will appear next to 2 `PagingLibraryViews`. Since those views wrap their `ErrorView` in a `ZStack` with a `Color.xxx` above it, there is a vertical offset where the `PagingLibraryViews`'s `ErrorView` is a tad higher than the `HomeView`. This isn't an issue it just looks weird. So I added a `Color.Clear` to the `HomeView` as well to ensure they're vertically positioned the same. I only added this since you will be able to switch between the `HomeView` and `PagingLibraryViews` very quickly so it's a lot more noticeable than if other views have a different position for their `ErrorView`.

---

### Offset Without `Color.Clear`
https://github.com/user-attachments/assets/5f738692-5c03-44d6-b90f-2fb9cec826a1

---

`ListRowButton` for tvOS, I made a change `onFocus` and `onDisabled`. `onDisabled` the buttons should darken 50%. `onFocus` the buttons would appropriately grow to mirror existing `.card` buttons but they would stand out. Default `.card` buttons in tvOS highlight lighter when they are focused while our custom `ListRowButton` did not. This made the buttons stand out more as custom and also made it less visible to see when they were focused. 

### `ListRowButton` Before

https://github.com/user-attachments/assets/6550d788-d570-49fe-bd3f-ccc530a5e4f0

### `ListRowButton` After

https://github.com/user-attachments/assets/2da01c69-1879-42ab-aa32-5f4f65ce92a5

---

### `ErrorView` In Action

https://github.com/user-attachments/assets/3c182da7-6a2c-40f0-a737-09986971759d

---

For reference, here are all of the iOS `ErrorView` locations. I put the ones that exist in tvOS in red and the ones that ARE tvOS in yellow:

<details>
<summary>`ErrorView` Locations pre-PR</summary>

Screenshot 2025-01-29 at 09 53 29

</details>


JPKribs:
One another note, while I'm working on these `ListRowButton`, do we want to switch out the "Sign Out" button? See below:

### Current

https://github.com/user-attachments/assets/b7878922-f9e5-4051-b74c-ac8b49e19b92

### Potential Change

https://github.com/user-attachments/assets/c5792ff5-f83e-4144-b040-7e38a3ee93b0

LePips:
Yes, let's please change the `Switch User` while we're here.

JPKribs:
> Yes, let's please change the `Switch User` while we're here.

I've committed this change

Swiftfin tvOS/Views/AppLoadingView.swift
```diff
@@ -18,10 +18,20 @@ struct AppLoadingView: View {
...
+            if !didFailMigration {
+                ProgressView()
+            }
+
             if didFailMigration {
-                Text("An internal error occurred.")
+                ErrorView(error: JellyfinAPIError("An internal error occurred."))
             }
         }
+        .topBarTrailing {
```
JPKribs, line 29: I brought this over from iOS. It doesn't seem to do anything and it's always disabled. It makes sense to pull this IMO but, if I'm pulling it from this PR I think it makes sense to pull this from iOS as well? Unless we have a usage for this we want to do in which case I think it should be on both platforms and we should TODO this so we know what we want to do later.

LePips: This view could be worked on a little bit but it rarely, if ever, should even be presented. If it does, it means we messed up badly. We should have a recovery state if a database migration couldn't happen - right now we just keep failing until the user re-installs.

LePips (approved):
`PrimaryButton` didn't feel necessary here as it was essentially just a stripped down `ListRowButton`. Looking at iOS, we could probably do the same there as well and just rename the current `ListRowButton` to `ColoredButton` or something. These are old views anyways and things have cleaned up since then.

Issue #1415, opened 2025-01-30
Error in « player difference » documentation ?


Jojolam44:
Hi,

In the player difference documentation page, there is a container section.
In this section, we can find this sentence : 

« AV1 is disabled by default but can be enabled for Native (AVKit) using Custom Device Profiles. »

Is it really AV1 or AVI ?

Thanks.

JPKribs:
Here is the code for determining playback for the Native (AVKit) player:

```
//
// Swiftfin is subject to the terms of the Mozilla Public
// License, v2.0. If a copy of the MPL was not distributed with this
// file, you can obtain one at https://mozilla.org/MPL/2.0/.
//
// Copyright (c) 2025 Jellyfin & Jellyfin Contributors
//

import Foundation
import JellyfinAPI

extension VideoPlayerType {

    // MARK: direct play

    @ArrayBuilder<DirectPlayProfile>
    static var _nativeDirectPlayProfiles: [DirectPlayProfile] {
        DirectPlayProfile(type: .video) {
            AudioCodec.aac
            AudioCodec.ac3
            AudioCodec.alac
            AudioCodec.eac3
            AudioCodec.flac
        } videoCodecs: {
            VideoCodec.h261
            VideoCodec.hevc
            VideoCodec.mpeg4
        } containers: {
            MediaContainer.mp4
        }

        DirectPlayProfile(type: .video) {
            AudioCodec.aac
            AudioCodec.ac3
            AudioCodec.alac
        } videoCodecs: {
            VideoCodec.h264
            VideoCodec.mpeg4
        } containers: {
            MediaContainer.m4v
        }

        DirectPlayProfile(type: .video) {
            AudioCodec.aac
            AudioCodec.ac3
            AudioCodec.alac
            AudioCodec.eac3
            AudioCodec.mp3
            AudioCodec.pcm_s16be
            AudioCodec.pcm_s16le
            AudioCodec.pcm_s24be
            AudioCodec.pcm_s24le
        } videoCodecs: {
            VideoCodec.h264
            VideoCodec.hevc
            VideoCodec.mjpeg
            VideoCodec.mpeg4
        } containers: {
            MediaContainer.mov
        }

        DirectPlayProfile(type: .video) {
            AudioCodec.aac
            AudioCodec.ac3
            AudioCodec.eac3
            AudioCodec.mp3
        } videoCodecs: {
            VideoCodec.h264
        } containers: {
            MediaContainer.mpegts
        }

[...]

PR #1416, opened 2025-01-30
[Meta] Video Player Documentation - Move AV1 from Video to Container Section

JPKribs:
This should be under containers not video. Confusing since AV**i** vs AV**1**. Added a note for the Apple SOC that first added HWA AV1 decoding as this should correspond with AVKit's ability to use it.

Pre-AV1 Decoders, I tested, and AV1 works but is VERY choppy so defaulting this to disabled is still the right move!

Resolves: https://github.com/jellyfin/Swiftfin/issues/1415

### DIFF
--- a/Documentation/players.md
+++ b/Documentation/players.md
@@ -93,7 +93,7 @@ Swiftfin offers two player options: Swiftfin (VLCKit) and Native (AVKit). The Sw
 | **VP8**     | ✅                | ❌             |
 | **VP9**     | ✅                | ❌             |
 
-- AV1 is disabled by default but can be enabled for Native (AVKit) using Custom Device Profiles. Enabling AV1 may result in a [poor experience for non-A17 SOCs and higher](https://en.wikipedia.org/wiki/Apple_A17).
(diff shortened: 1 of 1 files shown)


Documentation/players.md
```diff
@@ -86,14 +85,16 @@ Swiftfin offers two player options: Swiftfin (VLCKit) and Native (AVKit). The Sw
...
-| **AV1**     | ✅                | ❌             |
+| **AV1**     | ✅                | 🟡 Limited support |
 | **H.264**   | ✅                | ✅             |
 | **H.265**   | ✅                | ✅             |
 | **MPEG-2**  | ✅                | ❌             |
 | **MPEG-4**  | ✅                | ✅             |
 | **VP8**     | ✅                | ❌             |
 | **VP9**     | ✅                | ❌             |
 
+- AV1 is disabled by default but can be enabled for Native (AVKit) using Custom Device Profiles. Enabling AV1 may result in a [poor experience for non-A17 SOCs and higher](https://en.wikipedia.org/wiki/Apple_A17).
```
LePips, line 96:
```suggestion
- AV1 is disabled by default but can be enabled for Native (AVKit) using Custom Device Profiles. Enabling AV1 may result in a [poor experience for SOCs prior to A17](https://en.wikipedia.org/wiki/Apple_A17).
```

LePips (changes requested):

LePips (approved):
Fast change! And thank you to @Jojolam44 for reporting.

Issue #1327, opened 2024-11-25
Languages missing from app settings


felix920506:
### Describe the bug

Description:

Some languages that are in the Swiftfin weblate are not present in the language settings of the app. For example Traditional Chinese, Dutch, Ukrainian

Reproduction Steps:
1. Download Swiftfin on App store
2. Start Swiftfin for the first time
3. Leave swiftfin and go to settings app, find swiftfin and select language

### Application version

1.2

### Where did you install the app from?

App Store

### Device information

iPad Pro 2021

### OS version

iPadOS 17.5.1

### Jellyfin server version

Server is not required for reproduction of this issue

samglt:
Looks like many languages are in the repo but haven't been added to the Xcode project. I'll submit a PR later today.

Given that there are 46 languages in the repo and only 16 of them are actually user accessible, I think there should at least be some documentation somewhere saying to remember to add new localizations in Xcode when merging new translations from Weblate. 

I'm not sure how practical this would be to automate since it involves the .xcodeproj file but it would be nice if it was

JPKribs:
Resolved by #1417

PR #1417, opened 2025-01-30
Add existing localizations to Xcode project

samglt:
Fixing: #1327

About 30 localizations that have been merged from Weblate were not added into the Xcode project so they weren't actually usable

For reference the way I added them from the UI was:
Right click the Translations folder > Add Files to "Swiftfin" > Select the .lproj folders > Add > Reference files in place & Create groups & check both targets > Finish


LePips (approved):
Thank you very much. Interesting that Weblate doesn't automatically add these to the project, which I thought that it did. I don't know how this could be automated right now, but it isn't a bad idea.

Just brainstorming: I have been following [Tuist](https://github.com/tuist/tuist) for a while but haven't actually used it. Maybe it can help us in this situation.

PR #1419, opened 2025-02-04
[tvOS] App Settings & Splashscreen

chickdan:
Currently this is using the same AppSettingsView brought over from iOS. I was going to make the view shared but it doesn't look like a pattern used in project so far so I have kept them separate. The big difference here is the sign-out on background setting is still disabled due to the lack of a date picker on tvOS. I did find a potential [library](https://github.com/ViacomInc/TVOSPicker) but wanted input first.

There also was some earlier work in this view that was commented out which included resetting user settings and removing all servers, does that need to stay for future completion?

<details><summary>Toggling Splashscreen:</summary>
<p>

https://github.com/user-attachments/assets/99319d70-c7e2-4a46-96bf-00c18bb1d00e

</p>
</details> 

<details><summary>Signout on Close</summary>
<p>

https://github.com/user-attachments/assets/e7e4233b-845f-4b2e-9127-e23c7710135c

</p>
</details>


Shared/Coordinators/AppSettingsCoordinator.swift
```diff
@@ -17,18 +17,14 @@ final class AppSettingsCoordinator: NavigationCoordinatable {
     @Root
     var start = makeStart
 
+    @Route(.modal)
```
JPKribs, line 20: I think we'll still want to keep these separate since Modal vs Push for iOS and tvOS.

Swiftfin tvOS/Views/AppSettingsView/AppSettingsView.swift
```diff
@@ -0,0 +1,100 @@
...
+//                        CaseIterablePicker(
+//                            L10n.appearance,
+//                            selection: $appearance
+//                        )
+//                    }
+                }
+
+                Section {
+
+                    Toggle("Use splashscreen", isOn: $selectUserUseSplashscreen)
```
JPKribs, line 68: Localization needed. I don't know if iOS is the same but let's change it there too if it is.

Swiftfin tvOS/Views/AppSettingsView/AppSettingsView.swift
```diff
@@ -0,0 +1,100 @@
...
+                            }
+
+                            ForEach(viewModel.servers) { server in
+                                Text(server.name)
+                                    .tag(SelectUserServerSelection.server(id: server.id))
+                            }
+                        }
+                    }
+                } header: {
+                    Text("Splashscreen")
```
JPKribs, line 85: Localize

Swiftfin tvOS/Views/AppSettingsView/AppSettingsView.swift
```diff
@@ -0,0 +1,100 @@
...
+                } footer: {
+                    if selectUserUseSplashscreen {
+                        Text("When All Servers is selected, use the splashscreen from a single server or a random server")
```
JPKribs, line 88: Localize

Swiftfin tvOS/Views/AppSettingsView/Components/SignOutIntervalSection.swift
```diff
@@ -0,0 +1,72 @@
...
+        private var signOutOnBackground
+        @Default(.signOutOnClose)
+        private var signOutOnClose
+
+        @State
+        private var isEditingBackgroundSignOutInterval: Bool = false
+
+        var body: some View {
+            Section {
+                Toggle("Sign out on close", isOn: $signOutOnClose)
```
JPKribs, line 28: Localize

Swiftfin tvOS/Views/AppSettingsView/AppSettingsView.swift
```diff
@@ -0,0 +1,100 @@
...
+                Section(L10n.accessibility) {
+
+                    // TODO: supposedly supported but not working
+//                    ChevronButton(L10n.appIcon)
+//                        .onSelect {
+//                                router.route(to: \.appIconSelector, viewModel)
+//                        }
+
+                    // Disabled until tvOS can alter app appearance
+//                    if !selectUserUseSplashscreen {
```
JPKribs, line 58: I'll need to defer to @LePips for this but I believe tvOS will only ever be Dark mode? So this may not be required.

JPKribs (commented):
First pass, looks good! Mostly just some localizations we should clean up. Some of those strings might be on the iOS side as well. In which case, can we update this in those spots as well?

Shared/Coordinators/AppSettingsCoordinator.swift
chickdan: That was a detail that I did not notice, thanks!

Swiftfin tvOS/Views/AppSettingsView/AppSettingsView.swift
chickdan: Localized both iOS and tvOS!

JPKribs: Thanks for doing that!

Swiftfin tvOS/Views/AppSettingsView/Components/SignOutIntervalSection.swift
```diff
@@ -0,0 +1,72 @@
...
+        private var isEditingBackgroundSignOutInterval: Bool = false
+
+        var body: some View {
+            Section {
+                Toggle(L10n.signoutClose, isOn: $signOutOnClose)
+            } footer: {
+                Text(L10n.signoutCloseFooter)
+            }
+
+            // TODO: need to consider date picker options to re-enable
```
JPKribs, line 33: I think for existing date entry I've been using the `ChevronActionButton` Input? View? I'm traveling for work so I don't have it easily accessible but I'm using it for the Max Days in Next Up.

I'd argue, for now, that should be usable for this part of the PR. I think something nicer/cleaner would be better but having a 'Minutes' input so this functionality can exist beats a TODO IMO. 

Eventually, I'd like to replace my other Date/Time usages with a picker closer to iOS but that Chevron Input should do the trick for now (unless someone else says otherwise haha). I'd argue that's the last major signin feature that needs to be ported over from iOS 🎉 so I'm inclined to make sure it gets added in here as well 😆

Swiftfin tvOS/Views/AppSettingsView/AppSettingsView.swift
LePips: Correct that tvOS will most likely be forever in dark mode as locking down a theme isn't uncommon.

Swiftfin tvOS/Views/AppSettingsView/Components/SignOutIntervalSection.swift
LePips: It is a bummer that tvOS doesn't have `DatePicker` or just a general carousel-picker. I just quickly googled and saw [this picker](https://github.com/ViacomInc/TVOSPicker), which we can definitely take a look at using.

LePips (approved):
I removed the `Accessibility` section since we won't be supporting switching app icons or light mode. Alternate app icon support is different on tvOS, at least when I looked at it years ago.

I also implemented the `Menu` for the server selection for the splash screen to use.

Issue #1337, opened 2024-12-02
Display movie rating on poster


foorschtbar:
**Describe the feature you'd like**

I first had to open a [issue](#1335) and get a tip on where to find the movie rating. I suggest displaying the rating in the same place as in the WebUI.

**Additional context**

Current poster without rating
![image](https://github.com/user-attachments/assets/ee5a6868-61b3-43a7-95a5-9592a6e217f8)

WebUI with rating on the poster
![image](https://github.com/user-attachments/assets/b5d6470c-ef8e-4f77-8c77-fecd21eed6a6)

samglt:
Made a quick change and I think it fits into that area well

![Image](https://github.com/user-attachments/assets/e6b0d8a8-55c2-4055-93c5-cce522a4206d)
![Image](https://github.com/user-attachments/assets/98a32bac-ce2e-4816-965b-d0c8969a8344)

PR #1420, opened 2025-02-06
[iOS] Show critic & community ratings alongside attributes

samglt:
Adding a feature requested in: #1337 

This shows critic and audience ratings above the fold in a similar place to where the Jellyfin web UI shows them.

Examples:
Image
Image

### DIFF
--- a/Swiftfin/Views/ItemView/Components/AttributeHStack.swift
+++ b/Swiftfin/Views/ItemView/Components/AttributeHStack.swift
@@ -17,6 +17,33 @@ extension ItemView {
 
         var body: some View {
             HStack {
+                if let criticRating = viewModel.item.criticRating {
+                    HStack(spacing: 2) {
+                        Group {
+                            if criticRating >= 60 {
+                                Image(.tomatoFresh)
+                                    .symbolRenderingMode(.hierarchical)
+                            } else {
+                                Image(.tomatoRotten)
+                            }
+                        }
+                        .font(.caption2)
+
+                        Text("\(criticRating, specifier: "%.0f")")
+                    }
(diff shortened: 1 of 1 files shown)


JPKribs:
I like how this looks! I thought about something similar but I was worried we're getting too cluttered on this bar. When all of the attributes are active, we're at the edge off the button:

Screenshot 2025-02-06 at 08 59 13

The way these attributes are set up, the font sized is influenced by iOS settings and accessibility settings. This is what this looks like before we get to the 'larger' options in iOS:

Screenshot 2025-02-06 at 09 02 26

I think for UI, @LePips is 100% the better resource for knowing what this should look like but that would be my concern with this change.

Lastly, I know there is a vocal group who do not like having community ratings on their media. I would argue we should create a setting for enabling this location so people who don't like the ratings can toggle them off.

samglt:
Thanks for catching the text sizing thing. Given that, I don't think these ratings can be put in the attributes stack. Maybe they could be in a separate HStack just above/below the attributes if that doesn't clutter up the area too much, or in a section below like Genres, Cast, etc. I'll try some different placements and post screenshots.

About people who don't want to see ratings, I'm one of those people; I didn't know it was big deal for others. I'll add a setting for that.

LePips (approved):
Thanks for the work! Good catch thinking about accessibility sizing @JPKribs. I think the overflow issue will probably use something like [this project](https://github.com/ksemianov/WrappingHStack), or a custom `Layout` when iOS 15 is dropped.

Putting these in these attributes is good for now but we may want to think about an alternate design. I thought putting it down below was good enough but some people do like to see them.

PR #1422, opened 2025-02-07
[tvOS] Mirror iOS Ratings + Attribute Settings

JPKribs:
### Summary

This mirrors the changes for https://github.com/jellyfin/Swiftfin/pull/1420 in tvOS. Pure Copy + Paste. Shout out to @samglt for doing the hard part and creating a layout that looks native and polished!

This section of tvOS has a lot of remaining horizontal room so luckily I don't think we have to worry about potential spacing issues that might arise in iOS. My only concern in this area is that the ratings are relatively smaller compared to the size of the tv screen so it might be more easily 'lost' for those looking for it. If that doesn't concern us, I think @samglt's design on this looks very polished and at home on tvOS!

In addition to copying over these attributes, I added 2 settings on both iOS and tvOS to turn Critic or Community ratings off. There is a certain amount of chatter online that some people are not fans of seeing ratings of their media so this gives them the ability to turn this off. I split this out in case people want one but not the other.

### Outstanding Discussion

This is a small change with only the toggle being new. Do we still want these ratings in the 'About' section of the `ItemView` as well? That 'About' section the description and ratings _(now both duplicates from above)_ and the media info. IMO, I think this section _(in another PR)_ should just be the media info and maybe we break it out a little more? Totally open to discussion, not a part of this PR.

Only item for this PR is do we want to remove the ratings down there as well with the toggle? Or, do we want to remove ratings altogether now they're in the attributes?

### Screenshots

<details>

<summary> tvOS New Rating</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-02-07 at 12 00 29

</details>

<details>

<summary> iOS New Setting</summary>

Simulator Screenshot - iPhone 16 Pro - 2025-02-07 at 12 16 09

</details>

### DIFF
--- a/Swiftfin tvOS/Views/ItemView/Components/AttributeHStack.swift
+++ b/Swiftfin tvOS/Views/ItemView/Components/AttributeHStack.swift
@@ -22,7 +22,7 @@ extension ItemView {
                     getAttribute(attribute)
                 }
             }
-            .foregroundColor(Color(UIColor.darkGray))
+            .foregroundStyle(Color(UIColor.darkGray))
         }
 
         @ViewBuilder
--- a/Swiftfin tvOS/Views/ItemView/Components/AttributeHStack.swift
+++ b/Swiftfin tvOS/Views/ItemView/Components/AttributeHStack.swift
@@ -6,26 +6,30 @@
 // Copyright (c) 2025 Jellyfin & Jellyfin Contributors
 //
 
-import Defaults
 import SwiftUI
 
 extension ItemView {
-
     struct AttributesHStack: View {
-
         @ObservedObject
         var viewModel: ItemViewModel
 
-        @StoredValue(.User.enableCriticRatings)
-        private var enableCriticRatings
-
-        @StoredValue(.User.enableCommunityRatings)
-        private var enableCommunityRatings
+        @StoredValue(.User.itemViewAttributes)
+        private var itemViewAttributes
 
         var body: some View {
             HStack(spacing: 25) {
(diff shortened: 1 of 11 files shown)


Swiftfin tvOS/Views/ItemView/Components/AttributeHStack.swift
```diff
@@ -6,6 +6,7 @@
 // Copyright (c) 2025 Jellyfin & Jellyfin Contributors
 //
 
+import Defaults
```
LePips, line 9: Import not needed.

LePips (changes requested):
I am wary of making future settings that increases granularity but am fine that attributes can be customizable. We should investigate some general solution in the future.

LePips:
For the `About` section, I don't have an answer right now but think that the entire section could be customizable in the future.

JPKribs:
> I am wary of making future settings that increases granularity

Sorry, I feel like I always do this... Do you mean, less settings altogether or a more broad approach? I was tinkering with a more broad approach before I committed this that is simultaneously more granular but also a more general solution:

I'd be happy to commit this branch if you would prefer this route. Otherwise, I can just commit the `Defaults` change.

https://github.com/user-attachments/assets/4b16a593-de94-4a37-b2c9-72b2c154c39d

LePips (approved):
We're on the same page about how these are customized. Now that the overlay section is allowing some customization, some sizing considerations need to happen. For example, if no attributes are selected there is a but of blank space where the view used to be.

PR #1423, opened 2025-02-07
[tvOS] Fix ItemType Libraries

JPKribs:
### Summary

In https://github.com/jellyfin/Swiftfin/pull/1411 we drop the `isRecursive` flag. This is only a requirement for `itemType` libraries. I've added it to the bool logic so if the `filter.itemType.isNotEmpty` then make it recursive. Resolves the issue where these libraries are folders instead of the items. See screenshot.

### Screenshot

Fixes:
Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-02-07 at 12 28 01

### DIFF
--- a/Shared/Objects/LibraryParent/LibraryParent.swift
+++ b/Shared/Objects/LibraryParent/LibraryParent.swift
@@ -50,7 +50,6 @@ extension LibraryParent {
             parameters.parentID = id
         case .folder:
             parameters.parentID = id
-            parameters.isRecursive = nil
         case .person:
             parameters.personIDs = [id]
         case .studio:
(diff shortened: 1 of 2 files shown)


LePips (changes requested):
Good catch. I think to more closely match the logic prior to #1411, we should instead set `parameters.isRecursive = true` by default and in the default implementation of `setParentParameters` the `parameters.isRecursive = true` can be removed from there.

JPKribs:
> Good catch. I think to more closely match the logic prior to #1411, we should instead set `parameters.isRecursive = true` by default and in the default implementation of `setParentParameters` the `parameters.isRecursive = true` can be removed from there.

These changes should now be in the latest commit.

LePips (approved):

PR #1424, opened 2025-02-08
Make UnmaskSecureField respond to dynamic type

samglt:
UnmaskSecureField didn't change in size with dynamic type changes, making it stick out in the UI


LePips (approved):
Thanks for the accessibility consideration! For some correctness sake, I just added getting the font from the environment instead.

Issue #609, opened 2022-10-19
Support for playlist


DarS2323:
**Describe the feature you'd like**
Would be great to see/manage playlist. 

Would be good if playlist/collection could be added to the home screen

PR #1428, opened 2025-02-09
[iOS & tvOS] Playlists - Library & Item View

JPKribs:
### Summary

Part 1 of 2 for: #609 

This PR adds Playlists to Swiftfin. This PR does not allow editing Playlists or adding items to Playlists. I intend to add that functionality to a later PR. This starts https://github.com/jellyfin/Swiftfin/issues/609 but I wouldn't consider #609 resolved until I can add the ability to add/remove items from a playlist.

I found that Playlists and Collections are both fairly similar. When I created the ViewModel for Playlists, it was almost identical to Collections. So, I renamed Collections to Lists since this applies to Collections and Lists now, and I added in a single check to ensure we're not allowing media types that are not supported by Swiftfin. This should have been the case prior since Collections can have Audio, Books, Photos, etc. I've added a check so only Movies, Episodes, Series, and Collections can be in Collections/Playlists. This is a bit of a lie since Collections can have Series/Collections but Playlists cannot but a Playlist handles this by turning the Series/Season/Collections into it's component Movies/Episodes instead. This shouldn't impact us since this is on the Server but it is the only real "quirk" of using a single ViewModel for both Playlists and Collections. It just means a Playlist is more permissive for items that don't exist in Playlists.

I added a check on Playlists/Collections so the PlayButton should be functional instead of being hidden/grayed out. The logic I can using it to find the first item that is NOT played as use that as the play button. If there are no items that are not played, instead just use the first item found. I think it could be argued maybe it should be:

1. Resumable Items
2. Unplayed Items
3. First Item

Because the PlayButton works on Collections (and Playlists) I moved both Collections & Playlists to use the CinematicScrollView instead of a custom, just the image.

```
        ItemView.CinematicScrollView(viewModel: viewModel) {
            ContentView(viewModel: viewModel)
        }
```

Finally, for the ItemEditorView, I found that on 10.10 metadata, images, etc. can only be edited by administrators so I have the locked in the same way as standard items. Users with the canDelete are administrators with access to the Playlist and the owner of the Playlist. So, Playlists are deletable from Swiftfin by the qualified parties but there is no ability to edit them except by an Admin. This mirrors web with the exception of a new API for making a Playlist public which we do not have access to on 10.8.

Please find screenshots below. Let me know if there are any questions! In terms of real changes, this can be boiled down to:

1. Collections, add logic for assigning the PlayButtonItem.
2. Enable the Playlist ItemType for MediaViewModel
3. Have the Switch Case for which ItemView to use, use CollectionItemView for items of the type 'Playlist'
4. Change Collections (and now Playlists) to use the standard ItemView since the PlayButton is now functional  (See 1.)
5. Lock Collections (and now Playlists) to only show child types that Swiftfin can utilize
6. Rename Collection items to be List... items since they are used for Collections and Playlists.

### Screenshots

<details>

<summary>The Media View - User Has Playlists vs Does not Have Playlists</summary>

Playlists vs No Playlist

</details>

<details>

<summary>Playlist PagingLibraryView</summary>

Playlist - PagingView

</details>

<details>

<summary>Scrolling a Playlist</summary>

Note the multiple ItemTypes of Episodes & Movies:

https://github.com/user-attachments/assets/74d2ee43-688f-4e7f-9edf-0e2f5bb2318b

</details>

<details>

<summary>Deleting a Playlist</summary>

iOS:

https://github.com/user-attachments/assets/81da4f45-4d30-4dcf-b611-20b2c3248b7b

tvOS:

Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-02-09 at 12 57 47

</details>

<details>

<summary>New Resume/Play Button for Collections</summary>

iOS:

Collection - Resume

tvOS:

Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-02-09 at 12 37 55

</details>


JPKribs:
Some QA type info:

- Marking a playlist as played mirrors web and marks playlist items as played.
- Favoriting a playlist only favorites that playlist object. All playlist items do no have their favorite status updated from a playlist. This mirrors web as well.
- Favorited playlists appear in the Favorites library.
- At this time, playlists use whatever poster type is dictated from the settings (Poster/Landscape) It looks like Jellyfin wants playlists to use Square so they are stretched. I'm up for resolving this but that seems like a larger item than this PR and I wanted to keep this PR from ballooning too much.
- Mirroring web, non-adminstrators can create/delete playlists but cannot update their metadata/images. Very weird IMO but that appears to be how it's set up on a Jellyfin side so that would need to be resolved there first.

JPKribs:
This was an idea I had for a revision. Instead of grouping everything together, collections/playlists can have various item types in them. on Web, these items are split out by group and IMO that looks better than grouping it. This is how this revision looks:

https://github.com/user-attachments/assets/d0215dc4-cd40-46c8-b1d4-7e1b427cab41

JPKribs (commented):
I've added a few comments in a review just to help answer some potential questions looking back through this myself. I think this is ready to go with only 4 questions/preferences that might need discussion. Otherwise, for the most part, this is primarily using existing `Views` and `ViewModels` so I feel good about minimizing impact.

1. Are we fine with Playlists and Collections both sharing the same UI and ViewModel? If I were to make these separate, they would be very similar so, to me, it made sense to combine.
2. Do we like splitting the item rows in Collections/Playlists by `BaseItemKind`? Should there be a setting to toggle displaying by type or merged? I have a more in-depth comment for this in the review.
3. Should the Edit/Delete logic be simplified to lean on Server's permissions or should this continue using the local preferences as well? I have a more in-depth comment for this in the review.
4. Is there an easy way to do `Localization` and `Plural Localization` without just having 2 localizations?

LePips:
I have largely not taken a look at this since I don't want it in 1.3, and want to focus on the fixes for that update, instead deferring to 1.4.

JPKribs:
> I have largely not taken a look at this since I don't want it in 1.3, and want to focus on the fixes for that update, instead deferring to 1.4.

No worries at all. Part of why I left the PR comments are for me to remember why I did these as well. I'll keep this merge-able with Main but otherwise I'm happy leaving this until 1.4+.

JPKribs:
Since I based this PR off the idea that I would be able to do Playlists and Collections from the same layout, this is going to be pretty hard to untangle. I've just branched this off into a new PR that can be found here:

https://github.com/jellyfin/Swiftfin/pull/1541

Issue #1096, opened 2024-06-11
unable to select different movie file when multiple are merged


C2gl:
### Describe the bug

on swiftfin on ipados, i have the option to select what file version of the movie i want to watch (same for show episode) but it selects the one by default and selecting the other one does nothing 

### Application version

1.1

### Where did you install the app from?

App Store

### Device information

ipad 9th generation 

### OS version

ios 17.5.1

### Jellyfin server version

10.9.6

warnerbryce:
Can confirm this in the lastest beta version 1.3.
![Image](https://github.com/user-attachments/assets/6c1e1b7b-d42c-430a-86df-b76a862581bb)
If i click on 4K, it stays on 2K. 
If i open the jellyfin iOS app i don’t have this problem, the choice is made in a different way. 
![Image](https://github.com/user-attachments/assets/fdb86691-6abd-453e-9eb9-190d272d6095)
On this screenshot, it was on 2K by default, i clicked on 4K and double checked, 4K version was still selected.

mxsrm:
This also affects the newest TestFlight version.

JPKribs:
This appears to be disabled in code, on line 83, here:

https://github.com/jellyfin/Swiftfin/blob/main/Swiftfin/Views/ItemView/Components/ActionButtonHStack.swift#L83

Re-enabling this is not as simple as un-commenting out this item as it looks like this was disabled when `ItemViewModel` was moved to being a `stateful` object. Adding this to the `ItemViewModel` looks as though it should resolve it but I was wary of touching this as it likely interferes with https://github.com/jellyfin/Swiftfin/pull/1203 which is a much broader and important update to how item playback is handled in Swiftfin.

I can keep an eye on this as 1203 is completed but I have a hunch that this area of Playback will be changed in that greater PR. I believe the goal is 1.3 (the next update) should contain 1203.

warnerbryce:
I updated the Jellyfin iOS app today and it’s working. 
Just some little anoyance : 
I select a version, i click on PLAY, the video is loading but at the same time the box for selecting the version revert back to it’s old position. 
The next time you want to watch the movie, you have to remember which version you wanted to watch

JPKribs:
> I updated the Jellyfin iOS app today and it’s working. 
> Just some little anoyance : 
> I select a version, i click on PLAY, the video is loading but at the same time the box for selecting the version revert back to it’s old position. 
[...]

PR #1429, opened 2025-02-09
[iOS & tvOS] Fix Version Selection

JPKribs:
### Summary

Please let me know if this interferes with https://github.com/jellyfin/Swiftfin/pull/1203 and I can close this out.

Resolves: https://github.com/jellyfin/Swiftfin/issues/1096

This adds a function to `ItemViewModel` to allow you to update the `SelectedMediaSource` statefully. This resolves the issue that the version selection did not update the value prior.

Additionally, since tvOS did not have versioning, I added a `Version` menu to the ellipsis if there is more than one version present. This is the only item I may need some advice on. I just added it to ellipsis menu that exists for refresh/delete. Would we prefer this to be more of a standalone menu like how it is on iOS? I can absolutely make that change as needed but, as a warning, that likely means we're out of room for more buttons for this spot on tvOS. We can definitely go to another row but for the first row, 3 buttons + ellipsis is likely our limit.

### Screenshots

I don't do multiple versions on my server so I just manually merged two random movies. Hence the two totally different movies. This did assist in testing as I could confirm that the selected movie was the one being played.

<details>
<summary>New Menu (tvOS)</summary>

Screenshot 2025-02-09 at 16 24 24

Screenshot 2025-02-09 at 16 22 33

</details>

<details>
<summary>iOS Fix</summary>

Simulator Screenshot - iPhone 16 Pro - 2025-02-09 at 16 26 13

</details>


JPKribs:
This is the alternative approach where this is next to the existing buttons:

Screenshot 2025-02-11 at 09 56 12

Screenshot 2025-02-11 at 09 56 18

Swiftfin tvOS/Views/ItemView/Components/ActionButtons/ActionButtonHStack.swift
```diff
@@ -64,6 +64,13 @@ extension ItemView {
             }
         }
 
+        // MARK: - Select Item(s)
+
+        private var multipleVersions: Bool {
```
LePips, line 69:
```suggestion
        private var hasMultipleVersions: Bool {
```

Swiftfin tvOS/Views/ItemView/Components/ActionButtons/SelectVersionButton.swift
```diff
@@ -0,0 +1,55 @@
...
+    struct SelectVersionButton: View {
+
+        @ObservedObject
+        var viewModel: ItemViewModel
+
+        // MARK: - Body
+
+        var body: some View {
+            Menu {
+                if let playButtonItem = viewModel.playButtonItem,
```
LePips, line 23: Just pass in the view model as well as the array of media sources so these checks don't need to be made.

LePips (changes requested):
This doesn't directly conflict with #1203 and thanks for fixing this. I actually do like the variant with the 3 buttons and then the menu if necessary, so we can switch to that.

JPKribs:
> I actually do like the variant with the 3 buttons and then the menu if necessary, so we can switch to that.

Sounds good! I can make that swap. As a more general question, what would be the route for other future buttons like "Trailer(s)" or "Add to Playlist"? Would those go down to a second row? Or, is that a "We'll get to it when we need it." Kind of question? 

I only ask because I was wondering about the adding/creating part of https://github.com/jellyfin/Swiftfin/pull/1428

LePips:
Ah, I forgot about trailers. We'll get there when we need it.

JPKribs:
No worries! I've made the change for this PR. Below are the various states that can exist:

Standard, Standard with Refresh/Delete Menu, Standard with Multiple Versions, and Multiple Versions with Refresh/Delete Menu.

Screenshot 2025-02-16 at 16 05 49

Below is what our drop down will show as. As a note, I just merged two totally unrelated movies to test this so the reality it would be like "Theatrical" and "Director's Cut" but in my case it's just 2 different movies haha.

Screenshot 2025-02-16 at 16 09 18

LePips (approved):
To not crowd the bottom stack we can take a look in the future of making this menu on the trailing side of the play button. We also need a way to surface what version is currently selected without going into the menu, perhaps with the version's title in the play button underneath the play label, although I don't know how that will look.

JPKribs:
> To not crowd the bottom stack we can take a look in the future of making this menu on the trailing side of the play button. We also need a way to surface what version is currently selected without going into the menu, perhaps with the version's title in the play button underneath the play label, although I don't know how that will look.

Sounds good! I agree that providing some visibility would be helpful. I wanted to look at the play button in another PR. Specifically, bringing back the resume/play from beginning options. I could look at this as well. That being said, I was likely going to hold off until after https://github.com/jellyfin/Swiftfin/pull/1203 just so you wouldn't have to deal with weird merge conflicts where possible.

PR #1435, opened 2025-02-16
[tvOS] Handle Empty Episode Selector

JPKribs:
### Summary

This is a very rare occurrence, but I found that if you have empty folder for a season, you can lock the episode selector until you re-enter the `ItemView`.

The issue is, without episodes, the episode selector does not exist. Because of the `FocusGuide`, there is no way to go down to Cast & Crew without episodes. Additionally, because there is no set area for episodes, going back to a season that contains episodes, the selector is tiny and shoved outside the screen.

This PR creates a "No episodes" filler text when episodes are not available along with a refresh button in case they want to update this season.

### Issue

https://github.com/user-attachments/assets/182337b4-cb25-4f8a-a005-469e36b091e6

### Resolution

https://github.com/user-attachments/assets/016476fb-659e-4b18-83fc-17967a78d9c1

### DIFF
--- a/Swiftfin tvOS/Views/ItemView/Components/EpisodeSelector/Components/EpisodeHStack.swift
+++ b/Swiftfin tvOS/Views/ItemView/Components/EpisodeSelector/Components/EpisodeHStack.swift
@@ -61,11 +61,27 @@ extension SeriesEpisodeSelector {
             }
         }
 
+        @ViewBuilder
+        private var emptyView: some View {
+            VStack {
+                Text(L10n.noEpisodesAvailable)
+                Button(L10n.retry) {
+                    viewModel.send(.refresh)
+                }
+                .buttonStyle(.bordered)
+                .focused($focusedEpisodeID, equals: "RefreshButton")
+            }
+        }
+
         var body: some View {
-            WrappedView {
+            ZStack {
                 switch viewModel.state {
                 case .content:
-                    contentView(viewModel: viewModel)
+                    if viewModel.elements.isEmpty {
+                         emptyView
+                     } else {
+                         contentView(viewModel: viewModel)
+                     }
                 case let .error(error):
                     ErrorHStack(viewModel: viewModel, error: error)
                 case .initial, .refreshing:
@@ -76,7 +92,7 @@ extension SeriesEpisodeSelector {
             .focusGuide(
                 focusGuide,
                 tag: "episodes",
-                onContentFocus: { focusedEpisodeID = lastFocusedEpisodeID },
+                onContentFocus: { focusedEpisodeID = viewModel.elements.isEmpty ? "RefreshButton" : lastFocusedEpisodeID },
                 top: "seasons"
             )
             .onChange(of: viewModel.id) {


LePips (changes requested):
Finally building out this view, I would like it to match the how we do it on iOS: using the same card design for all views. Overall, this will need to consider and work properly with focusing.

LePips:
Yes, like that

JPKribs:
Okay, I think I understand the `FocusGuide` now. I went through iOS and tried to mirror everything that made sense. Overall, it's a fairly minimal change. I've made the following changes:

1. There is now an `EmptyCard` when the `ViewModel` is empty.
2. I added a `ProgressView()` to the `LoadingCard`. It's not a requirement I just think it looks nice. Let me know if that's too busy or hurts performance too much and I can change that to something else like a static `"progress.indicator"`.
3. The `ErrorCard` did not work. You could not select it and it was not a normal card. This is now a regular card similar to `EmptyCard` or `LoadingCard`
4. I added padding to the bottom of the selector. This mirrors the same padding that exists on the `SeasonsHStack`. 
5. Added the ability to focus the `ErrorCard` to allow refreshing and `EmptyCard` to allow you to get to the "Cast and Crew" Section.

### Regular Season

https://github.com/user-attachments/assets/cab9bd21-d42c-4999-be2e-3257ba8a6090

### Empty Season:

https://github.com/user-attachments/assets/eaeebc53-ede6-4589-b576-4b19c00869f3

### Error Season:

https://github.com/user-attachments/assets/d07a8cd5-e5b0-4555-aa9b-29114f8bf636

### Loading Season:

https://github.com/user-attachments/assets/c25a342a-0a0f-4eb8-8943-9ba17b9af888

JPKribs:
I'm happier with where this is now. While I'm working on this, please let me know if there is anything else that needs to be adjusted in this area. It's a very close mirror to iOS at this point. Since the `...HStack` are all their own `struct` I had to pass in the `FocusState`. Let me know if that is not the correct approach.

The only item that doesn't have the ability to focus is the `LoadingCard`. This is because the `LoadingHStack1 is called from another location so I don't know if we want to use a `FocusState` there as well. Additionally, loading shouldn't exist for that long. Either it will load fully or error fairly quickly.

Only item I have a question on is the `LoadingCard` can stack a few times. From the videos in my last comment, you can see those stack up to 4 cards. Do we want this or should this only ever have 1 card?

LePips (changes requested):
> doesn't have the ability to focus is the `LoadingCard`

We should have a solution where this entire view can be used regardless of this episode selector, even when loading. Instead of the 2-4 random count of loading cards, which was mainly used as a skeleton view design, having a single card like the other non-content states have to allow better focusing could be a good idea.

JPKribs:
> > doesn't have the ability to focus is the `LoadingCard`
> 
> We should have a solution where this entire view can be used regardless of this episode selector, even when loading. Instead of the 2-4 random count of loading cards, which was mainly used as a skeleton view design, having a single card like the other non-content states have to allow better focusing could be a good idea.

Sounds good. I've made this change oh my latest commit. Everything looks and works as expected but I wasn't sure how to handle this:

```
    @FocusState
    private var focusedSection: String?

...

        VStack(spacing: 0) {
            SeasonsHStack(viewModel: viewModel, selection: $selection)
                .environmentObject(parentFocusGuide)

            if let selectionViewModel {
                EpisodeHStack(viewModel: selectionViewModel, playButtonItem: viewModel.playButtonItem)
                    .environmentObject(parentFocusGuide)
            } else {
                **LoadingHStack(focusedEpisodeID: $focusedSection)
                    .environmentObject(parentFocusGuide)
                    .focused($focusedSection, equals: "LoadingCard")**
            }
        }
```

I added a focus state but just for that? I'm not sure that's the right way to go. In testing, it all works correctly but the `FocusGuide` is still not my best spot. So if there is a different way to handle this please let me know and I can make adjustments.

This is what the `LoadingCard` and `LoadingHStack` now looks like:


https://github.com/user-attachments/assets/dd61fbf9-7e73-464c-985c-dd556b921083

Swiftfin.xcodeproj/project.pbxproj
LePips, line 1: A lot of changes to this file for supported platforms and catalyst support that we don't need right now.

Swiftfin tvOS/Views/ItemView/Components/EpisodeSelector/EpisodeSelector.swift
```diff
@@ -36,7 +49,9 @@ struct SeriesEpisodeSelector: View {
                 EpisodeHStack(viewModel: selectionViewModel, playButtonItem: viewModel.playButtonItem)
                     .environmentObject(parentFocusGuide)
             } else {
-                LoadingHStack()
+                LoadingHStack(focusedEpisodeID: $focusedSection)
```
LePips, line 52: We don't need this anymore since the `EpisodeHStack` will show the `LoadingHStack` internally.

LePips (changes requested):
Looking good, now just some clean up.

Swiftfin.xcodeproj/project.pbxproj
JPKribs: Ah shoot. Yeah, I was screwing around with building for MacOS and I forgot to untoggle this. I'll update that back.

Swiftfin tvOS/Views/ItemView/Components/EpisodeSelector/EpisodeSelector.swift
JPKribs: Gotcha. To make this change I could have gone two directions:

1. Change `EpisodeHStack` to accept `SeasonItemViewModel?` 
2. Default `selectionViewModel` to an empty `SeasonItemViewModel` if it's missing

I chose option 2 but let me know if I should have gone the other direction:

```
    private var selectionViewModel: SeasonItemViewModel {
        viewModel.seasons.first(where: { $0.id == selection }) ?? SeasonItemViewModel(season: BaseItemDto())
    }
```

LePips: We already have logic to set the correct selected season and that view model is passed into the `EpisodeHStack` anyways.

JPKribs: Gotcha, my bad. I was trying to handle this at the `selectionViewModel` but I understand what you mean now. Sorry about my confusion! This should be resolved.

LePips (approved):
Looks great! Some polish could be that if the loading or error cards and transitions to a new state it will stay focused on the card, instead of it hopping to the seasons.

PR #1436, opened 2025-02-17
[WIP] [iOS & tvOS] Playlists - Add / Create Views

### LINKED ISSUE #609: Support for playlist

JPKribs:
### Summary

Resolves: #609 
Continues: https://github.com/jellyfin/Swiftfin/pull/1428

Creates an "Add to Playlist" view that is accessible from the `ItemView`. Allows users to add an item to a playlist OR add an item to a new playlist. The latter creates the playlist at the same time as the item is added.

### In-Progress

1. https://github.com/jellyfin/Swiftfin/pull/1428 needs to wrap up first so I can re-use the `ListItemViewModel` and poster changes.
2. Do we want Adding to Playlists in the menu ellipsis at the top like I have it now or should this be with the other actions buttons?

### Videos

**iOS:**

https://github.com/user-attachments/assets/b12586f7-1eed-4807-aac6-04ae5e24ff6d

**tvOS:**

https://github.com/user-attachments/assets/db09db00-5fb0-4ef1-9598-3ee502b2f0fe


JPKribs:
This is "done" but with my rework of https://github.com/jellyfin/Swiftfin/pull/1428 this needs to be reworked as well. I am going to wait until after https://github.com/jellyfin/Swiftfin/pull/1541 is resolved to avoid conflicts and lower some of the clutter. I'm going to keep this branch for reference but my final version of this will likely look different. Especially since we now have 10.10 so the ViewModel needs a rework regardless.

PR #1438, opened 2025-02-20
[iOS & tvOS] Use `AboutView` for Series Poster Routing

JPKribs:
### Summary

This is a fairly small change and might not make sense. For episodes, this removes the "Series" section. Instead, if the `ViewModel.Item.Type` is `.episode`, the `ImageCard` routes the user to the Series. Otherwise, this continues to just be a poster. Works for both iOS and tvOS.

This removes the mostly empty "Series" section that kind of looks out of place with a single item. The only downside is it may not be directly obvious this button routes to the Series.

This change makes sense to me as I keep tapping this poster thinking it will take me back to the series. Let me know if this is too ambiguous. My idea for making this less ambiguous is, if the item is not an episode, just move focus back to the top of the page. That way it's always a button but we're not routing anyone from the `ItemView` back to the same `ItemView`.

### Video (tvOS)

https://github.com/user-attachments/assets/221ffd62-1d3e-452f-b4d8-c99fec0deb2a

### DIFF
--- a/Swiftfin/Views/ItemView/iOS/EpisodeItemView/EpisodeItemContentView.swift
+++ b/Swiftfin/Views/ItemView/iOS/EpisodeItemView/EpisodeItemContentView.swift
@@ -78,21 +78,6 @@ extension EpisodeItemView {
                     RowDivider()
                 }
 
-                // MARK: Series
-
-                // TODO: have different way to get to series item
-                //       - about view poster?
-                if let seriesItem = viewModel.seriesItem {
-                    PosterHStack(
-                        title: L10n.series,
-                        type: .portrait,
-                        items: [seriesItem]
-                    )
-                    .onSelect { item in
-                        router.route(to: \.item, item)
-                    }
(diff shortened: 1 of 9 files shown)


Swiftfin/Views/ItemView/iOS/EpisodeItemView/EpisodeItemContentView.swift
```diff
@@ -78,21 +78,6 @@ extension EpisodeItemView {
                     RowDivider()
                 }
 
-                // MARK: Series
-
-                // TODO: have different way to get to series item
-                //       - about view poster?
```
JPKribs, line 84: This was the comment I was basing this change on. Let me know if this was the right route!

LePips (approved):
This is fine but how this is communicated is a concern. An additional route to the series from episodes could be making the series name on the views a `Button`, as tapping the series name isn't an uncommon interaction.

PR #1441, opened 2025-02-25
[iOS] Fix Clipping Add User Button

### LINKED ISSUE #1425: [iOS] Grid in UserSelectView gets cut off with more than one user
The grid of user icons in the UserSelectView gets cut off and doesn't scroll when more than two icons are shown. This looks like it was introduced in #1328.

I think the issue is that something in the UserGridButton is using trackingSize() while the VGrid's .scrollIfLargerThanContainer() also uses trackingSize() to determine its frame size. Maybe the GeometryPreferenceKey gets set to multiple values within the same tree at the same time and breaks the scrollIfLargerContainer.

I can't untangle exactly which change would have caused this to happen though since the previous commit also has uses of trackingSize inside the UserGridButton.

Image

chickdan:
Not entirely sure why this is necessary but it was the only difference I can spot after the change in #1328

Fixes #1425

### DIFF
--- a/Swiftfin tvOS/Views/SelectUserView/Components/UserGridButton.swift
+++ b/Swiftfin tvOS/Views/SelectUserView/Components/UserGridButton.swift
@@ -63,6 +63,8 @@ extension SelectUserView {
                 ),
                 pipeline: .Swiftfin.local
             )
+            .aspectRatio(1, contentMode: .fill)
+            .clipShape(.circle)
             .overlay {
                 if isEditing {
                     ZStack(alignment: .bottom) {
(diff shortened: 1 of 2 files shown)


JPKribs:
That seems to resolve it for me as well! Would you be able to mirror that on tvOS as we:

Swiftfin tvOS/Views/SelectUserView/Components/UserGridButton.swift

Looks like the save fix there on L66 & L67.

I'll defer to @LePips if there is any Swift reason that shouldn't be there but otherwise it looks good to me!

LePips (approved):
Hm the "clipping" is due to the `ScrollIfLargerThanContainer` setting the frame and the size not being reported correctly, as well as some oddities with my current `on[Size|Frame]Changed` wrappers. Unknown why setting the aspect ratio and clip shape again fixes it.

I played around with some clean up but will defer that to later.

Issue #1256, opened 2024-10-06
Can't select subtitles, it jumps back to none


ikheetjeff:
### Describe the bug

Thank you all for this fantastic app.

When I click on a subtitle, I expect it to load. However, that doesn’t happen: it switches back to "none" and doesn’t display any subtitles. This occurs with all SUBRIP languages in this video. However, it doesn’t happen with all series/movies. I haven’t been able to find a reason why it doesn’t work with this file but does with other films/series.

I’ve tried the following:
- Reinstalled the app
- Created a new folder + file name without metadata
- Used a different subtitle file in the same language
- Used a different subtitle file in another language
- The file also had PGS subtitles, which I removed using MKVToolNix and loaded only the SUBRIP: still didn’t work.
- Converted the video to H264 using Handbrake. (idk why but yeah i wanted to try everything before making a issue here)

When I installed an earlier TestFlight build (1.2 (3)), the problem was resolved. The issue is also not present in the App Store version. 

I suspect it is related to this PR: https://github.com/jellyfin/Swiftfin/pull/1219
Hopefully, @cebrusfs, you can take a look at this due to your previous commit. My apologies if this tag is not appreciated.

This video shows the bug: (TestFlight 1.2 (4))

https://github.com/user-attachments/assets/d776c40a-5a65-460b-bf07-8975981ee151

This is how it should work: (TestFlight 1.2 (3) / App Store version)

https://github.com/user-attachments/assets/150b29ab-406b-4382-ad2a-ed7784527d7b

Original media file (added the SUBRIP manualy in jellyfin)

```
General
Complete name                            : The.Walking.Dead.Daryl.Dixon.S01E02.Alouette.EAC3.5.1.1080p.Bluray.x265-NLSubs-SJK.mkv
Format                                   : Matroska
Format version                           : Version 4
File size                                : 1.08 GiB
Duration                                 : 1 h 0 min
Overall bit rate                         : 2 559 kb/s
Frame rate                               : 23.976 FPS
Encoded date                             : 2024-02-24 19:18:12 UTC
Writing application                      : HandBrake 1.6.1 2023012300
Writing library                          : Lavf59.27.100
ErrorDetectionType                       : Per level 1

Video
ID                                       : 1
Format                                   : HEVC
Format/Info                              : High Efficiency Video Coding
Format profile                           : Main@L4@Main
Codec ID                                 : V_MPEGH/ISO/HEVC
Duration                                 : 1 h 0 min
[...]

PR #1311, opened 2024-11-11
[iOS] Fix #1256

Gondnat:
This PR is used to fix "Can't select subtitles, it jumps back to none #1256"

There are two parts to fix it:
1. The SubtitleProfile in request should be  as below, the old profile will get  broken  *DeliveryUrl* for external subtitles.  
```
{Format: "vtt", Method: "External"}
```

The broken URL lost the extension name ".vtt" after Stream.

Borken URL
```
"DeliveryUrl": "/Videos/7c4730fd-90f1-e75e-656c-fe7da8fb22aa/7c4730fd90f1e75e656cfe7da8fb22aa/Subtitles/0/0/Stream?api_key=953dccded7e2455ba0088f62cf322dfc",
```

Normal URL
```
"DeliveryUrl": "/Videos/7c4730fd-90f1-e75e-656c-fe7da8fb22aa/7c4730fd90f1e75e656cfe7da8fb22aa/Subtitles/0/0/Stream.vtt?api_key=953dccded7e2455ba0088f62cf322dfc",
```

2. To make embedded subtitles work well, they should reduce as same as  embedded audio  in function *adjustAudioForExternalSubtitles*

PR #1445, opened 2025-03-03
[iOS] Fix External Subtitle Selection

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1256
Original: https://github.com/jellyfin/Swiftfin/pull/1311

This PR updates the offsets for Audio Tracks and Subtitles. In my testing, this enables the usage of subtitles, both internal and external, for both Direct Play and Transcodes:

**Direct Play**

Jellyfin provides the full file with all video & audio tracks included. External tracks from Jellyfin are provided at the _beginning_ while VLCKit needs them at the _end_. To resolve the external track issue, we move all internal track indexes to the beginning and all external tracks to the end. As a result, all tracks are playable. I found external audio was playable but only if it did not require transcoding.

**Transcodes**

Jellyfin provides a transcoded file that only contains 1 video track and 1 audio track. Subtitle tracks are accessible using index 2+. To resolve the external subtitle track issue, I start the subtitles at index 2 and count up with each one. It's very simple but in testing, this was working without fail.

To make this work fully, there is a lot of reworking that I think #1203 is already covering. The gist of this is we would need to better follow how it's done here:

https://github.com/jellyfin/jellyfin-web/blob/master/src/components/playback/playbackmanager.js#L1691

On Jellyfin-Web, transcodes don't try to change the selected audio track because transcodes don't _have_ a second audio track. Instead, they create a new PlaybackURL with the selected track and switch the user over to that using the `changeStream` function. This then takes the previous position in playback and uses that as the starting position for the new PlaybackURL. Subtitles are accessible via their own DeliveryURLs so those can be selected without requiring a new PlaybackURL. VLCKit complicates this a bit as it's purely a File/M3U handler. So, we would need to use the UI to switch playback URLs.

I would argue, for now, the solution I provided gets us ~70% of the way there with minimal impact to #1203 & the `VideoPlaybackManager`. It's far from a perfect solution but that would require us to build out the `VideoPlaybackManager` which is already in process in #1203.

---

### Current

| Type          | Internal Subtitles | Internal Audio | External Subtitles | External Audio |
|---------------|-----------------|--------------------|----------------|----------------|
| Transcode     | ❌              | 🟡*                 | ❌             | ❌             |
| DirectPlay     | ✅              | ✅                 | ❌             | ❌             |

*The default track for audio will play but cannot be changed.

---

### This PR

| Type          | Internal Subtitles | Internal Audio | External Subtitles | External Audio |
|---------------|-----------------|--------------------|----------------|----------------|
| Transcode     | ✅              | 🟡<sub>1</sub>                 | ✅             | ❌<sub>3</sub>             |
| DirectPlay     | ✅              | ✅                 | ✅             | 🟡<sub>2</sub>             |

1. The default track for audio will play but cannot be changed.
2. External tracks can be selected and played so long as they do not need to be encoded.
3. This fails due to the issue with 2)

---

### Unknowns

I don't personally use external subtitles, preferring to encode everything into MKV. As a result, all of my testing was using files I created specifically for this. This leaves us with some potential unknowns and test cases that need to be validated:

1. Does having or not having https://github.com/jellyfin/jellyfin-plugin-subtitleextract change the result?

2. ~Does the number of external tracks impact anything?~
**Tested and Verified Working.**

3. ~Does having no internal subtitles change anything?~
**Tested and Verified Working.**

### DIFF
--- a/Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto+VideoPlayerViewModel.swift
+++ b/Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto+VideoPlayerViewModel.swift
@@ -131,6 +131,6 @@ extension BaseItemDto {
         let testBitrate = testSizeBits / testDuration
 
         /// Exceeding 500 mbps will produce an invalid URL
-        return min(Int(testBitrate), 500_000_000)
+        return min(Int(testBitrate), PlaybackBitrate.max.rawValue)
     }
 }
--- a/Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto+VideoPlayerViewModel.swift
+++ b/Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto+VideoPlayerViewModel.swift
@@ -125,11 +125,12 @@ extension BaseItemDto {
         let userSession = Container.shared.currentUserSession()!
 
         let testStartTime = Date()
-        try await userSession.client.send(Paths.getBitrateTestBytes(size: testSize))
+        _ = try await userSession.client.send(Paths.getBitrateTestBytes(size: testSize))
         let testDuration = Date().timeIntervalSince(testStartTime)
         let testSizeBits = Double(testSize * 8)
         let testBitrate = testSizeBits / testDuration
 
-        return Int(testBitrate)
+        /// Exceeding 500 mbps will produce an invalid URL
+        return min(Int(testBitrate), 500_000_000)
     }
 }
--- a/Shared/Extensions/JellyfinAPI/MediaStream.swift
+++ b/Shared/Extensions/JellyfinAPI/MediaStream.swift
@@ -230,7 +230,7 @@ extension [MediaStream] {
         guard allSatisfy({ $0.type == .audio }) else { return self }
 
         let embeddedAudioCount = filter { $0.isExternal == false }.count
-        let externalAudioCount = filter { $0.isExternal ?? false }.count
+        let externalAudioCount = filter { $0.isExternal == true }.count
 
         var mediaStreams = self
 
--- a/Shared/Extensions/JellyfinAPI/MediaStream.swift
+++ b/Shared/Extensions/JellyfinAPI/MediaStream.swift
@@ -219,7 +219,7 @@ extension [MediaStream] {
     /// What is this?
     /// Jellyfin maintains a sequential indexing system across all tracks with external at the beginning
     /// VLCKit handles external tracks (added separately to the media) by appending them to the end of the track list
-    /// These functions change the order of the track list to reflect VLCKit's expectated order
+    /// These functions change the order of the track list to reflect VLCKit's expected order
 
     /// Adjust track indexes based on the a count of external tracks
     func adjustedAudio(
--- a/Shared/Extensions/JellyfinAPI/MediaStream.swift
+++ b/Shared/Extensions/JellyfinAPI/MediaStream.swift
@@ -257,7 +257,7 @@ extension [MediaStream] {
                 // Update the working copy with updated audio track indexes
                 mediaStreams[i] = copy
             default:
-                // TODO: Figure out tracks.
+                // TODO: Figure out audio tracks.
                 break
             }
         }
--- a/Shared/Extensions/JellyfinAPI/MediaStream.swift
+++ b/Shared/Extensions/JellyfinAPI/MediaStream.swift
@@ -223,13 +223,14 @@ extension [MediaStream] {
     /// For transcode stream type:
     ///   Only the first internal video track and the first internal audio track are included, in that order.
     /// In both cases, external tracks are appended in their original order with indexes continuing after internal tracks.
-    func adjustedTrackIndexes(for streamType: StreamType) -> [MediaStream] {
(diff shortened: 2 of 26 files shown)


JPKribs:
@LePips I have tested this several ways with both DirectPlay and Transcode and I believe this fixes our issue. I have a branch where I was working on the Transcode audio track changes that mirror the [web function](https://github.com/jellyfin/jellyfin-web/blob/master/src/components/playback/playbackmanager.js#L1691): `changeStream` but I ended up with a lot in the `PlaybackManager`. This PR is not a perfect solution but it resolves https://github.com/jellyfin/Swiftfin/issues/1256 with the minimal impact to https://github.com/jellyfin/Swiftfin/pull/1203.

I exclusively encode to mkv with embedded subtitles so I had to invent files to test this. I feel confident I am covering all my bases but I know people use Jellyfin dramatically differently between servers. So, any additional testing is always appreciated!

JPKribs:
Unrelated, in my testing, I found that using `Auto` playback bitrate would throw an error if the playback bitrate exceeded 500mbps. I made a change to this to just use our `Maximum` playback bitrate. This really only impacts it if you are directly playing from your server's host machine but I added this because it was killing me with testing and it's probably also good to add a catch for that.

LePips (changes requested):
I still have an issue with a basic movie:
- direct play
- no embedded subtitles
- a single external subtitle

I don't know when I can dive into the technicalities of whether what is here is correct, but something is off.

JPKribs:
> I still have an issue with a basic movie

Good catch! I'll create a file to mirror that and I can try to work that out today.

JPKribs:
Trying to wrap my head around this it made more sense as a table:

Scenario | Playback Type | Media Configuration | Subtitle Status | Audio Status | Notes
-- | -- | -- | -- | -- | --
1 | Transcode | Embedded Subtitles Only | ✅ | N/A |  
2 | Transcode | Embedded Subtitles + External Subtitles | ✅ | N/A |  
3 | Transcode | External Subtitles Only | ✅ | N/A |  
4 | Transcode | Embedded Audio Only | N/A | 🟡 | Only plays the first track
5 | Transcode | External Audio Only | N/A | 🟡 | Only plays the first track
6 | DirectPlay | Embedded Subtitles Only | ✅ | N/A |  
7 | DirectPlay | Embedded Subtitles + External Subtitles | ❌ | N/A |  
8 | DirectPlay | External Subtitles Only | ✅ | N/A |  
9 | DirectPlay | Embedded Audio Only | N/A | ✅ |  
10 | DirectPlay | External Audio Only | N/A | 🟡 | Only if it does not require transcoding
11 | DirectPlay | Embedded Audio + Embedded Subtitles | ✅ | ✅ |  
12 | DirectPlay | Embedded Audio + External Subtitles | ✅ | ✅ |  
13 | DirectPlay | External Audio + Embedded Subtitles | ✅ | 🟡 | Audio only works if it doesn't require transcoding
14 | DirectPlay | External Audio + External Subtitles | ✅ | 🟡 | Audio only works if it doesn't require transcoding
15 | Transcode | Embedded Audio + Embedded Subtitles | ✅ | ✅ |  
16 | Transcode | Embedded Audio + External Subtitles | ✅ | ✅ |  
17 | Transcode | External Audio + Embedded Subtitles | ✅ | 🟡 | Audio only plays the first track
18 | Transcode | External Audio + External Subtitles | ✅ | 🟡 | Audio only plays the first track

✅ Working correctly
❌ Not working
🟡 Partially working or with limitations

---

At this time, I have no embedded subtitles working with external subtitles but that broken embedded + external subtitles.

JPKribs:
@LePips This is close. I've tested this with the files in matrix I provided. The only thing that does not seem to work are Transcodes with DVDSUBS. I believe this is because `SupportsExternalStream = false`??? I'm unsure why this would be but this appears to work for PSG, SRT, ASS but DVDSUB doesn't want to work for transcodes. Those other types, `SupportsExternalStream = true` so it feels like that's related?

Other than that, all other instances appear to be working correctly from my testing. I've reworked the logic a lot and frankly, it actually makes sense now:

---

### Direct Play

1. Order all of the Internal Tracks starting at 0. Order them as Video > Audio > Subtitles > Other(?). Retain the order of the Internal Tracks counting up
2. Order all of the External Tracks starting at the previous index from the internal tracks. Retain the order of the External Tracks counting up

---

### Transcoding

1. Drop all video tracks except for the first one. **(Jellyfin Transcodes contain a single Video Track)**
2. Drop all audio tracks except for our selected one. **(Jellyfin Transcodes contain a single Audio Track)**
3. Order all of the *remaining* Internal Tracks starting at 0. Order them as Video > Audio > Subtitles > Other(?). Retain the order of the Internal Tracks counting up
4. Order all of the External Tracks starting at the previous index from the internal tracks. Retain the order of the External Tracks counting up.

---

Please let me know if you can think of any additional testing I can do to help!

LePips (approved):
Thank you so much for your thorough research and testing of this! I wouldn't have been able to dedicate time for a while to work on this.

Per your testing results, I wouldn't mind putting the table from https://github.com/jellyfin/Swiftfin/pull/1445#issuecomment-2708964127 somewhere in our documentation here.

Even though I haven't tested as thoroughly as you have, I'm going to merge and upload to TestFlight so we can get feedback and possibly ship 1.3 soon.

Issue #1193, opened 2024-08-16
Season Scroller Doesn't Stay Scrolled


JPKribs:
### Describe the bug

This is such a small issue. This is purely just UI weirdness with no functional impact. See the video below for an example:

https://github.com/user-attachments/assets/7b02380f-b2f3-4971-9bf4-00be8d15d5cd

### Application version

1.1.1 & Main

### Where did you install the app from?

App Store

### Device information

tvOS Emulator

### OS version

tvOS 17.4

### Jellyfin server version

10.9.9

Issue #1266, opened 2024-10-09
Episodes list layout improperly when first enter a tv show on iPad


StevenWong12:
### Describe the bug

A tiny UI layout issue on iPad as demonstrated in title. Please refer to the video below.

https://github.com/user-attachments/assets/2020290b-2795-4f97-9297-e90e13bc6149

### Application version

1.2 & main

### Where did you install the app from?

App Store

### Device information

iPad (8th generation)

### OS version

iPadOS 17.6.1

### Jellyfin server version

10.9.3

JPKribs:
While I was working on another item, I realized what caused this to happen. This occurs when from the `CollectionHStack` when the view loads, it's supposed to scroll to the Next Up Episode for that series. If it's the first episode, this scrolls to index 0 and causes this. The fix is to just have it not scroll when the `PlayButtonItem` is index 0. I will add this fix to the PR with the other episode changes.

JPKribs:
This is resolved in 1.5

PR #1446, opened 2025-03-04
[tvOS] Season Selector Scrolling Bug

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1193

I've moved the `SeasonHStack` to its own file and added a scrollTo as well to have it scrollTo the `PlayButtonItem`'s Season. I was originally going to make this a `CollectionHStack` instead of `ScrollView -> HStack` but I was finding SwiftUI was better at handling changing the `SeasonButton` styling on selection without redrawing the view.

Other than this, I updated the `CollectionHStack` to have an ID to better enable the scrollTo. Finally, I cleaned up my previous temporary cards in the `EpisodeHStack` to be camelCase instead of TitleCase to better reflect the rest of Swiftfin. I also moved the `onContentFocus` logic to it's own function because I was getting some type-checker issues. I also added comments to explain the process in case anyone else is looking at it in the future.

**Issue:**

https://github.com/user-attachments/assets/7b02380f-b2f3-4971-9bf4-00be8d15d5cd

**Resolved:**

https://github.com/user-attachments/assets/eb3fb201-1474-43af-ad8c-de94c3c89fe6


JPKribs:
I'm able to resolve https://github.com/jellyfin/Swiftfin/issues/1193 by changing the `SeasonsHStack` into `CollectionHStack`. So far, it all works great with one exception. The `selection.wrappedValue == viewModel.id` doesn't seem to update? That is, the first season focused is highlighted white as normal but it never changes even when the selection changes. I'll need to work on that part because otherwise using the `CollectionHStack` seems to work a lot better!

```
extension SeriesEpisodeSelector {

    // MARK: SeasonsHStack

    struct SeasonsHStack: View {

        @EnvironmentObject
        private var focusGuide: FocusGuide

        @FocusState
        private var focusedSeason: SeasonItemViewModel.ID?

        @ObservedObject
        var viewModel: SeriesItemViewModel

        var selection: Binding<SeasonItemViewModel.ID?>

        @State
        private var didScrollToPlayButtonSeason = false

        @StateObject
        private var proxy = CollectionHStackProxy()

        // MARK: - Extracted helper methods

        private func scrollToSelectedSeason() {
            if let selectedID = selection.wrappedValue,
               let index = viewModel.seasons.firstIndex(where: { $0.id == selectedID }),
               index > 0
            {
                proxy.scrollTo(index: index, animated: false)
            }
        }

        private func onFocusSeasonChanged(_ newValue: SeasonItemViewModel.ID?) {
            guard let newValue = newValue else { return }
            selection.wrappedValue = newValue
        }

        // MARK: - Body

        var body: some View {
            CollectionHStack(
                uniqueElements: viewModel.seasons,
                id: \.id,
                variadicWidths: true
            ) { season in
                seasonButton(viewModel: season)
                    .id(season.id)
            }
            .scrollBehavior(.continuousLeadingEdge)
            .insets(horizontal: EdgeInsets.edgePadding)
            .itemSpacing(EdgeInsets.edgePadding)
            .proxy(proxy)
            .frame(height: 70)
            .padding(.top)
            .padding(.bottom, 45)
            .onChange(of: focusedSeason) { _, newValue in
                onFocusSeasonChanged(newValue)
            }
            .focusGuide(
                focusGuide,
                tag: "seasons",
                onContentFocus: { focusedSeason = selection.wrappedValue },
                top: "top",
                bottom: "episodes"
            )
            .onFirstAppear {
                guard !didScrollToPlayButtonSeason else { return }
                didScrollToPlayButtonSeason = true

                DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
                    scrollToSelectedSeason()
                }
            }
        }

        // MARK: - Season Button

        private func seasonButton(viewModel: SeasonItemViewModel) -> some View {
            Button {} label: {
                Text(viewModel.season.displayTitle)
                    .fixedSize()
                    .font(.headline)
                    .fontWeight(.semibold)
                    .padding(.vertical, 10)
                    .padding(.horizontal, 20)
                    .if(selection.wrappedValue == viewModel.id) { text in
                        text
                            .background(Color.white)
                            .foregroundColor(.black)
                    }
            }
            .buttonStyle(.card)
            .focused($focusedSeason, equals: viewModel.id)
        }
    }
}
```

JPKribs:
This also resolves https://github.com/jellyfin/Swiftfin/issues/1266. When using `proxy.scrollTo(element: playButtonItem, animated: false)`, if the `playButtonItem` is index 0, it scrolls the bar over into the middle of the screen. I am resolving this for iOS as well.

Currently, neither iOS nor tvOS scrolls but changing to `proxy.scrollTo(index:)` re-enables this. I've added a check for index < 0 to prevent #1266.

JPKribs:
I changed the `SeasonHStack` spacing to mirror the `EpisodeHStack` both between items and between the `HStack` and the edges. This was a aesthetic choice that I can undo if we prefer how this originally looked.

Swiftfin tvOS/Views/ItemView/Components/EpisodeSelector/Components/HStacks/EpisodeHStack.swift
```diff
@@ -57,11 +57,44 @@ extension SeriesEpisodeSelector {
 
                 DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
                     guard let playButtonItem else { return }
-                    proxy.scrollTo(element: playButtonItem, animated: false)
```
LePips, line 60: I fixed the underlying issue in the `CollectionHStack` package, so now this line should look like:

```swift
proxy.scrollTo(id: playButtonItem.unwrappedIDHashOrZero)
```

Swiftfin/Views/ItemView/Components/EpisodeSelector/Components/EpisodeHStack.swift
```diff
@@ -43,10 +43,14 @@ extension SeriesEpisodeSelector {
                 guard !didScrollToPlayButtonItem else { return }
                 didScrollToPlayButtonItem = true
 
-                // good enough?
                 DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
                     guard let playButtonItem else { return }
-                    proxy.scrollTo(element: playButtonItem, animated: false)
```
LePips, line 49: Same here.

LePips (changes requested):

Swiftfin tvOS/Views/ItemView/Components/EpisodeSelector/Components/HStacks/EpisodeHStack.swift
JPKribs: Thank you for doing that! I'm wondering if https://github.com/jellyfin/Swiftfin/issues/1266 is better handled at the package as well? The issue occurs when scrollTo(index: 0) occurs. Since you're starting at 0, it does this weird centering.

I'm now realizing my current fix on this PR is no good because sometimes you genuinely do need to scroll to 0. I'll need to think about that.

LePips: I believe that is caused by the layouts getting an odd offset if there's a single element, so yes that would happen within the package instead.

https://github.com/LePips/CollectionHStack/blob/03dc666e8b20ec216fda60f55ccc0eeaabbc5fad/Sources/CollectionHStack/Layouts/ContinuousLeadingEdgeFlowLayout.swift#L20

JPKribs: I've added the `id: \.unwrappedIDHashOrZero` to these `CollectionHStacks`, and I reverted the `CollectionHStacks`'s `ScrollTo`s back to using the element. For now, I am removing #1266 from this PR. I can try and take a shot at this over in CollectionHStack but my UIKit knowledge is pretty limited.

JPKribs:
Okay, I narrowed this PR down a lot more. I updated the description to match what is being resolved. This should purely resolve https://github.com/jellyfin/Swiftfin/issues/1193 now with the added bonus of adding a scrollTo latest season.

LePips (approved):

PR #1447, opened 2025-03-05
[tvOS] Sign-In Flow - User Auto Time-Out

chickdan:
Utilizing https://github.com/ViacomInc/TVOSPicker mentioned in https://github.com/jellyfin/Swiftfin/pull/1419 this PR closely follows the iOS implementation.

https://github.com/user-attachments/assets/8eee3ad3-36c0-4c76-93a5-5a9b54c4e9d9


JPKribs:
In one of the other PR that I have open, I wanted to do something similar to the blurred full screen and was running into a similar error. So, I ended up adding the view to the coordinator and calling it with `.fullScreen` instead of `.push` or `.modal`.

I was basing it off of the item overview, which does this same thing:

https://github.com/jellyfin/Swiftfin/blob/main/Swiftfin%20tvOS/Views/ItemOverviewView.swift

chickdan:
Thank you for the suggestion, it is no longer crashing! Still some work left to do though.

LePips (approved):
This is good for now, albeit simple since we're using another package. I realize that implementing a ticker like this probably isn't difficult in SwiftUI and in today's age of asking an AI to do it.

I'm not against investigating our own picker for this type of data as well as building out the view itself so it isn't so plain. I also thought that per-minute was probably a bit too much and thought 5 minutes would have been better (would also change iOS). Thoughts?

PR #1450, opened 2025-03-14
[Meta] Library & Playback Documentation

JPKribs:
### Summary

This PR adds our Track Selection table to our `player.md` documentation. This is updated from #1445 to include the current state of the Native Player as well. I think that could be replaced with a single sentence instead of a table but I thought I would start with the table and we could trim back if we didn't like it. 

I know some of the player track selection needs to be re-reviewed after the Video Player Manager PR but for now I thought I would add this in to assist in fielding some Issues that get created and give a centralized place for people to review current state.

On this note, I created a page and a table for Library Support. To the best of my ability, I tried to draft up notes to assist in explaining the **WHY** current library types are not in Swiftfin. I want to note: These are **MY UNDERSTANDING** of the situation and do not reflect the project as a whole. I'd love to have these accurately reflect the actual current state but I thought I would start here to get the ball rolling. Easy to review and revise than start from scratch! The reason I added this is I see we get a decent number of Issues reported with similar library support questions so I wanted to include to help alleviate and centralize that response.

For the Library section, please don't hesitate to adjust or remove what doesn't fit @LePips! If we decide this part doesn't make sense, I'm fine to pull that from this PR as well. I didn't link the library section through anywhere but, if we believe this is valuable, this is an item we could link to either the player.md or on the main page.

Let me know!

### DIFF
--- a/Documentation/players.md
+++ b/Documentation/players.md
@@ -102,15 +102,15 @@ Swiftfin offers two player options: Swiftfin (VLCKit) and Native (AVKit). The Sw
 | Subtitle Format | Swiftfin (VLCKit) | Native (AVKit) |
 |----------------|-------------------|----------------|
 | **ASS**        | ✅                | ❌             |
-| **CC_DEC**     | ✅                | 🟡             |
+| **CC_DEC**     | ✅                | ✅             |
 | **DVBSub**     | ✅                | ❌             |
 | **DVDSub**     | ✅                | ❌             |
 | **PGSSub**     | ✅                | ❌             |
 | **SRT**        | ✅                | ❌             |
 | **SSA**        | ✅                | ❌             |
 | **Teletext**   | ✅                | ❌             |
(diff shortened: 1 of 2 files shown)


Documentation/players.md
```diff
@@ -102,15 +102,15 @@ Swiftfin offers two player options: Swiftfin (VLCKit) and Native (AVKit). The Sw
 | Subtitle Format | Swiftfin (VLCKit) | Native (AVKit) |
 |----------------|-------------------|----------------|
 | **ASS**        | ✅                | ❌             |
-| **CC_DEC**     | ✅                | 🟡             |
+| **CC_DEC**     | ✅                | ✅             |
```
JPKribs, line 105: Updated these as they *are* supported for Native Player. Then, the Track Selection section notes they cannot be selected. I can go either way on this.

LePips (approved):
Looks good! Just have to keep up to date if we're going to track PRs in documentation.

PR #1451, opened 2025-03-15
[tvOS] "Native" Styled Menu Button

JPKribs:
### Summary

The goal of this PR is to make a Menu Button, similar to `ListRowButton` or `ChevronButton`. This way, the style of our tvOS Menus. This led me to create a `ListRowMenu`. The goal will be to use this Menu in place of our existing menus.

Below are a video example of the new button next to existing buttons for comparison. I think the animation is just **barely** off but it's very very close to native. I've also replaced the Menus in the 3 locations that I know they exist in tvOS. If there are any I'm missing please let me know!

### Video

In this video is a Chevron Button, this new `ListRowMenu`, and our previous Menu that we were using. I've tried to exactly match our animations to default `.card` buttons but I still have a bit of work to do.

https://github.com/user-attachments/assets/9d915a4e-adb4-4f27-b1bf-4394373b6e4d

---

### AppSettingsView

---

**Before**

https://github.com/user-attachments/assets/e3bd4d62-0cf1-4192-bcba-0fd75cdb6691

---

**After**

I also changed the `SignOutIntervalSection` to use a `ChevronButton` to route to `Duration`.

https://github.com/user-attachments/assets/d06e9bad-39fb-41e2-bae2-43cbe3d903c1

---

### PlaybackQualitySettingsView

---

**Bitrate Picker**

This wasn't a menu before but this was an Inline picker which looks great but is very hard to use with this many selections. It's very easy to click past an option.

https://github.com/user-attachments/assets/2b3fa5cb-ffeb-4024-ac5e-21e0a03f16bc


LePips (commented):
I'm glad that you were able to mimic the native focused styling.

If you think there is more work to do or you would like to remove `InlineEnumToggle` here that would be just fine. We can also implement the functionality to pass in a binding to a `CaseIterable` value and it will automatically implement a `Picker`.

JPKribs:
> We can also implement the functionality to pass in a binding to a `CaseIterable` value and it will automatically implement a `Picker`.

I've added this to my latest commit and used this for the `PlaybackQualitySettingsView`. It works right but let me know if those inits weren't what you were expecting. The `init<ItemType>` is pretty new to me so no hard feelings if I'm completely wrong there.

As for the `InlineEnumToggle`, I actually think those are better for **MOST** enums. I think the only enums where it's worse are ones like `PlaybackQualitySettingsView` where there are 10+ bitrate options so cycling through and potentially missing one is greater. That being said, I think the dropdown mirrors iOS better so I can go either way. I think we only have 3 menu/large enums in tvOS settings.

LePips:
I would very much like to get rid of `InlineEnumToggle` - it will become our most basic data selector above `Toggle`. For selections that require more detail we can still use the current `EnumPickerView`, like for fonts where we would like to show the actual font with the name.

JPKribs:
Haha sounds good. I've made that update on my end! Here is what this looks like now for things likes Player selection:

https://github.com/user-attachments/assets/06fc88bc-f2b5-4dec-916d-e90d03075066


---


Let me know if there is anything else that would be good to add here! I'm 99% happy with that menu button. I've pinpointed what is off. The focus looks fine but the defocus, the text seems to defocus at a slower rate than the rest of the button. You really have to be staring at it to notice it but I think that's the only point that looks off to me.

LePips (approved):

LePips:
I noticed something subtle as well and we're now in the uncanny valley. I'm overall good with this design but improvements are always welcome.

PR #1453, opened 2025-03-17
[iOS] External display wide ratio support

quangDecember:
### Summary

Current SwiftFin app only mirror directly what's on iPad screen to external display, therefore when plugging in a monitor or projector, there's huge black bars on the sides. 

### Changes

Use built-in function supported by AVKit, applying to native player

### After 

Main display
IMG_2781
External display
IMG_2780

P/s: if this small ticket works I'm happy to continue working on the VideoPlayer struct (the VLC player)


LePips (approved):
Thank you for the contribution! I would heavily recommend not working on the current state of other video player, as that is rewritten in #1203.

PR #1455, opened 2025-03-17
Drop iOS 15

LePips:
Drop iOS 15 support.

See [this discussion](https://github.com/jellyfin/Swiftfin/discussions/1564).


eisams:
FYI: There's no devices that would lose support if you were to require tvOS 18 compared to the current tvOS 15, so you could consider bumping the minimum further if that makes overall development easier

Screenshot 2025-03-19 at 17 28 23

https://en.wikipedia.org/wiki/TvOS#Supported_OS_releases

LePips:
Thank you, I have made that observation before: https://github.com/jellyfin/Swiftfin/pull/1076#issuecomment-2542035586

When we drop iOS 16, probably later this year to keep up with an n-2 support policy, we can make the jump to iOS 18 due to the supported devices. tvOS will most likely happen sooner.

JPKribs (approved):
🎉

Issue #1157, opened 2024-07-23
PosterButton Context Menus


JPKribs:
**Describe the feature you'd like**

I would be happy to work on this but I wanted to make a request here to make sure I wasn't stepping on anyone's toes before I started in earnest. The goal for this feature would be to copy item options to be accessible from the pagingLibraryView instead of needing to open them up to access the settings. This would be in addition to options on the ItemView. 

An example of this would be going over your list of movies and realizing that one was already watched so you can mark it as played without having to click into it. 

### iOS / iPadOS
1. Long press on a Movies, Series, or Episode
2. Get all options available for selection without entering the actual Media Item (Play, Shuffle, Played, Favorite, etc.)

### tvOS
1. Hold down the select button on a Movies, Series, or Episode
2. Get Non-Playback options (Favorite & Mark as Played)
-----
1. Hold down the play button on a Movies, Series, or Episode
2. Get Playback options (Play / Play from Beginning / Resume & Shuffle)

### Example Images

**Series**

<img width="232" alt="Screenshot 2024-07-23 at 14 00 55" src="https://github.com/user-attachments/assets/603e5d1e-617f-4a07-9e94-39323e2db8f1">

**Movie**

<img width="260" alt="Screenshot 2024-07-23 at 14 00 28" src="https://github.com/user-attachments/assets/9cd661a5-afe2-4b99-8a97-fed7d184c1c7">

**Additional context**
I was looking at this on my end and I already have some of these actions figured out for iOS. I see that there was previously some work for the resume vs play from beginning that I can help sort out first. Shuffle and Playlists are just filler for now as more of a down-the-road option along with downloads. I already have the Play (For Movies), Favorite, and Mark as Played options working but they don't update the posterButton so I would still need to figure out how to refresh that.

LePips:
Yes, I've always wanted more context menu options. The main thing that has prevented it in the past from attaching it to every item is just how to update the item throughout the app. This should be possible now though by just having us multi-purpose the notification for an item being updated for view models that hold an `OrderedSet` of items.

LePips:
Considering closed as #2075 has implemented the architecture for context menus. More options can be added.

PR #1457, opened 2025-03-19
[WIP] [iOS] `PosterButton` - Context Menus

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1157

This PR creates context menus on the PosterButtons for iOS. This PR may expand to tvOS but I want to make sure I am doing this correctly/efficiently before I duplicate that work in tvOS.

The approach that I took was to use modifiers, identical to the `onSelect` modifier. This way, only the required context menu options would be used. I built out all of the context menus that I thought we would see:

*Functional*
- onToggleIsPlayed
- onToggleIsFavorite
- onRefresh
- onDelete

*Not Functional but Currently Possible*
- onPlay
- onRestart

*Not Functional & Requires Additional Work*
- onShuffle
- onAddToPlaylist
- onDownload

The reason I built this all out instead of using using the existing `.contextMenu` was to avoid different areas of the app accidentally creating differing designs for the same button. Building them onto the `PosterButton` ensures that all of these functions are identical throughout the app.

At this time, I have the following questions:

1. Is making these modifiers the correct call? Should this just be in the existing `.contextMenu` instead?
2. Where do these context menus need to exist? I only have `PagingLibraryView` for now but I could see doing this on the `HomeView` as well.
3. Are all of the functions above wanted? Something like `onPlay` vs `onRestart` only really saves a single click while `onDelete` could get a user into trouble much quicker than our current deletion options.
4. This is going to require notifications to update individual `PosterButton` I think? For now, Played and Favorite work but the `PosterButton` does not update for it. This is purely POC but I think that's a bigger part of this.

---

### WIP Screenshot

Simulator Screenshot - iPhone 16 Pro - 2025-03-26 at 21 26 22

Simulator Screenshot - iPhone 16 Pro - 2025-03-26 at 21 41 48

Simulator Screenshot - iPhone 16 Pro - 2025-03-26 at 21 43 10

Simulator Screenshot - iPhone 16 Pro - 2025-03-26 at 21 44 33

### DIFF
--- a/Shared/Components/PosterButton/Components/ContextMenu.swift
+++ b/Shared/Components/PosterButton/Components/ContextMenu.swift
@@ -12,15 +12,20 @@ import SwiftUI
 
 extension PosterButton {
 
-    struct ItemContextMenu: View {
+    struct ContextMenu: View {
 
         // MARK: - Current User Session
 
         @Injected(\.currentUserSession)
         private var userSession
 
+        #if os(tvOS)
         @EnvironmentObject
-        private var mainRouter: MainCoordinator.Router
+        private var router: ItemCoordinator.Router
+        #else
+        @EnvironmentObject
+        private var router: MainCoordinator.Router
+        #endif
 
         // MARK: - PosterButton ContextMenu Variables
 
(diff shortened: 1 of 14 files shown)


JPKribs:
Having now seen:

https://github.com/jellyfin/Swiftfin/pull/1484

I believe the approach that I am currently taking a in this PR is incorrect. I'm going to retain this branch just in case but I'm going to close out this PR.

I know this is the second time that I've opened and closed a branch related to context menus. It's still something that I'd like to do, but the poster button is used in a lot of locations. So, doing it correctly is important. Also, I think context menus are a nice quality of life, but are hard to prioritize over some other items.

PR #1462, opened 2025-03-25
[WIP] [iOS & tvOS] Internal Text Subtitle Offset

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1256#issuecomment-2738771373
Similar issue to: https://github.com/jellyfin/jellyfin-webos/issues/85

I don't totally understand this. In my testing on the original fix (https://github.com/jellyfin/Swiftfin/pull/1445) I guess I missed this because I was only using English subtitles and all of my internal subtitles are PGS. The issue appears to be that, when there are _ONLY_ internal SUBRIP subtitles, the first and last subtitles have swapped `DeliveryURLs`. I do not understand _WHY_ this is happening but I am able to recreate this consistently when testing with internal SUBRIP subtitle only. External Text-Based subtitles appear to work without issue. Internal Text-Based subtitles appear to work without issue IF there are external subtitles OR a graphic PGS subtitle included. 

This logic switches the first and last subtitle tracks IF there are only text based subtitles internally. I will need to do some testing since I don't have a lot of files available to test this. Just getting this out there while I test this.

This PR also drops subtitles from transcoding where `.isSupportsExternalStream` is false since those are not selectable.


JPKribs:
This is passing all my testing but I honestly have no idea WHY this happening unless there is something wonky with the API? Sorry to bother @LePips, but would you be able to test this out on your end and make sure I'm not missing anything? The fact this varies based on subtitle type makes be nervous I'm missing the bigger picture here...

JPKribs:
I'm closing this out until after https://github.com/jellyfin/Swiftfin/pull/1463. I don't think that is going to resolve it but this issue is too inconsistent to properly replicate and test. Maybe the API changes resolve this but I'll keep an eye on this.

PR #1465, opened 2025-03-30
Stateful - Set cleanup and `final` classes

LePips:
In #1203 I've been needing to remove `lastAction` and have been meaning to do some other cleanup with `Stateful`, as well as making the necessary view models `final` and removing `final` from stored variables as they aren't necessary.


JPKribs (approved):
Makes sense!

PR #1467, opened 2025-03-30
[iOS 16] - Replace `WrappingHStack` for `AttributeView`

JPKribs:
### Summary

Requires: https://github.com/jellyfin/Swiftfin/pull/1455

This PR replaces `WrappingHStack` with a Layout for iOS 16. This layout should roughly mirror `WrappingHStack` but with some additional flexibility. This is tested and functional but I'll need to revisit this after iOS 15 is dropped.

This also fixes the layout not aligning correctly on iPadOS introduced by `WrappingHStack`.

### Screenshots

<details>
<summary>Cinematic - Normal & Large Text</summary>

Cinematic - Normal Text

Cinematic - Large Text

</details>

<details>
<summary>Logo - Normal & Large Text</summary>

Logo - Normal Text

Logo - Large Text

</details>

<details>
<summary>Compact - Normal & Large Text</summary>

Compact - Normal Text

Compact - Large Text

</details>


JPKribs:
I still don't totally understand layouts. This was a ton of trial and error and AI really failed me haha. That being said, I am happy with this and think this is an improvement over the existing `WrappingHStack`.

I chose to fit everything possible in the bottom row, then start filling the top row. Which was harder than needed because I still needed to retain the order to I'm reversing the array a couple times to get that working right to make sure the top row is the smaller row PLUS the order still goes top to bottom in order.

The impact of this is that sometimes the size is *just* big enough that only 1 item gets pushed up so it can look a little weird. looking. I personally think it looks... fine... but here are some alternatives:

1. If 1 item needs to go up, move 2 items up. Make 2 the minimum for a second row.
2. Make row 1 and 2 even so they should be *roughly* the same size if 2 rows are required.

Swiftfin/Views/ItemView/Components/AttributeView/AttributeLayout.swift
```diff
@@ -0,0 +1,179 @@
...
+                rowHeight: bottomHeight,
+                y: bounds.minY + (topRow.isEmpty ? 0 : topHeight + lineSpacing),
+                alignment: alignment
+            )
+        }
+
+        // MARK: - Calculate Row Width
+
+        private func calculateRowWidth(indices: [Int], sizes: [CGSize]) -> CGFloat {
+            guard !indices.isEmpty else { return 0 }
```
chickdan, line 75:
```suggestion
            guard indices.isNotEmpty else { return 0 }
```

Swiftfin/Views/ItemView/Components/AttributeView/AttributesView.swift
```diff
@@ -7,33 +7,41 @@
...
+
         // MARK: - Body
 
         var body: some View {
             let badges = computeBadges()
-            if !badges.isEmpty {
-                WrappingHStack(badges, id: \.self, alignment: .center, spacing: .constant(8), lineSpacing: 8) { badgeItem in
-                    badgeItem
+
+            AttributeLayout(alignment: alignment, spacing: 8, lineSpacing: 8) {
```
chickdan, line 34: Might I suggest to wrap this in an `if #available` conditional so that this can be merged without bumping to iOS 16.

chickdan: Alrighty, only suggesting so there aren't too many lingering PRs is all :slight_smile:

JPKribs: No worries! I'm with you but this is a rare one I'm find leaving out there for the changeover. Otherwise, it's making the change then leaving myself a note to update this when iOS15 is dropped. I did that for 10.8 -> 10.10 and found myself spending a lot longer than I'd like just going over old notes haha.

JPKribs:
@LePips do we want this? I think you already cleaned up the issues with my `WrappingHStack` usage in https://github.com/jellyfin/Swiftfin/pull/1543. So, this isn't as urgent as it would have been. Cleaning up this PR would let us drop `WrappingHStack` but we'd have to maintain the `Layout`. It gives us more control over how the attributes stack but we'd have to decide if we see that payoff.

I can go either way! I just wanted to check with you before squashing these build issues.

LePips:
I would very much still like this, I'm not worried about maintaining it. There are nuances about the current implementation that are abstracted away with `Layout`.

Swiftfin/Views/ItemView/Components/AttributeView/AttributeLayout.swift
```diff
@@ -0,0 +1,179 @@
...
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2025 Jellyfin & Jellyfin Contributors
+//
+
+import SwiftUI
+
+extension ItemView {
+
+    struct AttributeLayout: Layout {
```
LePips, line 13: Please rename this `FlowLayout` and move it to a separate file. I think we can also get a simpler implementation asking Claude or another AI - at least, they give me something relatively simpler and easier to follow.

LePips (changes requested):

Swiftfin/Views/ItemView/Components/AttributeView/AttributeLayout.swift
JPKribs: I think there is like a 90% chance I bork this in Git so if I randomly open another PR it's because something went wrong lol

JPKribs: I borked it... New PR: https://github.com/jellyfin/Swiftfin/pull/1569... Sorry about the extra hassle, I still get myself into Git trouble when branches get too far behind.

PR #1468, opened 2025-04-01
[tvOS] Move AddUserButton to the bottom bar of Select User View

hevey:
This PR cleans up the Select User View as mentioned in the Quality of Life section in the tvOS status up discussion post. 

It change two things. The first added an add user button to the bottom bar of view and replaces the Add User button in the GridView with a No User View when there is no logged in users. Screenshots for both are below. 

The new Add User Button
image

No User View
image


JPKribs:
Thank you for your commit! This looks great! Getting that button to match the menu style buttons on the bottom row was something I was struggling with so this is a good start IMO.

I have a couple layout changes if you wouldn't mind:

1. Can we change this `addUser` button to be iconOnly? Just to give us some symmetry when all buttons are present.
2. Would we be able to use our original `addUserButton`? The 2 reasons being to mirror how this looks on iOS and that button had some logic built in to handle `.server()` & `.all` as a dropdown so that helps us avoid the issue where the login often isn't totally clear which server it's the login for.

Example:
Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-04-01 at 10 50 28

Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-04-01 at 10 56 09

---

We have one issue that comes up from this that we want to address. For whatever reason, selecting "Switch User" from the settings, on first load we end up with this:

Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-04-01 at 10 58 28

Which originally was resolved here: https://github.com/jellyfin/Swiftfin/pull/1441 but for some reason that Add User button is structural?? Would you be able to look at this as well? It only seems to happen on first load into the view. Changing the Server to .all then back resolves this. I believe it's resolved because it adds the Server Name under the user, redrawing the view, and fixing this.

Altogether, this looks great! Please let me know if you have any questions/comments on the 3 items above. 

Again, thank you for helping out with Swiftfin!!!

hevey:
> I have a couple layout changes if you wouldn't mind:
> 
> 1. Can we change this `addUser` button to be iconOnly? Just to give us some symmetry when all buttons are present. Also, the 400 width helps account for some longer server names or servers that are pure IP address:ports.

No probs, will swap these about 
> 2. Would we be able to use our original `addUserButton`? The 2 reasons being to mirror how this looks on iOS and that button had some logic built in to handle `.server()` & `.all` as a dropdown so that helps us avoid the issue where the login often isn't totally clear which server it's the login for.

To clarify do you want a button still in the grid view or just a view to mimic the look of the button?

> We have one issue that comes up from this that we want to address. For whatever reason, selecting "Switch User" from the settings, on first load we end up with this:
> 
> Which originally was resolved here: #1441 but for some reason that Add User button is structural?? Would you be able to look at this as well? It only seems to happen on first load into the view. Changing the Server to .all then back resolves this. I believe it's resolved because it adds the Server Name under the user, redrawing the view, and fixing this.

easy can certainly look at what is causing this

JPKribs:
> No worries, just some clarifications would you like the add button to be in the grid still when no user is logged in but to remove it and have the add user button in the bottom menu when there is a logged in user on a server? I can re-work the bottom button to handle the all server logic also if this is the case so it works similarly to the grid button.

I didn't think about that but I think that is a good call. That button to add users when there are existing users. Otherwise, if there are no users we can use the `addUserButton`. Personally, I think just your button looks a lot cleaner but the Wife-Test says having a more prominent button on the empty grid is valuable since it does a good job of directing the user. 

I think this is similar to our iOS admin items. If there are no items, we should have a prominent "ADD" button but if there are items, we should use the cleaner button tucked away. See below for a loosely related version on iOS:

Screenshot 2025-04-01 at 16 14 58

---

> easy can certainly look at what is causing this

I really appreciate that! Let me know if there is anything I can provide to help out!

hevey:
> > No worries, just some clarifications would you like the add button to be in the grid still when no user is logged in but to remove it and have the add user button in the bottom menu when there is a logged in user on a server? I can re-work the bottom button to handle the all server logic also if this is the case so it works similarly to the grid button.
> 
> I didn't think about that but I think that is a good call. That button to add users when there are existing users. Otherwise, if there are no users we can use the `addUserButton`. Personally, I think just your button looks a lot cleaner but the Wife-Test says having a more prominent button on the empty grid is valuable since it does a good job of directing the user.
> 
> I think this is similar to our iOS admin items. If there are no items, we should have a prominent "ADD" button but if there are items, we should use the cleaner button tucked away. See below for a loosely related version on iOS:
> 
> Screenshot 2025-04-01 at 16 14 58
> > easy can certainly look at what is causing this
> 
> I really appreciate that! Let me know if there is anything I can provide to help out!

Alright I will roll with this and have a crack at it all tonight.

hevey:
ok so I have reworked the bottom bar now to allow for the server button to be the original size
Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-04-02 at 21 01 00

when the server view doesn't have a logged in server the new bottom bar button is removed and the original AddUser Button in the grid is added
Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-04-02 at 21 01 06

Also when the the all servers option is used the new AddUserBottomButton provides the same server picker menu
Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-04-02 at 21 01 20

JPKribs:
> At present I don't think I've introduced it so wanted to know whether this has been observed before. I think it might be related to some media not loading in on the top row initially after a login.

I've seen that as well prior to this. I have not taken the time to investigate it but I know there are some bigger ambitions for the `HomeView` which should also incorporate that fix.

For this PR, this looks awesome! I like the bottom button switches to a Menu for .all as well. That's a good touch. I'll take a deeper look at this tomorrow but I think we're approaching the point where we can run this all by the expert (@LePips).

hevey:
I just pushed a smaller clean up commit so nothing feature wise has changed. I haven't had a chance to look at the UI clipping bug yet but if I get an opportunity tonight I will. 

If this gets merged before I get to it, I am happy to look at it and any of the are UI bugs I have seen as a separate PR

Shared/Strings/Strings.swift
```diff
@@ -756,6 +756,8 @@ internal enum L10n {
   internal static let lockedFields = L10n.tr("Localizable", "lockedFields", fallback: "Locked Fields")
   /// Locked users
   internal static let lockedUsers = L10n.tr("Localizable", "lockedUsers", fallback: "Locked users")
+  /// Login
+  internal static let login = L10n.tr("Localizable", "login", fallback: "Login")
```
JPKribs, line 760: Since we're not using this anymore let's pull it.

JPKribs (changes requested):
This looks good to me! One change removing that extra label we're not using anymore. 

I'll defer to LePips for the final review but from my end this looks good.

LePips:
Apologies for my large changes. My first comment was going to be "please remove the grid item stuff because that should not only be used for users" and would be a lot cleaner but that is a lot of work, so I did it myself instead of getting you to read through this large work. Then, I got sucked into a meeting I had to listen to for hours and cleaned up a lot of stuff from my initial implementations while I was at it.

I'll be doing some last check ups and testing soon.

Shared/Strings/Strings.swift
hevey: All sorted

hevey:
> Apologies for my large changes. My first comment was going to be "please remove the grid item stuff because that should now only be used for users" and would be a lot cleaner but that is a lot of work, so I did it myself instead of burdening you to work through that. Then, I got sucked into a multi-hour listen-only meeting and cleaned up a lot of stuff from my initial implementations while I was at it.
> 
> I'll be doing some last check ups and testing soon.

No worries, that makes sense now I think about it. 
Just cleaning up the localisation I noticed now when picking a server with no user logged in I have lost the bottom menu. 

Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-04-05 at 08 23 56

LePips:
Apologies I've ballooned this, this view required a lot of cleanup and more I can think about. Just took advantage of some free time.

Thank you for the contribution!

There is a SwiftUI bug with larger add user grid button that adds some padding compared to just being a button, but besides that please test and check if it works as expected.

JPKribs (approved):

PR #1473, opened 2025-04-05
[tvOS] Various Setting Cleanup

JPKribs:
### Summary

This PR I localize some "Strings" for both iOS and tvOS. 

Some of the setting views had `.withDescriptionTopPadding()` and some did not. I've updated this so it's consistent across the board to you don't get that weird vertical jittering when the top row of two views is just barely off.

The Sign-In View, every button had `.bottom` padding except the "Sign In" button, resulting in in being just a touch narrower than the other buttons on this view. All buttons have have the same padding to ensure their size is consistent.

Finally, and biggest, I am once again trying to tackle this the menu button I created here: https://github.com/jellyfin/Swiftfin/pull/1451

It's not perfect but I synced up the scaling animation so the text and the button background don't reset to default state at different timings. I also changed it from Spring to EaseInOut and I think that mirrors the other buttons a little better.

### Before:

https://github.com/user-attachments/assets/c1ab19ca-1e15-46b0-b4ed-ef627b90981b

### After:

https://github.com/user-attachments/assets/da651255-1dc3-46da-b26a-5902e284dfb4

### DIFF
--- a/Swiftfin tvOS/Components/ListRowMenu.swift
+++ b/Swiftfin tvOS/Components/ListRowMenu.swift
@@ -6,14 +6,6 @@
 // Copyright (c) 2025 Jellyfin & Jellyfin Contributors
 //
 
-//
-// Swiftfin is subject to the terms of the Mozilla Public
-// License, v2.0. If a copy of the MPL was not distributed with this
-// file, you can obtain one at https://mozilla.org/MPL/2.0/.
-//
-// Copyright (c) 2025 Jellyfin & Jellyfin Contributors
-//
-
 import SwiftUI
 
 struct ListRowMenu<Content: View, Subtitle: View>: View {
--- a/Swiftfin tvOS/Components/ListRowMenu.swift
+++ b/Swiftfin tvOS/Components/ListRowMenu.swift
@@ -27,19 +27,19 @@ struct ListRowMenu<Content: View, Subtitle: View>: View {
         Menu(content: content) {
             HStack {
                 title
-                    .foregroundColor(isFocused ? .black : .white)
+                    .foregroundStyle(isFocused ? .black : .white)
                     .padding(.leading, 4)
 
                 Spacer()
 
                 if let subtitle {
                     subtitle
-                        .foregroundColor(isFocused ? .black : .secondary)
+                        .foregroundStyle(isFocused ? .black : .secondary)
                 }
 
                 Image(systemName: "chevron.up.chevron.down")
                     .font(.body.weight(.regular))
-                    .foregroundColor(isFocused ? .black : .secondary)
+                    .foregroundStyle(isFocused ? .black : .secondary)
                     .brightness(isFocused ? 0.4 : 0)
             }
             .padding(.horizontal)
--- a/Swiftfin tvOS/Components/ListRowMenu.swift
+++ b/Swiftfin tvOS/Components/ListRowMenu.swift
@@ -36,33 +36,21 @@ struct ListRowMenu<Content: View, Subtitle: View>: View {
 
     private var labelView: some View {
         HStack {
-            ZStack {
-                title.foregroundColor(.black)
-                title.foregroundColor(.white)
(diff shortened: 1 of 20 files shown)


Swiftfin/Views/SettingsView/CustomDeviceProfileSettingsView/CustomDeviceProfileSettingsView.swift
```diff
@@ -56,7 +56,7 @@ struct CustomDeviceProfileSettingsView: View {
             Section(L10n.profiles) {
 
                 if customProfiles.isEmpty {
-                    Button("Add profile") {
+                    Button(L10n.addProfile) {
```
LePips, line 59: Let's just use `L10n.add` instead.

Swiftfin tvOS/Views/SettingsView/CustomDeviceProfileSettingsView/CustomDeviceProfileSettingsView.swift
```diff
@@ -58,14 +58,14 @@ struct CustomDeviceProfileSettingsView: View {
...
-                            Label("No profiles defined. Playback issues may occur.", systemImage: "exclamationmark.circle.fill")
+                            Label(L10n.noDeviceProfileWarning, systemImage: "exclamationmark.circle.fill")
                         }
                     }
                 }
 
                 Section {
                     if customProfiles.isEmpty {
-                        Button("Add profile") {
+                        Button(L10n.addProfile) {
```
LePips, line 68: Let's just use L10n.add instead.

Swiftfin tvOS/Components/ListRowMenu.swift
```diff
@@ -25,32 +25,31 @@ struct ListRowMenu<Content: View, Subtitle: View>: View {
...
-                    }
+                Spacer()
 
-                    Image(systemName: "chevron.up.chevron.down")
-                        .font(.body.weight(.regular))
-                        .foregroundStyle(isFocused ? Color.black : Color.secondary)
-                        .brightness(isFocused ? 0.4 : 0)
+                if let subtitle {
+                    subtitle
+                        .foregroundColor(isFocused ? .black : .secondary)
```
LePips, line 37: Use `foregroundStyle`.

Swiftfin tvOS/Components/ListRowMenu.swift
```diff
@@ -25,32 +25,31 @@ struct ListRowMenu<Content: View, Subtitle: View>: View {
...
-
-                HStack {
-                    title
-                        .foregroundStyle(isFocused ? Color.black : Color.white)
-                        .padding(.leading, 4)
-
-                    Spacer()
+            HStack {
+                title
+                    .foregroundColor(isFocused ? .black : .white)
```
LePips, line 30: Use `foregroundStyle`.

Swiftfin tvOS/Components/ListRowMenu.swift
```diff
@@ -25,32 +25,31 @@ struct ListRowMenu<Content: View, Subtitle: View>: View {
...
-                        .brightness(isFocused ? 0.4 : 0)
+                if let subtitle {
+                    subtitle
+                        .foregroundColor(isFocused ? .black : .secondary)
                 }
-                .padding(.horizontal)
+
+                Image(systemName: "chevron.up.chevron.down")
+                    .font(.body.weight(.regular))
+                    .foregroundColor(isFocused ? .black : .secondary)
```
LePips, line 42: Use `foregroundStyle`.

LePips (changes requested):
Per the top comment of `SplitFormWindowView` where I previously questioned if the top padding was necessary, it looks like it is. Remove the function and just always add that padding instead.

JPKribs:
> Remove the function and just always add that padding instead.

Sounds good! This should all be in my latest commit.

LePips (approved):
Thanks for the cleanup!

PR #1475, opened 2025-04-06
[tvOS] External Trailer Fix

JPKribs:
### Summary

Today, I build the Main tvOS build on my actual Apple TV and I found that external trailers always failed to open. I needed to add this to info.plist:

```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<array>
	<string>youtube</string>
</array>
</plist>
```

This is IF we want to proceed with this. I've tested this on a real device and can confirm this now works. I wasn't even aware there was a way to build on a remote Apple TV because I never looked into it...

I think the two major reasons we wouldn't want to do this are: 
1. Not all external trailers are going to be YouTube. Do we want to add URL schema for all various sources that come up? This opens us up to the requests for "Enable Trailers for X Platform" which adds some additional maintenance.
2. From a FOSS perspective, linking into Google's YouTube can feel a bit taboo. I don't personally see the issue but I can understand why that might less than ideal to some folks.

Let me know if we want to proceed!

### Proof

https://github.com/user-attachments/assets/1864c765-88a7-40de-ab4a-bed7c5737034

### DIFF
--- a/Swiftfin tvOS/Resources/Info.plist
+++ b/Swiftfin tvOS/Resources/Info.plist
@@ -18,6 +18,10 @@
 	<string>$(CURRENT_PROJECT_VERSION)</string>
 	<key>CFBundledisplayTitle</key>
 	<string>Jellyfin</string>
+	<key>LSApplicationQueriesSchemes</key>
+	<array>
+		<string>youtube</string>
+	</array>
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
 	<key>NSAppTransportSecurity</key>
--- a/Swiftfin tvOS/Views/ItemView/Components/ActionButtonHStack/Components/TrailerMenu.swift
+++ b/Swiftfin tvOS/Views/ItemView/Components/ActionButtonHStack/Components/TrailerMenu.swift
@@ -136,11 +136,22 @@ extension ItemView {
         // MARK: - Play: External Trailer
 
         private func playExternalTrailer(_ trailer: MediaURL) {
-            if let url = URL(string: trailer.url), UIApplication.shared.canOpenURL(url) {
+            if let trailerUrlString = trailer.url {
+                let youtubeDeepLink = "youtube://\(trailerUrlString)"
+                if let youtubeURL = URL(string: youtubeDeepLink), UIApplication.shared.canOpenURL(youtubeURL) {
+                    UIApplication.shared.open(youtubeURL) { success in
+                        if !success {
+                            error = JellyfinAPIError(L10n.unableToOpenTrailer)
+                        }
+                    }
+                    return
+                }
+            }
+            if let urlString = trailer.url, let url = URL(string: urlString), UIApplication.shared.canOpenURL(url) {
                 UIApplication.shared.open(url) { success in
-                    guard !success else { return }
-
-                    error = JellyfinAPIError(L10n.unableToOpenTrailer)
+                    if !success {
+                        error = JellyfinAPIError(L10n.unableToOpenTrailer)
+                    }
                 }
             } else {
                 error = JellyfinAPIError(L10n.unableToOpenTrailer)


JPKribs:
@LePips no rush since this only impacts tvOS / Main. I've built this out more. Instead of hardcoding YouTube, I made a real `DeepLinkURL` Struct for this. Testing and working on a real world Apple TV. I only really tested Vimeo and YouTube.

I'm honestly up in the air whether we want to support this type of linking. Entirely your call on this. I've built this out more so adding more in the future / maintaining is easier but this is a bit messy.

JPKribs:
> I'm curious what happens if those apps are not installed... will the user be prompted to install them?

Right now it errors with: "Unable to open trailer in ApplicationName"

I tested that with a URL to Vimeo when I uninstalled. I could make it more of a call to action or even link in the app store but then we'd have to keep those links up to date if they ever change. I don't imagine App Store URLs change all that much though?

dkanada:
Whether or not we should link to external apps for trailers feels like more of a server question. Since the server currently supports it I'd say it's a net positive that iOS and tvOS include support. It would be similar to linking IMDb for a movie when the external ID exists on the server.

JPKribs:
> Whether or not we should link to external apps for trailers feels like more of a server question. Since the server currently supports it I'd say it's a net positive that iOS and tvOS include support. It would be similar to linking IMDb for a movie when the external ID exists on the server.

That makes complete sense! Most of my hesitance is that would make Swiftfin one of the few apps that "requires" another app downloaded to work. Web, trailers embed in the same browser window so it feels very cohesive. To add to it, I don't think I've used an Apple TV app that opens another app? I was clicking through this morning and I think Swiftfin would be the only app that has this type of handoff. Not that it's a problem just part of why I wanted to tread carefully when looking at this.

dkanada:
Your comments also make sense - it's exactly why I don't bother with trailers in my own case haha. For what it's worth I'm certain the Android TV client supports YouTube but I'm not sure about other services. I also believe some Apple TV+ integrations link out to external apps if you select a show only on their service but I haven't looked at those in a while.

JPKribs:

LePips:
Just waiting to personally test on my Apple TV.

JPKribs:
I just did some cleanup to fix the complexity warning. Please let me know if we'd prefer the original "complex" route instead. This new route doesn't trigger the complexity warning but I worry about the nested loop.

The goal/issue this is used for is identifying the app that the trailer URL should open. For example, parsing the URL "youtube.com/abc" and identifying this is YouTube and attaching the prefix "YouTube://" to the URL so tvOS knows to use the YouTube app. This also needs to account for youtu.be, m.youtube.com, etc. On first call, I create my map to make this easier to query. I could also manually create this map. The reason I have this create on call was to reduce complexity in maintaining/expanding this list.

This is only called once a trailer on tvOS is selected. I am completely open to alternative structures!

geonizeli:
@LePips maybe you forgot about this PR. Don't want to disturb but I am anxious about the next tvOS release since we don't have TestFlight for tvOS.

JPKribs:
> @LePips maybe you forgot about this PR. Don't want to disturb but I am anxious about the next tvOS release since we don't have TestFlight for tvOS.

This one is more on me. This can't merge with Main right now so I need to do some cleanup. Additionally, there are some things I can improve on for our final version.

I'm out of pocket for a bit so I won't have a chance to get this back on track but I'll request another review when this is actually ready again.

JPKribs:
Sorry about that delay! @LePips ready when you are! Tested on real hardware using Vimeo and YouTube. I tried to make `ExternalTrailerSource`.`SourceType` as easy to maintain/add to as possible. I think the cost of the maintainability is the lookup to go from External URL to formatted youtube://URL could be more performant if we just left everything in prebuilt maps rather than arrays. I believe making this more maintainable to add/change items in the enum >> an extra loop when starting a trailer.

But please let me know if you have any recommendations for this!

LePips (changes requested):
Please retest with my changes, but they should largely be the same with different names.

I'm also iffy on the untested source schemes. For example, I've seen elsewhere that Netflix has used `nflx://`, Amazon Video has used `aiv://`, or that all of these sources support deep linking into their apps to begin with. What sources are you using to determine the schemes and support of deeplinking in these apps?

I would only care to keep the sources we know that work and/or would mostly be pulled by metadata providers (for example, I highly doubt the prevalence of trailers from Twitch).

JPKribs:
I *had* a source in my notes but am now getting:

Screenshot 2025-11-01 at 11 49 49

I agree that we should only use vetted links. I'm loading this up on my tv now and I can test everything out to make sure we have something more sound than a dead link I found on Reddit! I'll test this now and get back to you with that tested.

JPKribs:
@LePips is this how you would recommend this? This functionally works but I wasn't sure if your route you'd prefer 2 entries like this or another route. IF we wanted to keep Twitch, we'd need [something similar](https://dev.twitch.tv/docs/mobile-deeplinks/). *I'm also game to drop Twitch or at least limit to like clips*

```swift
    static let appleTV_Show = Self(
        displayTitle: "Apple TV",
        deepLinkScheme: "com.apple.tv://",
        hostPatterns: ["tv.apple.com"],
        requiresPathValidation: true,
        requiredPathComponent: "/show/"
    )

    static let appleTV_Movie = Self(
        displayTitle: "Apple TV",
        deepLinkScheme: "com.apple.tv://",
        hostPatterns: ["tv.apple.com"],
        requiresPathValidation: true,
        requiredPathComponent: "/movie/"
    )
```

JPKribs:
Okay, I've tested what I could. Part of my issue is that some of these I am able to get it to open the apps but I can't check if it routes correctly without a paid account so I can't validate from there. Some of these ([Amazon Prime](https://www.amazonforum.com/s/question/0D56Q0000AXbgHYSQZ/is-deeplinking-to-prime-video-app-content-from-apple-tvos-app-possible)) have deeplinks for iOS that do not work on tvOS and there isn't good documentation for finding it.

This left me with Apple TV, Vimeo, and Youtube. I'm personally fine trimming the list to those 3 and, if it comes up in the future, we have the infrastructure to add onto this. I feel as though Youtube is our primary concern since I believe that's the only external source Jellyfin will get by default? I could be wrong about that but with my non-fancy library everything is youtube.

Vimeo, I added some REGEX so the format can come in like a normal URL and get transformed into the appropriate `vimeo://video/id` as opposed to having to format it like that to begin with.

I've included videos of testing below. Please note that the black, unlit part is my TV dying, unrelated to Swiftfin:

<details>

<summary>Videos for AppleTV Show/Movie, Vimeo, and Youtube.</summary>


https://github.com/user-attachments/assets/fe5fb8d6-e989-42b8-a52a-79e3d7e62812


https://github.com/user-attachments/assets/f78f5e52-03bc-44b2-a10a-5563c5e6f1fa


https://github.com/user-attachments/assets/8b9fa077-aa6c-48dd-a1c3-9ecf0c57fe00


https://github.com/user-attachments/assets/855e6115-2d14-4b02-a523-56fcde327945



</details>

holow29:
Instead of relying on deep links, can we just use Universal Links (for apps/sites that support them)? https://developer.apple.com/documentation/xcode/allowing-apps-and-websites-to-link-to-your-content/

JPKribs:
> Instead of relying on deep links, can we just use Universal Links (for apps/sites that support them)? https://developer.apple.com/documentation/xcode/allowing-apps-and-websites-to-link-to-your-content/

We are already using Universal Links in iOS and they do not work on tvOS for things like YouTube Trailers. This PR was for tvOS only.

Issue #1477, opened 2025-04-06
[tvOS] (master) Crash on first login


sjorge:
### Describe the bug

(Reference: https://github.com/jellyfin/Swiftfin/discussions/1294#discussioncomment-12742078)

1. build the 'Swiftfin tvOS' project
2. run in simulator | run on device (I had it on both)
3. select your discovered server on the right (not sure if it matters)
4. login with an admin account (not sure if it needs to be an admin account)
5. Debugger hangs here
![](https://private-user-images.githubusercontent.com/379665/430713975-23e2231e-bccb-4fdb-8cc2-bfe30581456d.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NDM5NjEyNzksIm5iZiI6MTc0Mzk2MDk3OSwicGF0aCI6Ii8zNzk2NjUvNDMwNzEzOTc1LTIzZTIyMzFlLWJjY2ItNGZkYi04Y2MyLWJmZTMwNTgxNDU2ZC5wbmc_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjUwNDA2JTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI1MDQwNlQxNzM2MTlaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT1hY2RmNGExOGVhOGU5N2QyMzNjMTk1NmIwMjRlOGQ3NjdiMDUzYzA0MTFiMTY0MDRmNjJmOTc1NDBhMmMwNTcxJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCJ9.Bd0E8cljtkoaA1iFbG8YkQr2cqYckojv5Hs_2SF9mjw)

You can get past this by force closing the app or rebuilding again.

If you force close the app, start it without the debugger it just crashes. **This only happens the very first time if there was no previous server configured**.

[...]

PR #1481, opened 2025-04-06
[tvOS] Fix First Login Crash

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1477

tvOS seems to crash on first login due to a memory allocation issue. By wrapping these operations in `DispatchQueue.main.async`, we ensure that:

1. **UI Updates Happen on the Main Thread**: SwiftUI requires all UI updates to happen on the main thread. When we modify Defaults or post notifications that trigger UI updates, doing this on a background thread can cause crashes.

2. **Thread Safety for Shared Resources**: `Container.shared.currentUserSession.reset()` deals with a shared resource that might be accessed from multiple threads. The async wrapper ensures thread-safe access.

3. **Memory Access Synchronization**: The crash was an `EXC_BAD_ACCESS` error, which often happens when memory is deallocated on one thread while being accessed on another. The async wrapper ensures proper synchronization.

4. **Deferred Execution**: By deferring these operations to the next run loop cycle, we give any pending UI updates a chance to complete before changing application state.

(According to Claude) tvOS is particularly sensitive to threading issues because:

1. Its rendering pipeline has different timing characteristics than iOS
2. Memory management is more aggressive due to limited resources
3. The focus system adds additional complexity to the UI update cycle

---

Does it make sense to, where possible, move these operations to the `viewModel` instead?

### DIFF
--- a/Swiftfin tvOS/Views/UserSignInView/UserSignInView.swift
+++ b/Swiftfin tvOS/Views/UserSignInView/UserSignInView.swift
@@ -68,6 +68,7 @@ struct UserSignInView: View {
 
     // MARK: - Handle Sign In
 
+    @MainActor
     private func handleSignIn(_ event: UserSignInViewModel.Event) {
         switch event {
         case let .duplicateUser(duplicateUser):
@@ -78,11 +79,9 @@ struct UserSignInView: View {
         case let .signedIn(user):
             router.dismissCoordinator()
 
-            DispatchQueue.main.async {
-                Defaults[.lastSignedInUserID] = .signedIn(userID: user.id)
-                Container.shared.currentUserSession.reset()
-                Notifications[.didSignIn].post()
-            }
+            Defaults[.lastSignedInUserID] = .signedIn(userID: user.id)
+            Container.shared.currentUserSession.reset()
+            Notifications[.didSignIn].post()
         }
     }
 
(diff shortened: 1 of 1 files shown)


JPKribs:
Should we also add this for iOS? Are we just getting lucky this doesn't happen on iOS or is it actually structurally different enough this should be fine?

```
            DispatchQueue.main.async {
                Defaults[.lastSignedInUserID] = .signedIn(userID: user.id)
                Container.shared.currentUserSession.reset()
                Notifications[.didSignIn].post()
            }
```

sjorge:
this does indeed resolve #1477 for me

LePips (approved):
> Does it make sense to, where possible, move these operations to the viewModel instead?

No, as I've adopted the architecture that events that should cause view transitions should happen in views as reactions to received events.

JPKribs:
> No, as I've adopted the architecture that events that should cause view transitions should happen in views as reactions to received events.

Gotcha! That makes sense. Sorry, all the architecture side of things is beyond me lol. Does this `MainActor` need to exist on iOS as well or is this purely tvOS quirkiness?

LePips:
I will have to investigate more threading issues on tvOS as I think this is somewhat also related to why https://github.com/jellyfin/Swiftfin/pull/1090#discussion_r1788712254 was necessary.

PR #1482, opened 2025-04-07
Clean Up `SelectUserView`

LePips:
A lot of cleanup in these views and refining behaviors when enabled and selected.


JPKribs (approved):
Everything looks good to me and functions as expected when tested!

PR #1483, opened 2025-04-07
[Meta] Cleanup Jellyfin Version Badge

JPKribs:
### Summary

Jellyfin badge has 3 digits when API should only change for X.Y. So, X.Y.Z isn't required but can be added back in if it is required. This is so completely unimportant but I cannot boast that the SDK is up-to-date if the version is off by 1!

Alternatively, I can update this to 10.10.7 but then you can expect a PR from me in 6-8 weeks for 10.10.8 😆

### Before
Screenshot 2025-04-06 at 22 26 10

### After
Screenshot 2025-04-06 at 22 24 03

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
   <h1>Swiftfin</h1>
   <img src="https://img.shields.io/badge/iOS-15+-red"/>
   <img src="https://img.shields.io/badge/tvOS-17+-red"/>
-  <img src="https://img.shields.io/badge/Jellyfin-10.10.6-9962be"/>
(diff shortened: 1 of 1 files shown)


JPKribs:
On another note, could we bump the TestFlight? I feel like we got a lot of good things knocked out this week that would be nice to get some testing behind. From my end, I want to start confirming `ActiveSessions` doesn't have anything I was missing due to the SDK versioning.

sjorge:
Sticking to the major.minor seems good to me, there should be no API breakage **within** a minor release as I understand it.

LePips (approved):
> bump the TestFlight

I will be doing so shortly.

PR #1484, opened 2025-04-07
Cleanup `ChevronButton`

LePips:
Cleans up API for `ChevronButton` to largely match `Button` and removes `ChevronAlertButton` in favor of condensing into special `ChevronButton` behavior.


JPKribs (approved):
Big fan of this change! Looks great! Tested from my end and the chevron buttons / chevron alert buttons work as expected.

LePips:
We can definitely take a look at that.

PR #1485, opened 2025-04-07
[iOS] Admin Dashboard - User Activity

JPKribs:
### Summary

Long time no see, Admin Dashboard!

With 10.10 now available, I have reopened: https://github.com/jellyfin/Swiftfin/discussions/1283

This PR creates the Activity Log & Details view in the Admin Dashboard. This only allows filtering based on the items included in the API: https://api.jellyfin.org/#tag/ActivityLog/operation/GetLogEntries

This means, we can only edit based on `minDate` and `hasUserId`. I've added these filters. Jellyfin-Web seems to have a lot of additional options. I can try to mimic those but it would be filtering the arrays manually on devices. I'd love for this to be closer to Jellyfin-Web since I think they did a really good job but I also didn't want to pull 300+ records onto the device to get the 10 valid ones so performance is tricky to mirror.

That being said, I think I've covered everything we can do natively with the API. Please find a video below.

### Video

https://github.com/user-attachments/assets/4bb49e4b-cc11-42a1-aab2-b708828d7768

### Screenshot

<details>
<summary>Resetting minDate to nil</summary> 

simulator_screenshot_4CAB451A-770E-4EC1-BAA5-3C55B462C722

</details>


Swiftfin/Views/AdminDashboardView/Components/MediaItemSection.swift
```diff
@@ -0,0 +1,76 @@
...
+//
+// Copyright (c) 2025 Jellyfin & Jellyfin Contributors
+//
+
+import JellyfinAPI
+import SwiftUI
+
+extension AdminDashboardView {
+
+    struct MediaItemSection: View {
```
JPKribs, line 14: I made this into it's own extension since I'm now using this in another location.

Swiftfin/Views/AdminDashboardView/ServerActivity/ServerActivityView/ServerActivityView.swift
```diff
@@ -0,0 +1,199 @@
...
+
+    @EnvironmentObject
+    private var router: AdminDashboardCoordinator.Router
+
+    // MARK: - State Objects
+
+    @StateObject
+    private var viewModel = ServerActivitiesViewModel()
+    @StateObject
+    private var userViewModel = ServerUsersViewModel()
```
JPKribs, line 25: Instead of doing a getUser for every record, I get all of the users then match to that list of users. This reduces the number of API called at the expense of using more memory. Let me know if this isn't the right call.

Swiftfin/Views/AdminDashboardView/ServerActivity/ServerActivityView/ServerActivityView.swift
```diff
@@ -0,0 +1,199 @@
...
+                .listRowSeparator(.hidden)
+                .listRowInsets(.zero)
+        } else {
+            CollectionVGrid(
+                uniqueElements: viewModel.elements,
+                id: \.unwrappedIDHashOrZero,
+                layout: .columns(1)
+            ) { entry in
+
+                let logViewModel = ServerActivityDetailViewModel(
```
JPKribs, line 86: This should contain the ActivityLogEntry + User + Item. The Item is only pulled from the Server when the ViewModel is refreshed. The VM is only refreshed in ServerActivityDetailsView since we don't need the Item on the Record List level. We DO need the user though so that's why I am passing in the full user.

This part was the hardest portion to wrap my head around without doing 1,000+ API calls. But please let me know if there is a better way to do this!

JPKribs (commented):
I think what I have is good. These are the 3 items that might require some additional detail/insight. Please let me know if you have any questions!

Shared/Extensions/Date.swift
```diff
@@ -0,0 +1,25 @@
...
+// Copyright (c) 2025 Jellyfin & Jellyfin Contributors
+//
+
+import Foundation
+
+extension Date {
+
+    // MARK: - Return the Date at 00:00:00
+
+    var dateOnly: Date {
```
JPKribs, line 15: Last item. I made this because otherwise the filter would return only items at the current time forward on the selected date. This made it hard to select filter by today.

I ended up only using this twice so I can make this local to the work I was doing if we don't thing there is another usage for this elsewhere.

JPKribs:

LePips (approved):
I largely moved the list of users held into the `ServerActivityViewModel` to ensure that we get the users first and then load the logs. While not breaking, it is possible for the logs to load before the users.

PR #1488, opened 2025-04-08
[iOS] Admin Dashboard - QuickConnect Other User

JPKribs:
### Summary

In 10.10, you are able to authenticate a user using QuickConnect on their behalf. This was done by adding an optional `userID` parameter: https://api.jellyfin.org/#tag/QuickConnect/operation/AuthorizeQuickConnect

This PR re-uses the `QuickConnectAuthorizeView` and adds an optional `userID` to the Initializer. When called without a `userID`, this authenticates the current user (unchanged) but when a `userID` is provided, it authorizes that user.

The `QuickConnectAuthorizeView` is otherwise unchanged. I updated the dismiss to be router agnostic since it's `SettingsCoordinator` vs `AdminDashboardCoordinator`.  

If there are any `QuickConnectAuthorizeView` changes that we want to do while I'm in there lemme know! Otherwise, this was a surprisingly small change.

### Screenshot

This just takes you to the `QuickConnectAuthorizeView` but with the `userID`

<details>

<summary>New Menu Option</summary>

Simulator Screenshot - iPhone 16 Pro - 2025-04-08 at 10 33 45

</details>

<details>

<summary>Revised Quick Connect View</summary>

Simulator Screenshot - iPhone 16 Pro - 2025-04-11 at 22 40 41

</details>


JPKribs:
One final change. The feedback I got was someone might accidentally log someone into the wrong account. So I've added the following to the page as an indicator of who is being logged in via QuickConnect.

IMO the text is a little wordy so I'm open to suggestions!

Simulator Screenshot - iPhone 16 Pro - 2025-04-08 at 13 03 41

Swiftfin/Views/SettingsView/UserProfileSettingsView/QuickConnectAuthorizeView.swift
```diff
@@ -45,10 +49,60 @@ struct QuickConnectAuthorizeView: View {
     @State
     private var error: Error? = nil
 
+    // MARK: - Initialize
+
+    init(_ user: UserDto? = nil) {
```
JPKribs, line 54: This is really really sloppy... This works and I can't think of a better way to do this but it's messier than I would like. Open to ideas on doing this better!

JPKribs:

Shared/Strings/Strings.swift
```diff
@@ -92,14 +92,8 @@ internal enum L10n {
   internal static let allLanguages = L10n.tr("Localizable", "allLanguages", fallback: "All languages")
   /// All Media
   internal static let allMedia = L10n.tr("Localizable", "allMedia", fallback: "All Media")
-  /// Allow collection management
-  internal static let allowCollectionManagement = L10n.tr("Localizable", "allowCollectionManagement", fallback: "Allow collection management")
```
JPKribs, line 96: These are less related to this PR... This verbiage just made more concise sense while I was fixing up localization for this PR.

LePips (approved):
I just moved the `user` injection to the view model instead of the view. Looks great!

PR #1490, opened 2025-04-10
Get Localizations from Jellyfin & Apple

JPKribs:
### Summary

I am working on allowing subtitle uploading, editing, and searching OpenSubtitles in [this PR](https://github.com/JPKribs/Swiftfin/tree/subtitles).

Currently, we are pulling the possible ISO 3 letter codes for Languages and Countries from Apple's built in list. For 99% of uses that is fine but I recently found that Jellyfin actually has all of the valid country/language codes here: https://api.jellyfin.org/#tag/Localization

I don't know if that was added in 10.10 or I missed it when I first made the `LanguagePicker`.

This PR updates the `LanguagePicker` and `CountryPicker` to init from either the country code that is used by most items or by using a full `CultureDto` and `CountryInfo` as needed. These pickers now get all of their values from Jellyfin so this should result in only valid items for Jellyfin in the dropdown. I've also updated the Picker names to reflect they are using ISO codes.

### Impact

This doesn't really change anything for this PR. The only major difference is Apple has some localizations that Jellyfin can't handle so this prevent that potential error. Additionally, Jellyfin provides `DisplayName` which allows locals to actually alphabetize correctly opposed to the Apple ones which are in the language which means localization ends up a bit strange.

The real impact of this is later in [Subtitle Management](https://github.com/JPKribs/Swiftfin/tree/subtitles) and eventually being able to set default/fallback subtitles at a Swiftfin level.

The `CountryISOPicker` will likely only ever be used in this one location but I think it's valuable to keep our [localization options valid entries for Jellyfin.](https://api.jellyfin.org/#tag/Localization)


LePips (changes requested):
At least for this specific feature, I'm doubting how necessary this is. The endpoint used for this just gets the [contents of a static ISO6392 file](https://github.com/jellyfin/jellyfin/blob/86b81c912d5070e497de8ed222ae4882c768fddd/Emby.Server.Implementations/Localization/iso6392.txt#L4) and doesn't provide localization.

Foundation provides some kind of superset of these ISO language codes and localization for free, which we already use.

JPKribs:
Primarily, my thinking was that:

1. If Jellyfin provides it, we should use that since we can confirm it's always going to align with the .NET localization on server.

2. All of the localization and name are returned in the server language. Apple likes to return them in the language that was provided. Or, at least the way I previously had it set up did it that way. So, the difference between Germany and Deutschland. While this isn't necessarily an issue, it does make it a little bit quirky to use for something like a country or language picker. Like, if I know this show was filmed in Germany I'm going to scroll to the G to look for it opposed to with the Apple one I should go to D.

3. This allows us to use things like `CountryInfo` and `CultureDto`. That being said, neither of those appear to really have any usage right now, so it might just be something extra for no real benefit.

I don't think either of these are the wrong approach, but I just thought I would throw this out there as a more Jellyfin-approach. If we would prefer to use Apple, I would be fine using that as well!

The reason that I started this was that Apple returned back a two letter code while Jellyfin uses a two letter code for MOST things, and a three letter code for subtitles. So, for subtitle editing / defaulting, I would need to update the language picker specifically to also return the three letter code as well.

Entirely your call on this!

LePips:
1 - The only issue I foresee is that if another client uses the list as their only data source and errors out 
2 - The values from the server aren't localized in any way. They are just parsed against the `|` delimited from the file on the server. We are a client and should present the titles as such in our user's language. A good example is iOS's selection of changing the OS language.

JPKribs:
That makes sense!

Let me update this PR to use Foundation again. Regardless, I need to update the pickers to alphabetize better and allow 3 letter codes.

JPKribs:
When I use the iOS 16+ Foundation call to:

`Locale.Language.systemLanguages.compactMap { $0.languageCode?.identifier(.alpha2) }`

The full localization list is only 36 languages.

When I use:

```
    let identifiers = Locale.availableIdentifiers
    let codes = identifiers.compactMap { id in
        Locale(identifier: id).languageCode
    }
    return Set(codes)
```

I get 284 languages but there are still some not available that exist in Jellyfin:

Screenshot 2025-04-23 at 13 45 07

There are also some that exist in Swiftfin that don't exist in Jellyfin. 

Which approach do we prefer for this? The iOS 16+ route let's us use 3 letter codes which we need to subtitles but we lose most of our localizations. The earlier route only had 2 letter codes but works on more versions of iOS. We would need an alternative approach for subtitle editing, uploading, or setting defaults to get 3 digits.

There is also very likely an API that I'm just not aware of that does this better but those were the 2 routes I found in Foundation

LePips:
Sorry, coming back to this and guess seeing the issue of unsupported values across clients immediately, I'm fine with getting the locales from the server. We could possibly mix the server values along with the Foundation values to get our localization.

JPKribs:
> Sorry, coming back to this and guess seeing the issue of unsupported values across clients immediately, I'm fine with getting the locales from the server. We could possibly mix the server values along with the Foundation values to get our localization.

No worries. It looks like iOS 16 deprecates and adds new localization options that include alpha3 so I'll try again after that. I'd rather focus on one implementation than a split 15/16+.

---

Edit: This should be ready for review now post iOS 16

JPKribs:
Taking a look at this with:

```
guard #available(iOS 16, *) else {
    return [emptyCultureDto] + viewModel.cultures
        .sorted { getDisplayName(for: $0) < getDisplayName(for: $1) }
}
```

This seems like our most robust solution as this contains more options than just Jellyfin or just Apple but the UI noticeably hitches when we scroll to where these pickers exist.

I am putting this into "In-Progress" to work on this since I think this would be good to nail down but any efficiency recommendations are appreciated. I think our Main version of these pickers are totally fine so I'm not worried about rushing this in.

Shared/Components/Localization/CountryPicker.swift
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    // MARK: - Picker Title
+
+    private let title: String
+
+    // MARK: - ISO Language Codes
+
+    @Binding
+    private var twoLetterISORegion: String?
```
LePips: I'm confused why we need processing with the two and three letter codes instead of just using the `CountryInfo/CultureDto` directly.

LePips (changes requested):

Shared/Components/Localization/CountryPicker.swift
JPKribs: In an earlier version of this I was use using the `CountryInfo` and `CultureDto` but I added the letter codes to make this easier to bind. Unfortunately, Jellyfin gives use these objects but doesn't seem to use them anywhere? All of the Country/Language codes expected in Jellyfin API calls are either 2 or 3 letter codes. 

My thinking was allow binding on those values to make that easier to use opposed to requiring usages of this convert from letters to `CountryInfo`/`CultureDto` then back again.

LePips: It's not necessarily about whether the server uses them or not, these views are just selections into these lists of items to get the relevant codes for usage elsewhere. Performing these other selections and searching into the lists for the objects is inefficient.

We can still provide a string-based binding API like this work currently does, however we would just select the relevant object that matches the initial code and then update the binding on the object selection.

JPKribs: Gotcha, that makes sense. I will take a crack at those revisions tomorrow!

JPKribs: Let me know if that is better! I tried a few ways. Initing from the partial object was harder than I'd like. Just since you have to go from `EN` -> `(EN, ENG, English)` so I couldn't figure out a way to do this without some `onChange`. Total open to revisions!

LePips (changes requested):
Apologies, I may have led you astray. I've implemented just getting using the server's values and then getting the localized value for the `displayTitle` from the relevant ISO code. I also realize defaulting to English/US wouldn't be appropriate and just default to `nil` instead.

Per the ordering of items, we will leave it as the order given to us from the server which is the order determined by the underlying files defining the values. While we've typically "ordered by English", the server does as well - [but with some oddities](https://github.com/jellyfin/jellyfin/blob/23c25289dae40f292f0ffcf8c83f8ea2f9eebd61/Emby.Server.Implementations/Localization/countries.json#L81) per the "official" name of some countries. I don't consider this an issue.

I only haven't worked on the tvOS side, which I request that you complete.

JPKribs:
> I only haven't worked on the tvOS side, which I request that you complete.

Thank you for doing all this! I can take a look at tvOS tomorrow and make sure that all looks good.

I like the `BaseFetchViewModel` change! It's much much cleaner. I believe the Parental Ratings are very similar. I'll look to move those over in a later PR.

JPKribs:
tvOS is done. Only thing I did that might be weird is the `onChange` I moved into the `#if os()` checks to get rid of the deprecation warning for tvOS. I left a note to change this on iOS 17.

LePips (approved):
Thanks!

PR #1491, opened 2025-04-12
[Meta] Fix Webplates

JPKribs:
### Summary

Localization cleanup script was setting the encoding to UTF-8 while other localization are encoded to UTF-16LE. I **THINK** this is where our Webplate errors are originating. I have no idea why this is only coming up now but the project still builds and all of our localizations are now encoded the same way.

### Screenshots

Before:
Screenshot 2025-04-11 at 18 29 18

After:
Screenshot 2025-04-11 at 18 38 36


JPKribs:
@LePips I'm not positive this works but this was the only thing I saw that stood out as weird. Per the Webplate issue:

Screenshot 2025-04-11 at 19 59 10

thornbill (approved):

Issue #1470, opened 2025-04-04
Played/unplayed filters not functioning correctly on Shows/Collections libraries


5fish:
### Describe the bug

Currently, using the “Unplayed” filter on a library will show all items, including completed items.
The “Played” filter instead shows no results at all, even if there are completed items in the library.

### Application version

1.3(8)

### Where did you install the app from?

TestFlight

### Device information

iPhone 16 Pro

### OS version

iOS 18.3.2

### Jellyfin server version

10.10.6

JPKribs:
As a note:

Seems to work correctly for Movies but NOT for Shows and Collections. Seems to be related to `isFolder` objects.

quangDecember:
@JPKribs I'd like to take this ticket

My first impression is that the request we're sending when applying filter:

```
curl --request GET \
  --url 'https://jellyfin.REDACTED_REDACTED/Users/REDACTED_REDACTED/Items?startIndex=0&limit=50&recursive=true&sortOrder=Ascending&parentId= REDACTED_REDACTED&fields=MediaSources&fields=Overview&fields=ParentId&fields=Taglines&includeItemTypes=Movie&includeItemTypes=Series&includeItemTypes=BoxSet&filters=IsPlayed&sortBy=SortName&enableUserData=true' \
  --header 'Accept: application/json' \
  --header 'Authorization: MediaBrowser Client=Swiftfin iOS, Version=1.0.0, Device=iPhone15ProMax, Token= REDACTED_REDACTED, DeviceId=iOS_ REDACTED_REDACTED'
```

the result will be correct when we remove `recursive` parameter

JPKribs:
Thank you for pitching in!!

That isRecursive is a good find. I would be cautious when testing to make sure nothing's getting dramatically changed. Specifically tvOS. Last time I removed this it caused issues:

https://github.com/jellyfin/Swiftfin/pull/1423

So this might end up needing some revision figuring out ItemType libraries (Movies & TV Shows tabs on tvOS) as well as resolving filtering.

Let me know if you have any questions I can help field!

quangDecember:
thanks @JPKribs I'm moving implementation discussion to PR
https://github.com/jellyfin/Swiftfin/pull/1495#issuecomment-2798953724

PR #1495, opened 2025-04-12
Played/Unplayed status for "Collection" and "TV Shows"

quangDecember:
fixing #1470 
set `recursive` to `false` for TVShows & Collection 

https://github.com/user-attachments/assets/9f824dca-f5eb-456e-9307-6c455c56513d

### DIFF
--- a/Shared/ViewModels/LibraryViewModel/ItemLibraryViewModel.swift
+++ b/Shared/ViewModels/LibraryViewModel/ItemLibraryViewModel.swift
@@ -57,7 +57,7 @@ final class ItemLibraryViewModel: PagingLibraryViewModel<BaseItemDto> {
 
         // Default values, expected to be overridden
         // by parent or filters
-        parameters.isRecursive = true
+        parameters.isRecursive = isParentTypeTVShowsCollection() ? false : true
         parameters.includeItemTypes = BaseItemKind.supportedCases
         parameters.sortOrder = [.ascending]
         parameters.sortBy = [ItemSortBy.name.rawValue]
@@ -122,5 +122,12 @@ final class ItemLibraryViewModel: PagingLibraryViewModel<BaseItemDto> {
         let response = try? await userSession.client.send(request)
 
         return response?.value.items?.first
+    }
+
+    func isParentTypeTVShowsCollection() -> Bool {
+        if let parent = parent as? BaseItemDto, let collectionType = parent.collectionType {
+            return collectionType == .tvshows || collectionType == .boxsets
+        }
+        return false
     }
 }


quangDecember:
https://github.com/jellyfin/Swiftfin/issues/1470#issuecomment-2798927393

@JPKribs FYI: `isFolder` is `true` and parent type is `JellyfinAPI.BaseItemKind.collectionFolder` for Movies, Collection & TVShows so I gotta use collection type to check & limit the scope

Shared/ViewModels/LibraryViewModel/ItemLibraryViewModel.swift
```diff
@@ -57,7 +57,7 @@ final class ItemLibraryViewModel: PagingLibraryViewModel<BaseItemDto> {
 
         // Default values, expected to be overridden
         // by parent or filters
-        parameters.isRecursive = true
+        parameters.isRecursive = isParentTypeTVShowsCollection() ? false : true
```
JPKribs, line 60: IMO, I think let's just change this to:

Add another extension and add a new var to [CollectionType](https://github.com/jellyfin/Swiftfin/blob/main/Shared/Extensions/JellyfinAPI/CollectionType.swift) and call it something like:

```
var baseItemType: [CollectionType]
    return [.tvshows, .boxsets]
```

Then, in `ItemLibraryViewModel`, we can call

`parameters.isRecursive = !CollectionType.baseItemType.contains(collectionType)`

This way, the `CollectionType` data lives on the `CollectionType` enum. Additionally, as we move onto Music, Playlists, Mixed Libraries, etc. can add items that need this same treatment there.

Then, we can remove `isParentTypeTVShowsCollection` below. This better separates `CollectionType` logic from the `ViewModel` logic.

That being said, I'll defer to @LePips if they have a different route for this as well!

Shared/ViewModels/LibraryViewModel/ItemLibraryViewModel.swift
```diff
@@ -123,4 +123,11 @@ final class ItemLibraryViewModel: PagingLibraryViewModel<BaseItemDto> {
 
         return response?.value.items?.first
     }
+
+    func isParentTypeTVShowsCollection() -> Bool {
```
JPKribs, line 127: We can remove this in favor of the `CollectionType` change.

JPKribs (changes requested):
This looks good! A little bit of reorganization to make this cleaner in the future. I think there is a 100% chance we're going to need to expand the items where `isRecursive == false` so these changes should give us some additional flexibility.

I've tested this from both iOS and tvOS and this appears to work correctly. Thank you for looking into this!! I think we're at a point in the release cycle where these types of polish fall down the list a little more than they should. So this type of troubleshooting is a great assist!

Thank you for your PR and please let me know if you have any questions on the 2 requested changes!

Shared/ViewModels/LibraryViewModel/ItemLibraryViewModel.swift
quangDecember: Updated

JPKribs:
Looks good! Tested and working as expected!

I moved everything down a bit and tacked on a comment so, down the road, we can remember what and why we made these changes.

Thank you for troubleshooting this and coming up with a clean solution! I'll defer to @LePips for a final look over but this looks good to me!

JPKribs (dismissed):
Looks good! Tested and working as expected!

I moved everything down a bit and tacked on a comment so, down the road, we can remember what and why we made these changes.

Thank you for troubleshooting this and coming up with a clean solution! I'll defer to @LePips for a final look over but this looks good to me!

LePips (changes requested):
Apologies, but the issue is actually elsewhere and a bit more nuanced. While the query is built on the server, there is some odd interplay between the `isPlayed` filter and `includeItemTypes` that I tried to track down but didn't have enough time.

The reason for the behavior exhibited in this comment: https://github.com/jellyfin/Swiftfin/issues/1470#issuecomment-2781505629

Is due to this line:
https://github.com/jellyfin/Swiftfin/blob/8b58c52a877b99a131ff4b9b4332ffbca80eb6c2/Shared/Objects/LibraryParent/LibraryParent.swift#L36

Change it to `[.series]` and it will work for tv show libraries, and not movies or boxsets. For an unknown reason, it will just take the first item from the `includeItemTypes` and filter the server query from that, but I was unable to track down where.

---

The more correct fix on my branch, which makes me think I probably need to refactor `LibraryParent` a bit more.

https://github.com/jellyfin/Swiftfin/compare/main...LePips:SwiftFin:supported-item-types-based-on-parent

LePips:
I would like to have this fixed for 1.3 which I hope to get out soon.

quangDecember:
> I would like to have this fixed for 1.3 which I hope to get out soon.

sorry I've been busy the last 1.5 month. 

I checked your branch: it seems to works fine for me on tv shows, but not boxsets.
reason: the network request for your branch is having `Items?startIndex=0&limit=50&recursive=false&sortOrder=Ascending&parentId=9d7ad6afe9afa2dab1a2f6e00ad28fa6&fields=MediaSources&fields=Overview&fields=ParentId&fields=Taglines&includeItemTypes=Movie&includeItemTypes=Series&includeItemTypes=BoxSet&filters=IsPlayed&sortBy=SortName&enableUserData=true"` where recursive=false.



> Apologies, but the issue is actually elsewhere and a bit more nuanced. While the query is built on the server, there is some odd interplay between the `isPlayed` filter and `includeItemTypes` that I tried to track down but didn't have enough time.
> 
> The reason for the behavior exhibited in this comment: [#1470 (comment)](https://github.com/jellyfin/Swiftfin/issues/1470#issuecomment-2781505629)
> 
> Is due to this line:
> 
> https://github.com/jellyfin/Swiftfin/blob/8b58c52a877b99a131ff4b9b4332ffbca80eb6c2/Shared/Objects/LibraryParent/LibraryParent.swift#L36
> 
> Change it to `[.series]` and it will work for tv show libraries, and not movies or boxsets. For an unknown reason, it will just take the first item from the `includeItemTypes` and filter the server query from that, but I was unable to track down where.
> 
> The more correct fix on my branch, which makes me think I probably need to refactor `LibraryParent` a bit more.
> 
> [main...LePips:SwiftFin:supported-item-types-based-on-parent](https://github.com/jellyfin/Swiftfin/compare/main...LePips:SwiftFin:supported-item-types-based-on-parent)

can you explain me again what can I fix here? merge your branch to mine and adapt?

quangDecember:
> > I would like to have this fixed for 1.3 which I hope to get out soon.
> 
> sorry I've been busy the last 1.5 month.
> 
> I checked your branch: it seems to works fine for me on tv shows, but not boxsets. reason: the network request for your branch is having `Items?startIndex=0&limit=50&recursive=false&sortOrder=Ascending&parentId=9d7ad6afe9afa2dab1a2f6e00ad28fa6&fields=MediaSources&fields=Overview&fields=ParentId&fields=Taglines&includeItemTypes=Movie&includeItemTypes=Series&includeItemTypes=BoxSet&filters=IsPlayed&sortBy=SortName&enableUserData=true"` where recursive=false.
> 
> > Apologies, but the issue is actually elsewhere and a bit more nuanced. While the query is built on the server, there is some odd interplay between the `isPlayed` filter and `includeItemTypes` that I tried to track down but didn't have enough time.
> > The reason for the behavior exhibited in this comment: [#1470 (comment)](https://github.com/jellyfin/Swiftfin/issues/1470#issuecomment-2781505629)
> > Is due to this line:
> > https://github.com/jellyfin/Swiftfin/blob/8b58c52a877b99a131ff4b9b4332ffbca80eb6c2/Shared/Objects/LibraryParent/LibraryParent.swift#L36
> > 
> > Change it to `[.series]` and it will work for tv show libraries, and not movies or boxsets. For an unknown reason, it will just take the first item from the `includeItemTypes` and filter the server query from that, but I was unable to track down where.
> > The more correct fix on my branch, which makes me think I probably need to refactor `LibraryParent` a bit more.
> > [main...LePips:SwiftFin:supported-item-types-based-on-parent](https://github.com/jellyfin/Swiftfin/compare/main...LePips:SwiftFin:supported-item-types-based-on-parent)
> 
> can you explain me again what can I fix here? merge your branch to mine and adapt?

actually nvm I see your branch is based on my branch, let me check that out first

quangDecember:
@LePips 
I added back 2 things for the "collection" to work: recursive=false and includeItemTypes=Movie,Series,Boxset

PR #1499, opened 2025-04-15
[iOS & tvOS] Cleanup Permission Validation

JPKribs:
### Summary

This starts by cleaning up **device** permissions into 3 categories:

1. `enableItemEditing` - Can the user edit Metadata, Subtitles, or Lyrics for an item? This enables those settings on Device.
2. `enableItemDeletion` - Can the user delete items from libraries? This enables those settings on Device.
3. `enableCollectionManagement` - Can the user delete or edit Metadata for collections? This enables those settings on Device.

This removes the `enableSubtitleManagement` and `enableLyricsManagement` as they were not in use and they are kind fo redundant and clunky.

---

Then, I moved individual permission checks to the `UserSession` instead of manually checking these each time. There, there are 2 types of checks:

1. Does the Server allow this user to perform this action more generally? This is used to determine if the `CustomizeSettings` button/toggle for that type is shown.
2. Does the user have permissions to the specific item they are looking at to edit / delete? This is used to determine if the user has both Server permissions to perform an action and permissions to the item itself. For example, you cannot edit lyrics on a movie since there are none. You cannot edit subtitles on a series since those exist on the episode/show level

### Impact

There should be no effective changes but this should hopefully make this easier to manage. Additionally, I wanted to add this prior to subtitle editing. This sets that up to be a cleaner implementation.


Swiftfin/Views/ItemEditorView/ItemEditorView.swift
```diff
@@ -12,15 +12,30 @@ import SwiftUI
...
     @EnvironmentObject
     private var router: ItemEditorCoordinator.Router
 
     @ObservedObject
     var viewModel: ItemViewModel
 
+    // MARK: - Can Edit Metadata
+
+    private var canEditMetadata: Bool {
+        viewModel.userSession?.user.permissions.items.canEditMetadata(item: viewModel.item) == true
```
LePips, line 24: These don't need the `?` since it's an implicitly unwrapped optional on the view model.

Swiftfin tvOS/Views/SettingsView/CustomizeViewsSettings/Components/Sections/ItemSection.swift
```diff
@@ -41,6 +41,10 @@ extension CustomizeViewsSettings {
 
                 ListRowMenu(L10n.enabledTrailers, selection: $enabledTrailers)
 
+                /// Enable Refreshing & Deleting Collections
+                if userSession?.user.permissions.items.canManageCollections ?? false {
```
LePips, line 45: Make these `== true` instead of `?? false`, for boolean values I want to be moving towards that instead.

Shared/Objects/UserPermissions.swift
```diff
@@ -20,11 +20,24 @@ struct UserPermissions {
 
     struct UserItemPermissions {
 
+        @StoredValue(.User.enableItemDeletion)
```
LePips, line 23: Do not use these property wrappers in this way. Instead, get the values via the subscript in the `init`.

Shared/Extensions/JellyfinAPI/BaseItemKind.swift
```diff
@@ -29,3 +29,10 @@ extension BaseItemKind: ItemFilter {
         rawValue
     }
 }
+
+extension BaseItemKind {
+
+    static var identifiableTypes: [BaseItemKind] {
```
LePips, line 35:
```suggestion
    static var itemIdentifiableCases: [BaseItemKind] {
```

LePips (changes requested):

LePips (approved):

Issue #1461, opened 2025-03-25
Play button in collections does nothing


Belanthiel:
### Describe the bug

Can see collections as expected, and can play individual movies within collections.

When clicking on the Play button to play the entire collection nothing happens.  I can see the button depress on the screen, but it does absolutely nothing.

The expectation is that the Play (play all) button should in fact cause the collection to play.

### Application version

1.0.1 (70)

### Where did you install the app from?

App Store

### Device information

Apple TV 4K

### OS version

TVOS 18.3.1

### Jellyfin server version

10.10.6

JPKribs:
For now this is intentional. I've updated this logic in https://github.com/jellyfin/Swiftfin/pull/1428. This will not be included in the 1.3 release but should be included for 1.4.

JPKribs:
iOS and iPadOS are ahead of tvOS. You can follow our progress towards the next tvOS release here:

https://github.com/jellyfin/Swiftfin/discussions/1294

I can't give the specifics of why this was disabled by default as it pre-dates my involvement in Swiftfin. I'm sorry for the inconvenience but this should be resolved in that PR so there is a solution on the horizon 🙂

PR #1500, opened 2025-04-15
[iOS & tvOS] Complete Collections' `ItemView`

JPKribs:
### Summary

Originally part of https://github.com/jellyfin/Swiftfin/pull/1428

Resolves: https://github.com/jellyfin/Swiftfin/issues/1461

---

This PR splits out the Collection changes from #1428 since to reduce the side of that PR and reduce the scope to purely Playlists and Collections for each of these PRs respectively. This PR makes two major changes to the Collection `ItemView` to fill out the view more and make Collections more usable in Swiftfin.

1. Categorize and separate items based on `BaseItemKind`. This better mirrors Jellyfin-Web and makes the view easier to find items of specific types.

2. Update the `PlayButton` to be active and play the first available item in the Collection. This goes in the following order:

> - If there are unplayed items, play the first unplayed item.
> - If all items are played, play the first item.
> - If there are no playable items, disable the button*

*This should only occur if the item types are not valid in Swiftfin or your Collection contains nested Collections with no other items available. We could look through this but I worry that opens us up to Collection A contains Collection B which owns Collection A... and we loop forever.

---

The bulk of the work is a copy/paste mirror of other `ItemView` and localization of `BaseItemKind`. I don't know if `pluralDisplayTitle` is the correct approach.

The bulk of the unique changes are in `CollectionItemViewModel` for assigning the `PlayButtonItem`.

--

### Videos

**iOS**

https://github.com/user-attachments/assets/cb26760a-da58-482e-83b3-22ee913b30a3

**tvOS**

https://github.com/user-attachments/assets/cf160da7-7c7c-440f-94a3-6fa30a901746

**Resumable Content**

https://github.com/user-attachments/assets/065e4d8f-94e0-44d4-9012-2b098ff43c7c


Shared/Extensions/JellyfinAPI/BaseItemKind.swift
```diff
@@ -24,8 +24,164 @@ extension BaseItemKind: SupportedCaseIterable {
...
+            return L10n.video
+        case .year:
+            return L10n.year
+        }
+    }
+}
+
+extension BaseItemKind {
+
+    var pluralDisplayTitle: String {
```
JPKribs, line 109: I don't think this is right...

Shared/ViewModels/ItemViewModel/CollectionItemViewModel.swift
```diff
@@ -37,4 +45,83 @@ final class CollectionItemViewModel: ItemViewModel {
...
+        guard let seriesID = series.id else {
+            return nil
+        }
+
+        let request = Paths.getSeasons(seriesID: seriesID)
+        let response = try await userSession.client.send(request)
+
+        if let unplayedSeason = response.value.items?.first(where: { $0.userData?.isPlayed == false }) {
+            return try await getFirstItem(season: unplayedSeason)
+        } else if let firstSeason = response.value.items?.first {
```
chickdan, line 92: If the first season is unplayed, wouldn't it be the `unplayedSeason`? If so, I'm not sure this conditional will be hit.

JPKribs: Find first available season that is unplayed. If there are no unplayed seasons, fallback to the first season. This *should* mirror the existing `playButtonItem` logic on the `SeriesItemViewModel`

Swiftfin tvOS/Views/ItemView/CollectionItemView/CollectionItemContentView.swift
```diff
@@ -26,16 +27,28 @@ extension CollectionItemView {
...
-                    PosterHStack(
-                        title: L10n.items,
-                        type: .portrait,
-                        items: viewModel.collectionItems
-                    )
-                    .onSelect { item in
-                        router.route(to: \.item, item)
+
+                    ForEach(BaseItemKind.allCases, id: \.self) { sectionType in
+                        let sectionItems = viewModel.collectionItems.filter { $0.type == sectionType }
```
LePips, line 32: Instead, have the `collectionItems` be an `OrderedDictionary<BaseItemKind, [BaseItemDto]>`

LePips (changes requested):
Collections are different in that they are like a library of grouped items, which has made it being an "item view" with a play button weird to me. However, making it also a library outright isn't correct either.

I'm good with the implementation we've had for a while now but I do expect a new design for collections that would essentially remove the playback ability from these views.

Shared/ViewModels/ItemViewModel/CollectionItemViewModel.swift
```diff
@@ -9,22 +9,30 @@
...
         await MainActor.run {
             self.collectionItems = collectionItems
+            self.playButtonItem = playButtonItem
         }
     }
 
-    private func getCollectionItems() async throws -> [BaseItemDto] {
+    // MARK: - Get Collection Items
 
+    private func getCollectionItems() async throws -> OrderedDictionary<BaseItemKind, [BaseItemDto]> {
```
LePips, line 35: These will always be ordered by whatever is given to us by the server and isn't defined by us, is that what we want? We could alternatively sort by the `BaseItemKind.rawValue` to somewhat ensure that the sections are always in English alphabetical order.

Shared/ViewModels/ItemViewModel/CollectionItemViewModel.swift
```diff
@@ -35,6 +43,89 @@ final class CollectionItemViewModel: ItemViewModel {
...
+            by: \.type
+        )
+
+        return groupedItems
+            .compactKeys()
+    }
+
+    // MARK: - Get Play Button Items
+
+    private func getPlayButtonItem(_ items: OrderedDictionary<BaseItemKind, [BaseItemDto]>) async throws -> BaseItemDto? {
```
LePips, line 59: We are doing a lot of processing on the device to get this first played item and isn't that efficient. The removal of playback from the views for this may come sooner rather than later.

Shared/ViewModels/ItemViewModel/CollectionItemViewModel.swift
JPKribs: It depends on whether we think Playlists and Collections are similar items. Structurally, they're both just an array of `BaseItemDto` so they could re-use a lot of the same views. But, if we want to lean away from Collections having a `PlayButtonItem` it's likely that we would want to separate those. I think Playlists we'd want to retain the order from Server but Collections, if we're creating something new and separate, might be cool to had a "Order by" option. 

Some collections I might want it by year. Some I might want by kind. Some, the server order might be intentional.

I don't want to balloon this PR but I'd be happy to look at this if that's a better fit?

JPKribs: Here is a really rough around the edges idea:

https://github.com/user-attachments/assets/f1a6adef-6fcb-460d-a322-2b3aad1e4123

LePips: My comment was specifically asking about how we want to order Movies, Genres, Studios, and sections of items. Collections do have a customizable ordering, which is synced with the server, for the items themselves which we will have functionality for.

JPKribs: My bad. Ordered in latest commit.

Shared/ViewModels/ItemViewModel/CollectionItemViewModel.swift
JPKribs: Below is how this looks removing. These look fine IMO with the exception of Compact Poster.

****

https://github.com/user-attachments/assets/178a7178-cce6-44a8-8db9-033e9605e486

****

https://github.com/user-attachments/assets/344f0eeb-3a6e-4cce-916b-7c1ff2ad712c

<details>

<summary>Unplayable Collections on iPhone</summary>

Simulator Screenshot - iPhone 16 Pro - 2025-04-30 at 15 21 52

Simulator Screenshot - iPhone 16 Pro - 2025-04-30 at 15 22 47

Simulator Screenshot - iPhone 16 Pro - 2025-04-30 at 15 23 06

</details>

<details>

<summary>Playable Movies on iPhone (For Comparison)</summary>

Simulator Screenshot - iPhone 16 Pro - 2025-04-30 at 15 23 23

Simulator Screenshot - iPhone 16 Pro - 2025-04-30 at 15 23 34

Simulator Screenshot - iPhone 16 Pro - 2025-04-30 at 15 23 45


</details>

LePips (approved):
Thanks for finishing these views. On tvOS, the absence of the play button interrupts the focus but we can find another fix for that later.

Issue #1091, opened 2024-06-10
Make 10.9 Server Minimum


LePips:
The 10.9 schema brought a lot of updates to endpoints and deprecated many endpoints. While 10.9 server is 10.8 client compatible, there are issues when a 10.9 generated-schema wants to work with 10.8 server.

Just need to come up with a good time within the 2024 calendar year to make 10.9 servers a minimum, and we should also probably add a badge/indicator to the README for it. Probably do at the same time as the iOS 16 minimum release.

Previous PR for reference: https://github.com/jellyfin/Swiftfin/pull/1072

SputNikPlop:
Would the bump to iOS 16 be possible after the iOS 18 release?

JPKribs:
This is somewhat complete since are now on 10.10. Should we document this anywhere to close this out or is the 10.10 badge in the README sufficient?

PR #1501, opened 2025-04-16
[WIP] [iOS & tvOS] Check API / Server Version

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1091

---

**This PR is a starting point but does not fully resolve the issue. If there are folks looking to help out with Swiftfin this would be a great intermediate-level PR to help out. If you are interested in taking this on, please let me know in the comments!Otherwise, I can work on this later down the road.**

---

The goal of this PR is to check the Jellyfin Version and compare it against the current API version we are using for Swiftfin. If the version is too old (or too new!), it should warn the user that their API version might result in a degraded experience. The current version results in an error but I think it should be a warning since *most* APIs aren't going to cause issues between like 10.9 and 10.10 but it's still worth letting the user know without fully blocking usage. If they do run into issues, they will have a warning / alert to give them the first troubleshooting step.

iOS was very simple. I added a check in `ServerCheckViewModel` and throw an error if the Server version is too low.

tvOS there was no `ServerCheckView` so I had to update the coordinator to add one. Because iOS and tvOS have differing coordinators, there isn't a clean way to get to the `SettingsView` from `ServerCheckView`. For now, I've added a `Switch Users` option but I am open to this changing. This was just the easiest route to preventing soft-locking Swiftin on tvOS.

---

### TODO

- [ ] Perform the same check when adding the Jellyfin Server. This data should be available in the [Public Server Info](https://api.jellyfin.org/#tag/System/operation/GetPublicSystemInfo). Have this produce a warning. Allow them to add the server to Swiftfin but alert them that some features may not work correctly.
- [ ] Perform the same check when signing into a Jellyfin User or Switching users. Either should be fine. Basically, when going from `SelectUserView` to `HomeView`, we should be validating that their API version is compliant with Swiftfin.
- [ ] Make a less intrusive UI. Right now, I am using the `ServerCheckView` and fully erroring out using the `ErrorView`. It makes more sense for this to be a non-error view. Instead, just have a `WarningView` on the `ServerCheckView` that says "Your version of Jellyfin is out of date. Some features may not perform correctly." with an `Okay` button. This would require a change to `ServerCheckView` to return a new event for `serverVersionWarning` or something like that.
- [ ] Add a badge to the `SettingsView` or something as a persistent badge that their version version is out of date.
- [ ] Add a route to the `SettingsView` for tvOS instead of just a "Switch Users" option. This should better mirror iOS.
- [ ] General cleanup and localization. This current version is rushed. A final version will need to be localized in the error messages.

---

### Screenshots [WIP]

Both of these are created artificially changing the minimum version. The actual minimum version is 10.10 based on API.

<details>

<summary>tvOS</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-04-16 at 13 14 17

</details>

<details>

<summary>iOS</summary>

Simulator Screenshot - iPhone 16 Pro - 2025-04-16 at 11 25 15

</details>

### DIFF
--- a/Shared/Coordinators/MainCoordinator/tvOSMainCoordinator.swift
+++ b/Shared/Coordinators/MainCoordinator/tvOSMainCoordinator.swift
@@ -29,6 +29,8 @@ final class MainCoordinator: NavigationCoordinatable {
     var mainTab = makeMainTab
     @Root
     var selectUser = makeSelectUser
+    @Root
+    var serverCheck = makeServerCheck
 
     init() {
 
@@ -41,7 +43,7 @@ final class MainCoordinator: NavigationCoordinatable {
                 if Container.shared.currentUserSession() != nil {
                     await MainActor.run {
                         withAnimation(.linear(duration: 0.1)) {
-                            let _ = root(\.mainTab)
+                            let _ = root(\.serverCheck)
                         }
                     }
                 } else {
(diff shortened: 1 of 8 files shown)


JPKribs:
We should be able to get the SDK version from the SDK now per:

https://github.com/jellyfin/jellyfin-sdk-swift/pull/45

We should use this instead of the static value in the plist.info

PR #1504, opened 2025-04-18
[iOS & tvOS] Improve rounded corner performance

### LINKED ISSUE #1110: Super choppy scrolling on M4 iPad Pro
### Describe the bug

I see a huge performance regression while browsing the library. It looks like 30fps refresh. This is on the latest M4 iPad Pro with latest 17.5.1 and Swiftin 1.1. Other apps do not have this issue. Uninstall/reinstall didn't fix the issue. 

### Application version

1.1

App Store

### Device information

iPad Pro 2024

### OS version

17.5.1

### Jellyfin server version

10.9.4

samglt:
Rounding all four corners of a View using a bezier path (which RoundedCorner uses) performs worse than a RoundedRectangle and results in poor performance on screens that have many rounded objects like the home screen or the Moves/Series screens. Note that UnevenRoundedRectangle performs as poorly as RoundedCorner.

In my testing this is more obvious on an iPad with lots of posters visible on screen, but it also causes more subtle hitching on my phone.

Mostly fixes #1110, but the posterShadow effect also contributes to the hitches more subtly.

### DIFF
--- a/Shared/Extensions/ViewExtensions/ViewExtensions.swift
+++ b/Shared/Extensions/ViewExtensions/ViewExtensions.swift
@@ -161,8 +161,13 @@ extension View {
         shadow(radius: 4, y: 2)
     }
 
+    @ViewBuilder
     func cornerRadius(_ radius: CGFloat, corners: UIRectCorner) -> some View {
-        clipShape(RoundedCorner(radius: radius, corners: corners))
+        if corners == .allCorners {
+            clipShape(RoundedRectangle(cornerRadius: radius))
+        } else {
+            clipShape(RoundedCorner(radius: radius, corners: corners))
+        }
     }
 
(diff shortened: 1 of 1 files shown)


JPKribs:
Tested and performance does seem to be improved! I'm not aware of what the wider implications of this change could be (if any).

I'll defer to @LePips for sign off

chickdan:
Interesting! I didn't actually know about RoundedCorner and have always used RoundedRectangle 😅 I wish there was an easier way to know which approaches are more performant since SwiftUI is so flexible.

LePips (approved):
Thank you and this is good for now. When we drop iOS 15 we can just move to the new option that provides corner customization.

- https://developer.apple.com/documentation/swiftui/shape/rect(topleadingradius:bottomleadingradius:bottomtrailingradius:toptrailingradius:style:)

Issue #430, opened 2022-05-09
Pause/Resume playback when taking/ending calls


noranraskin:
### Describe the bug

When taking a call on iPad while playing media with SwiftFin, playback just continues and I have to pause manually.
**Desired behavior:**
Media playback pauses automatically when getting a call and continues after ending the call.

### Application version

1.0.0 (70)

### Where did you install the app from?

TestFlight

### Device information

iPad Air 4. Generation

### OS version

iPadOS 15.4.1

### Jellyfin server version

10.7.7

DAcodedBEAT:
Hello, this still looks to be an issue today on Swiftfin version `1.1 (78)`.

samglt:
On my devices this also causes playback to stop working and an infinite hang once you exit the video (see attached demo). Observing the AVAudioSession interruptionNotification and handling it similar to Apple's NowPlayable demo fixes this hang. I see that #1203 is in progress and will solve this problem eventually, but since that's delayed and won't reach iOS 15 I'll put up a small PR just to fix this issue.

https://github.com/user-attachments/assets/c46a0096-c9c3-4f7e-a359-0e69b8c385f2

JPKribs:
This should be resolved by https://github.com/jellyfin/Swiftfin/pull/1581 and should be available in our latest TestFlight release.

PR #1506, opened 2025-04-20
[iOS] Handle interruption notifications

samglt:
Fixes issue #430 where interruptions break the player and causes the app to hang indefinitely. I know there's a big refactor coming in #1203 but if that's not going to be included in the next version or in iOS15, I think it's worth fixing this before then.


LePips (approved):
I'm picking up my video player work again and while the implementation into the manager will be redone, I will merge so that it's there for me to reference.

PR #1507, opened 2025-04-21
[iOS & tvOS] Explicitly request transparent logos

### LINKED ISSUE #1240: Image Transparency
### Describe the bug

Looks like Swiftfin is handling image transparency not quite the same as the Jellyfin web app:

This is how it appears in the Jellyfin web app:

image

This is how it appears in Swiftin (both tvOS and iOS):

image

For reference, this is for The Circle (https://www.imdb.com/title/tt9581768/)

### Application version

1.0.1

App Store

### Device information

Apple TV 4K (1st Generation)

### OS version

tvOS 17.6.1

### Jellyfin server version

10.9.11

samglt:
Fixes #1240. Currently with Jellyfin's API, logos that are stored as SVG's on the server are served as non-transparent jpegs unless we explicitly request a transparent format like webp.

Using png since I see no indication that tvOS supports webp.

### DIFF
--- a/Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto+Images.swift
+++ b/Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto+Images.swift
@@ -101,7 +101,8 @@ extension BaseItemDto {
         let parameters = Paths.GetItemImageParameters(
             maxWidth: scaleWidth,
             maxHeight: scaleHeight,
-            tag: tag
+            tag: tag,
+            format: type == .logo ? .png : nil
         )
 
         let request = Paths.getItemImage(


thornbill:
iirc if you don't specify any parameters that requires the server to transform the image (looks like you are specifying a max height and width here) the server will return the image in the original format. `tag` should **always** be included to avoid caching issues when the image is changed.

P.S. You likely want to use _fill_ height/width instead of "max" to avoid images appearing zoomed. https://github.com/jellyfin/jellyfin/pull/5556

LePips:
> iirc if you don't specify any parameters that requires the server to transform the image ... the server will return the image in the original format

Testing against a simple svg doesn't seem to get me these results, the server's image processing will always perform its query of client supported formats with a jpg default.

LePips (approved):
Thank you for your investigation of this issue as it caused me to also go in depth of the image processing that we should expect from the server.

This solution is good for now but I will be taking a look at our image requests soon to have the proper `Accept` header, fix some logging issues, and provide ways to clear the caches.

JPKribs:
This appears to have been resolved from the [server side as well](https://github.com/jellyfin/jellyfin/pull/14160) (Thank you for the report @samglt!). Do we need a second post-10.11 commit for this item as well use SVG or should this current resolution be fine?

PR #1510, opened 2025-04-23
Implement localization for hardcoded strings

rasko-dev:
Looking to find and replace as many hardcoded strings as I can with localized variants.


rasko-dev:
I came across the file Swiftfin/Views/VideoPlayer/VideoPlayer+KeyCommands.swift , which contains a lot of blocks like `KeyCommandAction(
                title: "Decrease Playback Speed",
                input: "[",
                modifierFlags: .command
            ) `
I'm wondering if these titles are something that should be localized? I'm not entirely sure what these are or where/if they're shown, so asking for you to decide.

JPKribs:
> I came across the file Swiftfin/Views/VideoPlayer/VideoPlayer+KeyCommands.swift , which contains a lot of blocks like `KeyCommandAction( title: "Decrease Playback Speed", input: "[", modifierFlags: .command ) `
> I'm wondering if these titles are something that should be localized? I'm not entirely sure what these are or where/if they're shown, so asking for you to decide.

Thanks for checking! I just looked at your PR and you are making quick work of this so thank you so much for your help!!

For the KeyCommandAction, the title should be localized but the rest should be left. For an example, please see line 104 from `VideoPlayerKeyCommandsModifier`:

```
            KeyCommandAction(
                title: L10n.jumpBackward,
                input: UIKeyCommand.inputLeftArrow
            )       
```

For you example, "Decrease Playback Speed" should be localized but "]" does not need to be.

rasko-dev:
I've gotten good progress on this and feel like I'm about done, however I have a few questions:

- In `Shared/Extensions/JellyfinAPI/MediaStream.swift` there are several properties about media, for example:
```Swift
if let value = type {
    properties.append(.init(title: "Type", subtitle: value.rawValue))
}

if let value = codec {
    properties.append(.init(title: "Codec", subtitle: value))
}
```
Are these "title" fields used internally and therefore can't easily be localized, or should they be localized? They are shown in the app when viewing a media item, at the bottom under About > Media > select a video/audio track.

- I've also skipped over anything relating to the download function, as this appears to still be a big work in progress, so a lot of things might be changed. I'll localize things related to download if you wish, however it feels like a lot of things might be changed in the future regarding this feature.

Other than these two questions I feel like this is completed, and just needs a quick check to make sure I haven't changed any strings I shouldn't have.

Swiftfin/Views/PagingLibraryView/Components/LibraryViewTypeToggle.swift
```diff
@@ -82,7 +82,7 @@ extension PagingLibraryView {
                 }
 
                 if viewType == .list, UIDevice.isPad {
-                    Stepper("Columns: \(listColumnCount)", value: $listColumnCount, in: 1 ... 3)
+                    Stepper(L10n.columns + ": \(listColumnCount)", value: $listColumnCount, in: 1 ... 3)
```
JPKribs, line 85: Can we create a `columnsWithCount()` or something like that instead? The only reason I say that is I'm unsure if every language is Noun + Count. Some might be Count + Name.

JPKribs:
> I've gotten good progress on this and feel like I'm about done, however I have a few questions:
> 
> 
> 
> - In `Shared/Extensions/JellyfinAPI/MediaStream.swift` there are several properties about media, for example:
> 
> ```Swift
> 
> if let value = type {
> 
>     properties.append(.init(title: "Type", subtitle: value.rawValue))
> 
> }
> 
> 
> 
> if let value = codec {
> 
>     properties.append(.init(title: "Codec", subtitle: value))
> 
> }
> 
> ```
> 
> Are these "title" fields used internally and therefore can't easily be localized, or should they be localized? They are shown in the app when viewing a media item, at the bottom under About > Media > select a video/audio track.
> 
> 
> 
> - I've also skipped over anything relating to the download function, as this appears to still be a big work in progress, so a lot of things might be changed. I'll localize things related to download if you wish, however it feels like a lot of things might be changed in the future regarding this feature.
> 
> 
> 
> Other than these two questions I feel like this is completed, and just needs a quick check to make sure I haven't changed any strings I shouldn't have.

You are correct, those aren't localized. Those media stream are part of the MediaStream properties. So it's a key:value pair so localizing it would create issues.

Downloads we can leave for now since many of those localizations will change before it's actually ready.

I left 1 comment but otherwise this looks great! Thank you again for doing this, this is a huge help to making Swiftfin more accessible!

rasko-dev:
Alright, I've implemented the change you requested, as well as changing out the ellipsis character `\u{2026}` with `...` for the `... See more` string. This seems in line with the already localized string `Learn more...`, although let me know if I'm overlooking something here.

Other than that I'd say this is ready, I've combed through the app and codebase and I'm struggling to find more unlocalized strings, although there could still be some left. So unless you have any other requests/comments I'm marking this as ready for review. Just a note that I have not tested these changes at all as I don't have any means to do so.

JPKribs:
> edit: also I've noted that the client log does not have localized buttons such as "close", "Network", "Logs", "All" however this seems to be an imported package and I don't see any easy way to do so, however since this is a technical feature this might not be a nessecary item.

Those logs are from an external package so we don't have control over those. Additionally, error logs that mirror the development team's language don't hurt us too badly as all of the github issues and documentation will have the same language, making it easier to google / search.

I appreciate that you also fixed capitalization while you did this! That should prevent conflicts with https://github.com/jellyfin/Swiftfin/pull/1511 so that's a great help!

JPKribs (dismissed):
This looks great! Thank you for helping with this!

---

Going to fix the merge conflict. It's going to look weird for a second but this should be resolved shortly

rasko-dev:
It appears the localizable.strings file is missing a semicolon at line 416 😅

JPKribs:
> It appears the localizable.strings file is missing a semicolon at line 416 😅

Good catch! There is annoying issue right now depending on when you branched Swiftfin the English localizations can default back to ASCII instead of UTF-16LE. That was why this had a merge conflict. The contents was good but the encoding was off. Nothing you did wrong at all just an annoying thing we found recently. Normally, just pulling it down and copying over to a UTF-16LE file is enough but I somehow lost a semicolon in the process?

Should be good now. Sorry about that!

Screenshot 2025-04-25 at 14 13 10

JPKribs (approved):
Approved! I just had to break it first lol

Shared/Components/TruncatedText.swift
```diff
@@ -31,7 +31,7 @@ struct TruncatedText: View {
 
     private var isTruncatedBinding: Binding<Bool>
     private var onSeeMore: () -> Void
-    private let seeMoreText = "\u{2026} See More"
+    private let seeMoreText = L10n.ellipsisSeeMore
```
LePips, line 34: For now, I would ask we don't localize the strings with an ellipsis explicitly. This is because I will probably remove the ellipsis from these strings and design altogether and trying to get it accurately represented in the strings file and ensuring it's localized in all languages is just more work. I also want to minimize strings as possible.

So, for these just concatenate them: `"\u{2026}" + L10n.ellipsisSeeMore`

Swiftfin/Components/ListTitleSection.swift
```diff
@@ -33,7 +33,7 @@ struct ListTitleSection: View {
                 }
 
                 if let onLearnMore {
-                    Button("Learn More\u{2026}", action: onLearnMore)
```
LePips, line 36: Same ellipsis argument.

Swiftfin/Components/ListTitleSection.swift
```diff
@@ -99,7 +99,7 @@ struct InsetGroupedListHeader<Content: View>: View {
                 }
 
                 if onLearnMore != nil {
-                    Text("Learn More\u{2026}")
+                    Text(L10n.learnMoreEllipsis)
```
LePips, line 102: Same ellipsis argument.

LePips (changes requested):
Thank you for your contribution! Just request some strings are reverted.

LePips (approved):

PR #1511, opened 2025-04-23
Convert all English UI text from Title Case to Sentence case

xsiviso:
Part A:
- [x] Convert all English UI text from Title Case to Sentence case
- [x] Ensure proper nouns remain capitalized (e.g., Jellyfin, Swiftfin, GitHub, etc.)

Part B:
- [x] Verify that navigation titles use the localizedCapitalized() function to ensure view titles are capitalized appropriately


xsiviso:
For Part B: Do you have to add `localizedCapitalize` to a `.navigationTitle()`?
for example: change `.navigationTitle(L10n.device)` to  `.navigationTitle(L10n.device.localizedCapitalized)`

JPKribs:
This looks great so far! Looking at the changes, I'm realizing I might have been our biggest capitalization offender here haha.

For Part B, yes, something like: `.navigationTitle(L10n.home.localizedCapitalized)`. If the title is a single word, I don't believe this is a requirement. There are also some very rare occasions where we might need another string to localize a title. A good example of this might be:

`"liveTVAccess" = "Live TV access"`

`.navigationTitle(L10n.liveTVAccess.localizedCapitalized)`

Which results in the title annoyingly being set to "Live Tv Access". In this case, we may need to create an alternative string as `"liveTVAccessTitle" = "Live TV Access"` or `"TVAccess" = "TV Access"`.

I believe I did that prior with that exact example. This is the part that gets to be more of a judgement call or a 'change and test'. 

**I would hold off and hopefully @LePips can give you a more definitive answer to handling `navigationTitle`. I'm not an iOS developer by trade so they can give us a better, more industry standard answer for how they want to handle `navigationTitle` casing.**

Issue #178, opened 2021-10-09
Add support for downloading subtitles


edenman:
**Describe the feature you'd like**
Ability to search for and download subtitles using the OpenSubtitles plugin.  This exists in the Expo client but I didn't see it tracked in this project.

LePips:
Plug-ins are something that I've thought about but will be very far out.

farverio:
Just wanted to voice support for this feature as well. Something very helpful that infuse offers and is particularly nice for non native speakers

PR #1513, opened 2025-04-24
Subtitle Management

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/178

This PR enables managing subtitles. This completes the long journey for Admin Functionality I started [here](https://github.com/jellyfin/Swiftfin/discussions/1283). Other items will definitely come up that need to get added but I feel comfortable closing out that thread with this change as feature items will be more "maintenance". So, less need for a huge, multi-part thread.

This PR enables the following functionality:

1. View all subtitles for a file. Separated into embedded and external.
2. Select a subtitle for subtitle details
3. Enable deletion of **multiple\*** external subtitles
4. Allow uploading a new **text-based\*\*** subtitle
5. Allow searching Subtitle Providers and selecting **multiple\*** subtitles for download

\*Unique to Swiftin. Jellyfin-Web (and the API) only allow single item changes.
\*\*Image based subtitles cannot be uploaded via the API. Mirrors Jellyfin-Web.

---

### Quirks

I am using the `MediaStream` to get the subtitles for an item. For this reason, I need to reload the item when a subtitle is added or delete to re-get all of the media streams. The reason I cannot just append/remove subtitles is the index can change pretty dramatically based on actions. So reloading the item ensures that I always have the correct index which is especially important when deleting.

The delete function on server is for a single index. So, when deleting multiple I need to:

1. Delete Index N
2. Re-get Item
3. Delete Index N + 1
4. Repeat

It's not super efficient but this view likely isn't getting a ton of traffic (I doubt this is a feature even I will use haha) so I think the performance hit is fine to better ensure that we are always getting the appropriate index.

I've been testing this for a while and I don't know if there is a better way to do this but I am open to suggestions!

---

iOS allows for deleting subtitles, uploading text subtitles, or searching for subtitles.

tvOS allows for searching for subtitles.

---

### Video

<details> 

<summary>tvOS</summary> 

**Uploading**

https://github.com/user-attachments/assets/38551099-832f-4aec-b78f-b9a105227e94

</details>

<details> 

<summary>iOS</summary> 

**Management**

https://github.com/user-attachments/assets/a1a12cf9-68ff-4532-93d1-b0b97c6aa6a1

**Uploading**

https://github.com/user-attachments/assets/74697fd8-8fe2-45ce-a7d8-ddc9d2f48c7a

</details>


Shared/ViewModels/ItemAdministration/ItemSubtitlesViewModel.swift
```diff
@@ -0,0 +1,353 @@
...
+                    self.eventSubject.send(.error(JellyfinAPIError(error.localizedDescription)))
+                }
+            }
+        }
+        .asAnyCancellable()
+    }
+
+    // MARK: - Delete Subtitle
+
+    private func deleteSubtitles(mediaStreams: Set<MediaStream>) async throws {
```
JPKribs, line 246: I'd appreciate a second set of eyes testing this. I was leaning towards only allowing 1 subtitle to be deleted at a time (mirroring web) but this has been working for all of my testing. That being said, I don't want to be responsible for people having the wrong subtitle delete. The downside of this is the performance hit of having to pull the tracks each time to re-validate the ID for deletion.

LePips:
As a note, this is too big for me to properly review for 1.3 as now I want to get that out soon.

JPKribs:
This should be Post-1.3 and iOS 16 ready now! iOS 16 let me add a nice QOL change where I now default the subtitle search and upload language to the user language with the new `.alpha3` API.

This should be ready to go but I think this is lower priority as this is a fairly large change. Luckily, aside from new strings, this is entirely self-contained so this shouldn't have any conflicts with other commits.

Swiftfin/Views/ItemEditorView/ItemSubtitles/ItemSubtitlesView.swift
```diff
@@ -0,0 +1,245 @@
...
+                }
+            }
+        }
+    }
+
+    // MARK: - Subtitle Detail Sheet
+
+    @ViewBuilder
+    private var expandedSubtitleSheet: some View {
+        NavigationView {
```
LePips: Now that we're iOS 16 min, use `NavigationStack`.

Swiftfin/Views/ItemEditorView/ItemSubtitles/ItemSubtitlesView.swift
```diff
@@ -0,0 +1,245 @@
...
+                    .navigationBarCloseButton {
+                        expandedSubtitle = nil
+                    }
+            }
+        }
+    }
+
+    // MARK: - Add/Remove Selected Subtitle
+
+    private func toggleSelection(for subtitle: MediaStream) {
```
LePips: Use the `Set.toggle` function instead. There are other places throughout that it can be used instead.

LePips (changes requested):
Upon initial review, please change the API usages:
```
foregroundColor -> foregroundStyle
navigationBarTitle -> navigationTitle
```

As well as using `if let` shorthand where appropriate.

On iOS, I would also prefer when viewing a subtitle's contents that it was pushed instead of a separate modal.

JPKribs:
> foregroundColor -> foregroundStyle
> navigationBarTitle -> navigationTitle

Is `.navigationBarTitleDisplayMode(...)` still fine?

> On iOS, I would also prefer when viewing a subtitle's contents that it was pushed instead of a separate modal.

When you say contents do you mean the `MediaStreamInfoView`? Or the whole `ItemSubtitlesView`?

LePips:
> Is .navigationBarTitleDisplayMode(...) still fine?

Yes. `navigationBarTitle` has been deprecated for a while now for `navigationTitle`.

> When you say contents do you mean the MediaStreamInfoView?

Ah, I missed what this view actually was. Ignore that comment and we can evaluate this again when we replace Stinsen.

JPKribs:
> Yes. `navigationBarTitle` has been deprecated for a while now for `navigationTitle`.

Apologies, I was asking about the navigationBarTitle**DisplayMode**. I've been setting it to `.inline`. Is that no longer applicable moving away from `navigationBarTitle`?

> we can evaluate this again when we replace Stinsen.

I'll add in a TODO there so I can remember to clean that up post-Stinsen!

JPKribs:
Ugh, that `AlphabetizeStrings.swift` build issue drives me crazy. The issue is sometimes the subtitles get saved back to git has `.utf8`. You can fix the build issues by changing the read portion of `AlphabetizeStrings.swift` to read it in as `.utf8` instead of `.utf16`. If you keep the write it at the end as `.utf16`, you can reverse the `.utf8` change afterwards.

There's gotta be a better way to handle this situation, I feel like the subtitles get encoded back to UTF8 all the time and I don't really know why.

LePips:
> navigationBarTitleDisplayMode

Yes, we need to still use `navigationBarTitleDisplayMode`.

> AlphabetizeStrings.swift build issue

I think something must have happened during your merge from main, as adding new strings fresh from main does not have any build issues. We should not be changing the encoding of the file, this will not allow it to be read for localization.

I would recommend a rebase of your work on top of main instead.

JPKribs:
> I would recommend a rebase of your work on top of main instead.

I'll look to do that later this week! I think I'm a bit burnt on code from the weekend so I might take a couple days to look at it.

If I open a new branch it's because I failed to re-base on Main haha

---

### Edit:

image

PR #1518, opened 2025-04-26
[tvOS] Fix Next Up Overlapping with TabBar

### LINKED ISSUE #1479: [tvOS] (master) The top bar clips over the 'Next Up' posters (when set to landscape)
### Describe the bug

(Reference: https://github.com/jellyfin/Swiftfin/discussions/1294#discussioncomment-12742078)

1. Setup next up section to have landscape instead of posters

Image

### Application version

master

None

### Device information

Apple TV 4K (Ethernet)

### OS version

tvOS 18.3.1

### Jellyfin server version

10.10.7

chickdan:
The issue here is that the ZStack has `.ignoresSafeArea()` which allows for the CinematicResumeView to overlap with the TabBar but if there is no CinematicResumeView then ignoring the safe area allows the NextUpView to move all of the way up to the top edge. The fix here is to simply apply `.safeAreaPadding()` when there is no CinematicResumeView.

I had considered the possibility of adding this to the PosterHStack inside of `NextUpView` but decided this is most likely a one-off issue so a one-off solution will suffice.

The value here is rather arbitrary so there is very likely a better/more dynamic way at determining the amount, I was originally thinking of using GeometryReader but I made the mistake of not taking notes and have now forgotten the idea I had :/

>[!NOTE]
>This issue occurs whether the posters are portrait or landscape

Screenshot 2025-04-26 at 3 23 42 PM

closes #1479

### DIFF
--- a/Swiftfin tvOS/Views/HomeView/HomeView.swift
+++ b/Swiftfin tvOS/Views/HomeView/HomeView.swift
@@ -39,7 +39,9 @@ struct HomeView: View {
                     if showRecentlyAdded {
                         CinematicRecentlyAddedView(viewModel: viewModel.recentlyAddedViewModel)
                     }
+
                     NextUpView(viewModel: viewModel.nextUpViewModel)
+                        .safeAreaPadding(.top, 150)
                 }
 
                 ForEach(viewModel.libraries) { viewModel in


JPKribs:
I _think_ this is good but tvOS `HomeView` is weird for testing. I remember when I was last looking at this it wasn't appropriately `stateful` so the order things load in can cause trouble. Especially with the cinematic items up top. That'd be my only concern.

IMO, this should be good and we should try to get `HomeView` stateful soon before the TestFlight regardless of if we do the rework / skeleton loading. 

For this PR, I'll test this and make sure it doesn't break if things load in out of order

sjorge:
I added this PR to the build I am running and it does seem to fix https://github.com/jellyfin/Swiftfin/issues/1479 for me 🚀

LePips (approved):
I'm okay with this for now but my ultimate solution is to have a design for the home view that doesn't strongly type based on `_Resume_`, `_RecentlyAdded_`, `_NextUp_` and so on: https://github.com/jellyfin/Swiftfin/pull/1382#issuecomment-2584316878

That just means this view/work would be overridden when that happens.

Issue #1229, opened 2024-09-05
Siri Remote, play pause with the big center button.


ChristopheBis:
Hi,

Is there a way to have the behaviour of the center button of the Siri Remote working like the other apps (play /pause)?
The behaviour is the same for Apple TV, prime, YouTube but swift fin is different and frustrating, I have to press the real play pause button which is weird.

Thank you

StarFox2873:
bump

vordenken:
Yes, please! I started using Swiftfin on tvOS and thought the app was broken because I couldn't pause the video.
I never used the "play/pause" button before. This would match every other video players behaviour on tvOS.

OlivierAlbertini:
Hi,

I did it locally, my children were upset xd. Are you still interested ?

JPKribs:
This item should be resolved as of 1.5 on tvOS. When the overlay is present, selecting the progress bar will pause/play content.

PR #1520, opened 2025-04-26
[tvOS] Select button now pauses/plays

ryanh2o3:
Fixes issue #1229 

This PR modified the select button handling to pause/play content while still displaying the overlay.
I think this behaviour is more in line with what other tvOS apps normally do. 
I tested this on Apple TV 4K via both the iOS app remote and LG Magic remote.


ryanh2o3:
Alternatively I could have it do the same thing as the pause/play button and just display the overlay with no timeout. 
`PressCommandAction(title: L10n.pressDownForMenu, press: .select) {
                        if videoPlayerManager.state == .playing {
                            videoPlayerManager.proxy.pause()
                            withAnimation(.linear(duration: 0.3)) {
                                isPresentingOverlay = true
                            }
                        } else if videoPlayerManager.state == .paused {
                            videoPlayerManager.proxy.play()
                            withAnimation(.linear(duration: 0.3)) {
                                isPresentingOverlay = false
                            }
                        }
                    }`

ryanh2o3:
Just realised this might conflict with #1203

LePips:
The files for this are indeed all overwritten in #1203, so idk exactly what to do with this since it won't make a release before that work is done.

ryanh2o3:
No worries, I'll just close the pr.

Issue #554, opened 2022-09-03
Replace UDPBroadcast


LePips:
Replace UDPBroadcast with a different socket framework, picking between [SwiftNIO](https://github.com/apple/swift-nio) (more bare metal) and [BlueSocket](https://github.com/Kitura/BlueSocket) (has basic TCP/UDP sockets, just what we need). All we need for this is a basic UDP broadcast/multicast pings and listener for server pongs.

Needs to work along with https://github.com/jellyfin/Swiftfin/issues/408, if applicable.

Should think about moving this overall functionality to the sdk instead.

LePips:
I have a simple replacement using BlueSocket however it would require the multicast entitlement. This is a problem when it comes to development.

anthonylavado:
I have applied for the multicast entitlement, just in case.

anthonylavado:
Updating here - we have received the entitlement from the App Store.

DT3264:
Any updates around this one? / what work is required to be able to work on this issue?

LePips:
There is no technical issue, just a local replacement would be needed.

Issue #408, opened 2022-04-06
Support Websocket Remote Control


anthonylavado:
**Describe the feature you'd like**
It would be great if the player would be able to respond to the remote control events that can be sent from other clients. Such commands include play/pause, seeking to a different timestamp, and even displaying a message onscreen in a notification toast.

**Additional context**
It's my understanding that the remote control functions use websockets and the `sessions` endpoint on the server.

Attached is a picture of what the remote control function looks like in the web interface on an iPhone. In this case, it was controlling playback of a recorded show in Firefox on desktop.

![image](https://user-images.githubusercontent.com/596194/162043210-5c168b04-9566-4bc8-a0b1-e5ed3791d7ed.jpeg)

Schlauer-Hax:
Whats the status on this? Is there already a websocket connection or will this have to be written from scratch?

dechamps:
I would also very much like to see this, especially "Play on [device]" support (#1299). I use Swiftfin on an Apple TV 4K but I don't own any other Apple device. This means the only way to search for something is to use the on-screen keyboard with the remote. That gets tedious fast. It would be much easier if it was possible to search for the media on a separate Jellyfin client (e.g. Android phone) and then just use the "Play on" feature to "cast" it to the Apple TV.

PR #1533, opened 2025-05-06
Use Server Discovery from SDK

JPKribs:
### Summary

Resolves(?): https://github.com/jellyfin/Swiftfin/issues/554
Relies on: https://github.com/jellyfin/jellyfin-sdk-swift/pull/47

---

All of the work for this PR is done, but this branch currently uses [my own branch with the ServerDiscovery changes.](https://github.com/JPKribs/jellyfin-sdk-swift/tree/serverDiscovery) for the JellyfinSDK. When we have a final version of [the ServerDiscovery changes](https://github.com/jellyfin/jellyfin-sdk-swift/pull/47), I will update this PR to use the updated 0.5.2(?) Swift SDK. For now, this version of the SDK and this PR can be used to validate the changes for [the SDK changes](https://github.com/jellyfin/jellyfin-sdk-swift/pull/47).

---

The primary goal of this PR is to replace UDPBroadcast with Apple's [Swift-NIO](https://github.com/apple/swift-nio).

A secondary goal of this PR is to make the JellyfinDiscovery process more robust. None of these are huge issues but I thought I would address them while I was here:

1. The previous `deinit` for `ServerDiscovery` was flaky. As a result, if you opened the `ConnectToServerView` then closed it, then reopened it, JellyfinDiscovery would break until the app restarted. 
2. Due to 1), I wanted to try and make my implementation "recover" on failure. I've thrown a lot at this and it appears to work how I expect so I feel this make it more robust and less likely to break. 
3. IF JellyfinDiscovery breaks, there is currently no error or notice to the user that is the case. I've added some more visible notice so the user can be alerted that JellyfinDiscovery is no longer proceeding.
4. There was no visible indication that JellyfinDiscovery was running/searching. I just added this to the existing `ProgressView` so the user could see there was something running.
5. This last item is very rare but if you duplicates a Jellyfin server the ServerID is the same between both devices. The result is that the first to respond shows up in the list since `ServerDiscoveryResponse` is `Identifiable`. I changed the `ForEach` to used `\.Self` instead to resolve this but the other route would be to make the `id: serverID + address` or something like that but `\.Self` was the easiest route in case we decided we didn't want this item. Realistically, 99% of people are never going to have this issue. I only have this because I cloned my server for a Dev server / testing EFCore.

---

<details>

<summary>New Error Message on Crash</summary>

The red "Error" is shown if there a problem then the `localizedDescription` is shown under. In this case, Channel not ready. Swiftfin then tries again in 12 seconds per the Discovery Broadcast Timer. I was considering a version of this with a timer.

Simulator Screenshot - iPhone 16 Pro - 2025-05-06 at 13 16 12

**With Timer (Not in PR)**

https://github.com/user-attachments/assets/952459ac-3cf7-452d-b287-02f94a97ebfd

</details>

---

### Why NIO over BlueSocket?

I've selected SwiftNIO over BlueSocket for our UDP broadcast/multicast implementation primarily because it's backed by Apple, and I have more faith that this will get updates than other packages. When I looked at [Bluesocket](https://github.com/Kitura/BlueSocket), I was concerned that the last release was in 2021.

The usage for NIO was a little more... manual(?) than I would have liked but it looks like NIO is very granular so future usage for things like WebSocketting shouldn't be roadblocked due to lack of functionality. In my brief research, I found that SwiftNIO is described as having "excellent" WebSocket support (Thinking forward to #408). I originally did this with the `Network` package but I decided NIO would be good to get the ball rolling with and learning so I could potentially look at #408 down the road. Probably after #1203 since I don't really want to touch playback code until it's settled haha.


JPKribs:
Functional for both IPv4 and IPv6. This recovers if there are issues and appropriately displays errors / loading.

The only thing I don't love.... I ran into this error: https://github.com/apple/swift-nio/issues/1494. It didn't impact anything or cause issues but to get that error to go away I needed to add:

```
    private static func getDeviceIPv4Address() -> String? {
        var ptr: UnsafeMutablePointer<ifaddrs>?
        defer { ptr.flatMap(freeifaddrs) }
        guard getifaddrs(&ptr) == 0, let first = ptr else { return nil }

        for cur in sequence(first: first, next: { $0.pointee.ifa_next }) {
            let flags = Int32(cur.pointee.ifa_flags)
            guard flags & IFF_UP != 0, flags & IFF_LOOPBACK == 0 else { continue }
            guard cur.pointee.ifa_addr.pointee.sa_family == sa_family_t(AF_INET) else { continue }

            var addr = cur.pointee.ifa_addr.withMemoryRebound(
                to: sockaddr_in.self, capacity: 1
            ) { $0.pointee.sin_addr }
            var buf = [CChar](repeating: 0, count: Int(INET_ADDRSTRLEN))
            inet_ntop(AF_INET, &addr, &buf, socklen_t(INET_ADDRSTRLEN))
            return String(cString: buf)
        }
        return nil
    }
```

It's just so unnecessary since we kind of DO want to bind to all sources. For example, tvOS with Ethernet, it would make the most sense to bind to all sources at 0.0.0.0?

That's the only part of this that seems off. Otherwise, I am happy with this PR.

sjorge:
Ran a build with this, not sure how to test though. I removed the server and it showed up as a discovered server I could add.

Is it only used for that?

JPKribs:
> Is it only used for that?

For now, yes. So, if i did this correctly, it should be unnoticeable. Hopefully a little more robust but I don’t think people sit on this screen and try to crash it like I did. So mostly, does network discovery still work.

This PR is more about switching packages so we can start looking at web socketing for things like Sync Play. We’re still a long ways from that point but this is a distant step 1

sjorge:
My network is dualstack, but it just lists it with the fqdn... is there a way to check without a packet trace to see if it was using IPv6?

JPKribs:
I think that would require putting the JF server on only the IPv6? At least, that's what I was doing to test it. What I did was, I put the server on exclusively IPV6 and did the same for my laptop with the emulator.

IPv6 is frankly something I haven't really worked with a lot so I very well could have done it incorrectly to test it

JPKribs:
That's awesome! Thank you for the test! There was no reason that _shouldn't_ have worked, but I'm so green on V6 that it's just good to have a second set of eyes. I appreciate the help!

LePips:
This is an amazing thing to upgrade. I've previously tried otherwise with Starscreen, NIO, or Bluesocet but gave up due to other priorities.

> go in the SDK instead?

I actually would be more interested in this living in the SDK, like how we provide the `QuickConnect` object.

JPKribs:
Opened: https://github.com/jellyfin/jellyfin-sdk-swift/pull/47

JPKribs:
I'm trying to cleanup old PRs, so I think the game plan for this one moving forward:

1. WebSocket - More important and should determine our usage of NIO
2. If we are using NIO, this should be ready. If we are not using NIO for the websocket then this can be redone to not include that dependency. Whatever we use for the WebSocket should be used for this so that should make our decision
3. Merge to main and clean this up. I think the UI changes here are fine but I think I've learned a bit since last May so I will want to take another pass on this regardless.

JPKribs:
Closing this out. The main change of this is on the SDK and this is pretty dated now. I will pick this back up when we have the changes in the SDK to update the unified view for this.

Issue #1537, opened 2025-05-14
HTML Tags Appear in Overview Text


sagar5534:
## 🐛 Issue: HTML Tags Appear in Overview Text

I've noticed that **raw HTML tags (e.g., `<br>`, `<i>`, etc.) are appearing in the Overview labels** for Movies/Shows.

This likely stems from metadata providers who use basic HTML formatting in their descriptions. These HTML tags eventually get passed into Jellyfin and are displayed as-is, rather than being interpreted or stripped — resulting in a poor user experience.

---

## ✅ Proposed Fix (via PR)

I've opened a PR that resolves this by **sanitizing or converting the `overview` field**, depending on the context where it is displayed.

### 🔧 The fix includes two strategies:

1. **Sanitize (Remove HTML Tags)**  
   - For areas where clean, plain text is preferred — like truncated labels or condensed views — we simply strip HTML tags.  
   - This ensures no layout breaks and provides a more readable summary.

2. **Render as `AttributedString` (Rich Text Display)**  
   - In areas where there's more space (like full overview popups), we convert the HTML into an `AttributedString`.  
   - This preserves formatting like line breaks, italics, etc., giving a richer and more faithful display of the original metadata.

---

## ✨ Benefits

- Cleaner, more professional UI  
- Preserves intent of metadata providers when space allows  
- Prevents broken or ugly HTML from leaking into user-facing views

## ✨ Screenshots of Issue:
![Image](https://github.com/user-attachments/assets/290efbf9-17c6-4b43-86e5-dbd1d941ed3a)![Image](https://github.com/user-attachments/assets/c5586061-3520-4b45-9a95-27c6c854e896)
![Image](https://github.com/user-attachments/assets/2578e553-4493-44fe-81e7-3a50d4a9d580)

### Application version

Latest

### Where did you install the app from?

App Store

### Device information

IOS and TvOS

### OS version

Latest

### Jellyfin server version

Latest

PR #1538, opened 2025-05-14
Formatting overview label for HTML elements

sagar5534:
Resolves(?): https://github.com/jellyfin/Swiftfin/issues/1537

### 🔧 The fix includes two strategies:

1. **Sanitize (Remove HTML Tags)**  
   - For areas where clean, plain text is preferred — like truncated labels or condensed views — we simply strip HTML tags.  
   - This ensures no layout breaks and provides a more readable summary.

2. **Render as `AttributedString` (Rich Text Display)**  
   - In areas where there's more space (like full overview popups), we convert the HTML into an `AttributedString`.  
   - This preserves formatting like line breaks, italics, etc., giving a richer and more faithful display of the original metadata.

---

## ✨ Benefits

- Cleaner, more professional UI  
- Preserves intent of metadata providers when space allows  
- Prevents broken or ugly HTML from leaking into user-facing views

## ✨ Screenshots of fix:
Simulator Screenshot - iPhone 16 Pro - 2025-05-14 at 18 26 33
85)
Simulator Screenshot - iPhone 16 Pro - 2025-05-14 at 18 26 39

### DIFF
--- a/Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto.swift
+++ b/Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto.swift
@@ -273,4 +273,33 @@ extension BaseItemDto {
             nil
         }
     }
+
+    var formattedOverview: AttributedString? {
+        guard let overview else { return nil }
+        guard let data = overview.data(using: .utf8) else { return nil }
+
+        do {
+            let mutableAttributedString = try NSMutableAttributedString(
+                data: data,
+                options: [
+                    .documentType: NSAttributedString.DocumentType.html,
+                    .characterEncoding: String.Encoding.utf8.rawValue,
+                ],
+                documentAttributes: nil
+            )
+
+            // Remove font attributes from the entire range
+            let fullRange = NSRange(location: 0, length: mutableAttributedString.length)
+            mutableAttributedString.removeAttribute(.font, range: fullRange)
+
+            return AttributedString(mutableAttributedString)
+        } catch {
+            return nil
+        }
+    }
+
+    var cleanedOverview: String? {
+        guard let overview else { return nil }
+        return overview.replacingOccurrences(of: "<[^>]+>", with: "", options: .regularExpression)
+    }
 }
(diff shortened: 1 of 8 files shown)


JPKribs:
Hi, firstly thank you for your contribution! It's always awesome to see new contributors!

Just as a heads up, we have another open branch that also touches on this issue here: https://github.com/jellyfin/Swiftfin/pull/1067. There are some comments and history there that may be relevant to your PR as well. I'm not personally opposed to taking taking one PR over the other, especially since our original one has had some outstanding changes approaching the 1 year mark. Most likely, I'm going to lean on @LePips just to confirm we're going the right direction here but that PR should give you a good indication of potential issues/concerns we may have for this PR.

I'm out of the state this week for work so I won't personally have a chance to look at this until probably next week!

sjorge:
Seems to work fine on tvOS. At least in the short summary.

Never found a way to read the whole description, but that was already the case before this change too.

Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto.swift
```diff
@@ -273,4 +273,33 @@ extension BaseItemDto {
...
+                options: [
+                    .documentType: NSAttributedString.DocumentType.html,
+                    .characterEncoding: String.Encoding.utf8.rawValue,
+                ],
+                documentAttributes: nil
+            )
+
+            // Remove font attributes from the entire range
+            let fullRange = NSRange(location: 0, length: mutableAttributedString.length)
+            mutableAttributedString.removeAttribute(.font, range: fullRange)
```
LePips, line 293: This immediately removes all formatting.

LePips (changes requested):
Thank you for your contribution, but it still doesn't achieve the desired result when all formatting is immediately stripped. A final solution would have to look like a mashup between this and #1067 where we make a custom `UILabel`, get the environment `Font`, make the attributed string, and enumerate over it to apply bold, italics, and the environment font as desired.

I do like the consideration for a cleaned/parsed overview, and could move forward with that.

LePips:
If this has no movement soon, I will implement it with the enumerated formatting and the "cleaned" overview.

PR #1541, opened 2025-05-18
Add Playlists Support

JPKribs:
### Summary

Part 1 of 2 for: #609 
Replaces: https://github.com/jellyfin/Swiftfin/pull/1428

*This is a re-submission of https://github.com/jellyfin/Swiftfin/pull/1428. I made that PR based on the idea that Collections and Playlists could share a view. They behave just differently enough that I believe that they need separate views. This should look very similar to Collections but there are some differences.*

---

This PR adds Playlists to Swiftfin. This PR does not allow editing Playlists or adding items to Playlists. I intend to add that functionality to a later PR. This starts https://github.com/jellyfin/Swiftfin/issues/609 but I wouldn't consider #609 resolved until I can add the ability to add/remove items from a playlist. 

---

When I originally started this, I wanted to make it just look like the `CollectionView` and call it a day but I wasn't a huge fan of how this looked or functioned. So, I made it more of a list. Selecting the item takes you the `itemView` for that item. Selecting the play button starts playback.

Playlists automatically flatten themselves, so the `playlistView` has the `PlayButton` active. This selects either the first unplayed item OR, if all items are played, it selects the first item overall.

### Screenshots

<details>

<summary>iOS</summary>

Simulator Screenshot - iPhone 16 Pro - 2025-05-18 at 17 01 27

</details>

<details>

<summary>iPadOS</summary>

Simulator Screenshot - iPad Pro 13-inch (M4) - 2025-05-18 at 18 15 11

</details>

<details>

<summary>tvOS</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-05-18 at 18 16 15

</details>

### Video

https://github.com/user-attachments/assets/5490551e-95af-41a1-8739-b1c173aeccd8

### Why Not Use the `CollectionVGrid`?

I originally wanted to use that but I 1) don't know if I can move items list I can with a list and 2) couldn't get it to work since there is an Item header at the top of the view so it was conflicting with the `ScrollView`

### What Next?

Below are items I need to knock out in later PRs:

**PlaylistEditorViewModel**

- [ ] Create a `PlaylistEditorViewModel` for all of the editing functionality for playlists

**PlaylistView**

- [ ] Allow moving/reordering items in the `PlaylistView`
- [ ] Allow removing playlist items from the `PlaylistView`
- [ ] Allow creating playlists or adding items to existing playlists

PR #1543, opened 2025-05-20
Cleanup `ItemView`s

LePips:
During other work I realized again that these views can be cleaned up greatly, especially between iOS and iPadOS. iPadOS began with other views long ago when I thought they could have different layouts, but that hasn't come to fruition. This also brings some design unification I would like in 1.3.


JPKribs (changes requested):
This all looks good with the only wonky part being collections are back to just using the poster but scaling it down weirdly:

<details>

<summary>Example</summary>

Simulator Screenshot - iPhone 16 Pro - 2025-05-20 at 10 00 35
</details>

I don't think it's an issue to have collections not use the Cinematic/CompactPoster/Logo but would we be able to have this pull the .landscape for iOS? iPadOS looks to still be using the Cinematic same with all of the other item types.

LePips:
Thanks, that allowed me to do another sweep and fix some missed things. The current motivation for collections to use the simplified item view is because the default metadata providers don't get that much metadata to properly saturate one of the views with posters and logos.

However, I would be open to allowing item types, or specific media, to use certain view types selectable by the user.

JPKribs (dismissed):
> The current motivation for collections to use the simplified item view is because the default metadata providers don't get that much metadata to properly saturate one of the views with posters and logos.

I think that's likely a good call for [playlists](https://github.com/jellyfin/Swiftfin/pull/1541) as well. I like the idea of the full cinematic view but I agree it's best to gear views towards the most common use-cases.

Looks great! Thanks for humoring that change:

<details>

<summary>Updated View - Looks Good!</summary>

Simulator Screenshot - iPhone 16 Pro - 2025-05-20 at 12 54 27

</details>

JPKribs:
Not for this PR but just a note:

Screenshot 2025-05-20 at 13 03 30

When I moved all the `ActionButtons` to a single struct I think I need to use the `AccentColor.overlayColor` so we don't lose the check for lighter `AccentColor`. Also, I wonder if we want to add a divider or something when the `PlayButton` is missing to split the Attributes and the Action Buttons a little more?

LePips:
I would actually like to eventually update those buttons to a rectangular style. An affect of that is they wouldn't look so "floaty".

JPKribs:
> I would actually like to eventually update those buttons to a rectangular style. An affect of that is they wouldn't look so "floaty".

Similar to tvOS? Would that be reusable between both or would we want them to be different?

JPKribs (approved):

JPKribs:
@LePips I don't think this warrants a PR at this point but [I took a crack at the action buttons](https://github.com/JPKribs/Swiftfin/tree/actionButtons). I went the route of porting them over from tvOS. Is this kind of what you meant or something else?

<details>

<summary>Image 1</summary>

Simulator Screenshot - iPhone 16 Pro - 2025-05-21 at 15 41 46

</details>

<details>

<summary>Image 2</summary>

Simulator Screenshot - iPhone 16 Pro - 2025-05-21 at 15 41 58

</details>

<details>

<summary>Image 3</summary>

Simulator Screenshot - iPhone 16 Pro - 2025-05-21 at 15 42 32

</details>

<details>

<summary>Image 4</summary>

Simulator Screenshot - iPhone 16 Pro - 2025-05-21 at 15 44 34

</details>

<details>

<summary>Image 5</summary>

Simulator Screenshot - iPhone 16 Pro - 2025-05-21 at 15 45 10

</details>


<details>

<summary>iPadOS</summary>

simulator_screenshot_8ECD8033-FFE9-4961-AA87-27439FE17FF5

</details>

Issue #1532, opened 2025-05-06
Adopt VisionOS Circular Design


JPKribs:
**Describe the feature you'd like**

Starting an iOS 17, it looks like Apple has started adopting these circle backgrounds for their toolbar buttons. At this time, we don't really have a lot of toolbar buttons, so this wouldn't be insanely difficult to implement. It looks like just the icon with a material background. Please find a video below for more detail.

This would be an update for the back button, the close button for modal, and the navigation menu button. I can't think of any other locations that would need this update? Maybe, the plus icon in the `addItemImageView`?

Is this a direction we would want to go?

**Additional context**

Below is an example of the material backgrounds for the buttons and a very slick transition for the back button. I'm not sure how they did:

https://github.com/user-attachments/assets/c096600f-0d8e-4e11-b682-3dd8691d5645

JPKribs:
The part of the animation that I think is slick is they transition the white into the accent color as you scroll. Additionally, they transition from a circular background into a regular button once you reach the threshold (see left back button). 

Otherwise, this feels almost identical to the transition we already have!

LePips:
I did work on this a long time ago and I'll have to find the work again. This requires dipping into UIKit to set the navigation bar buttons correctly and performing the animations. I think I left off trying to make a friendly API and taking into account accessibility settings.

LePips:
Found it. I think I left off trying to animate it "just right" from the header chevron to the back button via expanding/shrinking, but should abandon that and instead just leave out the side like I think the App Store does it. This is more of a POC and I'd rather rewrite it using some iOS 16 APIs.

[LibraryRowButton.zip](https://github.com/user-attachments/files/20181164/LibraryRowButton.zip)

JPKribs:
On the easier note of this, what are your thoughts on the newer 'frosted glass' icons? It looks like Apple is using more of the accent color text and frosted circle as the background. That part doesn't look like that would be too difficult for us to change but it makes the accent coloring much more muted:

<details>

<summary>Us<summary>

![Image](https://github.com/user-attachments/assets/45a64494-619a-44f7-a95c-d0cfc829dab3)

</details>

<details>

<summary>Apple<summary>

![Image](https://github.com/user-attachments/assets/0b9d2d61-a2f7-4a04-b959-6cfd2b3a25cf)

</details>

<details>

<summary>Apple - Active Filter<summary>

[...]

PR #1545, opened 2025-05-21
Create an `OrnamentButton`

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1532

Breaking this out of another PR. Trying to do a better job splitting out items that can be standalone PRs. This is a simple button that tries to mirror's Apples newer VisionOS Ornaments style: https://developer.apple.com/design/human-interface-guidelines/ornaments

I think this roughly mirrors what Apple is using with the exception of the disabled version. I think apple just hides these when disabled but I don't have a real world example of this. Also, Apple doesn't have a shadow which I added since it gives it a layering/popping effect. Open to discussion

My thinking this would replace our Close Button and Menu Ellipsis. Down the road, it looks like this is where Apple puts their download button which I think looks nice.

Totally open to feedback! If we hate this, I'm also game to close it out. I just thought Apple's frosted buttons they've been using lately looked solid. I was going to use something like this in [PR #1541](https://github.com/jellyfin/Swiftfin/pull/1541) but I thought it would be better to put this in it's own PR so this got some discussion on it's own merit first.

---

### Screenshots

**Light Mode**

Screenshot 2025-05-21 at 00 23 22

**Dark Mode**

Screenshot 2025-05-21 at 00 23 49

### Videos

**Light Mode**

https://github.com/user-attachments/assets/1e15e1ae-33c4-4b83-b149-1283ce1bd429

**Dark Mode**

https://github.com/user-attachments/assets/ecf76f08-ee41-40ae-8556-09fe7c50bc54

### Apple Examples

**Music**

image

**Podcasts**

Screenshot 2025-05-20 at 22 41 57


thornbill:
The shadow doesn't seem to match with Apple's design.

JPKribs:
> The shadow doesn't seem to match with Apple's design.

Good call! I've cleaned that up in this PR and updated this in the main post.

LePips:
I do really like this attention to detail and can do a closer audit later, but I would like it implemented as a `ButtonStyle` instead of a custom button entirely.

JPKribs:
Good call. I've moved that over. This likely pushes this to after  iOS 16 since we would need the [ButtonMenuStyle](https://developer.apple.com/documentation/swiftui/buttonmenustyle) for the ellipsis menu. 

This is all in my latest commit and I think it's in a good spot but I have this:

```
                    if #available(iOS 16.0, *) {
                        Menu(L10n.options, systemImage: "ellipsis") {
                            items()
                        }
                        .menuStyle(.button)
                        .buttonStyle(.ornament)
                    } else {
                        Menu(L10n.options, systemImage: "ellipsis.circle") {
                            items()
                        }
                        .labelStyle(.iconOnly)
                        .backport
                        .fontWeight(.semibold)
                        .foregroundStyle(accentColor)
                    }
```

JPKribs:
I am going to close this out. It looks like the iOS changes at this year's WWDC mean this styling is likely a) now out of date and b) it *looks* like the glass styling of later iOS is going to be provided to us by Apple so there is no need for custom styling.

LePips:
Apologies for not commenting on this earlier. I do really like this, but I would like to find a way to unify the design of our buttons alongside the work in #1547, and porting it to existing buttons.

Also, we would still need this stuff even with iOS 26 releasing. Frankly, we will most likely opt out of the liquid glass design until we support iOS 26 as a minimum, but I have no guarantees since I haven't played with it yet.

https://www.donnywals.com/opting-your-app-out-of-the-liquid-glass-redesign-with-xcode-26/

JPKribs:
> Apologies for not commenting on this earlier. I do really like this, but I would like to find a way to unify the design of our buttons alongside the work in #1547, and porting it to existing buttons.
> 
> Also, we would still need this stuff even with iOS 26 releasing. Frankly, we will most likely opt out of the liquid glass design until we support iOS 26 as a minimum, but I have no guarantees since I haven't played with it yet.
> 
> https://www.donnywals.com/opting-your-app-out-of-the-liquid-glass-redesign-with-xcode-26/

No worries at all! I'll keep the branch alive if we want to pull from it but my thinking was this style was pretty short lived so it doesn't make a ton of sense to migrate to a new custom button if there is an Apple-native version on the horizon.

LePips:
With #1547, I would like to see these also as the `.material/tintedMaterial` button styles. However, now we will need to edit the button style to adhere to a container shape instead of assuming a rounded rectangle. There is a `TODO` documenting that.

JPKribs:
> I would like to see these also as the .material/tintedMaterial button styles.

I was thinking the same thing when I saw it. I'll take a look at this this afternoon

JPKribs:
Here is what has changed:

Migrate over to `Material` for`PillHStack`.
Migrate Close & Menu buttons in Navigation with `TintedMaterial`
Migrate filters over to `TintedMaterial` & update `TintedMaterial` to have an optional `isPressed`
Remove `isPressed` since the only usages are in `TintedMaterial`
Remove the static rounding on `TintedMaterial` to instead go in a case by case bases so I can do circles for some and rounded rectangles for others.

## Results:

### ItemViews

For `ItemViews`, the menu ellipsis and close buttons are new. Along with the Genre/Studio buttons are now using the `material` styling as well.

**Light:**

https://github.com/user-attachments/assets/4ff05d91-498c-40ab-ad02-2be6e9db710b

**Dark:**

https://github.com/user-attachments/assets/412dcdd7-89f4-457a-8cd7-d90ee54f5a90

### Active Session & Server Activity

Both of these locations had `isPressed` logic and `line.3.horizontal.decrease.circle`. I changed these to use `line.3.horizontal.decrease` and `Material` instead. Additionally, I removed the `isPressed` button style and replaced it with an optional `isPressed` option for `Material` since these were only used on buttons that are now `Material` 

https://github.com/user-attachments/assets/1e4eaa75-ffb1-4598-947b-f2ed5be1be44

JPKribs:
@LePips I'm going back over outstanding PRs to see what I can't close out and I picked this up again. What is the consensus on this item moving forward? With 26 and Liquid Glass, I understand that this is migrating buttons to a "dated" design language in more locations. I did have an idea today that we could wrap this in a build flag for `.material` on 26+ use the glass background otherwise use this styling? Or simply adding a .glass backport that uses this instead and using .glass everywhere with the backport?

I know that's a bit of a bigger ask for this PR but I'm interested to hear your thoughts! I know you have actual professional experience with Swift projects so I am curious how they are handling the Liquid Glass migrations.

LePips:
It is unfortunate that I've let this sit for so long. I did change how some of these styles are implemented in [#1752](https://github.com/jellyfin/Swiftfin/pull/1752/changes?mode=single#diff-4fa309993ba2274ef393424492afd9b5d9a8a605a53455dd55417a7ecbcf894aR60) to preserve button interactions by moving to a `LabelStyle` and preserve prevent the material changing the colors (too much) by not using `TintedMaterial` as much.

Transitioning to LG would look like implementing a version check to apply the style or not. However, since it's been so long and thinking about LG is a bit more prevalent, I'm a little less inclined to want to implement this style for previous platforms and we can avoid some of that work altogether.

JPKribs:
> It is unfortunate that I've let this sit for so long. I did change how some of these styles are implemented in [#1752](https://github.com/jellyfin/Swiftfin/pull/1752/changes?mode=single#diff-4fa309993ba2274ef393424492afd9b5d9a8a605a53455dd55417a7ecbcf894aR60) to preserve button interactions by moving to a `LabelStyle` and preserve prevent the material changing the colors (too much) by not using `TintedMaterial` as much.
> 
> Transitioning to LG would look like implementing a version check to apply the style or not. However, since it's been so long and thinking about LG is a bit more prevalent, I'm a little less inclined to want to implement this style for previous platforms and we can avoid some of that work altogether.

Sounds good! I think I started this PR like 1 week before Liquid Glass so my timing was off from the start lol. I will close out this PR in favor of your changes in #1752 and the eventual LG work!

PR #1546, opened 2025-05-21
optimize: Send media source id to playback info

lostb1t:
Jellyfin accepts the media source id. This way the return body will only include that source.

Smaller response body and massive speed gains if the sources are remote like a strm file because those files probe on playback info not when scanning

I kept the source matching on the response just in case an old server version doesn’t support it


JPKribs:
Hey, thank you for the first contribution! Always great to see new faces. Unfortunately, this will likely conflict with #1203 so this PR may not be viable to add. Primarily, this location [no longer exists in #1203](https://github.com/jellyfin/Swiftfin/pull/1203/files#diff-75fa94d6d17e1c4d2c6c7b77538b162ef5d055d83ab140552897a063f1100928) so this would need to be re-added after / during that PR *(if it hasn't been added already)*.

LePips:
I will not let this be blocked by #1203 and can take a look at this soon for 1.3.

lostb1t:
awesome!

LePips (approved):

LePips:

Issue #1016, opened 2024-04-09
Apple TV not playing audio correctly


Netlok:
### Describe the bug

When playing Blu-ray rips (h.264 video, DTS audio) surround sound only comes through as 2 channel. Once transcoded to AAC with Handbrake, 5.1 surround plays as expected, but only with "Native Player" turned on under 'Experimental Features'. If 'Native Player' is left off no audio plays at all.

VLC app on Apple TV plays surround correctly, leading me to believe it's not due to libvlc (issue #663). The closest other bug report I've found is issue #723 .

### Application version

1.0.1(70)

### Where did you install the app from?

App Store

### Device information

Apple TV 4k (Model No. A2843)

### OS version

TVOS 17.4 (21L227)

### Jellyfin server version

10.8.13

radityakurnianto:
I guess i got similar issue, when playing the videos, audios not sync with the videos

J0riz:
Issue indeed seems a lot like #723

I hope someone manage to build Swiftfin with a newer version of libvlc.

J0riz:
Good news. Audio works perfectly using VLCKit 3.6.0b11 on all channels with 5.1. 😄 

VLCKit 3.5.1 however still has the same audio problems as described here, in #663 and #723 

You need to modify the Cartfile to point to two json so it can request the newer VLCKit 3.6.0 sources.

e.g.;
```
{
  "3.6.0": "https://download.videolan.org/cocoapods/unstable/MobileVLCKit-3.6.0b11-fbffb2f1-6d82c5af.tar.xz",
}

{
  "3.6.0": "https://download.videolan.org/cocoapods/unstable/TVVLCKit-3.6.0b11-fbffb2f1-6d82c5af.tar.xz"",
}
```

The audio channel issue was fixed somewhere in VLCKit 3.6.0 which is currently also used in the official builds of VLC for tvOS.
Which explains no one has big audio channel issues anymore using the VLC app.

Building this with Swifin from Main branch however still has a lot of (known) interface issues including the video overlay not showing in full screen.

So to really get this issue resolved Swiftfin needs to be build on VLCKit 3.6.0 and also some interface issues nee to be resolved within Swiftin.

radityakurnianto:
Wow thats cool, swftfin team need to rebuild the project then resubmit to
store 👍


*Raditya Kurnianto**iOS Developer*


*linked.in <https://id.linkedin.com/in/radityakurnianto>*

[...]

Issue #723, opened 2023-03-12
Multi Channel Audio through Standard Video Player does not work


turtle12:
### Describe the bug

So far I could not get surround sound to work at all with the standard video player. I tried various 5.1 audio files with different codecs like Dolby Digital, AAC and so on and surround sound did not work for any of them. My sounhd system actually reports a multi ch pcm input passed through from my TV but all the sound is coming from the front left and right channels it seems.

When I switch to the experimental native player I get really good surround sound support that also plays correctly on my sound system, even dolby atmos via DD+.

### Application version

1.0.1

### Where did you install the app from?

App Store

### Device information

Apple TV 4K 2022 (A2843)

### OS version

tvOS 16.3.3

### Jellyfin server version

10.8.9

J0riz:
This issue seems to be the same as: https://github.com/jellyfin/Swiftfin/issues/663

The same issue with audio channels also exists in VLC app on Apple TV. So the cause seems to be an issue in libVLC.

With this audio file the left channel plays at left front speaker, left surround also plays on left front speaker. all the other channels go through center speaker:
- https://www2.iis.fraunhofer.de/AAC/ChID-BLITS-EBU-Narration.mp4

I can't yet find a related issue at libVLC bug tracker. However comments in this issue seems to have some similarities:
- https://code.videolan.org/videolan/VLCKit/-/issues/79

With the Experimental Native player in Jellyfin plays all 5.1 audio channels  at my amp work correctly. However than my screen stays black which makes the Experimental Native Player unusable.

I hope someone can find the exact cause within Jellyfin and/or libVLC.

mattyeark:
It appears this isn't a libVLC issue, or at least an issue within libVLC.
I used [these audio test files](https://drive.google.com/drive/folders/1JxmeedtAtgmoafXv9rroiDOS2vEX7N4b), and proceeded to test them with the VLC app on the Apple TV. They played fine (I have a 5.1 sound system), the side channels played on the side speakers, which is the correct behavior. On Swiftfin, they play on the front left and right speakers.
So, somehow Swiftfin is misusing libVLC? To be noted: even though the side channels get "squashed" into the front channels, the audio signal that the Apple TV sends to my amp is a 5.1 one... the side channels' audio is just nonexistent.

J0riz:
[...]

PR #1552, opened 2025-05-30
Reverting to VLCKit 3.5.0 | solving OPUS issue

nds2204:
Had issue with OPUS codec, due to regression in VLCKit 3.6.0.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,3 +1,3 @@
-binary "https://code.videolan.org/videolan/VLCKit/raw/master/Packaging/MobileVLCKit.json" == 3.6.0
-binary "https://code.videolan.org/videolan/VLCKit/raw/master/Packaging/TVVLCKit.json" == 3.6.0
+binary "https://code.videolan.org/videolan/VLCKit/raw/master/Packaging/MobileVLCKit.json" == 3.5.0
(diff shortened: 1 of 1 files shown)


JPKribs:
Hi, do you know if the AV1/OPUS regressions in [3.6.1b1](https://code.videolan.org/videolan/VLCKit/-/commit/8e652244ecf03b5b102c899671e4ce8e23d996ce) resolve this as well or only a full rollback to 3.5.0?

LePips (approved):
The issue looks to be the lidvlc version used for `3.6.0`:
https://www.reddit.com/r/VLC/comments/1dexl6y/some_opus_audio_tracks_to_videos_not_playing_in/

I'm fine to revert to `3.5.0` as I don't see much in the diffs that we directly benefitted from when we updated a while ago.

J0riz:
I understand it got reverted.

libVLC 3.6.0 however does fix some audio issues:
https://github.com/jellyfin/Swiftfin/issues/1016
https://github.com/jellyfin/Swiftfin/issues/723

LePips:
Ah, I thought I remembered audio reasons to keep 3.6 but GitHub search failed me prior to me doing this work. I'll have to think about how to go moving forward.

akagumo:
I am curious if it is possible to include both 3.5 and 3.6 and allow users to select? I need dts decoding from original BR rips and don't use opus.

PR #1556, opened 2025-06-01
Cinematic Episode Image

JPKribs:
### Summary

After [1543](https://github.com/jellyfin/Swiftfin/pull/1543), the Cinematic view on iPadOS was using the Episode Backdrop. It looks like most metadata providers for episodes don't provide backdrops so this always fails and falls back to `EmptyView()`. Similar to how we handle collection and [their more limited metadata](https://github.com/jellyfin/Swiftfin/pull/1543#issuecomment-2895493278). Just in case someone has a manual backdrop added, I have this PR do the following:

1. If the item is both an episode and does not contain a backdrop image, use the series backdrop image.
2. If the item is either a non-episode or *does* contain a backdrop image, use the item's backdrop image.

I've update this same logic on tvOS so this is consistent between the environments.

### DIFF
--- a/Swiftfin/Views/ItemView/ScrollViews/iPadOSCinematicScrollView.swift
+++ b/Swiftfin/Views/ItemView/ScrollViews/iPadOSCinematicScrollView.swift
@@ -48,8 +48,14 @@ extension ItemView {
             OffsetScrollView(
                 headerHeight: globalSize.isLandscape ? 0.75 : 0.6
             ) {
-                ImageView(viewModel.item.imageSource(.backdrop, maxWidth: 1920))
-                    .aspectRatio(1.77, contentMode: .fill)
+                // Use the Image Backdrop if provided. Otherwise, use the Series Backdrop.
+                if viewModel.item.type == .episode && viewModel.item.backdropImageTags == [] {
+                    ImageView(viewModel.item.seriesImageSource(.backdrop, maxWidth: 1920))
+                        .aspectRatio(1.77, contentMode: .fill)
+                } else {
+                    ImageView(viewModel.item.imageSource(.backdrop, maxWidth: 1920))
+                        .aspectRatio(1.77, contentMode: .fill)
+                }
             } overlay: {
                 VStack(spacing: 0) {
                     Spacer()
--- a/Swiftfin/Views/ItemView/ScrollViews/iPadOSCinematicScrollView.swift
+++ b/Swiftfin/Views/ItemView/ScrollViews/iPadOSCinematicScrollView.swift
@@ -48,14 +48,8 @@ extension ItemView {
             OffsetScrollView(
                 headerHeight: globalSize.isLandscape ? 0.75 : 0.6
             ) {
-                // Use the Image Backdrop if provided. Otherwise, use the Series Backdrop.
-                if viewModel.item.type == .episode && viewModel.item.backdropImageTags == [] {
-                    ImageView(viewModel.item.seriesImageSource(.backdrop, maxWidth: 1920))
-                        .aspectRatio(1.77, contentMode: .fill)
-                } else {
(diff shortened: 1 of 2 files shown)


sjorge:
Would be nice to use the episode thumbnail as a fallback before the series backdrop.

Most episode for me have thumbnails.

JPKribs:
I just did a quick search on mine. I use TVDB for my series, which I think is pretty standard for most people? I have no episodes that contain thumbnails. Maybe that that's a configuration change on my end? But, even going through some of the newer series that I have I don't have thumbnails as even optional metadata I can add without uploading my own.

I'm not opposed to it but my goal for this PR is to hopefully clean this up before the final 1.3 release. Adding other various image options would likely require settings and other configuration which would make this harder to be ready for the final release of 1.3.

sjorge:
AFAIK tmdb is the standard one, tvdb needs to be installed.

image

So it grabs it from tmdb, or from the file if it's embedded or it grabs a screenshot. Most seem to come from tmdb as far as I can tell. New shows even have a few options.

sjorge:
Hmm interesting they show up as Primary type in jellyfin, but that seems to refer to the thumbnails on tmdb itself, e.g. for the last if us it grabs these for s2 for me https://www.themoviedb.org/tv/100088-the-last-of-us/season/2

JPKribs:
The primary image is currently what is used for tvOS. Previously, it was what was used for iPadOS. In my experience, especially for older shows, these images tend to be lower resolution so scaling them as a cinematic background (1920x1080), IMO, does not look good. I'm sure newer shows look fine but I'm personally of the opinion that I would rather just see the show backdrop than a pixelated episode card.

As I mentioned, I'm up for whatever people wanna do but primarily this PR is to just prevent empty view for episodes. I'll defer to @LePips for any design goals as they have a more cohesive vision than I do haha. But, I want to reiterate that my goal is to try and keep this PR as small as concise as possible to prevent requiring other testing.

---

Thinking more, it might be cool to have a generic, Swiftfin background image we could use if nothing populated? Similar to how we have filler icon + color for empty posters. But I think that's for another PR.

sjorge:
> Thinking more, it might be cool to have a generic, Swiftfin background image we could use if nothing populated? Similar to how we have filler icon + color for empty posters. But I think that's for another PR.

That could also be a nice option! We could probably ask Erik in the ui-ux channel on discord if we want to go down that route?

Swiftfin/Views/ItemView/ScrollViews/iPadOSCinematicScrollView.swift
```diff
@@ -48,8 +48,14 @@ extension ItemView {
             OffsetScrollView(
                 headerHeight: globalSize.isLandscape ? 0.75 : 0.6
             ) {
-                ImageView(viewModel.item.imageSource(.backdrop, maxWidth: 1920))
```
LePips, line 51: Just use the `imageType` variable above instead of `.backdrop`.

LePips (changes requested):
The root of this issue is just that I defined the image type to be used but just forgot to use it during my refactor (I use the same pattern in `SimpleScrollView`).

Per the same work on tvOS, I just didn't touch tvOS in #1543 because I was focused on iOS and these views deserve similar cleanup, so I would request that work is moved to later.

Instead of checking explicitly if there are backdrop images, we can define the list of image sources that are used instead based on the type. `ImageView` will iterate until it successfully gets an image. This is where we would define also using the thumbnail image.

Per using the series image, I don't find it desirable for these item views - I would like that they only present their own metadata.

Swiftfin/Views/ItemView/ScrollViews/iPadOSCinematicScrollView.swift
JPKribs: Didn't notice that. This is updated.

LePips (approved):

Issue #1557, opened 2025-06-05
Cast and Genre


tobylibo:
### Describe the bug

In English:
Selecting an cast incorrectly displays the Genre. Previous versions have correctly displayed the movies/tvshows in which the actors have starred.

In Chinese:
选择后会错误地显示影片类型。之前的版本是能正确地显示该演员出演的电影/剧集的。

### Application version

1.3 (12)

### Where did you install the app from?

TestFlight

### Device information

iPhone 12 Pro, iPad Pro

### OS version

iOS 18.5

### Jellyfin server version

10.9.10

PR #1558, opened 2025-06-05
Cast & Crew Fix

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1557

Prior to #1495, we defaulted to Recursive for everything. After #1495, we default to Recursive false unless we can validate it's a `BaseItemDto` and not a `tvShow` or `collection`. The reason this causes issues is that selecting a cast member is not a `BaseItemDto` but a `BaseItemPerson` so this casting fails and, instead of defaulting to true it would default to false.

### Issue (Now Resolved in this PR)

https://github.com/user-attachments/assets/01f10f53-4cd1-47f1-aced-53f25a9de27a


LePips:
Thanks for the quick fix!

PR #1566, opened 2025-06-12
Remove 

JPKribs:
### Summary

_Goodbye iOS 15!_

This PR removes all pre-iOS 16 and tvOS 17 backports and conditional `#available`s. I've gone through and tested the views these were located on and everything appears normal. There is only one backport remaining for `scrollClipDisabled` which is iOS 17+.

We have two items that have been deprecated in iOS 16. These should both be resolved by https://github.com/jellyfin/Swiftfin/pull/1490.

Screenshot 2025-06-12 at 09 53 25

### Outstanding Items

We have 2 files that need a second look after this PR. 

1. `GaugeProgressStyle`. We have a note that a `gauge` `ProgressViewStyle` should exist in iOS 16 but I wasn't seeing it. I left this for now with a TODO to look at this later in case I missed something
2. `SelectUserView` has a note "Could use provided UI, but is iOS 16+". I'm not familiar with that so I didn't take a crack at this. For now, I've left the TODO and this can be addressed in another PR


Swiftfin/Components/CircularProgressView.swift
```diff
@@ -9,6 +9,7 @@
 import Defaults
 import SwiftUI
 
+// TODO: Can we remove this in iOS 16?
 // SwiftUI gauge style not available on iOS 15
```
LePips, line 13: I made this comment confusing. I would like to possible replace the underlying implementation of this type with a `Gauge` with the `.gaugeStyle(.accessoryCircularCapacity)` applied. Just change these comments for a `TODO` with that.

LePips (changes requested):
So beautiful, thank you!

LePips:
> "Could use provided UI, but is iOS 16+"

I think I was referring to [LocalAuthenticationView](https://developer.apple.com/documentation/localauthentication/localauthenticationview), but that's macOS only. So, IDK what I was thinking with that comment 🤷

Swiftfin/Components/CircularProgressView.swift
JPKribs: Oh! That makes sense! We only use that circular progress view in two places so moving to a gauge is really simple. This is what I have now (see below). I'd be in favor of making this swap in this PR if you don't feel like this is overkill for this PR:

```
                        Gauge(value: (observer.task.currentProgressPercentage ?? 0) / 100, in: 0 ... 1) {
                            EmptyView()
                        }
                        .gaugeStyle(.accessoryCircularCapacity)
                        .tint(Color.accentColor)
                        .scaleEffect(0.5)
                        .transition(.opacity.combined(with: .scale).animation(.bouncy))
                        .frame(width: 25, height: 25)
```

Screenshot 2025-06-12 at 10 47 43

JPKribs:
@LePips I also found one more. I missed it because it was #unavailable I was looking for #available. Is there a way to do this and just say "All iOS versions"? I got this working with the iOS16, iOS17... but that's just a chore.

```
    @ViewBuilder
    func listRowCornerRadius(_ radius: CGFloat) -> some View {
        if #unavailable(iOS 16) {
            introspect(.listCell, on: .iOS(.v15)) { cell in
                cell.layer.cornerRadius = radius
            }
        } else {
            introspect(
                .listCell,
                on: .iOS(.v16),
                .iOS(.v17),
                .iOS(.v18)
            ) { cell in
                cell.layer.cornerRadius = radius
            }
        }
    }
```

JPKribs:
I've added the note changes. Removed the [LocalAuthenticationView](https://developer.apple.com/documentation/localauthentication/localauthenticationview) reference.

I have 2 changed in here that need discussion:

1. Do we want to include the gauge changes in this PR or move this to another?
2. On `listRowCornerRadius` [is there a better way to just say "Do all OSes!"](https://github.com/jellyfin/Swiftfin/pull/1566#issuecomment-2967568998). I'm not familiar with `introspect(.listCell...` so this could be totally unnecessary

---

New Gauge:

Screenshot 2025-06-12 at 10 47 43

---

For reference, to find all items I did the following searches. Please feel free to let me know if this is incomplete:

- iOS15
- iOS 15
- .v15
- iOS16
- iOS 16
- .v16
- backport
- #available
- #unavailable

LePips:
> Do we want to include the gauge changes in this PR or move this to another?

Let's do that in another PR. But, we may also not do it if we prefer our design which allows for more customization, especially for thicker lines in those smaller row-contexts.

> "Do all the OSes!"

That's how the initial introspect used to work but it transitioned to explicit version adoption for good reason.

From [their README](https://github.com/siteline/swiftui-introspect#how-it-works):

> Although this introspection method is very solid and unlikely to break in itself, future OS releases require explicit opt-in for introspection (.iOS(.vXYZ)), given potential differences in underlying UIKit/AppKit view types between major OS versions.

Swiftfin/Views/AdminDashboardView/ServerTasks/ServerTasksView/Components/ServerTaskRow.swift
```diff
@@ -82,10 +81,14 @@ extension ServerTasksView {
                     Spacer()
 
                     if observer.state == .running {
-                        ProgressView(value: (observer.task.currentProgressPercentage ?? 0) / 100)
-                            .progressViewStyle(.gauge)
-                            .transition(.opacity.combined(with: .scale).animation(.bouncy))
-                            .frame(width: 25, height: 25)
+                        Gauge(value: (observer.task.currentProgressPercentage ?? 0) / 100, in: 0 ... 1) {
```
LePips: The point of the comment wasn't to replace `ProgressView`, but to replace the underlying implementation of `GaugeProgressStyle`. We would still use `ProgressView` with `.progressViewStyle(.gauge)` applied.

But as said elsewhere, do not do this work in this PR.

JPKribs: Sounds good! I'll put everything back for this.

JPKribs:
> That's how the initial introspect used to work but it transitioned to explicit version adoption for good reason.

Gotcha! Sounds good. Should we leave a note or something to make sure to update those as iOS 26 comes up or turn this into a var or setting somewhere so we can more easily update these in bulk? Or leave this for now?

---

Gauge changes reverted. Let me know about the above. Otherwise, I think this should be ready!

LePips (approved):
Looks great, thank you!

PR #1567, opened 2025-06-12
Admin Dashboard Cleanup

JPKribs:
### Summary

Now 1.3 is out, I am getting more feedback about what needs adjusting in the admin dashboard. This PR cleans up some irregularities in the admin dashboard. This resolves the following bugs:

1. When I first started, I was making a custom "Delete X" string for each confirmation. Later on, we just use a generic "Delete". I've updated the old usages to use the "Delete" to make this more cohesive and remove unnecessary strings. I've left the User strings in place since those are higher stakes. I also left the custom device deletion messages since it's a rare place where 'delete' doesn't really mean delete. For this reason, leaving the description helps provide clarity where other deletions are more straight forward.
2. From Jellyfin, Server Task naming is all over the place. I've added a `localizedCapitalized` to these server tasks so they at least show up as cohesive in Swiftfin. I know that's more of a Server issue but, IMO, it's a 1/2 a line change and it makes it look more intentional from our end.
3. Some Admin Dashboard titles were capitalized and some were not. I think there is a bigger title vs sentence case change that needs to be done but for now this resolves this descrepency. All titles are capitalized appropriately and I have removed unnecessary "duplicate" vs "Duplicate" strings that should make this easier to localize.
4. This is a much smaller change but the Active Session shows as "No Results." when there are no active session. I had a couple folks tell me the API failed thinking it was saying "No results returned". I've just updated this to the "None" we use in most other places since that's more descriptive of reality and mirrors other usages.


JPKribs:
The last item that might need some change. All other views allow bulk deletion except for the `APIKeyView`. IMO, who on earth is in a rush to delete a bunch of API keys(?) but for consistency, it might make sense to give this the same bulk deletion treatment as other Admin Dashboard views?

Open to feedback on that. It's not hard to do but the time/reward feels really really insignificant IMO so I'm feeling less inclined to change it.

I think outside of potentially cleaning up `APIKeyView`, I think that's the last post-1.3 cleanup I can think of?

Swiftfin/Views/AdminDashboardView/ServerTasks/EditServerTaskView/EditServerTaskView.swift
```diff
@@ -35,7 +35,7 @@ struct EditServerTaskView: View {
     var body: some View {
         List {
             ListTitleSection(
-                observer.task.name ?? L10n.unknown,
+                observer.task.name?.localizedCapitalized ?? L10n.unknown,
```
LePips: We should generally not capitalize strings that come from the server.

Swiftfin/Views/AdminDashboardView/ServerTasks/ServerTasksView/Components/ServerTaskRow.swift
```diff
@@ -32,7 +32,7 @@ extension ServerTasksView {
         private var taskView: some View {
             VStack(alignment: .leading, spacing: 4) {
 
-                Text(observer.task.name ?? L10n.unknown)
+                Text(observer.task.name?.localizedCapitalized ?? L10n.unknown)
```
LePips: Same here.

Swiftfin/Views/AdminDashboardView/ServerTasks/ServerTasksView/Components/ServerTaskRow.swift
```diff
@@ -95,7 +95,7 @@ extension ServerTasksView {
             .animation(.linear(duration: 0.1), value: observer.state)
             .foregroundStyle(.primary, .secondary)
             .confirmationDialog(
-                observer.task.name ?? .emptyDash,
+                observer.task.name?.localizedCapitalized ?? .emptyDash,
```
LePips: Same here. Also change fallback to `L10n.unknown`.

Swiftfin/Views/AdminDashboardView/ServerTasks/EditServerTaskView/EditServerTaskView.swift
JPKribs: That's fair. I can revert that. This was the reason I was doing this but I agree this is more of a server / plugin issue opposed to our issue. Just wish we were more consistent on the Server side with these:

simulator_screenshot_27FBFB43-BE6B-43F7-80B9-9DCEDC6355C8

JPKribs:
I've reverted the usage of `tasks.x.localizedCapitalized` in my last commit!

PR #1569, opened 2025-06-13
Replace `WrappingHStack` for `FlowLayout`

JPKribs:
### Summary

Original discussion and work: https://github.com/jellyfin/Swiftfin/pull/1467

This PR replaces `WrappingHStack` with a Layout for iOS 16. This layout should roughly mirror `WrappingHStack` but with some additional flexibility.

### Screenshots

<details>
<summary>Cinematic - Normal & Large Text</summary>

Cinematic - Normal Text

Cinematic - Large Text

</details>

<details>
<summary>Logo - Normal & Large Text</summary>

Logo - Normal Text

Logo - Large Text

</details>

<details>
<summary>Compact - Normal & Large Text</summary>

Compact - Normal Text

Compact - Large Text

</details>


JPKribs:
Coming from here: https://github.com/jellyfin/Swiftfin/pull/1467#discussion_r2144084084

This implementation is less overall code and should be easier to follow. I (with Claude) overly documented this Layout. I don't personally think this level of documentation is valuable for the final version of this but I thought I would leave it in to start off the review and make sure that we're all on the same page for what each section does.

This is the same as last time but I added a value for choosing how it populated. I have 2 options in there:

**Bottom Up:**

Screenshot 2025-06-12 at 21 31 34

**Top Down:**

Screenshot 2025-06-12 at 21 33 18

Regardless of the direction, the order of the original array is still preserved and ordered top to bottom, left to right.

I wanted to use [VerticalDirection](https://developer.apple.com/documentation/swiftui/verticaldirection) but that looks like iOS 18+. I've left a note to use that and clear out the extra enum IF we think it's valuable to have the ability to switch directions.

LePips (approved):
I've supplemented the layout to be more generic to any number of rows rather than a maximum of 2. We can take a look adding the functionality for setting max rows but we would also have to consider how it looks as elements grow wider and eventually clip out of the screen/its own size - which is what we want to avoid.

Some of the nuances for switching to a `Layout` over `WrappingHStack` is that we no longer need to build the array of attribute views and instead use a `ViewBuilder`. I also do some cleanup with how the attribute views are constructed.

LePips (approved):

PR #1574, opened 2025-06-14
10.10 - Create Version Warning

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1091

Bumps SDK version from 0.5.1 to 0.5.2.

Checks the Jellyfin Server version against the SDK version. Generates a persistent warning in the settings if the Server Version is below the SDK Version.

### Locations

SettingsView
EditServerView

### DIFF
--- a/Shared/SwiftfinStore/V2Schema/V2ServerModel.swift
+++ b/Shared/SwiftfinStore/V2Schema/V2ServerModel.swift
@@ -27,7 +27,6 @@ extension SwiftfinStore.V2 {
         @Field.Stored("id")
         var id: String = ""
 
-
         @Field.Relationship("users", inverse: \StoredUser.$server)
         var users: Set<StoredUser>
 


JPKribs:
@LePips No rush on this I'm going to be stepping away for the afternoon. I just wanted to explain my approach and make sure you agree before I start looking at CoreStore stuff.

**Lines are currently commented to get this to build**

https://github.com/jellyfin/Swiftfin/blob/1a99b35615bbb0f1a47617d77de499927861a1f2/Shared/SwiftfinStore/SwiftfinStore%2BServerState.swift

I am going to add a `let version` to the `ServerState` and add this to the `server` object in the `CoreStore`. This will get updated during `updateServerInfo()`. This way, we always just have the server version. The SDK version is static as well so the comparison would just be:

```
func isVersionCompatible() -> Bool {
         if let majorMinor = self.version?.majorMinor {
             return majorMinor >= JellyfinClient.sdkVersion.majorMinor
         } else {
             return false
         }
}
```

Since this is in the `userSession` and on the `serverState`, this would allow us to check this anywhere we need to without external calls. I am thinking this goes:

1. Server Settings (Done -  See Screenshots)
2. Edit Server View (Done -  See Screenshots)
3. Connect To Server View
4. Select User View
5. Maybe the Server Check View if the rest of these aren't too overkill

The downside is I think we'd need a V3 and a migration to do this? I'm not familiar enough with CoreStore to say for certain but we'd at least need a migration. The current, uncommented/active version of this on this PR calls the `publicSystemInfo` API each time. Not efficient at all but let's us put of a migration for now. I'm mostly doing that to test but it's an option if we want to avoid a CoreStore V3 until 1.4+ since I think this check might be helpful in a 1.3.1 to reduce new potential issues?

Let me know what you think!

LePips:
I think this is a good check, but we ought to only present a warning to administrators which means this would only be presented on the settings and server detail views. I would also only like it to present the warning state, not a successful check.

I will have to think about the migrations, because that is a bit heavy for this. We can take a look at just having it as a `StoredValue` for now.

JPKribs:
> I will have to think about the migrations, because that is a bit heavy for this. We can take a look at just having it as a `StoredValue` for now.

Gah! I didn't even realize we already store `PublicInfo` there. I can just use that instead. I've updated this to use that.

I currently have this check in the settings (only showing for when incompatible). I've added a more complete description of the issue on the `ServerDetailView`. So the user will see the ! icon:

Screenshot 2025-06-14 at 17 04 34

And if they click through to that view, they will see this:

Screenshot 2025-06-14 at 17 05 38

For clarity, I added the `ServerVersion` to this view. That way they can see their current version and, at the bottom, they can see the error and the version number required to make the error go away.

Is there anywhere else we want this to show up? I think this implementation is subtle enough that people wont feel like they're being blasted by the warning but it's easy to see if you are troubleshooting something. That is, going into settings you are gently alerted.

---

tvOS below:

Screenshot 2025-06-14 at 17 22 40

Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-06-14 at 17 22 49

LePips (approved):

PR #1576, opened 2025-06-15
Remove `TextPair/TextPairView` for `LabeledContent`

LePips:
Removes my `TextPair/TextPairView` to now use `LabeledContent`.


JPKribs (changes requested):
Looks good! I think you need to merge Main back in to get the build to work. I added [a TextPairView](https://github.com/JPKribs/Swiftfin/blob/865f75536610192dcc440c5f41cfda725ee5abf8/Swiftfin/Views/EditServerView.swift#L47-L52) in https://github.com/jellyfin/Swiftfin/pull/1574 which is giving us the build error.

JPKribs (approved):

PR #1582, opened 2025-06-16
Prioritize hevc over h264 for transcoding with native player

Cosmic-Skye:
Right now, native player will transcode to h264 over hevc if the server has both transcoding options enabled. For the native player, HEVC is objectively a better option due to size efficiency and compatibility with DoVi/HDR, among other things.

This PR prioritizes HEVC in the device profile when transcoding is required.

### DIFF
--- a/Shared/Objects/VideoPlayerType/VideoPlayerType+Native.swift
+++ b/Shared/Objects/VideoPlayerType/VideoPlayerType+Native.swift
@@ -110,8 +110,8 @@ extension VideoPlayerType {
             AudioCodec.eac3
             AudioCodec.flac
         } videoCodecs: {
-            VideoCodec.h264
             VideoCodec.hevc
+            VideoCodec.h264
             VideoCodec.mpeg4
         } containers: {
             MediaContainer.mp4


JPKribs:
If I knew the order mattered I wouldn't have just alphabetized these....

This same change should be made to Swfitfin player as well. With the exception of likely leaving AV1 over HEVC. Which, I think opens the wider question of "What should the full order be?". 

AVKit seems easy enough with the 3:

1. HEVC
2. H.264
3. MPEG4

The only note there is that we should add AV1 to the top of this list [once we target iOS 17+](https://github.com/jellyfin/Swiftfin/discussions/1564). 

For Swiftfin, this question is a bit harder. I would argue, at minimum:

1. AV1
2. HEVC
3. H.264
4. etc.

But I don't know if anyone has any strong feels about the etc. there. Likely, there is some preference towards VP9 but I'm really not familiar with anything outside the big 3.

holow29:
I don't think order matters much other than the ones you listed because once the server gets to its first allowed one (i.e. one that the server hasn't shifted out due to e.g. AV1 or HEVC being disabled serverside), it will just use that. In other words, the order only matters to find the first acceptable codec for transcoding. The other codecs are then in the array to ensure that they are passed through to the client. (In reality, any server's backstop is almost always going to be H264, so anything listed after that only matters for stream copy indicative purposes I think.)

I agree with your ordering JPKribs.

JPKribs:
~Now I'm thinking about why do we even have MPEG4 in there? I feel like this might be a valid PR to cleanup the transcode profiles. Make Native and Swiftfin the save codecs with:~

1. ~AV1 - _For now commented out on Native_~
2. ~HEVC~
3. ~H.264~

~Then, if folks need other transcode profiles we have the custom profile system in place for them to enable what is needed. I'll defer to @LePips but that is my opinion on what would be the cleanest route~

Ignore this. Read below vvv

holow29:
My understanding (when I first tried to redo the device profiles): The transcode profiles need to have the DirectPlay-supported codecs like mpeg4 otherwise they will get transcoded. For example, a file with mpeg4 video and dts audio will hit the transcoding profile on native player because dts is not supported for DirectPlay. The server will look and see if mpeg4 is in the transcoding profile’s videoCodec array and if not, it will transcode the video as well. I am not sure that the profiles need much or any cleanup other than a slight reordering here or there since I don’t think anything has changed much since I looked at them last other than maybe AV1.

LePips (approved):

PR #1583, opened 2025-06-16
Subtitle Management

JPKribs:
### Summary

Formerly: #1513
Resolves: https://github.com/jellyfin/Swiftfin/issues/178

This PR enables managing subtitles. This completes the long journey for Admin Functionality I started [here](https://github.com/jellyfin/Swiftfin/discussions/1283). Other items will definitely come up that need to get added but I feel comfortable closing out that thread with this change as feature items will be more "maintenance". So, less need for a huge, multi-part thread.

This PR enables the following functionality:

1. View all subtitles for a file. Separated into embedded and external.
2. Select a subtitle for subtitle details
3. Enable deletion of **multiple\*** external subtitles
4. Allow uploading a new **text-based\*\*** subtitle
5. Allow searching Subtitle Providers and selecting **multiple\*** subtitles for download

\*Unique to Swiftin. Jellyfin-Web (and the API) only allow single item changes.
\*\*Image based subtitles cannot be uploaded via the API. Mirrors Jellyfin-Web.

---

### Quirks

I am using the `MediaStream` to get the subtitles for an item. For this reason, I need to reload the item when a subtitle is added or delete to re-get all of the media streams. The reason I cannot just append/remove subtitles is the index can change pretty dramatically based on actions. So reloading the item ensures that I always have the correct index which is especially important when deleting.

The delete function on server is for a single index. So, when deleting multiple I need to:

1. Delete Index N
2. Re-get Item
3. Delete Index N + 1
4. Repeat

It's not super efficient but this view likely isn't getting a ton of traffic (I doubt this is a feature even I will use haha) so I think the performance hit is fine to better ensure that we are always getting the appropriate index.

I've been testing this for a while and I don't know if there is a better way to do this but I am open to suggestions!

---

iOS allows for deleting subtitles, uploading text subtitles, or searching for subtitles.

tvOS allows for searching for subtitles.

---

### Video

<details> 

<summary>tvOS</summary> 

**Uploading**

https://github.com/user-attachments/assets/38551099-832f-4aec-b78f-b9a105227e94

</details>

<details> 

<summary>iOS</summary> 

**Management**

https://github.com/user-attachments/assets/a1a12cf9-68ff-4532-93d1-b0b97c6aa6a1

**Uploading**

https://github.com/user-attachments/assets/74697fd8-8fe2-45ce-a7d8-ddc9d2f48c7a

</details>


Swiftfin/Views/ItemEditorView/ItemSubtitles/ItemSubtitlesView.swift
```diff
@@ -0,0 +1,214 @@
...
+                    }
+                    .buttonStyle(.toolbarPill(.red))
+                    .disabled(selectedSubtitles.isEmpty)
+                    .frame(maxWidth: .infinity, alignment: .trailing)
+                }
+            }
+        }
+        .navigationBarMenuButton(
+            isLoading: viewModel.backgroundStates.contains(.updating),
+            isHidden: isEditing || !hasSubtitles
```
LePips, line 123: We are preventing adding subtitles if there aren't any subtitles. We should match other instances where the list shows buttons to add content.

LePips (changes requested):

Swiftfin/Views/ItemEditorView/ItemSubtitles/ItemSubtitlesView.swift
JPKribs: I have two versions of this:

The last commit I have it shows the menu up top when there are no external subtitles if there are internal subtitles. If there are no subtitles, neither internal nor external, show the regular add button:

Screenshot 2025-07-18 at 21 46 09

The only option, since add/edit is only for external, show the add button if there are no external subtitles even if there are internal subtitles:

simulator_screenshot_96B3BF23-B247-4514-B561-B3475387D35D

Let me know if you want me to swap this out for the second option. I went with the first since it seems the most consistent with the other options we've done this for.

JPKribs: Also, I made this a menu instead of a button since there is a search and an upload option. Let me know if we want to adjust that.

Issue #1577, opened 2025-06-15
minor UI changes


iPleyGit:
Since you're more or less following the Apple TV+ design, I’m suggesting several UI adjustments to improve visibility.

On Apple TV+, when you're on a show's or movie's screen, most of the space is focused on the content being viewed — meaning the background image takes up almost the entire screen. In our design, a large portion of the screen is taken up by the list of episodes.
In my opinion, it would be ideal to enlarge the background image of the show or movie a bit more, so it resembles the current Apple TV+ layout.

<details>
  <summary>Apple TV+</summary>

 ![Image](https://github.com/user-attachments/assets/347eacaf-f75c-4eb3-b6fe-0e8f5ef944b5)

</details>

JPKribs:
For Item 1:

This exists as is called "Cinematic":

<details>

<summary>Example</summary>

![image](https://github.com/user-attachments/assets/960254e4-6b6e-4e28-b0a5-b2789cc6cbda)

</details>

You can access this via:

Settings > Customize > Items (top option)

![image](https://github.com/user-attachments/assets/e3990302-285a-416a-b631-3e050d1b1a21)

ClappTV:
> Para el artículo 1:
> 
> Esto existe como se llama "Cinemática":
> 
> Ejemplo
> Puede acceder a esto a través de:
> 
> Configuración > Personalizar > elementos (opción superior)
> 
> ![image](https://github.com/user-attachments/assets/e3990302-285a-416a-b631-3e050d1b1a21)

Thanks, that's exactly the mode I was asking about. Is there any reason not to keep it as the default?

LePips:
We can only use images provided to use by Jellyfin, which are provided from metadata providers. The images we get are basic images representing the backdrop, portrait, thumb, logo, etc.

The default view is what works well most of the time - just using the backdrop and logo with the title as the fallback. We can largely depend on a backdrop being provided and logos are relatively new as being provided by the server. If you have images that composition well with the cinematic view, that's great, and you can enable it. The only reason why the cinematic view exists is because the implementation was fairly easy.

Apple, on the other hand, curates and makes images specifically for their apps and should not be compared against in ways like this. You can see what they look like by looking up a show on: https://bendodson.com/projects/apple-tv-movies-artwork-finder/.

>  it hides the season selector and simply displays the text

I believe I've personally made the decision to just ignore this feature, as I'm fine with the selector being present even if there's just a single season. However, I could welcome the change.

[...]

PR #1585, opened 2025-06-16
Make Text Static on Single Season

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1577

Make the `SeasonSelector` static Text for single Season shows.

**Notes:**
- I mirrored the regular `SeasonSelector` spacing. I'm not sure if that's a good fit so it might be worth removing the horizontal spacing.
- I made this text also accent color. It looks good as `.primary` as well so we have alternatives. 

### Screenshot

<details>

<summary>Single vs Multiple Season(s)</summary>

Screenshot 2025-06-16 at 08 36 50

</details>

### DIFF
--- a/Swiftfin/Views/ItemView/Components/EpisodeSelector/EpisodeSelector.swift
+++ b/Swiftfin/Views/ItemView/Components/EpisodeSelector/EpisodeSelector.swift
@@ -32,10 +32,8 @@ struct SeriesEpisodeSelector: View {
            viewModel.seasons.count <= 1
         {
             Text(seasonDisplayName)
-                .font(.headline)
-                .foregroundStyle(Color.accentColor)
-                .padding(.vertical, 5)
-                .padding(.horizontal, 10)
+                .font(.title2)
+                .fontWeight(.semibold)
         } else {
             Menu {
                 ForEach(viewModel.seasons, id: \.season.id) { seasonViewModel in


LePips (changes requested):
I would like the text to be like Apple TV, where the static text would be larger and would be the text primary color. The accent color should largely be for "this can be selected".

Please also remove the padding.

JPKribs:
> I would like the text to be like Apple TV, where the static text would be larger and would be the text primary color. The accent color should largely be for "this can be selected".
> 
> Please also remove the padding.

Here is my revised version. I used the same font & weight as the `PillHStack` headers to match that:

<details>

<summary>New Button</summary>

Simulator Screenshot - iPhone 16 Pro - 2025-06-16 at 23 10 43

</details>

sjorge:
I did a build an was having a look as I have a lot of 1 season shows (anime) but it still showed the drop down.

After looking at the code it seems this only removes the drop downs when the seasons are named. Was this intentional?

JPKribs:
No but also having the text just be "-" also isn't ideal. But, I'm not really sure how you can have a season without a name. That value isn't optional and we have a fallback:

https://github.com/jellyfin/Swiftfin/blob/main/Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto.swift

It should always be a name or "Unknown"? Do you have empty season/specials folders? I know Jellyfin still counts those as seasons.

sjorge:
Probbaly just a static 'Season <#>' I guess you can heave a show with where you just have Season 1 but equally where you just have the last season in it...

Screenshot 2025-06-18 at 16 00 35

Screenshot 2025-06-18 at 16 00 59

Screenshot 2025-06-18 at 16 02 31

Screenshot 2025-06-18 at 16 02 54


I think just having the static text be 'Season <#>' is fine, jellyfin should know the season index at least. Let me grab screenshots from the same show on tvOS with current main branch.

Edit: tvOS (I also realized tvOS does not like it when only a series poster is available it's just black :(
Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-06-18 at 16 21 10
Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-06-18 at 16 20 50

LePips:
The issue wouldn't be that the seasons are named, but that the initial selection is `nil`.

LePips:
Frankly, I'm unable to reproduce the issue with my one season animes.

JPKribs:
@sjorge please try: https://github.com/jellyfin/Swiftfin/pull/1591.

I moved the season text to a computed value and removed the `if let` for the season name. Instead, I'm just basing this on total list of seasons in the `ViewModel`. So this should work even if the selection starts off as `nil`.

Still WIP for testing and I want to chain this with a fix for the layout under-assigning the vertical size of the frame:

Screenshot 2025-06-18 at 09 46 34

Issue #1580, opened 2025-06-15
Display information for unfinished content


iPleyGit:
### Describe the bug

When you have an episode halfway through, the top button still shows S1:E1. A clearer way would be to show something like “resume at 23 min” or something similar so people know there’s an episode in progress.
On another note, if I’m not mistaken, I think the button doesn’t update in real time, so you have to leave the screen and come back for things to appear updated.

PD: Thinking about it a bit more, maybe the button could be changed to say “Continue” and include a progress bar on the same button showing the remaining minutes.

<details>
  <summary>Swiftfin</summary>

![Image](https://github.com/user-attachments/assets/2871c8a5-4e38-417d-af98-312c6a283b7f)

</details>

<details>

### Application version

1.3

### Where did you install the app from?

TestFlight

### Device information

iPhone 13 Pro Max 

### OS version

18.5

### Jellyfin server version

10.11 and 10.10.7

JPKribs:
This exists but is a tiny bit bugged. The issue is, after leaving the content, we need to refresh the view to get the new playback position.

If you leave the view, come back it will show the time remaining of the current episode:

Example:

<img width="273" alt="Image" src="https://github.com/user-attachments/assets/3210dc47-2807-4527-a5e4-1f7c8348f84d" />

This bug can be tracked here:

https://github.com/jellyfin/Swiftfin/issues/1560

ClappTV:
I saw that issue, but it's not the same view. If you look closely, I’m on the screen where all the episodes are listed, and that image appears when you're on the episode itself

JPKribs:
> I saw that issue, but it's not the same view. If you look closely, I’m on the screen where all the episodes are listed, and that image appears when you're on the episode itself

You are totally right, apologies! Re-opening - The issue that I see as a potential is that we have `Compact Poster` as an ItemType. So, whatever text we are using needs to be usable on this level as well as the larger buttons. Meaning, this is more likely just using the button coloring to show time remaining without additional text. Especially since the time remaining can been found by scrolling down a hair and seeing the episode on the Episode Selector. 

I had a version of this from a while ago I can dust off but I found it didn't fit the Apple aesthetic very well.

ClappTV:
<details>
  <summary>continue button</summary>

![Image](https://github.com/user-attachments/assets/e81d2d55-24df-4af5-9740-e65432750587)

</details>

[...]

Issue #1560, opened 2025-06-08
The playback starts again from the beginning


tobylibo:
### Describe the bug

This bug has probably been around for a long time, and it's been bothering me ever since.

When the “Play” button is clicked, the video plays normally. If you close the playback window to return to the movie details and re-click the "Play" button, the video will play again from the beginning, not from the previous playback position. 

This bug is present in all versions.

### Application version

1.3 (13)

### Where did you install the app from?

TestFlight

### Device information

iPhone 12 Pro, iPad Pro

### OS version

iOS 18.5

### Jellyfin server version

10.10.7

tobylibo:
If you close the playback window to return to the movie details, the "Play" button shows "Play" instead of Remaining Time.

JPKribs:
If anyone is interested in looking at this:

This is because the `ItemViewModel` needs to appropriately update the `BaseItemDto` after playback closes/stops. Or, at least, it should update the `userData` for the item for `isPlayed` and `playbackPositionTicks` (or whatever the real name it) when playback is stopped. I haven't looked into it but I believe we need to kick off that notification from the end of playback so it may conflict with #1203.

If we're already kicking off that notification at the end of playback, this would just be updating the `ItemViewModel` to update the `userData` of the item when that notification is called.

PR #1581, opened 2025-06-16
Media Player

LePips:
This work was originally #1203.

---

This is an extensive rework of the video player and overall media management, removing the old strongly coupled `VideoPlayerManager` and its subclasses for a new `MediaPlayerManager`.  I last redid the video player thoroughly in #593 when I transitioned it from UIKit to SwiftUI. This is a more robust implementation that (coincidentally) mirrors the AVPlayer system.

## Media Player

A new architecture with `MediaPlayerManager` and `MediaPlayerItem`s has been implemented, where the manager holds the current session of a media item's playback (seconds progressed, playing/paused, error, etc.). It is up to a `MediaPlayerProxy` - the actual video or audio player - to drive media playback and report back to the manager relevant information.

## Video Player

The video player has been entirely redone.

- On iOS, now in portrait and landscape mode.
	- One step closer to https://github.com/jellyfin/Swiftfin/issues/321. Ensuring the player would work/function properly was my biggest concern.
- Pan-interactive overlay.
- Allow different rendering layers (VLC, AVPlayer, an MPV layer would finally fit in fairly easily).
	- I would interest a "dynamic" decision, as stated in https://github.com/jellyfin/Swiftfin/issues/1459. However, more investigation on how this would actually work is needed (mainly with subtitle selection).
- Further interaction with supplements below.

Since the video player has been strongly decoupled from the manager, this allows starting a new stream for the same media item. Potential benefits include:

- Changing playback quality
	- Implemented, but not surfaced to users for now.
	- https://github.com/jellyfin/Swiftfin/issues/184
- Subtitle settings (burn, embed, etc.)
	- https://github.com/jellyfin/Swiftfin/issues/699
	- Probably other issues requesting this.

### Live TV and Downloads

Live TV and downloads would previously implement other "video player managers". Now, they simply pass in the necessary playback item information to a player. Other abstractions make UI accommodations quite easy.

### Supplements and Observers

Observers and supplements are like plugins for the media manager, allowing players to present additional information and introduce more functionality during media playback. Supplements can provide some UI component to be presented to a player. Some implemented supplements are the media info, chapters, and episode queue.

Possible additional supplements and observers:
- Video playback info (VLC/AVPlayer/Jellyfin playback information)
	- https://github.com/jellyfin/Swiftfin/issues/339
- Change playback profiles or other media settings during playback
	- subtitle size/color/font
- Change media sources
- Additional queues of items
	- Playlists
	- Libraries
	- Collections
	- "Parts" for multi-part items
[...]

PR #1589, opened 2025-06-17
Refresh `ItemView` on End of Playback

JPKribs:
## Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1580
Resolves: https://github.com/jellyfin/Swiftfin/issues/1560

Potentially conflicts with: https://github.com/jellyfin/Swiftfin/pull/1581

Fixes the `...ItemView` not refreshing when leaving the View Player. There are three major changes in here:

1. `ItemViewModel` cleanup to move everything to overriding the `response` and removal of the deprecated `onRefresh()`. 
2. Migrate the `SeasonItemViewModel` to use `ItemViewModel`. This is because I don't think we should be calling a full refresh on `PagingLibraryViewModel` when an item updates. Moving this to `ItemViewModel` let's us keep all the `ItemView` refresh logic consistent. Later on, I will want to update `PagingLibraryViewModel` to refresh as well but this should be a `.update()` to the collection at the element location using the full `Element` as opposed to a full refresh or all elements.
3. In `VideoPlayerManager.sendStopReport()` I added our notifications. I think this can be cleaner. Part of our update issue is that the item sent need to update the Series, Season, and Item. Since we're only sending the `ID`, this was the simplest spot to call all three. Without this, there isn't a clean way to update the Series on an episode change or vice versa. Calling 3 notifications at once feels expensive but I'm not sure how to do this cleaner. Open to suggestion!

**This also resolves the issue on tvOS as well.**

### Movie:

https://github.com/user-attachments/assets/ca00e0a6-7a92-4131-bbc9-1c161fd81c62

### Show:

https://github.com/user-attachments/assets/94d26ff1-0b54-482b-aba8-3f24270c7946

### Episode:

https://github.com/user-attachments/assets/8927a09f-5fb8-4c43-bc2c-5527a6775ce6

### DIFF
--- a/Shared/ViewModels/ItemViewModel/SeriesItemViewModel.swift
+++ b/Shared/ViewModels/ItemViewModel/SeriesItemViewModel.swift
@@ -28,6 +28,7 @@ final class SeriesItemViewModel: ItemViewModel {
     // MARK: - Override Response
 
     override func respond(to action: ItemViewModel.Action) -> ItemViewModel.State {
+
         switch action {
         case .backgroundRefresh, .refresh:
             let parentState = super.respond(to: action)
@@ -67,12 +68,10 @@ final class SeriesItemViewModel: ItemViewModel {
                 }
             }
             .store(in: &cancellables)
-
-            return super.respond(to: action)
-
-        default:
-            return super.respond(to: action)
+        default: ()
         }
+
+        return super.respond(to: action)
     }
 
     // MARK: - Get Next Up Item
--- a/Shared/ViewModels/ItemViewModel/CollectionItemViewModel.swift
+++ b/Shared/ViewModels/ItemViewModel/CollectionItemViewModel.swift
@@ -44,11 +44,10 @@ final class CollectionItemViewModel: ItemViewModel {
                 }
             }
             .asAnyCancellable()
-            return super.respond(to: action)
-
-        default:
-            return super.respond(to: action)
+        default: ()
         }
+
+        return super.respond(to: action)
     }
 
     // MARK: - Get Collection Items
--- a/Shared/ViewModels/ItemViewModel/EpisodeItemViewModel.swift
+++ b/Shared/ViewModels/ItemViewModel/EpisodeItemViewModel.swift
@@ -37,12 +37,10 @@ final class EpisodeItemViewModel: ItemViewModel {
                 }
             }
             .asAnyCancellable()
-
-            return super.respond(to: action)
-
-        default:
-            return super.respond(to: action)
+        default: ()
         }
+
+        return super.respond(to: action)
     }
 
     // MARK: - Get Series Items


JPKribs:
Also, I did this to try and prevent publishing a change if the item hasn't changed. I don't know how much that matters since we're already getting the item(s) from the server already. I mostly did this to prevent redrawing views unnecessarily but I'm unsure if that's correct or if I misunderstand how the CollectionXStacks work?

```
if results.fullItem.id != self.item.id || results.fullItem != self.item {
    self.item = results.fullItem
}
```

JPKribs:
Just an idea, **_not_ going in this PR**, could we combine these two notifications? I'm sure there is a more clever way of doing this but this was what I came up with as a POC:

```
    /// - Payload: The new item with updated metadata.
    static var itemMetadataDidChange: Key<BaseItemDto> {
        Key("itemMetadataDidChange")
    }

    /// - Payload: The ID of the item that should refresh
    static var itemShouldRefreshMetadata: Key<String> {
        Key("itemShouldRefresh")
    }
```

Into a single Notification where the `ItemID` is required but the `BaseItemDto` is optional:

```
    /// - [Required] Payload: The updated ItemID
    /// - [Optional] Payload: The new BaseItemDto with updated metadata.
    static var itemMetadataDidChange: Key<(String, BaseItemDto?)> {
        Key("itemMetadataDidChange")
    }
```

The issue with having two similar notifications is that we need different approaches based on the data available to us. When we have the full `BaseItemDto`, we should just replace the old one with the updated version since we already have everything we need locally. However, sometimes we don't have the complete new item data and it makes more sense to refresh from the server instead. Having both notification types allows us to perform the most efficient action based on what data is available.

The server refresh approach is required for things like playback where the server has rules that determine when something gets marked as played or whether the playback position should be saved from a minimum threshold. In these cases, it's best to refresh from the server to ensure our information is accurate since our local assumptions could be wrong. 

The `BaseItemDto` replacement approach works well for things like editing or admin dashboard actions where we made the changes ourselves and already have the complete updated item locally. Since we have all the data and performed the changes, there's no need to make another server request when we can just update any existing references to use the new item we already possess.

Where, _I feel_, it makes sense to have this as a single notification is to ensure that we're not calling the same notification multiple times per `ViewModel`. This combination would prevent double-updating. Plus, I *believe* we can debounce these notifications too? Which seems like that would be a good additional as well.

I originally thought we could just use the `BaseItemDto` and `init` a new one with just the `id` but that would mean we would need to check for incomplete `BaseItemDto` as part of the sinks. That's why I thought the 2 variable notification would be better but :

1. I don't know if that is good practice / wont get us into issues.
2. I am sure there is a more clever solution I'm not considering here.

I thought I would throw this out there to get the ball rolling while I'm looking at notificaitons. Additionally, would be nice to clean up that Item notification TODO. Interested to hear other thoughts for this!

---

The end result of this would be, when subscribing to this notification we can have logic that runs based on the available information:
    
```
  Notifications[.itemMetadataDidChange]
  .publisher
  .sink { newItemID, newItem in
      guard newItemID == self.item.id else { return }

      Task {
          if let newItem, newItem != self.item {
              await self.send(.replace(newItem))
          } else {
              await self.send(.refresh)
          }
      }
  }
  .store(in: &cancellables)
```

So if the full item exists we can use it. Otherwise, we call the update from server.

Shared/ViewModels/ItemViewModel/SeriesItemViewModel.swift
```diff
@@ -21,36 +21,62 @@ final class SeriesItemViewModel: ItemViewModel {
...
+    // MARK: - Task
+
+    private var seriesItemTask: AnyCancellable?
+
+    // MARK: - Override Response
+
+    override func respond(to action: ItemViewModel.Action) -> ItemViewModel.State {
+        switch action {
+        case .backgroundRefresh, .refresh:
+            let parentState = super.respond(to: action)
```
LePips, line 34: Just have the `super.respond` at the end of the function instead.

Shared/ViewModels/ItemViewModel/SeasonItemViewModel.swift
```diff
@@ -6,40 +6,65 @@
...
+import Combine
 import Defaults
 import Foundation
 import JellyfinAPI
 
-// Since we don't view care to view seasons directly, this doesn't subclass from `ItemViewModel`.
-// If we ever care for viewing seasons directly, subclass from that and have the library view model
-// as a property.
-final class SeasonItemViewModel: PagingLibraryViewModel<BaseItemDto>, Identifiable {
+final class SeasonItemViewModel: ItemViewModel, Identifiable {
```
LePips: > This is because I don't think we should be calling a full refresh on PagingLibraryViewModel when an item updates.

The full refresh isn't due to the conformance to `PagingLibraryViewModel`, but how `SeriesItemViewModel` performs its background update - we are still doing a full refresh on updates for the season and the series.

I'm confused why you think we need to update the season item, the item itself is only used for its ID. I would still like conformance to `PagingLibraryViewModel` to investigate lazy loading the seasons.

> could we combine these two notifications?

It is correct that the notification system for these could be cleaned up. I have thoughts of other solutions but can work on that later.

While working on the video player, I've realized a lot about the separation between item data and user data. Item data rather rarely changes (description, blur hashes, children). The user data is what we actually care about regarding playback progress, so we should have a solution that involves `getItemUserData` as well.

Part of that solution also involves the sockets/messaging where user data information can be received, as well as a shared item cache I'll have to investigate thoroughly.

LePips (changes requested):

Shared/ViewModels/ItemViewModel/SeasonItemViewModel.swift
JPKribs: > This is because I don't think we should be calling a full refresh on PagingLibraryViewModel when an item updates.

Apologies, what I meant by this is we should probably be passing in the full `BaseItemDto` into this to replace the existing item opposed to passing in just the `ItemId`. My reasoning is we don't have a single item refresh on the `PagingLibraryViewModel`. Doing a full replacement shouldn't be too hard but to go from `ItemId` to a replaced value would likely require a new action on the `PagingLibraryViewModel`.

> I'm confused why you think we need to update the season item, the item itself is only used for its ID

I was getting a scenario where, on the `SeriesItemView`, this would get stuck. So, if I updated an item(s) in the Episode Picker this would result in the picker just showing nothing but blank items. Adding the Season update fixed this but it sounds like that might have been a red herring. I can try this without the season tomorrow and confirm

> Part of that solution also involves the sockets/messaging where user data information can be received, as well as a shared item cache I'll have to investigate thoroughly.

I'm also fine holding off on this change until after the player changes. Or, alternatively, doing a more minimal change with this PR to just remove the deprecated `onRefresh()` and enabling the commented out call to `ItemId` & `SeriesId`? 

From your requested changes it sounds like I'm a bit off-base with the full scope of this change so narrowing it to just the notification for now (and `onRefresh()` if desired). Then, I can take another crack at this after https://github.com/jellyfin/Swiftfin/pull/1581 and whatever the more final version of https://github.com/jellyfin/jellyfin-sdk-swift/pull/46 looks like?

Let me know what you think is best! If this PR feels like a bad fit, I'm not against starting over / dropping this either.

LePips: > passing in the full BaseItemDto

Yes, that is ideally a function that `PagingLibraryViewModel` would be able to do as well.

The other changes for posting the notifications, even in the player, and the deprecation handling is good and we can keep that.

JPKribs:
I've narrowed this scope to purely the `onRefresh()` removal and the notification setup for `ItemViewModel`. Removing the `SeasonId` from the notification, I did see some issues but I see the same issues with `SeasonId` enabled too so it's unrelated. I've removed `SeasonId` from the notification.

The issue I am seeing is It seems that when we update the `PosterHStack` on the `SeriesItemView` it redraws and there is a chance it just doesn't come back OR it comes back empty. You can recreate this with the `Mark as Played` button just marking and un-marking a season as played. I can get it to trigger like 5% of the time so it's not a big deal and outside the scope of this PR.

This should be good to go. I left a TODO for the notifications just to mark them easier for https://github.com/jellyfin/Swiftfin/pull/1581 but I can remove that if it isn't needed.

LePips (approved):

PR #1591, opened 2025-06-18
Fix Single Season Text on Nil Selection

JPKribs:
### Summary

You could enter this the `ItemView` with the selected season as `nil`. As a result, the single season would still be a picker when it should just be the new text.

This should force static text if there is only one season available. The text will appropriately update when the selected season is no longer `nil` since the title is a computed var now and used in both places were a season title would be seen on this view.


sjorge:
Hmm I don't see a different, is this one only use for Swiftfin iOS ?

JPKribs:
Correct. We do not change the season picker to static text on tvOS.

sjorge:
That explains it, looks OK in iOS simulator.

JPKribs:
> That explains it, looks OK in iOS simulator.

Were you previously only getting this issue in tvOS?

sjorge:
Yes, I wonder if it's because they are virtual seasons.

JPKribs:
I think there is a miscommunication about this issue then. If this was only ever a tvOS issue, that is expected as we are not making this change for tvOS. I personally don't think it adds much to that view and, on a bigger note, this would require a pretty major change to how we handle `focus`.

I'm going to close this out at this change is not necessary

PR #1602, opened 2025-06-22
Migrate to `NavigationStack`

### LINKED ISSUE #1034: Replace or work on Stinsen

LePips:
Closes #1034 

Migrates away from Stinsen to `NavigationStack`, implementing the root, tab, and navigation coordination objects with a new `Router` to use for all `View` routing.

New routes are now implemented as `NavigationRoute`s and can specify the style (push, sheet, fullscreen), the same as we did before with Stinsen. The routes are now in a single namespace which removes a lot of duplication, just at the cost of paying closer attention to what route is being used.

This work is just the basic migration with some other changes made if they were necessary for the new system. I've realized a lot of other cleanup that can happen and left some more `TODO`s. There are also other `.sheet` uses that should be migrated to being a `NavigationRoute`.

Using the new system, the `Search` tab will now focus on the search bar when the tab is selected again while searching.

This also lays the foundation for customizing the tabs as well as the iOS zoom navigation.

This does remove the previously existing deep linking construction from the root coordinator, but a new solution would not be difficult to implement either. The tvOS series label also gets weird again, but I defer that fix until later.


JPKribs:
Just for tracking, will this resolve https://github.com/jellyfin/Swiftfin/issues/1034 as well or is this purely `NavigationView` -> `NavigationStack`?

LePips:
Yes, this is that work. I'll provide a description soon.

LePips:
I just ask for a sanity check to ensure that nothing was missed or anything odd isn't happening with navigation.

JPKribs:
Sounds good! I'm out and about today, but I can test this this evening.

> anything odd isn't happening with navigation.

Should this resolve the state errors Stinson was giving us as well? Test iOS only or tvOS as well?

LePips:
It's both iOS and tvOS, and yes the state errors should largely be gone.

chickdan:
Nice! I was just looking into routing patterns with navigation stack yesterday with plans to do this work 😅

JPKribs:
iOS I ran through basically every scenario I can think of. No issues at all! This looks great.

tvOS builds on my end but for some reason there is a test error. I haven't had a chance to look into that more but it builds fine locally. 

I have a weird interaction on tvOS where if I go Home > Movie > Actor then I back out of the actor it takes me back to Home but that appears on Main as well. I think just needs to be it's own stack where now I think it adds to the existing stack.

Outside of that this looks great!

These are not breakers for this but some questions:

1. What about [half sheets](https://developer.apple.com/design/human-interface-guidelines/sheets)?
2. Do we need/want [popovers](https://developer.apple.com/design/human-interface-guidelines/popovers) while we're doing this?
3. Does any consideration need to be made for split-screen iPadOS? Not prodding, more just I don't know myself so I want to put that out there in case we need to account for this. Similar to the popover question.

---

After this is merged, I can set some time aside to find all the `.sheet` and migrate them to the new system. I think most of the sheets are my creation so I feel responsible.

---

But outside of the one quirk and the build issues this looks awesome! Once you get the tvOS build test to work just ping me and I can click the approve!

Shared/Coordinators/Navigation/NavigationCoordinator.swift
```diff
@@ -0,0 +1,57 @@
...
+    func push(
+        _ route: NavigationRoute,
+        style: NavigationRoute.TransitionStyle? = nil,
+        in namespace: Namespace.ID? = nil
+    ) {
+        let style = style ?? route.transitionStyle
+
+        #if os(tvOS)
+        switch style {
+        case .push, .sheet:
```
JPKribs, line 36: Half Sheets & Popovers?

JPKribs (changes requested):
Forgot to make this a review. Please just see my one comment above this one

LePips:
> half sheets

Presentation detents are determined on the view being presented.

<details>
<summary>`FilterView` w/ medium detents</summary>
Screenshot 2025-06-22 at 7 33 09 PM
</details>

> popovers

Popovers are not considered something navigable, they are to enhance the current view's content and should be restricted to just that view. That argument could also be made for the sheet instance for `LearnMore`, so we shouldn't transition that one over.

> split-screen iPadOS

`NavigationSplitView` is easily possible and I've played around with it locally, however there are layout issues with my `Collection_` packages I need to fix first.

On iOS 18+, we can now support the [`sidebarAdaptable`](https://developer.apple.com/documentation/swiftui/tabviewstyle/sidebaradaptable) tab view style.

JPKribs (approved):
Looks great! I've tested this latest version and everything works exactly the same as Main!

JPKribs:
@LePips no rush, but could we do a TestFlight after this? Start giving folks a chance to test the navigation changes so we can get ahead of anything that might've been missed

LePips:
I will after I get some local changes to poster buttons in.

Issue #1590, opened 2025-06-17
Create `PersonItemView`


JPKribs:
**Describe the feature you'd like**
Making an issue to track this while I work on this. Expecting this to be a larger PR so I'm probably looking at 1.5+ for this.

As we eventually start moving towards [media types that aren't movies and shows](https://github.com/jellyfin/Swiftfin/blob/main/Documentation/libraries.md), it begins to be more important to have a proper person view. Right now, selecting a person from Cast & Crew or the Search just takes you to the `PagingLIbraryView` of the items that person has been in. As Music, Artists, Books, etc. come to Swiftfin, it becomes more important to have views for looking and organizing by people.

There is a way to work this into the existing `ItemViewModel`. I have a working version of this now but Jellyfin has 2 datatypes (BaseItemPerson & BaseItemDTO) for people so I think I need to figure out a way to init from both.

**Additional context**

I'd love to just use the `CompactPosterScrollView` for this but that might end up being kind of hacky. I may look towards making person view for iOS, iPadOS, and tvOS that's separate but I'm 100% open to open to feedback! Below is my current working version for iOS:

![Image](https://github.com/user-attachments/assets/8f3d290c-d01f-4c9e-89b9-6bd4eebb6889)

https://github.com/user-attachments/assets/c925f4d7-4719-40fb-881a-f201ff88587e

https://github.com/user-attachments/assets/8612900b-91cc-44a5-b248-2dd9d43eb32c

JPKribs:
I think I'm `done` with this for now. There are a couple of other PRs I want to get wrapped up first which directly conflict with my changes needed here. For the sake of my sanity and reducing merge mess, I am going to sit on this one until at least https://github.com/jellyfin/Swiftfin/pull/1583 is merged. That way I don't need to edit the Metadata Editing settings twice.

For now, here is the branch with this work done. Outside of the design of the view (and potentially being able to re-use another view), this should be `ready` for iOS. Would still need to do the same work for tvOS but I'll do that as we get closer to needing this.

https://github.com/JPKribs/Swiftfin/tree/itemPersonView

LePips:
It is about time we should implement enhanced views for people. I've always thought to just do something similar to (what's now called) `ItemView.SimpleScrollView` and change the header to just the portrait image + name and the labeled details would go below the description. So, a little reorganization without the header image. However, seeing the header image being used here I think I like it as it fills in dead space that would naturally come from how little data we would be able to present.

[...]

Issue #1551, opened 2025-05-25
Collection items not paginated.


lostb1t:
When browsing to a collection, the call to the items endpoint is not limited. In my case i have big collections and this results in large payloads and timeouts.

PR #1606, opened 2025-06-23
Person `ItemView`

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1590
Resolves: https://github.com/jellyfin/Swiftfin/issues/1551 (Subject to change)

Creates an `ItemView` for people and artists. This can be used instead of the existing process where selecting a person goes to a `PagingLibraryView` with content from that person.

### iPadOS

Screenshot 2025-06-22 at 19 53 15

### Videos

**From Search**

https://github.com/user-attachments/assets/15f659a3-bd2d-4076-b415-d05197fb3988

---

**From Cast & Crew**

https://github.com/user-attachments/assets/e099f525-6dd0-48f4-9119-069b2e8b2270


JPKribs:
@LePips How would you prefer I handle this? All in 1 PR for iOS and tvOS or just iOS first then tvOS separate?

At this point, iOS & iPadOS _should_ be ready to go. I'm leaving this as WIP in case we want to add tvOS as well in this PR.

Shared/ViewModels/ItemViewModel/PersonItemViewModel.swift
```diff
@@ -0,0 +1,66 @@
...
+import Foundation
+import JellyfinAPI
+import OrderedCollections
+
+final class PersonItemViewModel: ItemViewModel {
+
+    // MARK: - Published Collection Items
+
+    @Published
+    private(set) var personItems: OrderedDictionary<BaseItemKind, [BaseItemDto]> = [:]
```
LePips: This is going to suffer from the same issue in #1551. To remedy both that issue and this one, we need to have a map between the `BaseItemKind` and corresponding `ItemLibraryViewModel`s.

Swiftfin/Views/ItemView/Components/ActionButtonHStack/ActionButtonHStack.swift
```diff
@@ -56,26 +56,30 @@ extension ItemView {
...
-                    item
-                        .foregroundStyle(
-                            .primary,
-                            accentColor
-                        )
-                }
-                .if(equalSpacing) { view in
-                    view.frame(maxWidth: .infinity)
+                /// Marking Persons and Artists as played doesn't do anything.
+                if viewModel.item.type != .person && viewModel.item.type != .musicArtist {
```
LePips, line 60: Make a property called `canBePlayed` that switches like this, default to `true`.

Swiftfin/Views/ItemView/ScrollViews/PersonScrollView.swift
LePips, line 1: We should not be copying entire views like this. Instead, just switch on the item's type to set the relevant content.

Swiftfin/Views/ItemView/ScrollViews/PersonScrollView.swift
```diff
@@ -0,0 +1,177 @@
...
+
+        @ViewBuilder
+        private var rightShelfView: some View {
+            VStack(alignment: .leading) {
+                Text(viewModel.item.displayTitle)
+                    .font(.title2)
+                    .fontWeight(.semibold)
+                    .foregroundColor(.white)
+
+                DotHStack {
```
LePips: I don't think I like all of this data in the shelf, try putting these below with the overview instead.

Swiftfin/Views/ItemView/ScrollViews/iPadOSPersonScrollView.swift
LePips, line 1: Same copying concern here.

Shared/Extensions/FormatStyle.swift
```diff
@@ -205,6 +205,29 @@ extension FormatStyle where Self == LastSeenFormatStyle {
...
+extension FormatStyle where Self == AgeFormatStyle {
+
+    static var age: AgeFormatStyle { AgeFormatStyle() }
+}
+
+struct AgeFormatStyle: FormatStyle {
+
+    private var death: Date?
+
+    private var today: Date {
```
LePips: Replace with just using `.now`.

Swiftfin/Views/ItemView/ScrollViews/PersonScrollView.swift
JPKribs: Sounds good. I had an earlier version of this where I was just switching inside of `CompactPortraitScrollView` on Person/Artist. Is that the route you see this going?

LePips: Yes

Shared/ViewModels/ItemViewModel/PersonItemViewModel.swift
JPKribs: I like that idea! I've tried to build on that but I'm hitting a bit of a wall. What I have below "works" but it means every person we return 36 `ItemLibraryViewModel`s even though only a couple will actually be used.

How can I efficiently determine the full scope of the `BaseItemKind`s that would exist in a Collection or under a Person without just calling 36 times to see if there is a single item for that item for that kind?

```
//
//
//

import Combine
import Foundation
import JellyfinAPI
import OrderedCollections

final class PersonItemViewModel: ItemViewModel {

    @Published
    private(set) var personItems: OrderedDictionary<BaseItemKind, ItemLibraryViewModel> = [:]

    private var personItemTask: AnyCancellable?

    override var presentPlayButton: Bool {
        false
    }

    override func respond(to action: ItemViewModel.Action) -> ItemViewModel.State {

        switch action {
        case .backgroundRefresh, .refresh:

            personItemTask?.cancel()

            personItemTask = Task { [weak self] in
                guard let self else { return }

                await MainActor.run {
                    self.personItems.removeAll()
                }

                do {
                    await MainActor.run {
                        self.personItems = self.getPersonItems()
                    }
                }
            }
            .asAnyCancellable()
        default: ()
        }

        return super.respond(to: action)
    }

    private func getPersonItems() async throws -> OrderedDictionary<BaseItemKind, ItemLibraryViewModel> {
        guard item.id != nil else {
            return [:]
        }

        let supportedKinds = BaseItemKind.supportedCases
        var allViewModels: [BaseItemKind: ItemLibraryViewModel] = [:]
        var completedViewModels: [BaseItemKind: ItemLibraryViewModel] = [:]

        for kind in supportedKinds {
            let viewModel = ItemLibraryViewModel(
                parent: item,
                filters: .init(itemTypes: [kind])
            )
            allViewModels[kind] = viewModel
        }

        await withTaskGroup(of: (BaseItemKind, Bool).self) { group in
            for (kind, viewModel) in allViewModels {
                group.addTask {
                    await withCheckedContinuation { continuation in
                        var cancellable: AnyCancellable?

                        cancellable = viewModel.$state
                            .sink { state in
                                if state == .content {
                                    cancellable?.cancel()
                                    continuation.resume(returning: (kind, state == .content))
                                }
                            }

                        Task { @MainActor in
                            viewModel.send(.refresh)
                        }
                    }
                }
            }

            for await (kind, isSuccess) in group {
                if isSuccess, let viewModel = allViewModels[kind], !viewModel.elements.isEmpty {
                    completedViewModels[kind] = viewModel
                }
            }
        }

        return OrderedDictionary(
            uniqueKeysWithValues: completedViewModels.sorted { $0.key.rawValue < $1.key.rawValue }
        )
    }
}
```

Swiftfin/Views/ItemView/ScrollViews/PersonScrollView.swift
JPKribs: Here is where I settled. I moved most of the data into the overview minus the age. Mostly just to have *sometime* up there besides the name. I also didn't think it looked good in the overview. Let me know if you want me to change this.

https://github.com/user-attachments/assets/661b6e7f-480d-43dc-94bd-11fcae05e737

https://github.com/user-attachments/assets/df46d069-1b26-4277-96f6-6abbc958469a

https://github.com/user-attachments/assets/e7b4ed30-956a-4394-b8d0-7fc42632e028

Shared/ViewModels/ItemViewModel/PersonItemViewModel.swift
JPKribs: I have this working in my latest version. In testing, this works really well! This tries for each supported base item type + episodes. Not ideal but not as bad as the original full 36 times for all cases. Hopefully you have an idea on how to reduce this but the response from the server is always just [] so bandwidth wise it doesn't seem _too_ bad.

I've updated the Collections to mirror this while I was doing this. In testing, this means that the page of the CollectionView and PersonView will contain 25 items and nothing more. I didn't think this was an issue as the see more shows all of this. I don't think this is too bad since this is for Collections and Persons only. Otherwise, this would be a `CollectionHStack` change to allow pagination there.

---

There was one bug where the Collections would say it contained every Collection item? Like, the Collection would return EVERY collection in the server and ignore the `ParentId` passed in. Not sure if that is an SDK bug or me failing to understand the API but if the `LibraryParent` is a `.boxset` and the `ItemType` filter is a `.boxset` it just returns ALL `.boxset`. I've disabled nested `.boxset` to resolve this. The loss is you cannot nest collections inside each other in Swiftfin like you can in Web. I've left a comment explaining it and it only seems to impact the Collection version of the ViewModel so I'm at a loss there.

Swiftfin/Views/ItemView/ScrollViews/PersonScrollView.swift
JPKribs: Ignore the missing person image in the searchview and on the about card at the bottom. I just resolved that item in the last commit.

LePips:
I've abstracted out the item type library fetching, which can also eventually use used for `Media/Favorites`. I've also accounted for some other interactions and brought over some work from my poster button cleanup.

> There was one bug where the Collections would say it contained every Collection item?

On the server if the requested item type only has a single element and is the boxset type, it nullifies the parent ID. I'm unsure why this is the case, but it wouldn't hurt to investigate if the web or other clients use this behavior and get rid of this behavior, or have us find a workaround.

https://github.com/jellyfin/jellyfin/blob/630846798d8302ea5ed805ade669158b7af2f123/Jellyfin.Api/Controllers/ItemsController.cs#L274

JPKribs:
>  it wouldn't hurt to investigate if the web or other clients use this behavior and get rid of this behavior, or have us find a workaround.

It looks like Jellyfin-Web just groups multiple items together so the call would have more than one `ItemType` filter in it:

https://github.com/jellyfin/jellyfin-web/blob/master/src/controllers/itemDetails/index.js#L1694C1-L1700C10

Testing, this is what this looks like:

image

Swiftfin/Views/ItemView/Components/OverviewView.swift
```diff
@@ -37,10 +40,32 @@ extension ItemView {
...
+                if let deathday = item.deathday?.formatted(date: .long, time: .omitted) {
+                    LabeledContent(
+                        "Died",
+                        value: deathday
+                    )
+                }
+
+                if let birthplace = item.birthplace {
+                    LabeledContent(
+                        "Birthplace",
```
JPKribs: Leaving myself a note: Localize!

LePips (approved):

Issue #1608, opened 2025-06-24
`.Sheet` Cleanup


JPKribs:
**Describe the feature you'd like**
This is an issue to track this progress. The goal is to find all instances of `.sheet` and migrate them to instead use the new Coordinators.

The `LearnMore` modals can be ignored as these `.sheet` are contextual information. This should purely be to remove the usage of `.sheet` for navigation.

**Additional context**
Creating this issue to link to our Help Wanted page: https://github.com/jellyfin/Swiftfin/discussions/1503

JPKribs:
Here is the full extent of what needs to be updated:

<img width="717" alt="Image" src="https://github.com/user-attachments/assets/dd7cccb8-521e-4272-bf27-b73f063638d5" />

PR #1609, opened 2025-06-24
Coordinator `.Sheet` Cleanup

JPKribs:
## Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1608

This PR resolves the `.sheet` routing found in `ItemEditorView` Image adding and Item Identification processes. I originally wanted to leave this for other folks to work on but:

1. I realized the `ItemImage` routing was non-functionality. This increased the priority for me.
2. Getting `ItemImage` routing working was more complicated than I originally thought it would be. So, I felt like I should take the lead

This migrates all `.sheet` to use the coordinator. As part of this, some of the `router.dismiss()` we were previously using needed to be migrated. I instead have this await the `viewModel.event` so we don't dismiss too early. I have recorded the processes below to confirm this is working as expected.

For `ServerActivityView`, I moved the filter from a `.sheet` to `ServerActivityFilterView`. This was probably the biggest change in terms of number of lines but this was almost a pure copy-paste.

I also migrated our last 2 `NavigationView` to `NavigationStack`.

## Assumptions

I left the following as stand alone `.sheet`:

1. `LearnMoreButton` - _Should_ be fine

I left the following as stand alone `.NavigationStack`:

1. `LearnMoreButton` - _Should_ be fine
2. `OrderedSectionSelectorView` - tvOS (didn't want to touch that yet)

## Outstanding Questions:

1. There is a TODO in `AppSettingsView`: `// TODO: Create NavigationRoute.appIconSelector`
2. I am getting this `Call to main actor-isolated instance method 'send' in a synchronous nonisolated context; this is an error in the Swift 6 language mode` in a few spots. Do we need to address this in advance?

### Image - Deletion, Remote Search, & Upload

https://github.com/user-attachments/assets/fd4d507b-5154-48d0-9917-edb521c57c97

---

### Identification

https://github.com/user-attachments/assets/e1cd983a-4149-4832-bdc0-23efeac156e9

---

### Activity Filters

https://github.com/user-attachments/assets/491fb996-a890-4de5-a966-c40f1e9a6fdf

### DIFF
--- a/Shared/Coordinators/Navigation/NavigationRoute/NavigationRoute+Item.swift
+++ b/Shared/Coordinators/Navigation/NavigationRoute/NavigationRoute+Item.swift
@@ -207,48 +207,27 @@ extension NavigationRoute {
         }
     }
 
-    static func itemImageDetails(
-        viewModel: ItemImagesViewModel,
-        imageInfo: ImageInfo
-    ) -> NavigationRoute {
+    static func itemImageDetails(viewModel: ItemImagesViewModel, imageInfo: ImageInfo) -> NavigationRoute {
         NavigationRoute(
             id: "itemImageDetails",
             style: .sheet
         ) {
             ItemImageDetailsView(
                 viewModel: viewModel,
-                imageSource: imageInfo.itemImageSource(
-                    itemID: viewModel.item.id!,
-                    client: viewModel.userSession.client
-                ),
-                index: imageInfo.imageIndex,
-                width: imageInfo.width,
-                height: imageInfo.height,
-                onDelete: {
-                    viewModel.send(.deleteImage(imageInfo))
-                }
+                imageInfo: imageInfo
             )
             .environment(\.isEditing, true)
         }
     }
 
-    static func itemSearchImageDetails(viewModel: ItemImagesViewModel, imageInfo: RemoteImageInfo) -> NavigationRoute {
+    static func itemSearchImageDetails(viewModel: ItemImagesViewModel, remoteImageInfo: RemoteImageInfo) -> NavigationRoute {
         NavigationRoute(
             id: "itemSearchImageDetails",
             style: .sheet
         ) {
             ItemImageDetailsView(
                 viewModel: viewModel,
-                imageSource: ImageSource(url: imageInfo.url?.url),
-                width: imageInfo.width,
-                height: imageInfo.height,
-                language: imageInfo.language,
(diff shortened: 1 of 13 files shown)


Shared/Coordinators/Navigation/NavigationRoute/NavigationRoute+Item.swift
```diff
@@ -159,6 +159,21 @@ extension NavigationRoute {
         }
     }
 
+    static func itemIdentify(
```
LePips: `identifyItem`

Shared/Coordinators/Navigation/NavigationRoute/NavigationRoute+Item.swift
```diff
@@ -167,6 +182,56 @@ extension NavigationRoute {
             ItemImagesView(viewModel: viewModel)
         }
     }
+
+    static func selectItemImage(viewModel: ItemImagesViewModel, imageType: ImageType) -> NavigationRoute {
```
LePips: Please sort these alphabetically within the other routes.

Swiftfin/Views/AdminDashboardView/ServerActivity/ServerActivityFilterView/ServerActivityFilterView.swift
```diff
@@ -0,0 +1,81 @@
...
+
+            Button(L10n.save) {
+                viewModel.minDate = startOfDay
+                router.dismiss()
+            }
+            .buttonStyle(.toolbarPill)
+            .disabled(viewModel.minDate != nil && startOfDay == viewModel.minDate)
+        }
+        .onFirstAppear {
+            if let minDate = viewModel.minDate {
```
LePips: Have this set in an `init` instead.

LePips (changes requested):
> There is a TODO in AppSettingsView

That looks like a leftover comment during my work, comment can be removed.

> I am getting this Call to main actor-isolated instance method ... Do we need to address this in advance?

Do not worry about this warning, for now.

Shared/Coordinators/Navigation/NavigationRoute/NavigationRoute+Item.swift
JPKribs: I moved this to `identifyItemResults` as to not conflict with the parent `identifyItem` route.

PR #1610, opened 2025-06-24
Cleanup Player Documentation

JPKribs:
### Summary

Update the video player documentation to better mirror https://jellyfin.org/docs/general/clients/codec-support/#video-compatibility.

1. Changes 🟡 to 🔶. This is not necessary but it mirrors Jellyfin's main documentation 🤷‍♂️.
2. Adds a note about power consumption and VLC.
3. Better mirrors our device profiles in code (Subtitles that are encoded, H.261, etc). I have not tested this for documentation but if we have this enabled in code it should be mirrored in documentation.
4. Moves some comments from inline to under the table.
5. Validated the Video formats to include depth. I have just mirrored what we have in our device profiles.

This is all based on the previous work of others in creating our profiles. Unfortunately, Apple doesn't seem to _have_ a simple table of AVKit and supported formats so I'm leaning a lot of testing and recording but this is far from foolproof.

I've made a note to review this after https://github.com/jellyfin/Swiftfin/pull/1581 as I suspect things may change.

### DIFF
--- a/Documentation/players.md
+++ b/Documentation/players.md
@@ -1,6 +1,6 @@
 # Player Differences
 
-Swiftfin offers two player options: Swiftfin (VLCKit) and Native (AVKit). The Swiftfin team recommends using Swiftfin (VLCKit) for optimal compatibility and features, though Native (AVKit) is also available for certain cases that benefit from Apple's native capabilities. All video, audio, and subtitle formats listed are supported for direct playback but may be repackaged based on container support. If transcoding is enabled on your server, any unsupported formats will be converted automatically.
+Swiftfin offers two player options: **Swiftfin** (VLCKit) and **Native** (AVKit). The Swiftfin team recommends using Swiftfin (VLCKit) for optimal compatibility and features, though Native (AVKit) is also available for certain cases that benefit from Apple's native capabilities. All video, audio, and subtitle formats listed are supported for direct playback but may be repackaged based on container support. If transcoding is enabled on your server, any unsupported formats will be converted automatically.
 
 ---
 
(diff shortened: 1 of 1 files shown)


Documentation/players.md
```diff
@@ -77,23 +83,37 @@ Swiftfin offers two player options: Swiftfin (VLCKit) and Native (AVKit). The Sw
...
+| [H.265 10Bit](https://caniuse.com/#feat=hevc)                                                  | ✅                | 🔶 [2]         |
+| [VP9](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Video_codecs#VP9)             | ✅                | ❌             |
+| [AV1](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Video_codecs#AV1)             | ✅                | 🔶 [3]         |
+
+**Notes**
 
-- AV1 is disabled by default but can be enabled for Native (AVKit) using Custom Device Profiles. Enabling AV1 may result in a [poor experience for SOCs prior to A17](https://en.wikipedia.org/wiki/Apple_A17).
+[1] MPEG-4 Part 2 support may vary depending on encoding parameters and iOS version.
+
+[2] HEVC decoding is supported on Apple devices with the A8X chip or newer and at least iOS 14. HEVC is only supported in MP4, M4V, and MOV containers.
```
holow29, line 112: Swiftfin has always targeted iOS 15+, right? I don't think we need to call out the iOS version in that case. Also, once the next release which drops iOS 15 lands, we can remove the warning about the A8X chip or relegate it to only apply to Swiftfin up to v1.3 since the only devices affected by this limitation are iPad Air 2 & iPad Mini 4 which don't support iPadOS 16. Don't recall any issue with Apple TV HD.

JPKribs: SoC's below A8X are able to get iOS 15.8. Until 1.4 where we are iOS16+, I still want this distinction included.

So, I do agree. With 1.4 I will drop the SoC and iOS designations since they will no longer be required. But, while 1.3 is still our active version I believe that this needs to be included.

Bastinator-school:
Hey im working on the PR  mentioned by @BotBlake 
Your player documentation used to mention differences between TvOS version and ios/ipadOS. Your new docs do not seem to make this distinction does this mean they now fully overlap? 
If they do ill change the pr to reference switftfin and not switftfin(ios) or switftfin(tvos)

BotBlake:
Hi there @Bastinator-school 
Glad to see you working on the PR, thanks for that!

Swiftfin (iOS) and Swiftfin (tvOS) are still two different 'parts' of the App. 
Can you refer the part of the documentation where you think that is not clear? 

I know that various places in the jellyfin.org Repository specifically mention the difference.

Kind regards and thank you again!

BotBlake:
Oh, I am an idiot 🗿
I thought this was the PR (I am on GitHub mobile)

I will check later what is going on in this specific PR 😅
Although my initial comment stands: iOS and tvOS use different implementations in the Swiftfin app. 

Sorry for the confusion 😅

Bastinator-school:
The reason i mention it here is because the current documentation Documentation/players.md which shows codec and container support does not mention what the difference is between TvOS implementation and ios implementation. So i do not know which codecs are supported where. The players.md does mention the native vs vlc but i cannot see whether the tvos can use one or both and if it fully supports them.

JPKribs:
The players and profiles will be the same between both iOS and tvOS. The only reason these are currently different is that our tvOS version on the App Store is an older build. You can follow our progress here: https://github.com/jellyfin/Swiftfin/discussions/1294

---

The current profiles where created here: https://github.com/jellyfin/Swiftfin/pull/519. Apple doesn't really restrict codecs based on hardware but more on OS version. Since all versions of Apple TV will be able to go to tvOS 26, there is nothing iOS can do that tvOS cannot. We have one restriction for AV1 that can be enabled manually via custom device profiles that we will *likely* remove after iOS 17 is out minimum version.

Issue #1487, opened 2025-04-08
Add transcode reason icons to sessions page


sjorge:
**Describe the feature you'd like**
The sessions page currently looks like

![Image](https://github.com/user-attachments/assets/029941cd-1ffc-46c6-bbe9-371ed4f56f72)

I can easily see it's transcoding but if I want to know why I have to tab it so it expands, it would be useful to have the icons appear next to transcode. I believe there are a max of 3 (video, audio, subs) so I think they should be able to fit next to the 'Transcoding' text.

(Icons I am talking about)

![Image](https://github.com/user-attachments/assets/c52933db-ccfc-4ef2-9cf5-6c6aeebc3ddb)

**Additional context**
n/a

JPKribs:
There is a potential of 5 concurrent reasons. Realistically, 3 will show but 5 is possible:

https://github.com/jellyfin/Swiftfin/blob/main/Shared/Extensions/JellyfinAPI/TranscodeReason.swift

I'm not opposed to this as an idea but it'd likely need to go onto a second row. The space after "Transcoding" can shrink considerably with localizations + longer tracks.

I likely won't look at this until after https://github.com/jellyfin/Swiftfin/discussions/1283. 

Additionally, this looks like a prime candidate for a `Layout` so I'll probably hold off until after #1455 as well.

JPKribs:
Starting looking at this using https://github.com/jellyfin/Swiftfin/pull/1569:

Here is what I have now. Still WIP and relies on https://github.com/jellyfin/Swiftfin/pull/1569 so any changes to that will likely change how this works:

| Scenario | Normal Font | Max Font |
|----------|-------------|----------|
| **Maximum Number of Reasons** | ![Image](https://github.com/user-attachments/assets/e84612a6-159f-4dbd-aacd-4e52e6c99281) | ![Image](https://github.com/user-attachments/assets/cdb11532-7f7c-46f5-8a6d-6bfbdf8bc3a0) |
| **Standard Transcode** | ![Image](https://github.com/user-attachments/assets/09df676f-8408-4441-a521-a7e9bc9ee19c) | ![Image](https://github.com/user-attachments/assets/cea3d780-cb87-4e7e-a012-3a411f3769d1) |

sjorge:
Since it seems quite often we get an extra line, perhaps just keep the play/pause button next to the timestamp and move all the transcode stuff to a new line?

JPKribs:
[...]

PR #1614, opened 2025-06-28
`ActiveSessionsView` Cleanup

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1487

The goal of this PR is to cleanup the `ActiveSessionsView`. When I first did this, I had just wrapped up my [100 Days of SwiftUI](https://www.hackingwithswift.com/100/swiftui). Now I'm approaching my 500th day of SwiftUI, I thought I would go back and make sure this was living up to it's potential.

- [X] Add Session Filters
- [X] Cleanup the `ViewModel` to better mirror other `ViewModel`s
- [x] Add transcode reasons to the session rows

### Filters

https://github.com/user-attachments/assets/87d97e1b-fbad-45c8-b041-60c22b268799

### Transcode Reasons

<details>

<summary>Maximum Reasons - Standard Font</summary>

Simulator Screenshot - iPhone 16 Pro - 2025-06-28 at 12 43 31

</details>

<details>

<summary>Maximum Reasons - Large Font</summary>

Simulator Screenshot - iPhone 16 Pro - 2025-06-28 at 12 43 46

</details>

<details>

<summary>Standard Reasons - Standard Font</summary>

Simulator Screenshot - iPhone 16 Pro - 2025-06-28 at 12 44 37

</details>

<details>

<summary>Standard Reasons - Large Font</summary>

Simulator Screenshot - iPhone 16 Pro - 2025-06-28 at 12 44 24

</details>


JPKribs:
One outstanding issue. When the `ViewModel` updates, it takes focus out of the `Menu`. This means, if you are trying to select a filter when the refresh occurs it will kick you out back to the view:

https://github.com/user-attachments/assets/e0e45cbc-d208-4292-80fe-0baae9da077a

Three routes I am considering, from least effort to most effort:

1. Move all of the filters to a `.sheet` via the coordinator. This will roughly mirror how the `PagingLibraryView` filters are handled. 
2. Add a modifier to the `menu` to pause the timer when pressed. Then figure out a good way to re-enable the timer when deselected. Likely requires a `ViewModifier`.
3. Create an `onVisible` `ViewModifier` that performs an action when an item is visible on screen. Would also need an `onHidden` action as well.

Unofficial 4th bandaid is to just push the refresh timer to update every 10 seconds up from 5. Functionally not too different but saves us some bandwidth and *most* filtering should be doable in ~10 seconds. It's not ideal at all but this would be the **easiest** route forward.

---

This issue will also be present when we move to a WebSocket but the WebSocket should also let us choose to update every 10 seconds as well if that's the route we want to pursue.

LePips:
2 is the least intrusive solution. I already know how to get the button pressed state and new functionality in iOS 16 fixes some issues with my solution with navigation toolbar buttons.

JPKribs:
> 2 is the least intrusive solution. I already know how to get the button pressed state and new functionality in iOS 16 fixes some issues with my solution with navigation toolbar buttons.

These work perfectly! Thank you for your help knocking that out! I took your same work for the filter style and used this for `ServerActivityView` since these are going to be pretty similar. Eventually, these are two of the WebSocket subscriptions you can do so getting this menu to not fold on update is important in both places. I left the `isFiltersPresented` logic out for now since `ServerActivityView` cannot use that until we are live updating that via WebSocket but the scaffolding and the filter layouts are the same.

Screenshot 2025-07-03 at 12 24 08

Testing, this all looks good to me! I think this is ready for a second look when you get a chance!

JPKribs:
Should this filter be done in the same way as `NavigationBarMenuButtonModifier` since we're using this layout in 2 places now? Help make this easier to adjust if anything comes up in the future?

LePips:
> Should this filter be done in the same way as NavigationBarMenuButtonModifier

I'm going to say not, let's not get in the habit of making a bunch of wrapper views.

LePips (approved):

PR #1616, opened 2025-06-29
Person `ItemView`

JPKribs:
### Summary

This PR is the tvOS version of [1606](https://github.com/jellyfin/Swiftfin/pull/1606).

In additional to adding the `PersonItemView`, I took this as an opportunity to clean up some of the TODOs that we created with [1606](https://github.com/jellyfin/Swiftfin/pull/1606). Primarily, this added some functions to `BaseItemDto` as opposed to manually setting them in the `...ItemView`.

The tvOS `PersonItemView` should mostly mirror `iPadOSCinematicScrollView`. I moved tvOS's overview to an `OverviewView` to mirror iOS. 

I found that the only difference between `PersonItemViewModel` & `CollectionItemViewModel` was removing the `.boxset`. In testing, we can resolve the `.boxset` issue using the following:

```
    private func getItems(for itemType: BaseItemKind) async -> ItemLibraryViewModel {

        /// This is hacky, but an item should NEVER contain a `UserView`. Adding this fixes the boxset quirk.
        let itemTypes = (itemType == .boxSet ? [.boxSet, .userView] : [itemType])

        let viewModel = ItemLibraryViewModel(
            parent: parent,
            filters: .init(itemTypes: itemTypes),
            pageSize: 20
        )
```

I don't *love* this as a solution but this is functional. The other alternative is to mirror Web and combine `[.person, .boxset, .playlist]` and just have an `Other Items` category on these pages? We can also just add logic to skip `.boxset` for `item.type == .boxset`. This all being said, this is the *only* real difference between `PersonItemViewModel` & `CollectionItemViewModel` so in my mind it makes sense to combine these. The other motivator is I believe this is a valid approach to `.playlist` as well.

### Questions / Issues

1. iOS we are able to use `See More` to handle collections and persons with more than 20 of a specific type. What is the best way to do this for tvOS?
2. I added the `See More` back to Episodes for iOS / iPadOS. Please let me know if this was intentionally left out.
3. Keeping the same width of `ActionButton`s, the person on tvOS might be a bit jarring. Let me know if this is too large. I think for Persons, it wouldn't be terrible to shrink this all down since we know that we're only every going to have 2 buttons. For now, this is just using the default sizing from other usages/

### Video

https://github.com/user-attachments/assets/df346100-4be6-4e52-82e3-fe1e08a5d497


JPKribs:
This is done with the big exception of the `SeeMore` on tvOS. I see there is a TODO for `TrailingContent` in `PosterHStack` for tvOS. I can work on that but I also don't know what `TrailingContent` tvOS should look like?

I have a _rough_ working version that looks like this:

Screenshot 2025-06-29 at 13 29 48

But I don't know if this is ideal. This roughly mirrors what exists in 1.0.1 on tvOS and I'm sure there is a good reason that this was removed.

Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto.swift
```diff
@@ -264,6 +264,28 @@ extension BaseItemDto {
         originalTitle != displayTitle ? originalTitle : nil
     }
 
+    /// Can this `BaseItemDto` be played
+    var presentPlayButton: Bool {
+        switch type {
+        case .aggregateFolder, .basePluginFolder, .boxSet, .folder, .genre, .musicGenre, .person, .photo, .photoAlbum, .studio,
```
LePips: It would be easier to grep these values if they were flipped: listing values explicitly that are true instead.

Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto.swift
```diff
@@ -264,6 +264,28 @@ extension BaseItemDto {
...
+            return false
+        default:
+            return true
+        }
+    }
+
+    /// Can this `BaseItemDto` be mark as played
+    var canBePlayed: Bool {
+        switch type {
+        case .aggregateFolder, .basePluginFolder, .folder, .genre, .musicGenre, .person, .photo, .photoAlbum, .studio, .userRootFolder,
```
LePips: Same here.

Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto.swift
```diff
@@ -291,4 +313,15 @@ extension BaseItemDto {
             nil
         }
     }
+
+    /// Does this `BaseItemDto` have `Genres`, `People`, `Studios`, or `Tags`
+    var hasComponents: Bool {
+        switch type {
+        case .aggregateFolder, .basePluginFolder, .channel, .folder, .genre, .liveTvChannel, .musicGenre, .person, .photo, .photoAlbum,
```
LePips: Same here.

Shared/ViewModels/ItemViewModel/TypeItemViewModel.swift
```diff
@@ -11,29 +11,19 @@ import Foundation
 import JellyfinAPI
 import OrderedCollections
 
-final class PersonItemViewModel: ItemViewModel {
+final class TypeItemViewModel: ItemViewModel {
```
LePips: Looking at this again this is just a simple wrapper around an `ItemTypeCollection`. We can instead move the `randomItem` to `ItemTypeCollection` and remove this class.

Swiftfin tvOS/Views/ItemView/Components/OverviewView.swift
```diff
@@ -0,0 +1,84 @@
...
+                        .lineLimit(taglineLineLimit)
+                }
+
+                if let itemOverview = item.overview {
+                    Text(itemOverview)
+                        .font(.subheadline)
+                        .lineLimit(overviewLineLimit)
+                }
+
+                if let birthday = item.birthday?.formatted(date: .long, time: .omitted) {
```
LePips, line 26: This looks unbalanced with the overview above these values, let's move them above the overview/tagline.

Swiftfin tvOS/Views/ItemView/PersonItemContentView.swift
```diff
@@ -0,0 +1,44 @@
...
+        private var router
+
+        @ObservedObject
+        var viewModel: TypeItemViewModel
+
+        var body: some View {
+            VStack(spacing: 0) {
+
+                ItemView.CinematicHeaderView(viewModel: viewModel)
+                    .frame(height: UIScreen.main.bounds.height - 150)
```
LePips: While we can get away with `UIScreen` usages a bit more on tvOS in layouts, let's avoid them going forward.

Shared/ViewModels/ItemViewModel/TypeItemViewModel.swift
JPKribs: To continue using this on the `ItemView`, this would still need to conform to `ItemViewModel`, correct? This is what I currently have for this. All functional and working on iOS and tvOS but I just wanted to make sure since I moved `randomItem` to `ItemTypeCollection` but am using the `TypeItemViewModel` since this conforms to `ItemViewModel` now. Please let me know if this should be called `ItemTypeCollection` instead:

```
//
//
//

import Combine
import Foundation
import JellyfinAPI
import OrderedCollections

final class TypeItemViewModel: ItemViewModel {

    // MARK: - Published Properties

    @ObservedPublisher
    var sections: OrderedDictionary<BaseItemKind, ItemLibraryViewModel>

    @Published
    private var elements: OrderedDictionary<BaseItemKind, ItemLibraryViewModel> = [:]

    // MARK: - Private Properties

    private let itemTypes: [BaseItemKind]
    private var loadTask: AnyCancellable?

    // MARK: - Init

    override init(item: BaseItemDto) {
        self.itemTypes = BaseItemKind.supportedCases
            .appending(.episode)
            .appending(.person)

        /// Temporarily initialize with empty value; will override after `super.init`
        self._sections = ObservedPublisher(wrappedValue: [:], observing: Just([:]).eraseToAnyPublisher())

        super.init(item: item)

        /// Now safe to bind to $elements
        self._sections = ObservedPublisher(
            wrappedValue: self.elements,
            observing: self.$elements
        )
    }

    // MARK: - Action Handling

    override func respond(to action: Action) -> State {
        switch action {
        case .refresh, .backgroundRefresh:
            loadSections()
        default:
            break
        }

        return super.respond(to: action)
    }

    // MARK: - Section Loading

    private func loadSections() {
        loadTask?.cancel()

        loadTask = Task {
            let newElements = await fetchSections()

            await MainActor.run {
                self.elements = newElements
            }
        }
        .asAnyCancellable()
    }

    private func fetchSections() async -> OrderedDictionary<BaseItemKind, ItemLibraryViewModel> {
        await withTaskGroup(of: (BaseItemKind, ItemLibraryViewModel).self) { group in
            for kind in itemTypes {
                group.addTask {
                    await (kind, self.loadItems(for: kind))
                }
            }

            let result = await group.reduce(
                into: OrderedDictionary<BaseItemKind, ItemLibraryViewModel>()
            ) { dict, entry in
                let (kind, viewModel) = entry
                if case .content = viewModel.state, viewModel.elements.isNotEmpty {
                    dict[kind] = viewModel
                }
            }

            return result.sortedKeys(using: \.rawValue)
        }
    }

    private func loadItems(for kind: BaseItemKind) async -> ItemLibraryViewModel {
        let typesToQuery = kind == .boxSet ? [.boxSet, .userView] : [kind]

        let vm = ItemLibraryViewModel(
            parent: item,
            filters: .init(itemTypes: typesToQuery),
            pageSize: 20
        )

        return await withCheckedContinuation { continuation in
            var cancellable: AnyCancellable?

            cancellable = vm.$state
                .filter { $0 != .initial && $0 != .refreshing }
                .sink { _ in
                    cancellable?.cancel()
                    continuation.resume(returning: vm)
                }

            Task { @MainActor in
                vm.send(.refresh)
            }
        }
    }

    // MARK: - Random Item Access

    func randomItem() -> BaseItemDto? {
        if elements.elements.count == 1 {
            return elements.elements.first?.value.elements.first
        }

        return elements
            .elements
            .shuffled()
            .filter { $0.key != .episode }
            .randomElement()?
            .value
            .elements
            .randomElement()
    }
}
```

LePips: Oh whoops, keep it as-is from before then. However, rename it back to `CollectionItemViewModel` just because I don't want `ItemType_` and `TypeItem_`. It's fine if people use it as well.

Swiftfin tvOS/Views/ItemView/PersonItemContentView.swift
JPKribs: Right now, all of the tvOS `...ItemContentView` are using that same `UIScreen.main.bounds.height - 150`. Would the best solution be to create a tvOS version of the `OffsetScrollView` and use that for all tvOS `...ItemContentView`s? If so, should I make that change in this PR or hold off and move that work to another PR?

I did a quick try at removing these `UIScreen.main.bounds.height - 150` but these are all nested in the `CinematicScrollView` so the sizing is all weird to do this from the `...ItemContentView` and would likely require we figure this out from the `CinematicScrollView` and all downstream `...ItemContentView` would inherit that spacing.

Haven't fully wrapped my head around what this would look like but I'm happy to look into this more if we want this as part of this PR. Otherwise, I can look at this at a later date for another PR.

JPKribs:
**Current State**

https://github.com/user-attachments/assets/90723d08-fa2c-406b-b2d4-f8743759f511

Re-ordered the `OverviewView` for tvOS and cleaned up the logo UIScreen. While I was there, if the logo goes to text I moved that down to 1 line and used the new `Marquee` text. See above.

With a death date:

Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-07-22 at 15 44 42

LePips (approved):
I took advantage of the fact that people won't have other details like other views so we could reuse the same collection item view.

Issue #1478, opened 2025-04-06
[tvOS] (master) Library name stays and covers posters in Media -> Library view


sjorge:
### Describe the bug

(Reference: https://github.com/jellyfin/Swiftfin/discussions/1294#discussioncomment-12742078)

1. Select 'Media' on the menu/tab bar
2. Select a library
3. Scroll down a bit...

![](https://private-user-images.githubusercontent.com/379665/430717172-cfe96a71-da27-4141-a07a-80490664eadd.jpg?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NDM5NjIxNDQsIm5iZiI6MTc0Mzk2MTg0NCwicGF0aCI6Ii8zNzk2NjUvNDMwNzE3MTcyLWNmZTk2YTcxLWRhMjctNDE0MS1hMDdhLTgwNDkwNjY0ZWFkZC5qcGc_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjUwNDA2JTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI1MDQwNlQxNzUwNDRaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT1lNjYzZGM5OThiMTgzZGRjZjk2NmFkZDczOTM2NDFiYTA3YzVjNjIxNzlkYzBkMjE2MjgxMjc0NzMwNzM4ZWYyJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCJ9.Ch0aZT99Cc0fkxfIHwehBPicxK4eoDSMVFo4Ig81A3Y)

### Application version

master

### Where did you install the app from?

None

### Device information

Apple TV 4K (Ethernet)

### OS version

tvOS 18.3.1

### Jellyfin server version

10.10.7

Issue #1105, opened 2024-06-13
Incorrect Size of Episode's Poster


fredmorais:
### Describe the bug

Loving this new update! Thank you for all the work.

1. Open a TV Show
2. When the first season loads, the posters are correctly sized
3. When I change to another season (2, 3, etc) the posters are oversized. This is fixed if I go back and choose another TV show or if I rotate the iPhone (also tested on iPad).
4. You can see this in the images below. The first image shows everything working fine on Season 1 and the second picture shows the over sizing problem.

![IMG_6046](https://github.com/jellyfin/Swiftfin/assets/40044095/d443706d-f5c2-47ed-b896-8743af569386)
![IMG_6047](https://github.com/jellyfin/Swiftfin/assets/40044095/27718fcb-c6e7-40d8-895c-7dedd96c4315)

### Application version

1.1 (78)

### Where did you install the app from?

App Store

### Device information

iPhone 15 Pro

### OS version

iOS 17.5.1

### Jellyfin server version

10.9.6

darrekk:
Might be a dupe of [this issue](https://github.com/jellyfin/Swiftfin/issues/1101).

JPKribs:
Can we confirm if this issue is still present on version 1.3? This is our current App Store version.

Issue #1031, opened 2024-04-19
Finalize Poster Sizing Design


LePips:
The new poster sizing for horizontal stacks and library views will use concrete column amounts (iOS, tvOS) or minimum sizes (iPadOS). I chose "good enough" values for everything but don't have a central design philosophy to back it up. Additionally, this may want to be customizable or change based on accessibility options.

For iOS:
- The 1.5/2/3 seem to be good enough however I didn't test much on smaller screen sizes so things might be too small. ~~On the other hand, for Plus models we may want to do smaller items (eg: 4 column for portrait)~~

For iPadOS:
- Again the minimum sizes just seemed good enough but there is a larger variation in screen sizes. There are also some views that use concrete columns and it may look weird. Additionally, landscape vs portrait screen orientation needs to be considered.

For tvOS:
- Easier since we would just need to pick a single design and clean it up.

See what other apps do on different screen sizes and accessibility options. This would also help a lot when other media gets added.

PR #1617, opened 2025-06-29
`PosterButton` refactoring and zoom transitions

LePips:
- Implements the `zoom` transition on iOS 18+. While the API is available on tvOS, it has no effect.
- `NavigationRoute` can now route with ad-hoc transitions.
- Refactors `PosterButton` and relevant views.
- Introduces item-environment bound `contextMenu` creation, allowing us to set a global context menu registry so all matching context menu targets could get the relevant context menu.
- Use `FocusedValue` for focused posters.
- tvOS tab bar now scrolls along with libraries.

There are a few noticeable bugs:
- The navigation bar will be in the incorrect appearance after dismissing an item view.
- Sometimes tapping on the screen after dismissing from a zoom transition will tap a view on the previously dismissed view.
- Layout issue with the item view on appearing.


JPKribs:
Doing some testing now! I wanted to check before I linked it but does this resolve the following issues:

https://github.com/jellyfin/Swiftfin/issues/1031
https://github.com/jellyfin/Swiftfin/issues/1105
https://github.com/jellyfin/Swiftfin/issues/1478

LePips:
This work was not intended to fix those issues.

- For #1031, I need to take into consideration the horizontal size class as well as investigate observing accessibility settings
- For #1105, I'm unsure since I can't reproduce.
- For #1478, this is not the same fix as the tab bar. 

> could we just remove the titles on those views?

Maybe, but navigation on tvOS overall needs a different solution that I have in mind.

JPKribs (approved):
Makes sense! With that in mind, this looks good!

PR #1623, opened 2025-07-04
[iOS & tvOS] Fetch correct server from SwiftfinStore when updating server info

### LINKED ISSUE #1619: Server names cannot being identified when add multiple server.
### Describe the bug

1. Swiftfin platform : iPhone XR iOS 18.5
2. Symptom:
  It is difficault to select right server when I added multiple servers because the server names seem to be  same.
4. Steps to reproduce :- Firstly, login to a Jellyfin server "http://server1:8096".
- Click Person icon at upper right corner at Home.
- Click Change User button
- Click server name at bottom of screen.
- Click "Add Server +" menu.
- Add a new Jellyfin server url like "http://server2:8096" and click "Connect".
- Take into user log-in to.  No matter what user login any user of these above two server.
- After then, you can see those SERVER NAMES LOOK AS SAME when you are trying to select one of the servers.

### Application version

1.3 (14)

App Store

### Device information

iPhone XR

### OS version

iOS 18.5

### Jellyfin server version

10.10.7

samglt:
When ServerState objects get updated info from the remote server, the ServerModel fetched from SwiftfinStore didn't have its id specified, so updateServerInfo would always update the one server with the info from whatever server Swiftfin most recently updated its info for. This doesn't have any effect when Swiftfin is only connected to one server, but when another server is added, it causes glitches in the server selection screen because two servers will sometimes share an ID.

Fixes #1619

### DIFF
--- a/Shared/SwiftfinStore/SwiftfinStore+ServerState.swift
+++ b/Shared/SwiftfinStore/SwiftfinStore+ServerState.swift
@@ -81,7 +81,7 @@ extension ServerState {
 
     func updateServerInfo() async throws {
         guard let server = try? SwiftfinStore.dataStack.fetchOne(
-            From<ServerModel>()
+            From<ServerModel>().where(Where(\.$id == id))
         ) else { return }
 
         let publicInfo = try await getPublicSystemInfo()


JPKribs:
I can attempt to test this but I haven't personally recreated this without duplicate servers. I will likely be leaning on @LePips and their expertise in this area as my access to multiple servers is limited. If I can recreate the original issue I will comment. Otherwise, if I continue to be unsuccessful on that front, I'll rely on area expertise.

Issue #1630, opened 2025-07-09
UX: Movie/TV Info: Tags under play button


yanbreu:
### Describe the bug

It looks like in some cases the movie tags like ratings, hdr, 5.1,… go under the play button. I only found this problem with this movie all other seem to work correctly.

Expected: No line break / all tags visible

<details><summary>iPhone View (Broken)</summary>
<p>

<img width="296" height="640" alt="Image" src="https://github.com/user-attachments/assets/fcb77c31-51b2-4d81-92a4-c27dedba0121" />

</p>
</details> 
<details><summary>iPad View (Working)</summary>
<p>

<img width="640" height="447" alt="Image" src="https://github.com/user-attachments/assets/c505cce0-d0a6-4579-8e80-be50794bdaaa" />

</p>
</details> 

### Application version

1.4(11)

### Where did you install the app from?

TestFlight

### Device information

iPhone 13 Pro Max

### OS version

iOS 18.5

### Jellyfin server version

10.10.7

PR #1632, opened 2025-07-09
Resolve `FlowLayout` Height Calculations

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1630

Using the `FlowLayout`, the height could be calculated incorrect when moving to 2 rows. This resulted in certain scenarios where the height provided by the `FlowLayout` was still only reserving 1 row even though it would push items into the second row. This would occur when there was "enough" room to fit everything into a single row for the whole screen (used to calculate the height) but there was NOT enough room to fit everything into a single row for the set width (when the `maxWidth` was set from the parent) available.

In testing, this caused the second row to slip under other UI elements since SwiftUI didn't allocate enough vertical space.

### Resolution

Fixed the height calculation to use the actual proposed width (not infinity) and ensured the cache properly recalculates the total size when bounds change. Now the layout correctly communicates its height requirements to SwiftUI regardless of parent constraints.


chickdan (approved):

JPKribs:
Moving this back to draft. I am still able to recreate this in some scenarios. Taking this back under review and I'll be back.

LePips (approved):

PR #1634, opened 2025-07-10
[tvOS] Add Marquee to handle long season names

samglt:
This adds [a package](https://github.com/samglt/Marquee) for marquee text and uses it to handle long season titles, which previously could take up all the horizontal space in tvOS. Related to the #1503 help wanted item.

Before:
before

After:
https://github.com/user-attachments/assets/f68195dd-db0a-4262-8723-ae71e54c7924


JPKribs:
The package seems fairly small. What are the advantages of an external package as opposed to using a more embedded solution? I only ask because this does create a new external dependency for Swiftfin. It's not the end of the world to add a new package but if it's small enough it would be nice to have more ownership of the Marquee text so we can make adjustments if we need anything custom.

I don't imagine this is going to update that frequently so I'm not too concerned about having to update/change this usage. I'm happy either way but I'm curious if you had any opinions on this as a direction?

samglt:
If you want it embedded it could be pared down even more, since there's a decent amount of logic that's not needed for just scrolling overflow text. Would save me from being a bottleneck for any issues down the line as well.

JPKribs:
> If you want it embedded it could be pared down even more, since there's a decent amount of logic that's not needed for just scrolling overflow text. Would save me from being a bottleneck for any issues down the line as well.

I think would be ideal just so we have a bit more ownership over this. Especially for any 'custom' customizations we may need. I think for us, there is some tvOS specific logic we will want (which you already added in this PR) for things like focus. While on the iOS side, we are likely already set. Our main (only?) usage of this would be scrolling `PlayButton` text when displaying a selected version out of multiple version options. There is an argument to be made that Marquee Text could be preferable for content whose title exceeds the line limit as opposed to cutting off like "Content Title: Subti...". But for now I think this is more valuable to focus on the known use cases then we/I can test out other options for this after we have our main use cases addressed! 

I am normally in favor of just using a package. However, for items where this is manageable enough for us to have our own version, I think that is preferable. Just so we're not pestering you/package maintainers for Swiftfin specific use cases & changes.

I'll defer to @LePips on this but overall I am happy with how this looks and I have no issues in my testing so far!

LePips (changes requested):
Thank you for working on this and I've seen your changes to the package for focus handling.

I think we can take the current package, slimming it down with the features we don't need, and have the following tweaks:

- A gradient-mask is applied on the leading/trailing edges. This masked gradient would also extend beyond the text's maximum bounds (like to the internal edge of the season button).
- Text isn't truncated if it's too long, instead being overflowed into the trailing gradient mask.

Another marquee solution that has those points is [Turbocharger's MarqueeText](https://github.com/nathantannar4/Turbocharger/blob/main/Sources/Turbocharger/Sources/View/MarqueeText.swift) but I'm not proposing we use it. It is a good reference thought to achieve the final implementation.

samglt:
I moved Marquee into shared components and added that gradient to the sides, and made it overflow when not active. 

https://github.com/user-attachments/assets/5c3b1949-a005-4532-9ab8-af40de40779a

JPKribs:
@samglt If you merge on Main now, that should clear up the linting errors.

samglt:
Got rid of the env variables, was one of those things you don't really think about just copy pasting stuff over. I also got rid of the () -> View & LocalizableStringKey inits since it looks like we're only passing strings in, both in this PR and in the PlayButton.

In the SeasonHStack I picked a 2s delay before it starts moving.

LePips (approved):

PR #1635, opened 2025-07-13
Only log failed CoreStoreLogger assertions

samglt:
Currently the console is full of critical logs from CoreStore because assertions are logged whether it's true or false. This just wraps the logger call so it only logs places where the assertion actually fails.

For reference, [here's a call to assert inside CoreStore](https://github.com/JohnEstropia/CoreStore/blob/104def4611b8f8c2067903b589d01d198d51e6d6/Sources/DataStack%2BQuerying.swift#L145).

### DIFF
--- a/Shared/Logging/SwiftfinCorestoreLogger.swift
+++ b/Shared/Logging/SwiftfinCorestoreLogger.swift
@@ -55,16 +55,15 @@ struct SwiftfinCorestoreLogger: CoreStoreLogger {
         lineNumber: Int,
         functionName: StaticString
     ) {
-        if !condition() {
-            logger.critical(
-                "\(message())",
-                metadata: nil,
-                source: "Corestore",
-                file: fileName.description,
-                function: functionName.description,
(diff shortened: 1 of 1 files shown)


Shared/Logging/SwiftfinCorestoreLogger.swift
```diff
@@ -55,14 +55,16 @@ struct SwiftfinCorestoreLogger: CoreStoreLogger {
...
     ) {
-        logger.critical(
-            "\(message())",
-            metadata: nil,
-            source: "Corestore",
-            file: fileName.description,
-            function: functionName.description,
-            line: UInt(lineNumber)
-        )
+        if !condition() {
```
LePips: Use a `guard` instead.

LePips (changes requested):
Apologies, meant to request previous comment and not approve.

LePips (approved):

LePips (approved):

PR #1636, opened 2025-07-13
[tvOS & iOS] Fix navigation to server edit screen

### LINKED ISSUE #1221: Unable to delete server on tvOS
### Describe the bug

Unlike in the iOS version when I long-press a server in the server list in tvOS, I don't seem to be presented with an option to remove the server from the list.

The steps I tried were as follows:
1. Click the settings (gear icon)
2. Click on switch user
3. Click the server
4. Long-press to remove the user
5. Go back to server list
6. Long-press to remove the server

### Application version

1.0.1 (70)

App Store

### Device information

Apple TV 4K 64GB (3rd generation)

### OS version

tvOS 17.6.1 (21M80)

### Jellyfin server version

10.9.7

samglt:
Re-fixes #1221 after a regression.

The `editServerFromSelectUser` route sets the `.isEditing` env variable true and allows server deletion from the edit view.

### DIFF
--- a/Swiftfin tvOS/Views/SelectUserView/Components/ServerSelectionMenu.swift
+++ b/Swiftfin tvOS/Views/SelectUserView/Components/ServerSelectionMenu.swift
@@ -81,7 +81,7 @@ extension SelectUserView {
                 Section {
                     if let selectedServer {
                         Button(L10n.editServer, systemImage: "server.rack") {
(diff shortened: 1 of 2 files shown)


JPKribs (approved):
Tested and working the same now as 1.3! Thank you for fixing this!

PR #1638, opened 2025-07-13
Fix `ItemView` Layout Issues for Uncached Images

JPKribs:
### Summary

Resolves the Layout Issue mentioned here: https://github.com/jellyfin/Swiftfin/pull/1617
> Layout issue with the item view on appearing

Fixes the `ItemView` loading in with all of the elements at the top of the screen then having those elements move down after the background image loads in. This issue is most apparent on iPadOS. I believe this is because we are requesting the image at a larger size so the loading occurs slower and this this adjustment is more visible. Additionally, since there is more screen, the movement of the items themselves is also just larger as the pixel distance from the top of the screen to the middle is larger. See below for an example of this:

https://github.com/user-attachments/assets/61b1f4d1-2a9b-4836-a0e2-c9fa5e513644

To resolve this, I updated the `OffsetScrollView` to appropriately reserve the area at the top of the view. This should occur even if the image is not loaded. This now looks like this:

https://github.com/user-attachments/assets/9e2c2eec-11f8-4f8e-92fa-040358ebef88

The original issue is common but not consistent. I've found this applies more to larger background images. Setting at image to ~10mb causes this issue every time. In my original example, 3rd Rock From The Sun has a huge 11mb image that creates this issue every time. 30 Rock has a more moderate 1mb image where this adjustment occurs on occasion if I am on slower networks. 

With this PR, I am unable to recreate this issue, even with artificially slowed network speeds. I'd appreciate a second set of eyes since that isn't a very 'official' way of testing this. This being said, I believe this resolves the issue but please let me know if there is a better way to go about reserving this area.


Swiftfin/Views/ItemView/Components/OffsetScrollView.swift
```diff
@@ -57,15 +57,17 @@ extension ItemView {
...
+                    if size.height > 0 {
+                        overlay
+                            .frame(height: headerHeight, alignment: .bottom)
+                            .overlay {
+                                Color.systemBackground
+                                    .opacity(headerOpacity)
+                            }
+                    } else {
+                        Color.clear
+                            .frame(height: UIScreen.main.bounds.height * heightRatio)
```
LePips: We shouldn't continue using `UIScreen`. I believe we should use `AlternativeLayoutView` here instead. If that doesn't work, there's a different layout issue present.

JPKribs: > We shouldn't continue using `UIScreen`

That's embarrassing I literally just made a PR to get rid of `UIScreen` in the `ItemView` 🤦‍♂️... 

I've removed `UIScreen` and moved us over to `AlternateLayoutView`. This looks correct to me. On Simulator this seems to chug a little but on the real iPad this seems fine. Positioning still appears correct with this change!

JPKribs:
Changes made! 

~I caught this PR up with Main and now I have some Linting errors from unrelated `ViewModels` I'm happy to clear those out but I wasn't sure I should since those are not related to this PR nor have they been changed in this PR.~

Re-running the lint fixed this. This can be ignored.

JPKribs:
> The alternative layout should be wrapping the overlay, not the entire scroll view.

You're right, sorry about that. This is resolved now.

PR #1639, opened 2025-07-13
Cleanup iPadOS `ItemView` TODOs

JPKribs:
### Summary

Drops usages of `UIDevice.main.bounds` in favor of `GeometryReader`. Additionally, changes the trailing 200 padding between the `Overview` and `PlayButton` and with a more dynamic 5% of total screen size. This reduces the effective spacing on iPad when vertical and helps ensure that the spacing isn't too large.

While working on this, I found that the Attributes almost always used 2 lines on the iPad Mini. Additionally, the `Genre` text would often get pushed to 2 lines on iPad Mini. I changed the `Genre` text & others to always be on a single line to avoid issues. Altogether, I wanted to cleanup how the current version of this looked on the iPad Mini. This is how this looked prior to this PR:

Simulator Screenshot - iPad mini (A17 Pro) - 2025-07-13 at 03 01 27

I've included various versions of how this looks now in this PR on both the iPad Mini (7.9") and iPad Pro (13") below for comparison.

### Outstanding Questions

1. I moved this calculation from `UIDevice.main.bounds` to `GeometryReader`. In my research, this **looked** like the correct move. Please let me know if this is not.
2. In my research, I also found `@Environment(\.horizontalSizeClass)` and `@Environment(\.verticalSizeClass)`. Right now, even the iPad Mini positioned vertically returns `.regular` so there's no real value I saw for this PR to use this. However, if we start allowing split screen for iPad, `@Environment(\.horizontalSizeClass)` can start to be used and will actually return `.compact` when split screen or in the side bar thing. Should I include this in this PR as a future looking item or should I hold off since there isn't a good way to test this?

### Screenshots

<details>

<summary>iPad Mini 7.9"- Vertical</summary>

Simulator Screenshot - iPad mini (A17 Pro) - 2025-07-13 at 02 53 46

</details>

<details>

<summary>iPad Mini 7.9" - Horizontal</summary>

Simulator Screenshot - iPad mini (A17 Pro) - 2025-07-13 at 02 53 50

</details>

<details>

<summary>iPad Pro 13" - Vertical</summary>

Simulator Screenshot - iPad Pro 13-inch (M4) - 2025-07-13 at 02 54 47

</details>

<details>

<summary>iPad Pro 13" - Horizontal</summary>

Simulator Screenshot - iPad Pro 13-inch (M4) - 2025-07-13 at 02 55 00

</details>


alexmomotiuk (approved):

Swiftfin/Views/ItemView/ScrollViews/iPadOSCinematicScrollView.swift
```diff
@@ -105,83 +98,92 @@ extension ItemView.iPadOSCinematicScrollView {
...
-                                if let playButtonitem = viewModel.playButtonItem, let runtime = playButtonitem.runTimeLabel {
-                                    Text(runtime)
+            GeometryReader { geometry in
+                HStack(alignment: .bottom) {
+
+                    VStack(alignment: .leading, spacing: 20) {
+
+                        ImageView(viewModel.item.imageSource(
+                            .logo,
+                            maxWidth: geometry.size.width * 0.4,
```
LePips: For the image size, I would like us to choose a static size so that only a single image is cached and used in the event of resizing as the image cache key is currently determined by the max width. We can choose something that "looks good enough".

LePips (changes requested):
> @Environment(\.horizontalSizeClass) ...@Environment(\.verticalSizeClass)

We will eventually use the size classes more, like for determining the poster column counts.

Swiftfin/Views/ItemView/ScrollViews/iPadOSCinematicScrollView.swift
JPKribs: Sounds good. I removed the width from the API call to prevent multiple caches. Then, I just left the same width calculation on the view's `.frame` with the existing `.fit` aspect ratio. This should enforce the 40% of max available width as a limitation without caching logos at various sizes.

JPKribs: Example images below: 

Mini:
Simulator Screenshot - iPad mini (A17 Pro) - 2025-07-17 at 12 38 00
Simulator Screenshot - iPad mini (A17 Pro) - 2025-07-17 at 12 38 05

Pro:
Simulator Screenshot - iPad Pro 13-inch (M4) - 2025-07-17 at 12 39 16
Simulator Screenshot - iPad Pro 13-inch (M4) - 2025-07-17 at 12 39 19

JPKribs:
Last change while working on https://github.com/jellyfin/Swiftfin/pull/1638.

I found that 0.6 for iPad was blurring too much. I changed this to 0.5. See below as before and after. Let me know if this is unwanted and I can revert this:

Screenshot 2025-07-17 at 21 10 25

LePips (approved):

PR #1646, opened 2025-07-15
[iOS] CinematicScrollView: Replace UIScreen bounds with GeometryReader

samglt:
Part of #1503's task replacing uses of UIScreen bounds.

The overall appearance is not significantly changed for either backdrop or poster images, and this incidentally fixes backdrop images being too low res and poster images extending too far down and obscuring the genre/year text.

<details><summary>Before & After (Backdrop)</summary>

</details>

<details><summary>Before & After (Poster)</summary>

</details>

### DIFF
--- a/Swiftfin/Views/ItemView/ScrollViews/CinematicScrollView.swift
+++ b/Swiftfin/Views/ItemView/ScrollViews/CinematicScrollView.swift
@@ -42,13 +42,15 @@ extension ItemView {
 
             let bottomColor = viewModel.item.blurHash(for: imageType)?.averageLinearColor ?? Color.secondarySystemFill
 
-            ImageView(viewModel.item.imageSource(
-                imageType,
-                maxWidth: UIScreen.main.bounds.width
-            ))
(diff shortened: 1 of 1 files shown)


samglt:
Looking again, this isn't quite right. The images are actually oversized because the GeometryReader proxy initially has zero size so the imageSource gets requested with a zero (unlimited) maxWidth.

samglt:
Now it only gets the imageSource when the GeometryReader's size is known, so the resolution is appropriate to the screen, and will be correct for both backdrop and poster images.

I've also realized that the height multiplier should be 0.8 to get the exact same appearance as before, but I think both the backdrop and poster tend to look better at 0.6

LePips (approved):
Thanks for taking a look at this. Having a static image size as [I comment on iPadOS](https://github.com/jellyfin/Swiftfin/pull/1639#discussion_r2213974298) isn't much of a concern here since this is iOS.

Taking a look at all of this again after so long makes me realize I may just combine what is currently the compact logo and cinematic views into a single view, and the user can then either make the image "larger" (what is currently cinematic), and still hide the logo.

Issue #1618, opened 2025-07-02
`PagingLibraryView` Request Smaller Images


JPKribs:
**Describe the feature you'd like**
Doing some profiling, I was finding that the reason my `PagingLibraryView` scrolled poorly was due to the size of my poster images (PNG). Swiftfin would request the image at it's full size and the result was me trying to load in 25x 10MB files at a time. 

I went to `Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto+Images.swift` and added the following change to cap the quality of the image to 90 and the performance increase was dramatic:

```
    private func _imageURL(
        _ type: ImageType,
        maxWidth: CGFloat?,
        maxHeight: CGFloat?,
        itemID: String,
        requireTag: Bool = true
    ) -> URL? {
        let scaleWidth = maxWidth == nil ? nil : UIScreen.main.scale(maxWidth!)
        let scaleHeight = maxHeight == nil ? nil : UIScreen.main.scale(maxHeight!)

        let tag = getImageTag(for: type)

        guard tag != nil || !requireTag else { return nil }

        // TODO: client passing for widget/shared group views?
        guard let client = Container.shared.currentUserSession()?.client else { return nil }

        let parameters = Paths.GetItemImageParameters(
            maxWidth: scaleWidth,
            maxHeight: scaleHeight,
>>>      quality: 90,
            tag: tag,
            format: type == .logo ? .png : nil,
        )

        let request = Paths.getItemImage(
            itemID: itemID,
            imageType: type.rawValue,
            parameters: parameters
        )

        return client.fullURL(with: request)
    }
```

The result of this was a decrease in my image from 10MB to ~250KB. There was some additional gains by requesting these images in JPG or WebP as well:

<img width="379" alt="Image" src="https://github.com/user-attachments/assets/73b9479d-a12a-4d60-8611-375999178d8c" />

I know that this isn't the correct place for that code as this downgrades ALL image quality. On the `ItemView`, I think having the full item image would be ideal. This change is best only on for like `ImagePosterSource` or something like that where we exclusively use this for the `PosterButton`. The other idea I had was a `Performance Mode` toggle that lowered image quality?

**Additional context**
Nuke is able to use the WebP image format as well. The performance difference between WebP and JPG in my testing is fairly small. If the source image was a JPG and a WebP was requested, the image size increased. The same was true when the source image was a WebP and a JPG was requested.

[...]

PR #1648, opened 2025-07-17
Limit Poster Image Quality

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1618

This PR adds a `quality` parameters to all `Poster` image source funcs. I've used this new func to set image quality to 70 for:

1. `PosterButton`s
2. Small width images (110, 70, 30) - Mostly found in the Admin Dashboard for things like People and the Active Sessions
3. `PagingLibraryView` list rows

I chose 70 because I cannot personally tell the difference in these locations as they are smaller and not the focus of the screen. I wouldn't expect to use this for `ItemView` backgrounds or anything like that but I believe this is valuable for locations where there are many of the same image type loaded in to better reduce bandwidth. `PagingLibraryView` being the big one.

Please let me know if there is anything I should do to approach this differently or if we want to try a different value besides 70. I have some testing in https://github.com/jellyfin/Swiftfin/issues/1618 folks can follow but it's kind of inconclusive. If the image does not reduce it's dimensions as part of the call, we can see some variance where images are made larger by increase quality. However, when I request the images with a set `maxWidth`, lower quality **ALWAYS** results in a smaller file size.

For this reason, I believe this PR should be a net benefit. However, we should probably not use this quality option for images without a set `maxWidth` or `maxHeight` as the resulting file may be larger.


LePips (approved):
One thing that wasn't considered in your testing was the change in sizes that we request. Lowering the quality will not always result in a lower file size simply due to how JPG compression works with artifacts and we really don't need to squeeze every byte out.

I've just chosen a baseline of 90 which cuts down a resized poster in about half in most cases.

PR #1649, opened 2025-07-18
Add Now Playing info center and remote control support for Swiftfin player

yuth:
- Add Now Playing info center integration with rich metadata
- Implement skip forward/backward commands with 15-second intervals
- Add timeline scrubbing support via changePlaybackPositionCommand
- Enable episode navigation with previous/next track commands for TV series
- Set rich media information including title, series, season, and episode details
- Load and display artwork from Jellyfin server with proper API authentication
- Real-time playback state updates for Control Center and Lock Screen
- Fix missing remote control functionality for iPhone users

The implementation includes:
- MPNowPlayingInfoCenter integration for rich media display
- Remote command handling for skip, scrub, play/pause, and navigation
- Async artwork loading with authentication via API key query parameter
- Episode-aware navigation that enables/disables based on adjacent episodes
- Real-time updates of playback position and state changes

PR #1650, opened 2025-07-18
fix: improve native player error handling and user feedback

yuth:
__Problem__

The native video player was crashing when attempting to play videos due to improper handling of media loading failures. The app would crash instead of gracefully handling errors like:
  - Invalid or unsupported video formats
  - Network connectivity issues
  - Server configuration problems
  - Corrupted media files

__Root Cause__

  The player was attempting to play videos immediately without:
  1. Properly loading the media asset first - trying to play before confirming the video was actually loadable
  2. Monitoring for playback failures - no error detection when videos failed to load
  3. Validating media compatibility - not checking if the video format was supported before 

__Solution__
  - Safe asset loading: Wait for video properties to load before attempting playback
  - Error detection: Monitor player status to catch loading failures before they cause crashes
  - Graceful error handling: Show user-friendly error messages instead of crashing
  - Format validation: Check if videos are playable before starting playback

### DIFF
--- a/Shared/ViewModels/VideoPlayerManager/VideoPlayerManager.swift
+++ b/Shared/ViewModels/VideoPlayerManager/VideoPlayerManager.swift
@@ -118,6 +118,7 @@ class VideoPlayerManager: ViewModel {
         }
 
         if newState == .stopped || newState == .ended {
+            hasSentStart = false
             sendStopReport()
         }
     }
@@ -156,7 +157,7 @@ class VideoPlayerManager: ViewModel {
                     // 3
                     previousItem = items[0]
                 }
-            } else {
+            } else if items.count >= 3 {
                 nextItem = items[2]
                 previousItem = items[0]
             }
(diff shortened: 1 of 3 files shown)


JPKribs:
Thank you for your contribution! Not to just reiterate the same thing as said [here](https://github.com/jellyfin/Swiftfin/pull/1649#issuecomment-3091136716), but we're on a bit of a holding pattern for playback changes while #1581 finishes up. 

There may be a good way to fit this in before but I'll leave that up to @LePips 's discretion. Since #1581 is a large change that touches a lot of places, we want to try and avoid creating new variables where possible to help make it more manageable to test and complete.

LePips:
Due to the [same comment here](https://github.com/jellyfin/Swiftfin/pull/1649#issuecomment-3093186936), I apologize again and will have to close this. However, proper state handling for error descriptions is a good idea I haven't implemented. If I do not to so in #1581, you are welcome to do so then.

On the other hand, I'm confused how you're experiencing crashes for the reasons stated, as AVPlayer will just set the corresponding error state instead of crashing. Could you provide a reproducible example in a sample project?

PR #1651, opened 2025-07-19
`PlayButton` Cleanup

JPKribs:
### Summary

The primary goal of this PR was to move the `PlayButton` to use the new logger. While I was there, I moved some repetitive items into a shared `func`. I also was finding that the context menu for resuming wasn't very intuitive. I frankly forgot that's how I set this up so I had to look at the code to remember where the resume button was. Instead, I am mirroring what other apps do and, on a resumable item, I am prompting the user via an alert for Resume, Play From Beginning, or Cancel.

On iOS, I think this is about a net neutral. It's more up-front and will result in more people being aware that we have a Play From Beginning functionality but appearance wise it's about the same. tvOS, IMO, this looks much better and works well.

Finally, as a last bit of cleanup, I removed the usages of `playbuttonitem == nil` and used a single var for `isValid`. Also, I added in the logic we'd need for `MarqueeText` to display the selected `mediaSource` once https://github.com/jellyfin/Swiftfin/pull/1634 is complete. This isn't a requirement and I can remove it if we either don't want it or feel like this should look different in a final version. All of the logic is there and commented out with a TODO for later.

The logger should also now use the correct logger usage for both iOS and tvOS along with producing an error message alert if there is ever an issue so the user is more aware if there isn't a valid media source.

### Videos

**iOS**

https://github.com/user-attachments/assets/21cfc0d0-de2a-4f2e-82c3-68590c2ba3fe

**tvOS**

https://github.com/user-attachments/assets/db7cfc8c-99ab-4e5e-941d-0912076f31dd

### DIFF
--- a/Swiftfin tvOS/Views/ItemView/Components/PlayButton/PlayButton.swift
+++ b/Swiftfin tvOS/Views/ItemView/Components/PlayButton/PlayButton.swift
@@ -6,16 +6,13 @@
 // Copyright (c) 2025 Jellyfin & Jellyfin Contributors
 //
 
-import Factory
 import JellyfinAPI
+import Logging
 import SwiftUI
 
 extension ItemView {
 
     struct PlayButton: View {
-
-        @Injected(\.logService)
-        private var logger
 
         @Router
         private var router
@@ -25,6 +22,16 @@ extension ItemView {
 
         @FocusState
         private var isFocused: Bool
+
+        private let logger = Logger.swiftfin()
+
+        // MARK: - Dialog States
+
+        @State
+        private var isPresentingResume = false
+
+        @State
+        private var error: Error?
 
         // MARK: - Media Sources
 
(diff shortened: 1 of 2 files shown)


LePips (changes requested):
I'm always down for more cleanup, but let's not add an additional click just for playing from the beginning. With my video player work, there will be a button to do so from the player.

JPKribs:
> let's not add an additional click just for playing from the beginning.

Sounds good. I have reverted this to the original `.contextMenu` for Play From Beginning. 

The other cleanup and logging changes are still in place. The bulk of the cleanup is moving the video player routing to a single func instead of almost duplicate calls from 2 places. Some of that predated me but I think I should have done that cleaner in https://github.com/jellyfin/Swiftfin/pull/1593 so making amends.

LePips (approved):

PR #1656, opened 2025-07-20
[iOS] Remove UIScreen from CompactLogo & CompactPoster views

samglt:
More UIScreen replacement changes from #1503 

I made the backdrop images consistent in resolution between the two views and also fixed CompactPoster's title growing off the screen for really long titles. Otherwise it's visually identical

~I could also make these use static sizes, but an image that's sharp on a new Pro Max phone would be too large and get aliased on an SE~

Picked a static width of 1320, which is the width of the 16PM.

  <details><summary>Before & After (Poster)</summary>

  

  

  </details>

  <details><summary>Before & After (Logo)</summary>

  

  

  </details>

### DIFF
--- a/Swiftfin/Views/ItemView/ScrollViews/CompactLogoScrollView.swift
+++ b/Swiftfin/Views/ItemView/ScrollViews/CompactLogoScrollView.swift
@@ -36,7 +36,7 @@ extension ItemView {
             GeometryReader { proxy in
                 if proxy.size.height.isZero { EmptyView() }
                 else {
-                    ImageView(viewModel.item.imageSource(.backdrop, maxWidth: proxy.size.height * 0.70 * 1.77))
+                    ImageView(viewModel.item.imageSource(.backdrop, maxWidth: 1320))
                         .aspectRatio(1.77, contentMode: .fill)
                         .frame(width: proxy.size.width, height: proxy.size.height * 0.70, alignment: .top)
                         .bottomEdgeGradient(bottomColor: bottomColor)
--- a/Swiftfin/Views/ItemView/ScrollViews/CompactLogoScrollView.swift
+++ b/Swiftfin/Views/ItemView/ScrollViews/CompactLogoScrollView.swift
@@ -38,7 +38,7 @@ extension ItemView {
                 else {
                     ImageView(viewModel.item.imageSource(.backdrop, maxWidth: proxy.size.height * 0.70 * 1.77))
                         .aspectRatio(1.77, contentMode: .fill)
-                        .frame(height: proxy.size.height * 0.70, alignment: .top)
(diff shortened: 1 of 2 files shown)


Swiftfin/Views/ItemView/ScrollViews/CompactLogoScrollView.swift
```diff
@@ -33,10 +33,15 @@ extension ItemView {
 
             let bottomColor = viewModel.item.blurHash(for: .backdrop)?.averageLinearColor ?? Color.secondarySystemFill
 
-            ImageView(viewModel.item.imageSource(.backdrop, maxHeight: UIScreen.main.bounds.height * 0.35))
-                .aspectRatio(1.77, contentMode: .fill)
-                .frame(height: UIScreen.main.bounds.height * 0.35)
-                .bottomEdgeGradient(bottomColor: bottomColor)
+            GeometryReader { proxy in
```
LePips, line 36: If we're going to have to do these checks often, let's make a `NonZeroGeometryReader` view instead that takes a content view and performs the check instead.

Swiftfin/Views/ItemView/ScrollViews/CompactLogoScrollView.swift
```diff
@@ -33,10 +33,15 @@ extension ItemView {
...
+                if proxy.size.height.isZero { EmptyView() }
+                else {
+                    ImageView(viewModel.item.imageSource(.backdrop, maxWidth: proxy.size.height * 0.70 * 1.77))
```
LePips: While I wasn't as concerned with a static width for the cinematic view, because I think we'll just remove it entirely at some point, let's do so here.

Additionally, it doesn't make sense to me to have the width property be determined by the height. The image dimensions are purely for memory management. If there are layout issues, they are caused by something else.

samglt: Ok, I think I got the reasons for doing this mixed together in my head.

We're picking a hard-coded max width/height for memory management and caching, and any visual issues caused by picking an overly large image for the device's screen are a UI issue shouldn't be dealt with by adjusting the size of image requested.

I was worried about aliasing when viewing large images on a small device, but taking a look at ImageView, we don't have any interpolation set, so turning that on will help with that.

Swiftfin/Views/ItemView/ScrollViews/CompactLogoScrollView.swift
samglt: Since the imagesource doesn't use the geometry reader to set its dimensions anymore, the if isn't needed

LePips (approved):

Issue #776, opened 2023-05-16
Multiple Users - Logged Out Between Sessions [TVOS]


Jpkribs:
I can't get this trigger 100% of the time but I think I narrowed it down to having multiple users on the AppleTV. Since adding another user profile on the AppleTV, all saved Jellyfin users will get wiped after the TV goes to sleep or reboots.

I believe this is related to how session data is stored since this only started once I added other users. On another AppleTV, I have a single user and I have not experienced this issue (yet). I could be totally off base but my if my guess right I think this could be resolved by:

https://github.com/jellyfin/Swiftfin/pull/661

https://github.com/jellyfin/Swiftfin/issues/258

holow29:
I haven't experienced this with multiple tvOS users. However, I haven't added more than one Jellyfin user.

extraslute:
I have experienced being logged out on a daily basis on my Apple TV 4K regardless of if there are multiple users or not. This doesn’t happen on my android tv nor any Rokus.

Jpkribs:
> I haven't experienced this with multiple tvOS users. However, I haven't added more than one Jellyfin user.

The multiple TVOS users could be a red herring. That's just when I started seeing it. I know other people who get this regardless of TVOS users.

Is there a way to collect TVOS logs that I could help look at/provide to better understand this?

LePips:
Sadly no logs would help, but I can investigate.

Jpkribs:
> Sadly no logs would help, but I can investigate.

Thank you! I don't know if it changes anything but it's a [2nd Gen 4K](https://support.apple.com/kb/SP845?viewlocale=en_US&locale=en_US)

Since there is one that was impacted and one that is not, I've been racking my brain thinking what the difference between them could be. The only other thing I could think of is that the one that gets logged out also has Infuse logged into the same server. I'm not sure how that would make a difference. Just hoping maybe something there might help!

IceHawk66:
> I have experienced being logged out on a daily basis on my Apple TV 4K regardless of if there are multiple users or not. This doesn’t happen on my android tv nor any Rokus.

I'm facing the same error on my apple tv 4k. Logged in yesterday (again) and I have no server today. Have to enter server + credentials again.

technicallymarv:
I have the same issue as well, have Swiftfin installed on 3 apple tv's with accounts for myself my wife and my kids and it seems almost daily the accounts get wiped and I have to reconnect to the server and login again.

bombyy:
[...]

Issue #1654, opened 2025-07-19
`CoreStore` vs `SwiftData`


JPKribs:
**Describe the feature you'd like**
Today, I was looking at: https://github.com/jellyfin/Swiftfin/issues/776. I was trying to figure out what caused this to occur so I was diving into `CoreStore` to see if maybe there was a configuration we could enable. I ended up finding a solution elsewhere related to `CoreData` that feels pretty hacky so I don't know if this is the right move. I can provide that below but I think this will need more time in the oven before I feel good about this. I'll provide that below but that's different from this Issue/Feature.

<details>

<summary>"Hacky" Solution</summary>

---

The main part of this solution is saving the SQL file in `libraryDirectory` specifically. Additionally, setting the file based on Apple's "[Mark Nonpurgeable Data as Excludable](https://developer.apple.com/documentation/foundation/optimizing-your-app-s-data-for-icloud-backup#Mark-Nonpurgeable-Data-as-Excludable)" documentation I was seeing this in a few locations as a workaround to keep items from being purged.

Finally, `.protectionKey: FileProtectionType.complete` should prevent changes to the file when booting or locked so that should prevent issues in the background.

To test, I have the Main branch of Swiftfin built and the AppStore version. VLC's tvOS app has a functionality to upload and cache data directly to the AppleTV. So, I use this to fill up the Apple TV's storage by caching several large files up there.

I then find that the AppStore Swiftfin requires a new login (empty storage) while the version with the code below does not.

---

```swift
//
// Swiftfin is subject to the terms of the Mozilla Public
// License, v2.0. If a copy of the MPL was not distributed with this
// file, you can obtain one at https://mozilla.org/MPL/2.0/.
//
// Copyright (c) 2025 Jellyfin & Jellyfin Contributors
//

import CoreStore
import Factory
import Foundation
import JellyfinAPI

typealias AnyStoredData = SwiftfinStore.V2.AnyData
typealias ServerModel = SwiftfinStore.V2.StoredServer
typealias UserModel = SwiftfinStore.V2.StoredUser

typealias ServerState = SwiftfinStore.State.Server
typealias UserState = SwiftfinStore.State.User

// MARK: Namespaces

extension Container {
    var dataStore: Factory<DataStack> { self { SwiftfinStore.dataStack }.singleton }
}

enum SwiftfinStore {

    /// Namespace for V1 objects
    enum V1 {}

    /// Namespace for V2 objects
    enum V2 {}

[...]

PR #1657, opened 2025-07-20
[WIP] Migrate tvOS from `.cachesDirectory` to `.applicationSupportDirectory`

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/776
Discussed: https://github.com/jellyfin/Swiftfin/issues/1654#issuecomment-3092617393

From some feedback, I believe this resolves our tvOS logout issue by moving our data from `.cachesDirectory` to `.applicationSupportDirectory`. This should mirror where this is places for iOS now. I have some build flags but I believe this should also work by using the same code for both iOS and tvOS. I wanted to start at this point to keep it separate since iOS is more up-to-date and I wanted to limit changes to the polished side of Swiftfin.

I have added a migration as well to move files from the `.cachesDirectory` to `.applicationSupportDirectory` if available. I put this in `Mappings` but please let me know if this is the incorrect move.

This is WIP for now. I am going to throw some more testing at this first. I've done some testing this morning but I want to test a bit more to be safe. Additionally, I want to test the migration as well. Specifically, file migration + upgrade from 1.0.1 to Main.

### DIFF
--- a/Shared/SwiftfinStore/SwiftfinStore+Mappings.swift
+++ b/Shared/SwiftfinStore/SwiftfinStore+Mappings.swift
@@ -50,4 +50,57 @@ extension SwiftfinStore.Mappings {
             ]
         )
     }()
+
+#if os(tvOS)
+
+    // MARK: - Migrate from Old tvOS Location to New
+
+    static func migrateFromOldLocation() throws {
+        let fileManager = FileManager.default
+        
+        let cachesDirectory = NSSearchPathForDirectoriesInDomains(
+            .cachesDirectory,
+            .userDomainMask,
+            true
+        ).first!
+        
+        let oldDatabaseDirectory = URL(fileURLWithPath: cachesDirectory)
+        
+        let applicationSupportDirectory = NSSearchPathForDirectoriesInDomains(
+            .applicationSupportDirectory,
+            .userDomainMask,
+            true
+        ).first!
+        
+        let newDatabaseDirectory = URL(fileURLWithPath: applicationSupportDirectory)
+            .appendingPathComponent("Swiftfin")
+        
+        let oldDatabasePath = oldDatabaseDirectory.appendingPathComponent("Swiftfin.sqlite")
+        let newDatabasePath = newDatabaseDirectory.appendingPathComponent("Swiftfin.sqlite")
+        
+        guard fileManager.fileExists(atPath: oldDatabasePath.path) else {
+            return
+        }
+        
+        guard !fileManager.fileExists(atPath: newDatabasePath.path) else {
+            try fileManager.removeItem(at: oldDatabasePath)
+            return
+        }
+        
+        try fileManager.createDirectory(
+            at: newDatabaseDirectory,
+            withIntermediateDirectories: true,
+            attributes: nil
+        )
+        
+        let oldFiles = try fileManager.contentsOfDirectory(at: oldDatabaseDirectory, includingPropertiesForKeys: nil)
+            .filter { $0.lastPathComponent.hasPrefix("Swiftfin") }
+        
+        for file in oldFiles {
+            let fileName = file.lastPathComponent
+            let destinationURL = newDatabaseDirectory.appendingPathComponent(fileName)
+            try fileManager.moveItem(at: file, to: destinationURL)
+        }
+    }
+#endif
 }
--- a/Shared/SwiftfinStore/SwiftfinStore.swift
+++ b/Shared/SwiftfinStore/SwiftfinStore.swift
@@ -51,10 +51,36 @@ extension SwiftfinStore {
     }()
 
     private static let storage: SQLiteStore = {
-        SQLiteStore(
+        #if os(tvOS)
+
+        /// Manually force the Application Support directory on tvOS to avoid purging
+        let applicationSupportDirectory = NSSearchPathForDirectoriesInDomains(
+            .applicationSupportDirectory,
+            .userDomainMask,
+            true
+        ).first!
+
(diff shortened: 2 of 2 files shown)


JPKribs:
Taking a break for today. 

The resulting new location seems good and in testing this seems to persist where 1.0.1 does not. The way I do my testing is I have 1.0.1 and this branch both loaded on the same tv. Then, I fill up the storage on the TV by uploading files via VLC. Both Swiftfins are closed during this. When I fully fill the storage, I have consistently found that 1.0.1 requires a new login while this branch does not. I don't think this is the most scientific route for this but this does seem to at least be a lot more resilient than 1.0.1.

My last issue is that I am working on migrating any existing data from the 1.0.1 / Main cache directory into this new directory. I swear, I have this original correct since this migrate moves my old files out of the expected location. So, if I go back to default my data is gone. But I'm either moving it to the wrong place (unlikely?) or I am moving this at the wrong time and there is some kind of conflict? 

Regardless, I'm going to set this down and come at this again later.

JPKribs:
@LePips I *think* this is good to go! I've successfully migrated 10 different simulators and confirmed they are working as expected. I adding a lot of logging for this but please let me know if it's done incorrectly or needs to be scaled up/down. Most of the logging I use in my day-to-day are for logs only I see so please let me know if there is a more standard way I should be approaching this.

I've been testing this by loading everything onto a physical AppleTV. I will have the AppStore 1.0.1 version logged in and this branch logged in. I will then let both of these go into the background / close Swiftfin. I use VLC's tvOS to upload videos into the AppleTV's storage until it is full. Once it's full, I check both Swiftfins and find that 1.0.1 have been 'logged out' where the data has been cleared with the cache and this branch is still active and logged in. 

Please let me know if there is a more scientific method for testing this. I know this route is kind of a pain but I've been testing this one for the day and this appears to have fixed the "log out" issue. Either that, or it's just more resilient? Regardless, I think you were right that this is the correct move for resolving this! 

Finally, In my testing, the logic I have for the the tvOS `SQLiteStore` should be identical to what iOS is doing with the CoreStore defaults. However, I wanted to minimize impact so I left iOS intact as-is. Please let me know if that isn't the right move. Below is the item I am referring to:

<details>

<summary>New SQLiteStore</summary>

```swift
    private static let storage: SQLiteStore = {

        let logger = Logging.Logger.swiftfin()

        let applicationSupportDirectory = FileManager.default.urls(
            for: .applicationSupportDirectory,
            in: .userDomainMask
        ).first!

        let databaseURL = applicationSupportDirectory.appendingPathComponent(bundleIdentifier, isDirectory: true)
            .appendingPathComponent("Swiftfin.sqlite")

        logger.info("Using the SQLite Database at: \(databaseURL.absoluteString)")

        #if os(tvOS)

        return SQLiteStore(
            fileURL: databaseURL,
            migrationMappingProviders: [Mappings.userV1_V2]
        )
        
        #else

        return SQLiteStore(
            fileName: "Swiftfin.sqlite",
            migrationMappingProviders: [Mappings.userV1_V2]
        )

        #endif
    }()
```

</details>

LePips:
Please test after my changes to see if it still works.

JPKribs:
> Please test after my changes to see if it still works.

I'll start testing now but I will likely need a couple hours get to run through it all. I'll reach out when I am done!

JPKribs:
@LePips 

I haven't done any testing for the logout portion of this. I realize that we're using the same locations so this should not require new testing. However, our migration was failing. It was strange, the SQLite file was being moved over but none of the logged in accounts were there. The issue was the `sqlite-wal` wasn't moving over:

Screenshot 2025-07-20 at 16 47 46

My understanding of the `sqlite-wal` is that this is just unprocessed transactions so this could just be that I am creating 1.0.1 then immediately attempting to migrate it, but IMO, this would make sense to move the the wal and shm over as well even if I don't believe the shm will impact anything. If anything, just to clean up after ourselves.

I added a commit that works in migrating these auxiliary files as well. I am only guarding on the SQLite since we only want to migrate the WAL and SHM if the SQLite moves so we're not mixing and matching auxiliary files.

I'm 100% sure you will have a better way of doing this than the `[URL]` haha. That being said, this latest commit works in resolving the migration portion of this. Please let me know if you need to me to test the logout portion again as well. I skipped this part since it's a slow test and we're using the same directories as prior but I'm not opposed if you feel it's warranted.

LePips (approved):
Thanks for testing!

sjorge:
Building current main for tvOS seems to result in a broken Swiftfin atm, it just hangs at a spinner. I think it's due to this PR.

The dev console logs:
[🔴 Error] DataStack+Migration#244:addStorage(_:completion:) Failed to load SQLite 'NSPersistentStore' metadata.
[🔴 Error] SwiftfinStore#103:setupDataStack() Failed creating datastack with: The operation couldn’t be completed. (com.corestore.error error 5.)

When running on hardware (it previously had another Swiftfin tvOS installed build from main a week or so ago.

Running it in a simulator is fine though (but that one was empty)

Edit: removing swiftfin and rebooting the tvOS box results in the same and installing again results in the same.
Edit2: building 41d470e0504de11f589e8f375a82374b9213248a (the commit before this merge on main works fine)

JPKribs:
@LePips I am so sorry, @sjorge is right... I swear I tested this in like 30 different ways but I must have missed something. Clean installs are screwing up on hardware.

I think is best to rollback while this is fresh. I can keep digging on my end to find a better route.

Sorry, I know that's a huge pain...

sjorge:
Glad I at least caught pretty quickly :) I try to bump my copy to main once a week or so.

LePips:
#1660

Issue #1111, opened 2024-06-15
External subtitles is the screen size


Daryan97:
### Describe the bug

1. Add an external subtitle to your movie/tv show.
2. Play the movie/tv show on Swittfin.
3. Select the subtitle.
4. When it shows the subtitle it is maximum size of the screen.
![IMG_9029](https://github.com/jellyfin/Swiftfin/assets/29106122/0a57e840-a357-42db-9e2d-8bbce5ba465d)

### Application version

1.1 (78)

### Where did you install the app from?

App Store

### Device information

iPhone 11 Pro

### OS version

iOS 17.2.1 (21C66)

### Jellyfin server version

10.9.6

guerda:
I have a similar issue but not as extreme. The subtitles are half the size of the screen.
I cannot change the size.
Also on ipad

SkyDiver7199:
The large "it" on screen just made me laugh out loud.

But I also face similar issue, tried changing fonts helps reduce a little, but still half of the screen. 
Smallest size I can go is 8

pvanaken:
Was this ever resolved as i am having the same issue on IOS

filoor:
Have the same issue on my iPhone

Mubelotix:
Not that big but subtitles are stuck to large font size. Settings are not applied

fongming:
I replicated this issue on the latest iOS version 1.3 downloaded from the Apple App Store. 

However, I also noted that setting the font size (Jellyfin app > Settings > Video Player > Subtitle Size) to be the smallest number (8) results in the largest subtitles, while setting “larger font sizes” (24) results in smaller sizes. 

I hope this will help make the app usable for more users while the developers fix the underlying bug.

samglt:
To fully fix this issue we'll want to limit the range of font sizes a little more (assuming there isn't something else going on).

For reference, here is what the largest font sizes look like (note that in 1.3, these would be 5, 4, 3, 2):

<details><summary>Sizes 20, 21, 22, 23</summary>
<p>

<img width="1334" height="750" alt="Image" src="https://github.com/user-attachments/assets/80cb5bad-fa0a-4d7b-9989-626dc2d878c5" />

<img width="1334" height="750" alt="Image" src="https://github.com/user-attachments/assets/b6107d22-a4b9-4904-b930-aaddef1c08f7" />

[...]

PR #1658, opened 2025-07-20
[iOS/tvOS] Fix subtitle size

### LINKED ISSUE #1633: Subtitle font size setting is backwards
### Describe the bug

The subtitle font size setting seems to be backwards—when I decrease the number, the subtitle size gets bigger and vice versa. 

### Application version

1.3

App Store

### Device information

iPhone 13

### OS version

iOS 18.4

### Jellyfin server version

10.10.7

samglt:
Fixes #1633 

The subtitle size setting is supposed to be inverted when fed to VLC, since VLCKit uses larger subtitle size numbers to mean smaller text (I think the text's height is 1/x of the screen's height), but when the subtitle size was set in the video player VM, it wasn't correctly inverted like it was in the .onChange in the actual View. This way the actual subtitles were (almost?) always the wrong size.

Additionally, I changed both locations from "24 - ..." to "25 - ..." so the range of actual VLC subtitle sizes would be 1-24 not 0-23, because 0 font size has strange behavior.

After making those changes I think 16 is a little big, so I lowered the default size to 9, which is the same size 16 used to be

Related to this, we might want to lower the maximum size presented in settings to something like 20-22 since at the highest sizes the text totally fills the screen and gets truncated to the point of being unreadable.

### DIFF
--- a/Shared/Services/SwiftfinDefaults.swift
+++ b/Shared/Services/SwiftfinDefaults.swift
@@ -237,7 +237,7 @@ extension Defaults.Keys {
 
             static let subtitleColor: Key<Color> = UserKey("subtitleColor", default: .white)
             static let subtitleFontName: Key<String> = UserKey("subtitleFontName", default: UIFont.systemFont(ofSize: 14).fontName)
-            static let subtitleSize: Key<Int> = UserKey("subtitleSize", default: 16)
+            static let subtitleSize: Key<Int> = UserKey("subtitleSize", default: 9)
         }
 
         enum Transition {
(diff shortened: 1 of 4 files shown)


JPKribs:
I would guess this also resolves: https://github.com/jellyfin/Swiftfin/issues/1111?

samglt:
I'd think so, but some of the reports in that thread are a little ambiguous.

JPKribs:
Once this hits a TestFlight, we can ask them to confirm if this is still an issue and close out from there. I'm like 95% sure this resolves #1111 as well but I agree it's too ambiguous to be certain.

Thank you for the fix BTW! You've been killing it lately with all the PRs!

LePips (approved):
I still saw subtitles go fullscreen, where only like 3 characters would fit, at the max height. I think this is a VLC issue for low values, but don't have the motivation to learn that codebase.

PR #1668, opened 2025-07-26
[iOS/tvOS] Fix corner radius performance regression

samglt:
I noticed that scrolling on my iPad was pretty choppy, and I saw that the refactoring in #1617 lost the use of `RoundedRectangle` when all the corner radii were the same. I know that the previous bezier path method of rounding corners was dropped with iOS 15 compatibility, but as far as I can tell the new `UnevenRoundedRectangle` is just a convenience wrapper in terms of performance.

Unfortunately due to the type checker I couldn't get it as compact as I'd like.


LePips (approved):
Ah apologies, I had missed your previous investigation in #1504.

PR #1669, opened 2025-07-29
[WIP] Inline Collapsable Text

chickdan:
This is an initial change to demo an Overview text that can expand/collapse in place rather than presenting a fullscreen modal to provide a better user experience. 

<details><summary>Before</summary>

https://github.com/user-attachments/assets/2bcaf0d3-07e3-4f7e-950c-f2aaa658d244

</details> 

<details><summary>After</summary>

https://github.com/user-attachments/assets/3f89aa30-c3fe-4a94-a37d-15a0deec5bcd

</details> 

The `CollapsableOverview` is from one of my projects that I tweaked a bit to get working here so this is not a fully fleshed out implementation but I'd like to get early feedback.


LePips:
Long ago when I made the item views as they are now, I did experiment with an inline collapsable feature. However, I found presenting a sheet, similar to the Apple TV app, more desirable for this specific context.

- On the item view, the overview is a smaller font to fit well within the context of presenting the item's contents all together. To focus on the overview, present the sheet in the larger font for readability.
- Overviews can be of any length, with animes known to be notoriously long. Having a line limit is one control for the item view, with the sheet presentation + scroll handling arbitrary lengths.
- We don't have to worry about animation, masking, or clipping views by just presenting a sheet.
- We don't have multiple methods of presenting the expanded overview (collapsable text here + sheet on the about card).

PR #1676, opened 2025-08-11
[iOS] Limit maximum subtitle size

samglt:
Belated piece of the fix for #1111. 

This just reduces the largest subtitle size option so you can't make the subtitles unreadably big. See [this for images](https://github.com/jellyfin/Swiftfin/issues/1111#issuecomment-3130849616). Now the biggest subtitle size is comparable to .body text at the largest accessibility size.

### DIFF
--- a/Swiftfin/Views/SettingsView/VideoPlayerSettingsView/Components/Sections/SubtitleSection.swift
+++ b/Swiftfin/Views/SettingsView/VideoPlayerSettingsView/Components/Sections/SubtitleSection.swift
@@ -30,7 +30,7 @@ extension VideoPlayerSettingsView {
                 BasicStepper(
                     L10n.subtitleSize,
                     value: $subtitleSize,
-                    range: 1 ... 24,
+                    range: 1 ... 20,
                     step: 1
                 )
 


Swiftfin/Views/SettingsView/VideoPlayerSettingsView/Components/Sections/SubtitleSection.swift
```diff
@@ -30,7 +30,7 @@ extension VideoPlayerSettingsView {
                 BasicStepper(
                     L10n.subtitleSize,
                     value: $subtitleSize,
-                    range: 1 ... 24,
+                    range: 1 ... 20,
```
LePips, line 33: Does the max size also need to change in the players? Don't worry about my refactor, I compare against main to ensure I keep the behaviors in line.

https://github.com/jellyfin/Swiftfin/blob/e71dc35ed14c367334f024cfeb245792b1b6f683/Swiftfin/Views/VideoPlayer/VideoPlayer.swift#L235

samglt: No, since that line is converting our subtitle size (lower is smaller) into VLC's subtitle size (lower is larger). My change means the largest user-facing size that can be selected (20) becomes 25 - 20 = 5, so VLC's subtitles can only be as large as 5 (which I believe is 1/5 of the height of the screen).

If we decide to allow even smaller subtitle size options in the future then we would have to increase that number by the same amount we increase the max of the range so that the largest subtitle size would stay the same visually.

LePips (approved):

PR #1677, opened 2025-08-13
Add episode-focused media navigator

AlexGustafsson:
In Jellyfin's web UI the user can chose between various sections when navigating media. There's "series", "recommended", and "episodes" - among others.

Jellyfin's web UI also has a user setting for controlling the default section of a library.

This makes Jellyfin great for episodal video where episodes are more "important" than the series themselves - such as for TV shows or internet video.

Today, Swiftfin only really supports the "series" section (as in, the media tab only shows collections and so on). 

This PR aims to change that by adding an option to show episodes of a library, no matter their parent. As a first step I'd like to show the absolute bare minimum code needed to get this done (breaking other features in the process) just to clearly show the intent / use case.

If the developers think it's a good idea, I'd be willing to put together actual code, with some guidance.

Before:

Series are shown.

image

After:

Episodes are shown. One example way to control the section would be through the "more" / "kebab" button.

image

### DIFF
--- a/Shared/ViewModels/LibraryViewModel/ItemLibraryViewModel.swift
+++ b/Shared/ViewModels/LibraryViewModel/ItemLibraryViewModel.swift
@@ -19,7 +19,9 @@ final class ItemLibraryViewModel: PagingLibraryViewModel<BaseItemDto> {
 
     override func get(page: Int) async throws -> [BaseItemDto] {
 
-        let parameters = itemParameters(for: page)
+        var parameters = itemParameters(for: page)
+        parameters.includeItemTypes = [.episode]
+        parameters.isRecursive = true
         let request = Paths.getItemsByUserID(userID: userSession.user.id, parameters: parameters)
         let response = try await userSession.client.send(request)
 
--- a/Shared/Objects/LibraryGroupingType.swift
+++ b/Shared/Objects/LibraryGroupingType.swift
@@ -1,0 +1,32 @@
+//
+// Swiftfin is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2025 Jellyfin & Jellyfin Contributors
+//
+
+enum LibraryGroupingType: String, CaseIterable, Displayable, Storable, SystemImageable {
+
+    case group
+    case ungrouped
+
+    // TODO: localize
+    var displayTitle: String {
+        switch self {
+        case .group:
+            "Group"
+        case .ungrouped:
+            "Ungroup"
+        }
+    }
+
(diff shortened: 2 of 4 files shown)


Shared/ViewModels/LibraryViewModel/ItemLibraryViewModel.swift
```diff
@@ -19,7 +19,9 @@ final class ItemLibraryViewModel: PagingLibraryViewModel<BaseItemDto> {
 
     override func get(page: Int) async throws -> [BaseItemDto] {
 
-        let parameters = itemParameters(for: page)
+        var parameters = itemParameters(for: page)
```
AlexGustafsson, line 22: This is essentially the key - fetching only episodes and making sure to fetch them recursively - just like the web UI.

AlexGustafsson:
As an aside (or for future reference rather), to make Swiftfin work better for this use case I had to sort by name (descending) as the names start off with the date. Sorting by created date does not seem to work, they seem to be nil.

I also disabled "series background" to ensure the episode's thumbnail was used.

JPKribs:
Just to confirm, the goal of this is to have libraries where it's a group by type as opposed to by library? I believe we have this built out already for tvOS for our Shows & Movies sections where we are grouping by Series & Movies.

In this case, we would want a library type (or toggle) that would create an ItemType library for Episodes that is recursive?

---

I'm not opposed to an idea like this but I think we'd need to be more generic about this. I could see this being valuable down the road for Photo Libraries seeing all photos instead of folders or Music Libraries that would want songs instead of artist > album > songs. For the latter, we'd want the ability to group by album and songs so this would need to be somewhat dynamic.

I think the idea is good but this should be built in a way that it would be reusable for all of these instances and also hidden from places like Movies & Collections where there is no reason/ability to ungroup. 

Finally, we would we need to be cognizant not to recreate https://github.com/jellyfin/Swiftfin/pull/1495.

---

I'll defer to @LePips but I think this is doable. That being said, this would likely be a larger undertaking than what you have so far. We'd need to test carefully since each time we touch the `PagingLibraryViewModel` it feels like some unexpected behavior arises.

LePips:
Thank you for pitching and implement a POC of the idea and I do think this is doable.

The more general implementation we should investigate is allowing a `LibraryParent` to define any number of groupings of its contents, rather than a single "grouped vs ungrouped", in this case `Shows` and `Episodes`. Then, when the library view model refreshes the library parent determines what items to retrieve, whether it's recursive or not, and possibly other settings when that grouping is changed.

Here's a POC I mocked during a meeting, realizing it would require much more refactoring than before. However, it's a general idea of the solution I would like implemented.

https://github.com/jellyfin/Swiftfin/compare/main...LePips:SwiftFin:library-grouping-poc

LePips:
I made an issue to further detail what this work would look like #1705.

PR #1678, opened 2025-08-13
[tvOS] Fix ChevronButton Title Alignment

chickdan:
On tvOS the `EmptyView` for the icon is pushing the title to the right so it looks strange compared to the other buttons.

Not the cleanest solution around so suggestions are welcome.

| Before | After |
|--------|--------|
| Screenshot 2025-08-13 at 5 13 03 PM | Screenshot 2025-08-13 at 5 14 42 PM |

### DIFF
--- a/Shared/Components/ChevronButton.swift
+++ b/Shared/Components/ChevronButton.swift
@@ -79,10 +79,15 @@ extension ChevronButton {
 
         let label: Label
         let action: () -> Void
+        let hasIcon: Bool
 
         var body: some View {
             Button(action: action) {
-                label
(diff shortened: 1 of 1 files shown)


JPKribs:
Thank you for looking at this! That's been bugging me for a while but I never invested the time to look into it lol

LePips (changes requested):
Thanks for taking a look, we can make this more concise with:

```swift
if Icon.self == EmptyView.self {
    label
        .labelStyle(.titleOnly)
} else {
    label
}
```

PR #1684, opened 2025-08-16
Cleanup `Notifications` & Series `UserData` Updates

JPKribs:
## Summary

This PR Cleans up notification / handling playback updates. Specifically, this resolves the issue created where there is a race condition where the `SeriesItemViewModel` will update and remove all of the seasons and re-add them. Then, the `SeasonItemViewModel` refreshes. This could be made to consistently trigger out of order resulting in an empty, un-inited `SeasonItemViewModel`. Instead, I only refresh the `SeriesItemViewModel.seasons` on a full refresh. This leaves the `SeasonItemViewModel` intact and already inited. Then, we can push updated to individual episodes to preserve the scroll position.

Finally, I tried to clean up `Notifications` for `BaseItemDto`. I renamed them to fit the current naming scheme of "didXHappen" where before these were "itemMetadataX". I've broken our `BaseItemDto` `Notifications` to the following:

``` 
    /// - Payload: The ID of the item that should refresh from server.
    static var doesItemRequireRefresh: Key<String> {
        Key("doesItemRequireRefresh")
    }

    /// 
    /// - Payload: The new item with updated metadata.
    static var didItemMetadataChange: Key<BaseItemDto> {
        Key("didItemMetadataChange")
    }

    /// 
    /// - Payload: The ID of the deleted item.
    static var didItemDeletionOccur: Key<String> {
        Key("didItemDeletionOccur")
    }
```

`doesItemRequireRefresh`: An item has been updated and should be refreshed from the server. This is useful for changes to `UserData` and image changes. 
- New images will have new tags which are part of the metadata and cannot be provided unless we have the full `BaseItemDto`. We also cannot be certain what these tags are without getting them from the server.
- `UserData` could be passed around _(see below)_ but often times the `UserData` we have disagrees with Server. If I say the episode is 95% complete, the server has a setting to round that to fully watched instead. Getting the `UserData` from server ensures this is correct. I think this is also something we can clean up later with the WebSocket.

`didItemMetadataChange`: An item has been updated and we have the full version of this. This item should either replace the active item or the active `PlayButtonItem` completely.

`didItemDeletionOccur`: An item has been deleted. Nothing matters except the ID so we can remove it.

---

I think there is room where we could have a `didItemUserDataChange` that pushes around changes to `BaseItemDto.userData`. However, `UserItemDataDto` has an `itemID` field that is almost **NEVER POPULATED**. So we'd either need to always update that to the respective `itemID` var before we call `didItemUserDataChange` OR we would have to call it as a Tuple like `(String, UserItemDataDto)`. Below would be the first route we ensure `UserItemDataDto.itemID` is populated:

```
    /// - Payload: The new item with updated userdata.
    static var didItemUserDataChange: Key<UserItemDataDto> {
        Key("didItemUserDataChange")
    }
```
---

To assist with refreshing items, I've added a refresh() to the `Poster` Protocol and create the appropriately funcs for `BaseItemPerson` and `BaseItemDto`. This allows doing element[i].refresh() opposed to (element[i] as BaseItemDto).refresh the adding this new `BaseItemDto` version of the element back into the array.

There is a way to also just do a `refreshUserData()` but that feels less relevant to all `Poster`.

Finally, I had 1 issue that was killing me where the `EpisodeCard` had an immutable episode so there was no way to get the refreshed `UserData`. I have a work around in a comment. Aside from this, this should be good to go!

---

### iOS

https://github.com/user-attachments/assets/0d1ec23e-a787-479c-a0fc-e72a28733eb8

## tvOS

https://github.com/user-attachments/assets/50211522-8379-48fc-b31b-5fa87775906f


Swiftfin/Views/ItemView/Components/EpisodeSelector/Components/EpisodeCard.swift
```diff
@@ -19,7 +19,16 @@ extension SeriesEpisodeSelector {
         @Router
         private var router
 
-        let episode: BaseItemDto
+        let id: String
+
+        @ObservedObject
+        var viewModel: PagingLibraryViewModel<BaseItemDto>
```
JPKribs: @LePips Is there a better way you would recommend doing this? The issue is, this PR stops refreshing the whole `SeasonItemViewModel`. This means the `SeasonEpisodeSelector` stays intact between playback updates.

I am using a notification to push the new episode userdata and this part is working. However, the parent `EpisodeHStack` and our `EpisodeCard` the `episode` (`BaseItemDto`) is immutable and does not update. So, even if we `RedrawOnNotificationView` the `UserData` updates aren't there.

This route works in refreshing the overlay on the impacted `EpisodeCard` without having to refresh the entire `EpisodeHStack`. This is important because refreshing the `EpisodeHStack` on tvOS will kick the focus out. Plus, I believe only updating the required items is ideal?

Long story short:

**Is there a a way to make the `CollectionHStack.Content` so it can refresh on change? Or a way to keep passing updates to items in the `CollectionHStack`? Or is this the best route?**

I believe I'm otherwise done with this PR but this refreshing has been driving me crazy.

Please see the video below to show what I mean:

###Old Issue

https://github.com/user-attachments/assets/c7518ffc-8c65-4b68-b868-19313aff0772

###With this `ViewModel` fix

https://github.com/user-attachments/assets/967e513f-2b02-4c98-93a8-56660bf2d935

Shared/Objects/Poster/Poster.swift
```diff
@@ -36,6 +36,8 @@ protocol Poster: Displayable, Hashable, LibraryIdentifiable, SystemImageable {
         maxWidth: CGFloat?,
         quality: Int?
     ) -> [ImageSource]
+
+    mutating func refresh() async throws
```
LePips: Please don't introduce an in-place mutation function. I consider it an anti-pattern and would like to be more explicit when value types are replaced.

Shared/ViewModels/LibraryViewModel/PagingLibraryViewModel.swift
```diff
@@ -137,13 +137,45 @@ class PagingLibraryViewModel<Element: Poster>: ViewModel, Eventful, Stateful {
...
+            .publisher
+            .receive(on: RunLoop.main)
+            .sink { [weak self] newItem in
+                guard let self = self, let newElement = newItem as? Element else { return }
+
+                if elements.ids.contains(newElement.unwrappedIDHashOrZero) && elements[id: newElement.unwrappedIDHashOrZero] != newElement {
+                    Task { @MainActor in
+                        self.elements[id: newElement.unwrappedIDHashOrZero] = newElement
+                    }
+                } else if newItem.type == .series {
```
LePips: Series items shouldn't require other logic to refresh in this context.

LePips (changes requested):
I haven't dived too deep into this, but this isn't the way to go. The addition of the in-place `refresh` method and then the fact each library will make the same duplicated call to get item metadata. We only need one place that calls the updated metadata and then broadcasts that to receivers.

Per the series item updating on episodes changing, just don't say the series item updated when an episode metadata changed.

JPKribs:
@LePips I have something that is functional and centralized. This isn't my final version of this but I was hoping to make sure this is more of what you are looking for. I've done DI in C# but I'm not finding a Swiftfin equivalent that I can mirror. So that part is weird but the `NotificationHandler` is at least functionally doing what the final version would do.

On that note, there isn't a batch way to get `UserData` for multiple `ItemID` so right now this is a loop. I'm trying to determine if doing a loop to get child `UserData` only is more or less performant than doing a single call for the full `BaseItemDto` with `Fields == nil` then using a `compactMap` to get only the `UserData`.

---

Here's the board I was basing this on. My premise is basically, if the data already exists for what you need to update, you call the `didItemXYZ`. Otherwise, calling `doesItemRequireRefresh` with the type will pass that to the handler and that will output the appropriate notifications for `didItemXYZ`. This centralizes work for things like getting `UserData` to reduce those calls and also handles more complicated scenarios like marking a whole series as played where all episodes get marked as played too. 

Screenshot 2025-08-24 at 01 49 49

---

Big outstanding issue for functionality is that the `CollectionHStack` / `CollectionVGrid` treat their elements as immutable so any items displayed using this don't update with the `UserData` change. You can make them update by scrolling them out of view where they are no longer visible then scrolling back and they're redrawn. There are hacky ways I've found to resolve this like decoupling the overlay from the poster but I think the solution is in the `CollectionHStack` / `CollectionVGrid` itself.

PR #1688, opened 2025-08-22
[tvOS] Migration `ActionButton`s to a `ButtonStyle`

JPKribs:
### Summary

When I first made the `ActionButton` on. tvOS it always bothered me that the `Menu` and `Button` variants were close but still visibly different. I realized that all the work I was doing was just attempting to recreate `.hoverEffect(.lift)` and that covered the focus & press animations. So, I turned the `ActionButton` into `ButtonStyle` that mirror iOS and just use the `.hoverEffect(.lift)` to mimic how `.card` looks.

While I was doing this, I increased the spacing between buttons a button to prevent the issue where, when you only have 2 `ActionButton` under the `playButton`, they would overlap. I did this by upping the padding from 20 to 30. This does NOT require any changes to the overall width as the most buttons we would have are:

`100 + 30 + 100 + 30 + 100 + 30 + 60 = 450` which was the original size of the `playButton` area.

Lastly, while I was moving the `playButton` to use the new `tintedMaterial` on tvOS, I changed this to use the `accentColor`. I am not sold on this change so I'd be happy to set this back to `.white`.

### Screenshots

Here is with the all buttons + versions. These buttons no longer have overlapped when focused:

Screenshot 2025-08-22 at 13 44 09

Focused:

Screenshot 2025-08-22 at 13 45 32

<details>

<summary>4 Options</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-08-22 at 13 46 50

</details>

<details>

<summary>3 Options</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-08-22 at 13 13 04

</details>

<details>

<summary>2 Options</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-08-22 at 13 13 58

</details>

### Video

https://github.com/user-attachments/assets/f28c2fa7-862b-46d9-bb12-3b07a20d1881


LePips (approved):
I've consolidated the button style and much of `PlayButton` to be consolidated, hopefully moving towards the whole button being for both platforms in the future.

PR #1696, opened 2025-09-02
[iOS/tvOS] Fix user icons not changing between servers

samglt:
Currently if you are connected to multiple servers and switch between them, your user icons don't change. It's a little confusing since if you're not paying attention to the user names it looks like you haven't switched servers:

https://github.com/user-attachments/assets/cfd1c131-aec9-457e-8686-589d6a9b6ed2

This also happens if you have multiple users in each server, so if you have 2 in one, and 3 in the other, the first 2 icons will stay the same as you swap between them.

The issue was that the ForEaches used to display the images was using the users' index in the list as their ID. The fix is to use the ID keypath that the views already have.

### DIFF
--- a/Shared/Components/CenteredLazyVGrid.swift
+++ b/Shared/Components/CenteredLazyVGrid.swift
@@ -111,6 +111,11 @@ extension CenteredLazyVGrid {
         let minimum: CGFloat
         let spacing: CGFloat
 
+        private var enumeratedId: KeyPath<EnumeratedSequence<Data>.Element, ID> {
+            let element = \EnumeratedSequence<Data>.Element.element
+            return element.appending(path: id)
+        }
+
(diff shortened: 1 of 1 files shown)


JPKribs:
Testing, this resolves this for me. This one bugged me so I'm happy to see there is a good fix! That being said, I'll defer to @LePips on whether this `enumeratedId` is the best route for this since this is used in a few places outside our User Selection.

samglt:
I didn't see anywhere else they're used, just user selection in iOS and tvOS. Both the views are just used in CenteredLazyVGrid, which isn't used anywhere else I can see.

LePips (approved):
Thanks, looks like I knew I was missing something because I passed in the id path but then never used it.

PR #1697, opened 2025-09-02
[iOS/tvOS] Fix subtitle font setting not working

### LINKED ISSUE #1122: Subtitle appearance settings ignored
## Description:

Despite choosing a different font than default and using the Swiftfin video player with SRT subtitles, the subtitles still look like the default appearance settings in VLC.

## To reproduce:

1. Set a different font in Settings > Video Player
2. Play an item with SRT subtitles enabled

## Screenshots:

IMG_0154
IMG_0153

### Application version

1.1

App Store

### Device information

iPad Air 13-inch (M2, 2024)

### OS version

iPadOS 18.0 (22A5282m)

### Jellyfin server version

10.9.6

samglt:
Fixes #1122 

For some reason UIFont(name: , size:) always returns Helvetica if the size is zero, even for nonsense font names, since the subtitle size isn't set by the UIFont, just picking a positive number makes it properly get the right font


JPKribs (approved):
This resolves this on my end as well! Thank you for your PR!

PR #1701, opened 2025-09-13
[iOS] Stop iOS native player calling sendStartReport twice

neriald:
### Summary

Resolves: iOS native player is calling sendStartReport twice which then causes app to continue sending progress report `sendProgressReport()` even after video playback is closed. This also causes server side transcoding job to continue endlessly.

VideoPlayerManager `onStateUpdated` method handles start report `sendStartReport()`, no need to send it from NativeVideoPlayer module too.

I have no means of testing tvOS, therefore I haven't touched the code there although it seems similar. Also I have no knowledge of LiveNativeVideoPlayer, so I left it untouched too.

### DIFF
--- a/Swiftfin/Views/VideoPlayer/NativeVideoPlayer.swift
+++ b/Swiftfin/Views/VideoPlayer/NativeVideoPlayer.swift
@@ -171,8 +171,6 @@ class UINativeVideoPlayerViewController: AVPlayerViewController {
 
     private func play() {
         player?.play()
-
-        videoPlayerManager.sendStartReport()
     }
 
     private func stop() {


JPKribs:
This is resolved in https://github.com/jellyfin/Swiftfin/pull/1581 which should be included in our next release. I do not believe this PR is required.

LePips:
Yeah, apologies for closing. #1581 is really close to completion. Really bad timing on my part.

neriald:
> Yeah, apologies for closing. #1581 is really close to completion. Really bad timing on my part.

Hey, no worries as long as a fix is on the way I don't mind. Thank you!

Issue #1603, opened 2025-06-22
[iOS 26] Blur on the start page for movies does not follow scrolling and has bad contrast


darkexplosiveqwx:
### Describe the bug

When on the main page for any show/episode and movie the blur does not follow when scrolling and causes bad contrast.

iPhone: https://github.com/user-attachments/assets/d8613e9c-4b26-4a6a-8a40-77cc8bdb1840
iPad: https://github.com/user-attachments/assets/d6835b7b-e1d6-42b5-a094-016fd3c9f25b

### Application version

1.3 (13)

### Where did you install the app from?

TestFlight

### Device information

iPhone 11, iPad Air M3 11in (2025)

### OS version

iOS and iPadOS 26.0 Developer Beta 1 (23A5260n)

### Jellyfin server version

10.10.7

JPKribs:
> iOS and iPadOS 26.0 Developer Beta 1 (23A5260n)

I am unable to recreate this issue on iOS 18. 

I'm inclined to say we should hold off on issue reports from Apple's dev builds **for now**. As we get closer to the final release, it will be worth revisiting to make sure we're ready. But, for now, dev builds are a moving target.

This _could_ be related to:

https://developer.apple.com/documentation/uikit/uiscreen/main

But I don't have a way to test iOS 26 at this time.

darkexplosiveqwx:
I'll report changes happening in further betas

LePips:
The introspection library used for this functionality is currently in beta getting the iOS 26 support. There are also many issues with Xcode 26 hindering that development.

darkexplosiveqwx:
What is the current state of this and the library?
On 09.09.2025 the Release Candidate of Apple Software version 26 got released with iOS build number 23A340. This could very well be the final release.

JPKribs:
> What is the current state of this and the library?

This is still in beta on their end. You can follow their progress there but, to avoid duplicate work, I believe it's best for us to hold off on this until there is either a final release or iOS 26 is out which would force our hand to use a beta release. Whichever comes first, but chasing beta versions can be time consuming as APIs and structure can change requiring reworks. So hopefully we have a final release soon.

https://github.com/siteline/SwiftUI-Introspect

PR #1706, opened 2025-09-16
Bump Introspect Version

JPKribs:
Resolves: https://github.com/jellyfin/Swiftfin/issues/1603

Introspect updated yesterday, resolving the scrolling bugs we were seeing in iOS 26.

This: `content.introspect(.scrollView, on: .iOS(.v16), .tvOS(.v16)` is now asynchronous so I made `Backport.scrollClipDisabled` on MainActor to resolve this. Let me know if this isn't correct.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
         run: brew install swiftgen
 
       - name: Set Xcode Version
-        run: sudo xcode-select -s "/Applications/Xcode_16.4.app"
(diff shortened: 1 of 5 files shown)


JPKribs:
@LePips do we need to update fastlane for XCode 26?

LePips:
Eventually, yes. It won't be an issue until ~early next year. However, right now I've only heard issues with how Xcode 26 is installed on the runner images, for some reason.

PR #1712, opened 2025-09-20
[iOS] Fix audio/subtitle tracks not changing

### LINKED ISSUE #1711: [iOS] Unable to select audio or subtitle tracks
Breaking this out of [this comment on 1581](https://github.com/jellyfin/Swiftfin/pull/1581#issuecomment-3313540886). Audio and subtitle track selection doesn't work in the new video player. Different tracks can be selected, but the actual audio/subtitles don't change.

samglt:
Fixes #1711 

Audio and subtitle tracks were connected to the selection UI from the item to the UI, but not the other way, so changing the track in UI did nothing. This just connects the SwiftUI State to the item via onChange. 

I thought it would be possible to modify .assign() to make it optionally two-way but I don't think that's possible going from Binding -> Published

### DIFF
--- a/Swiftfin/Views/VideoPlayerContainerView/PlaybackControls/Components/NavigationBar/ActionButtons/AudioActionButton.swift
+++ b/Swiftfin/Views/VideoPlayerContainerView/PlaybackControls/Components/NavigationBar/ActionButtons/AudioActionButton.swift
@@ -55,7 +55,8 @@ extension VideoPlayer.PlaybackControls.NavigationBar.ActionButtons {
                 }
                 .videoPlayerActionButtonTransition()
                 .assign(playbackItem.$selectedAudioStreamIndex, to: $selectedAudioStreamIndex)
-                .backport.onChange(of: selectedAudioStreamIndex) { _, newValue in
+                .backport
+                .onChange(of: selectedAudioStreamIndex) { _, newValue in
                     playbackItem.selectedAudioStreamIndex = newValue
                 }
             }
--- a/Swiftfin/Views/VideoPlayerContainerView/PlaybackControls/Components/NavigationBar/ActionButtons/AudioActionButton.swift
+++ b/Swiftfin/Views/VideoPlayerContainerView/PlaybackControls/Components/NavigationBar/ActionButtons/AudioActionButton.swift
@@ -55,6 +55,9 @@ extension VideoPlayer.PlaybackControls.NavigationBar.ActionButtons {
                 }
                 .videoPlayerActionButtonTransition()
                 .assign(playbackItem.$selectedAudioStreamIndex, to: $selectedAudioStreamIndex)
+                .backport.onChange(of: selectedAudioStreamIndex) { _, newValue in
+                    playbackItem.selectedAudioStreamIndex = newValue
+                }
             }
         }
     }
--- a/Swiftfin/Views/VideoPlayerContainerView/PlaybackControls/Components/NavigationBar/ActionButtons/SubtitleActionButton.swift
+++ b/Swiftfin/Views/VideoPlayerContainerView/PlaybackControls/Components/NavigationBar/ActionButtons/SubtitleActionButton.swift
@@ -55,7 +55,8 @@ extension VideoPlayer.PlaybackControls.NavigationBar.ActionButtons {
                 }
                 .videoPlayerActionButtonTransition()
                 .assign(playbackItem.$selectedSubtitleStreamIndex, to: $selectedSubtitleStreamIndex)
-                .backport.onChange(of: selectedSubtitleStreamIndex) { _, newValue in
+                .backport
+                .onChange(of: selectedSubtitleStreamIndex) { _, newValue in
                     playbackItem.selectedSubtitleStreamIndex = newValue
                 }
             }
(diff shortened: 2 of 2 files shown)


Swiftfin/Views/VideoPlayerContainerView/PlaybackControls/Components/NavigationBar/ActionButtons/AudioActionButton.swift
```diff
@@ -55,6 +55,9 @@ extension VideoPlayer.PlaybackControls.NavigationBar.ActionButtons {
                 }
                 .videoPlayerActionButtonTransition()
                 .assign(playbackItem.$selectedAudioStreamIndex, to: $selectedAudioStreamIndex)
+                .backport.onChange(of: selectedAudioStreamIndex) { _, newValue in
```
JPKribs: We've been keeping these on separate lines is you wouldn't mind the small change:

```
.backport
.onChange(of: selectedAudioStreamIndex) { _, newValue in
    playbackItem.selectedAudioStreamIndex = newValue
}
```

Swiftfin/Views/VideoPlayerContainerView/PlaybackControls/Components/NavigationBar/ActionButtons/SubtitleActionButton.swift
```diff
@@ -55,6 +55,9 @@ extension VideoPlayer.PlaybackControls.NavigationBar.ActionButtons {
                 }
                 .videoPlayerActionButtonTransition()
                 .assign(playbackItem.$selectedSubtitleStreamIndex, to: $selectedSubtitleStreamIndex)
+                .backport.onChange(of: selectedSubtitleStreamIndex) { _, newValue in
```
JPKribs: Same here as well

JPKribs:
@LePips This is tested and functional. This looks correct to me, but you're much more familiar with this area if you can give the sign-off.

LePips (approved):

JPKribs (approved):

DrHurt:
@JPKribs 

Can you please extend this to work with AVPlayer?  (Perhaps an interim solution to select track in advance like on the web player)

Eventually it’ll be nice if ffmpeg just maps & converts all tracks to webvtt for selection during playback.

JPKribs:
> @JPKribs
> 
> Can you please extend this to work with AVPlayer? (Perhaps an interim solution to select track in advance like on the web player)
> 
> Eventually it’ll be nice if ffmpeg just maps & converts all tracks to webvtt for selection during playback.

Aside from testing & approving work that others have done, I have had no real involvement in the player. Apologies, I am not your guy for this change. Additionally, VLC and AVPlayer are very different beasts so this would require some serious work to implement and part of why we don't have this feature at this time. If I remember correctly, Apple's requirements makes this a pain to work with using separate streams like how Jellyfin presents this. But, as I said, I haven't really done work on the player so I'd defer to the other parties who have more exposure in this area.

There is a lot of other work in getting VLC up to where we want it as the primary player. I would guess that we will get to this working with AVPlayer but there might be some distance between now and then just while we work on other items

DrHurt:
@JPKribs @LePips 

ffmpeg is able to output standards-compliant HLS with WebVTT compatible with iOS.

I had a look at the log and it seems Jellyfin currently omits the subtitles track altogether when remuxing.

I'm able to stream my mkv (h264+eac3+srt) directly to avplayer via the following command:
```
ffmpeg -i input.mkv 
-c:v copy
-c:a copy 
-c:s webvtt 
-f 'hls' -hls_time 6 -hls_playlist_type 'vod'  
-var_stream_map 'v:0,a:0,s:0,sgroup:subs,language:en-US,default:YES' 
-master_pl_name 'master.m3u8' -hls_segment_filename "%v_segment_%05d.ts" "%v.m3u8" -y
```

With the above command, I get on demand subtitles in avplayer (along with all the other benefits of avplayer like native Dolby Atmos + Vision, HomePods, reduce loud sounds.. etc).

I'm not familiar with how Jellyfin determines the syntax passed on to ffmpeg. 
I'm happy to open a ticket on the Jellyfin Server's GitHub if that's more appropriate.

LePips:
Currently, if you have media that is able to output WebVTT subtitles then the server is able to build an Apple-friendly HLS manifest and the `AVPlayerViewController` ("Native" player) will allow subtitle selection. However in my experience, they are for the most part temporally out of sync.

It is not unknown that the server doesn't build an Apple-friendly HLS manifest. However, this whole interaction between the server and what Apple expects has only led to confusion with the "Native" player by users, so it will be removed soon. While I do think it could be a worthwhile effort to improve the server HLS manifest construction, I don't have the time to do so.

DrHurt:
@LePips 

Thanks again for your reply.

I wouldn’t say I’m confused. I actually prefer the built-in avplayer (UI & rendering layer) even if it means remuxing or transcoding as it integrates better with Apple’s walled garden ecosystem (TV, HomePods, iPhone media controls etc)

Jellyfin already uses ffmpeg to remux compatible media so I suggest a small change to ffmpeg commands so that output would be more compatible with the built-in player for those who prefer it. 

I provided the command changes required to do so (tested on iOS & tvOS)

I took the time to test Jellyfin’s ffmpeg & report that it’s already capable of outputting Apple friendly WebVTT + HLS manifest with a command change.

I agree my solution would be easier to implement with a custom UI overlay instead of AVPlayerViewController (especially with complex multitrack media)

I’ll keep an eye on progress and report any issues.

LePips:
From a product design perspective, I have to consider how to design an app for video playback. Users shouldn't care about "what player they're using". I'm glad that technically-inclined users like you can understand the inner workings of the frameworks used, but the "confusion" mentioned above is about how normal users become confused on the lack of feature parity. It's unproductive to dismiss issues like these as implementation details, as I need to instead consider how things works as a product.

So while I understand the player experience provided by Apple is preferable in many ways, I will be removing the "player" in the near future. I strongly don't think I'll have it as an `Experimental` feature either, or introduce some other obscure settings page to enable it.

PR #1714, opened 2025-09-20
`Stateful` macro

### LINKED ISSUE #1596: Improvement for categories screen.
### Describe the bug

The screen where all the categories are shown sometimes has the title overlapping with the categories.
On the other hand, when you have custom images, it becomes difficult to know what’s in each category since there are no names displayed at the bottom of each one.

Image

### Application version

1.3

TestFlight

### Device information

iPhone 13 Pro Max 

### OS version

iOS 18.5

### Jellyfin server version

10.10.7

### LINKED ISSUE #1598: App crashes when selecting Media tab immediately after launching
### Describe the bug

If I select the media tab before the Home screen finishes loading, the app crashes.

Image

### Application version

commit hash a8e8588354f413948b892f08cdb5d4003bf258f0

None

### Device information

iPad Pro 11" (2nd gen)

### OS version

iOS 18.5

### Jellyfin server version

10.10.3

LePips:
Introduces a `Stateful` member macro I've been working on to streamline our stateful objects. This reduces the boilerplate necessary for the old `Stateful` and `Eventful` protocols greatly, and we will move over all old objects entirely. However, it's still very much in development, so there may be more changes done in the future and I still need to provide documentation. The macro package has Swift 6 concurrency checks applied - so we transitively get those checks on macro application. This should help greatly when we turn it on for the whole project ourselves, but also means that each migration won't be a similar amount of work (as there are view models that wrap/subclass other view models that would then need to be migrated, etc.).

I migrate over a small number of view models, varying in criticality of importance (server logs vs connecting to a server) to test the overall implementation. I also simplify some of the states that these objects would have, generally removing a `content` state instead of having the `initial` state just be the sentinel starting state. So, performing the migration would also constitute some cleanup of the entire intra-object state management.

This new system also allows cleanup of our error types. Using the essentially-an`Equatable`-string-box-`JellyfinAPIError` type should be phased out in preference of strongly-typed errors. Part of that migration will be to determine the many different domains that errors could be made.

---

The project is currently structured such that each platform has their own views and the view model/models are all shared. This was largely done due to platform differences in design and behavior and was an easy decision. However, through a combination of SwiftUI's maturity and my own learnings that have converged many systems together - this is largely no longer necessary.

There is complex logic that is copied between platform views (ie: `UserSignInView`), requiring careful consideration that it was performed correctly and the tvOS platform has fallen behind in "it hasn't been implemented on that platform" many times. The point of a declarative framework like SwiftUI is to be platform agnostic (at least where it doesn't leak). In retrospect, I may have taken it too far by creating the strong separation of views.

The `MediaView` was a small and simple enough view to showcase this merge. More complex views will require a more complex merging (like separate `ThisView+Content+[iOS|tvOS]` files), but using tools like `PlatformView`, `Engine`, and `Backport` it's possible now more than ever.

This work should largely also make a macOS client occur naturally.

---

Since these are such large and mostly management changes (vs a feature implementation), these would be good for contributions.


JPKribs (commented):
Tested and working as expected for impacted `ViewModel`s. This builds locally for tvOS. I am unsure why the build process is not working for github. That would be my only holdout. If that's fine/expected, this is good to go!

LePips:
Thanks, I just had to have CI skip macro validation. It's good to go now.

JPKribs (approved):

PR #1717, opened 2025-09-21
`Stateful` Macro - `ActiveSessionsViewModel`

JPKribs:
### Summary

This is the first PR starting to migrate `ViewModels` to use the new `Stateful` macro. This may not be entirely correct so I wanted to start with one `ViewModel` to iron out any quirks before I move to work in batches.

### Other Items

1. `.hourMinute` has been deprecated so I've moved the View to use Duration instead.
2. This `ViewModel` uses `refresh` and `backgroundRefresh`. I believe we can get away with just the single `refresh` but I wanted to leave this as intact as possible. Let me know if you would prefer I remove the `backgroundRefresh`.

### DIFF
--- a/Shared/ViewModels/AdminDashboard/ActiveSessionsViewModel.swift
+++ b/Shared/ViewModels/AdminDashboard/ActiveSessionsViewModel.swift
@@ -13,105 +13,57 @@ import OrderedCollections
 import SwiftUI
 
 @MainActor
-final class ActiveSessionsViewModel: ViewModel, Stateful {
+@Stateful
+final class ActiveSessionsViewModel: ViewModel {
 
-    // MARK: - Action
-
-    enum Action: Equatable {
+    @CasePathable
+    enum Action {
         case refresh
         case backgroundRefresh
-    }
 
-    // MARK: - BackgroundState
+        var transition: Transition {
+            .loop(.refreshing)
+        }
+    }
 
     enum BackgroundState: Hashable {
         case backgroundRefreshing
     }
 
-    // MARK: - State
-
-    enum State: Hashable {
-        case content
-        case error(JellyfinAPIError)
+    enum State {
         case initial
+        case error
+        case refreshing
     }
 
     @Published
     var activeWithinSeconds: Int? = 900 {
         didSet {
-            send(.refresh)
+            refresh()
         }
     }
 
     @Published
     var showSessionType: ActiveSessionFilter = .all {
         didSet {
-            send(.refresh)
+            refresh()
         }
     }
 
     @Published
-    var backgroundStates: Set<BackgroundState> = []
-    @Published
     var sessions: OrderedDictionary<String, BindingBox<SessionInfoDto?>> = [:]
-    @Published
-    var state: State = .initial
 
-    private var sessionTask: AnyCancellable?
+    @Function(\Action.Cases.refresh)
+    private func _refresh() async throws {
+        try await updateSessions()
+    }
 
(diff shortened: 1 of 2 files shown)


Shared/ViewModels/AdminDashboard/ActiveSessionsViewModel.swift
```diff
@@ -13,105 +13,57 @@ import OrderedCollections
...
         case backgroundRefresh
-    }
 
-    // MARK: - BackgroundState
+        var transition: Transition {
+            .loop(.refreshing)
+        }
+    }
 
     enum BackgroundState: Hashable {
```
LePips, line 29: `BackgroundState` no longer requires any conformances.

Shared/ViewModels/AdminDashboard/ActiveSessionsViewModel.swift
```diff
@@ -13,105 +13,57 @@ import OrderedCollections
...
-            return .initial
-        }
+    @Function(\Action.Cases.refresh)
+    private func _refresh() async throws {
+        try await updateSessions()
+    }
+
+    @Function(\Action.Cases.backgroundRefresh)
+    private func _backgroundRefresh() async throws {
+        backgroundStates.insert(.backgroundRefreshing)
```
LePips, line 63: For background states, these are specified in the `Action.transition` value. The state system will automatically insert and remove the background state as necessary.

```swift
var transition: Transition {
    switch self {
    case .refresh: .loop(.refreshing)
    case .backgroundRefresh: .background(.backgroundRefreshing)
    }
}
```

Swiftfin/Views/AdminDashboardView/ActiveSessions/ActiveSessionsView/ActiveSessionsView.swift
```diff
@@ -94,14 +94,14 @@ struct ActiveSessionsView: View {
...
         .onFirstAppear {
-            viewModel.send(.refresh)
+            viewModel.refresh()
         }
         .onReceive(timer) { _ in
             guard !isFiltersPresented else { return }
-            viewModel.send(.backgroundRefresh)
+            viewModel.backgroundRefresh()
         }
         .refreshable {
```
LePips, line 103: Looks like this can't be called so we can remove it.

Swiftfin/Views/AdminDashboardView/ActiveSessions/ActiveSessionsView/ActiveSessionsView.swift
```diff
@@ -117,33 +117,33 @@ struct ActiveSessionsView: View {
             .tag(nil as Int?)
 
             Label(
-                300.formatted(.hourMinute),
+                Duration.seconds(300).formatted(.units(allowed: [.hours, .minutes])),
```
LePips, line 120: Make an `hourMinuteAbbreviated` format style with the other `Duration.UnitsFormatStyle`s in `FormatStyle.swift`.

LePips (changes requested):
This a great first use of my new macro.

> This ViewModel uses refresh and backgroundRefresh. I believe we can get away with just the single refresh

This is a good case where the same functionality can both be a state transition and backgrounded. I'll take into consideration developing this feature but for now it's necessary to have 2 actions.

LePips:
> developing this feature

I did just that and it does clean things up nicely. There seem to also be issues with the CI overall, so I'll merge manually.

PR #1718, opened 2025-09-21
`Stateful` Macro - `APIKeysViewModel` & `AddServerUserViewModel`

JPKribs:
### Summary

I am starting with the Admin Dashboard for migration to `stateful`. This is because these were all my creation so I know them pretty well and where there might be some cleanup and also they are all iOS only so this minimizes the View changes required.

I wanted to do several of these in batch but I was finding a lot of cleanup that I could add along the way so I'm sticking to the first two I have done. I have migrated both `APIKeysViewModel` & `AddServerUserViewModel` to use the new `stateful` macro. I have also done the following changes:

1. On iOS, the `errorMessage` modifier should also do the error vibration/feedback. I've added this to do by default so we don't have to add this to all locations.
2. API Keys had a TODO to add an option to update a key. This deletes the key then recreates it for the same applications. This also included a new label. I would love to use events to update the list instead of doing a full refresh but the API APIs don't allow individual GETs nor do they return the newly created APIs. Not the end of the world since these are pretty light records but something I would like to do if there is an option in the future.
3. Added success vibration/feedback on successful update of an API. 
4. On deleting an API, I am just removing it from the array instead of refreshing. This is the only example where that is possible so it cleans up one unnecessary API call.
5. I updated the previously updated `ActiveSessionsViewModel` to use matching verbiage. So, `.refresh` & `.refreshing` vs `getActiveSessions`.


JPKribs:
@LePips let me know how many of these you want me to do at a time. One at a time feels like it would take up a lot of your time. Too many could get hard to review. I was thinking 3 or 4 if that makes sense to you. Smaller ones like this where there's a lot of other cleanup too.

LePips (approved):
Apologies, I started some cleanup to help explain a comment, which then led me to fixing a bug in the macro, which then led me to other things.

I am in no way stating that the same amount of cleanup must be made in future work. However, I don't recommend adding `MARK` comments in fairly simple contexts, like consistent property and object sites. I do find myself deleting them more often than adding them.

JPKribs:
> Apologies, I started some cleanup to help explain a comment, which then led me to fixing a bug in the macro, which then led me to other things. This work also made me rethink about considering `initial` the sentinel state and removing `content`, I don't think I liked how the api keys view would briefly show the "empty" state before fetching. I will most likely go back on this stance and reintroduce `.content` where I've previously removed it.
> 
> 
> 
> I am in no way stating that the same amount of cleanup must be made in future work. However, I don't recommend adding `MARK` comments in fairly simple contexts, like consistent property and object sites. I do find myself deleting them more often than adding them.

No worries! Thanks for the cleanup. I'll lean off the MARK moving forward. My goal is to be consistent between the view models that we move over to the macro. So, future ones, I'll use these for reference!

Issue #1709, opened 2025-09-20
[iOS] Rotation doesn't work in new media player


samglt:
As noted in #1581, there are some issues when rotating the new player view. Transcoded video doesn't properly fill the screen and locks up until you force quit the app, and on my device non-transcoded video also doesn't fill the screen, but you can still rotate back to portrait without quitting the app.

This happens on iOS 18 & 26 on my devices

<details><summary>Example</summary>
<p>

https://github.com/user-attachments/assets/9303c7e1-bb4a-4fa1-a6f7-2e77833e23d0

</p>
</details>

samglt:
Looking into it more, I think the screen filling issue is related to the NavigationRoute's .fullscreen transition style, since I can replace the [VideoPlayerViewShim here](https://github.com/jellyfin/Swiftfin/blob/3563a0605aec7de3b7c15550d167bbe81ddb4137/Shared/Coordinators/Navigation/NavigationRoute/NavigationRoute%2BMedia.swift#L84) with a Rectangle and the screen still doesn't fill on rotation. I can also use change .fullscreen to .push and the player rotates fine.

JPKribs:
I was looking at this for #1713 and I was noticing that the fullscreen for iOS is a lot busier compared to tvOS. This could be related to what's going on here:

NavigationInjectionView.swift:

```swift
.presentation(
    $coordinator.presentedFullScreen,
    transition: .zoomIfAvailable(
        options: .init(
            dimmingVisualEffect: .systemThickMaterialDark,
            options: .init(
                isInteractive: isPresentationInteractive
            )
        ),
        otherwise: .slide(.init(edge: .bottom), options: .init(isInteractive: isPresentationInteractive))
    )
) { routeBinding, _ in
    let vc = UIPreferencesHostingController {
        NavigationInjectionView(coordinator: .init()) {
            routeBinding.wrappedValue.destination
                .environment(\.presentationControllerShouldDismiss, $isPresentationInteractive)
        }
    }

    // TODO: presentation options for customizing background color, dimming effect, etc.
    vc.view.backgroundColor = .black

    return vc
}
```

samglt:
That's part of the new player change, it was simpler previously. I haven't checked what happens if you replace it with the previous commit

LePips:
On iOS/iPadOS `Transmission.PresentationLink` is used to get more complex behavior. In the app delegate, the top-most `UIPreferencesHostingController` is then used to determine the orientation mask.

[...]

PR #1719, opened 2025-09-22
Pin `Transmission` and fix severe hang

LePips:
Part of #1709 but not closing for now.

During development I used the `2.4.5` version of Transmission, pinning now as updating [past this version](https://github.com/jellyfin/Swiftfin/commit/e7d65fc423be1609c12f5de89c45d388b6a35819#diff-848025bfcbf263b40f46f4a2547f764c1261a2445d0fbfd4eb9717c6bd9403bbR287-L289) makes me experience the issue every time. I'll test some more and possibly make a POC to file an issue if needs be.

Part of the issue is also a severe hang with my `WithObservedObject` view.

### DIFF
--- a/PreferencesView/Package.swift
+++ b/PreferencesView/Package.swift
@@ -5,8 +5,8 @@ import PackageDescription
 let package = Package(
     name: "PreferencesView",
     platforms: [
-        .iOS(.v15),
-        .tvOS(.v15),
+        .iOS(.v16),
+        .tvOS(.v16),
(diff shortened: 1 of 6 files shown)


JPKribs (approved):
Tested on iPhone 13 Pro Max & iPhone 17 Pro (simulator). Tested on iPad M3 (simulator). Tested on both Transcode & DirectPlay. This has been resolved for these devices.

PR #1621, opened 2025-07-03
`ActiveSession` Manager

JPKribs:
### Summary

The core functionality of this that I want to ensure we have the options to send commands to an `ActiveSession`. There are 3 core commands that I already are covered in this PR. These are `Play/Pause`, `Stop`, & `Send a Message`:

Screenshot 2025-07-07 at 12 57 31

From the `ActiveSessionDetails` admins are able to send a message to a user, stop their playback or pause/play playback (if active). This should mirror what is available on Web.

PR #1720, opened 2025-09-22
`Stateful` Macro - Server Devices & Details

JPKribs:
### Summary

Updates Server `Devices` & `DeviceDetails`. There was a lot that I could cut from this as it was not being used. Specifically, device details aren't refreshable nor do we have the option to reload them. This comes from a static variable passed in and the only updating we do is to the custom name. The deviceDTO is checked on entry so there isn't a real world where there is anything that can be refreshed.

Only major change is if there is no custom name I am just going to show this with no custom name. This stops there from always being an active save button. I also removed the success alert when editing the custom name. This is not used anywhere else in Swiftfin so this is out of place. Instead, I am using the success vibration. This should be enough as this isn't a real change, just a label.

There might be more changes [later](https://github.com/jellyfin/Swiftfin/pull/1621) but those are likely in their own viewModel.


JPKribs:
@LePips This is done and working how it was prior to this change. One item I wanted to look into:

If I update the custom name for a devices, the `DevicesView` doesn't reflect it until I leave and come back. This seems a little overkill for a notification but I couldn't figure out a good way to pass this in as a binding without a lot of rework.

Shared/ViewModels/AdminDashboard/DeviceDetailViewModel.swift
```diff
@@ -11,96 +11,50 @@ import Foundation
...
 
-    enum BackgroundState: Hashable {
+    enum BackgroundState {
         case updating
     }
 
-    enum State: Hashable {
-        case initial
+    enum Event {
+        case error(JellyfinAPIError)
```
LePips: The error case no longer needs an associated type.

Shared/ViewModels/AdminDashboard/DeviceDetailViewModel.swift
```diff
@@ -11,96 +11,50 @@ import Foundation
...
+        case updatedCustomName
     }
 
-    @Published
-    var backgroundStates: Set<BackgroundState> = []
-    @Published
-    var state: State = .initial
+    enum State {
```
LePips: If there is only an `initial` state, the `State` enum actually doesn't need to be specified - it will automatically generate one.

Shared/ViewModels/AdminDashboard/DevicesViewModel.swift
```diff
@@ -12,121 +12,55 @@ import JellyfinAPI
...
+        case refreshing
     }
 
-    // MARK: - State
-
-    enum State: Hashable {
-        case initial
-        case content
+    enum Event {
         case error(JellyfinAPIError)
```
LePips: If there's a `State.error`, an `Event.error` is not necessary. `Event.error` would be used when the system doesn't enter a full error state, but may still publish errors (which we use with .alertMessage).

These `.error` cases just add the error variable and an error function, so having both would be redundant.

Shared/ViewModels/AdminDashboard/DevicesViewModel.swift
```diff
@@ -12,121 +12,55 @@ import JellyfinAPI
...
-    var state: State = .initial
+    private(set) var devices: [DeviceInfoDto] = []
 
-    var events: AnyPublisher<Event, Never> {
-        eventSubject
-            .receive(on: RunLoop.main)
-            .eraseToAnyPublisher()
+    @Function(\Action.Cases.refresh)
+    private func _refresh() async throws {
+        try await loadDevices()
```
LePips: Just have this function be the content of `loadDevices`, and then call `_refresh` where it was previously used.

JPKribs:
> @LePips This is done and working how it was prior to this change. One item I wanted to look into:
> 
> If I update the custom name for a devices, the `DevicesView` doesn't reflect it until I leave and come back. This seems a little overkill for a notification but I couldn't figure out a good way to pass this in as a binding without a lot of rework.

I actually think I figured out how to do this right. The `DeviceDetailsViewModel` is really only used to update the `customName` and nothing else. If I move the update function to the `DevicesViewModel`, I can pass this into the view as an `ObservedObject` and the update will pass back to the `DevicesView`. This also means I can delete `DeviceDetailsViewModel` and consolidate everything into the `DevicesViewModel`.

Only downside to this approach is you can enter the view without having a `DevicesViewModel`. So, I've added another `init` where one will be created for you if you do not have one. Since we are not refreshing, this shouldn't be a heavy init as there will be no data in it.

I've also updated the update to just take the id and the `DeviceUpdateDto` so future requirements we might have for updating can be handled by this same function. This is in place of building the `DeviceUpdateDto` ourselves from the `CustomName`

Swiftfin/Views/AdminDashboardView/ServerDevices/DeviceDetailsView/DeviceDetailsView.swift
```diff
@@ -12,101 +12,94 @@ import SwiftUI
...
+
+extension DeviceDetailsView {
+
+    init(device: DeviceInfoDto, viewModel: DevicesViewModel) {
+        self.device = device
+        self.viewModel = viewModel
+        self.temporaryCustomName = device.customName ?? ""
+    }
+
+    init(device: DeviceInfoDto) {
```
LePips: At least for now, we don't have a use case where device details wouldn't be presented without a view model - so we can remove this init. Additionally, this kind of functionality (having a default value) would have had to use a `StateObject` instead.

Swiftfin/Views/AdminDashboardView/ServerDevices/DeviceDetailsView/DeviceDetailsView.swift
```diff
@@ -12,101 +12,94 @@ import SwiftUI
...
-            .disabled(temporaryCustomName == viewModel.device.customName)
-        }
-        .alert(
-            L10n.success.text,
-            isPresented: $isPresentingSuccess
-        ) {
-            Button(L10n.dismiss, role: .cancel)
-        } message: {
-            Text(L10n.customDeviceNameSaved(temporaryCustomName))
+            .disabled(temporaryCustomName.isEmpty || temporaryCustomName == device.customName)
```
LePips: We should still allow the ability to remove the custom name so the button should still be enabled if it's empty. However, when it is empty, we need to send `nil` for the custom name. This is possible with the following map:

```swift
text: $temporaryCustomName.map(
    getter: { $0 ?? "" },
    setter: { $0.isEmpty ? nil : $0 }
)
```

Shared/ViewModels/AdminDashboard/DevicesViewModel.swift
```diff
@@ -12,136 +12,78 @@ import JellyfinAPI
...
+        }
     }
 
-    // MARK: - BackgroundState
-
-    enum BackgroundState: Hashable {
-        case refreshing
-        case updating
+    enum BackgroundState {
         case deleting
```
LePips: The `deleting` background state doesn't seem to have any UI components and can be removed.

Shared/ViewModels/AdminDashboard/DevicesViewModel.swift
```diff
@@ -152,11 +94,9 @@ final class DevicesViewModel: ViewModel, Eventful, Stateful {
         let request = Paths.deleteDevice(id: id)
         try await userSession.client.send(request)
 
-        try await loadDevices()
+        try await _refresh()
     }
 
-    // MARK: - Delete Devices
-
     private func deleteDevices(ids: [String]) async throws {
```
LePips: The contents of this function can be moved to where it's used.

Shared/ViewModels/AdminDashboard/DevicesViewModel.swift
```diff
@@ -152,11 +94,9 @@ final class DevicesViewModel: ViewModel, Eventful, Stateful {
         let request = Paths.deleteDevice(id: id)
         try await userSession.client.send(request)
 
-        try await loadDevices()
+        try await _refresh()
```
LePips: We should not be refreshing after we delete each individual item - we are refreshing after we do the entire list.

Shared/ViewModels/AdminDashboard/DevicesViewModel.swift
```diff
@@ -12,136 +12,78 @@ import JellyfinAPI
...
 
-        await MainActor.run {
-            self.devices = devices.sorted(using: \.dateLastActivity)
-                .reversed()
+        self.devices = devices.sorted(using: \.dateLastActivity)
+            .reversed()
+    }
+
+    @Function(\Action.Cases.update)
+    private func _updateDevice(_ id: String, _ options: DeviceOptionsDto) async throws {
```
LePips: Functions that have the `Function` macro applied don't _need_ to be underscored. The reason why some are underscored is because the `Stateful` macro will generate functions of the same name as the `Action` cases (ie: `Action.refresh` will generate a `refresh` function). So, using underscored names is necessary to resolve ambiguity errors - mainly in the case where I can't think of a better name for the `Function`-attached function than for it to be the same as the `Action` case. (ie: `Action.refresh -> _refresh`). These are the rules of the macros, for clarification.

What I'm pointing out here is that we have an action `Action.update`, which is attached via `Function` to a function named `_updateDevice`. By what I've explained above about the macro, the name of the function could just be `updateDevice`. However, we should pause whenever we have a different `Action` case name vs a function name and think about which name expresses better intent. If you feel like `updateDevice` would be better as a function to call, then that should be the action name. Or, if you feel like `update` is just enough, then this should probably just be `_update`. I'm leaning more towards the former.

We should take this thinking forward during all of the migrations.

Shared/ViewModels/AdminDashboard/DevicesViewModel.swift
```diff
@@ -12,136 +12,78 @@ import JellyfinAPI
...
-        case success
-    }
-
-    // MARK: - Action
-
-    enum Action: Equatable {
+    @CasePathable
+    enum Action {
         case refresh
         case delete(ids: [String])
```
LePips: Have this action take a `Set<String>` instead so we don't need to cast to an `Array` on the view side.

Shared/ViewModels/AdminDashboard/DevicesViewModel.swift
```diff
@@ -12,136 +12,78 @@ import JellyfinAPI
...
     }
 
-    // MARK: - State
+    enum Event {
+        case deleted
+        case updatedCustomName
+    }
 
-    enum State: Hashable {
+    enum State {
```
LePips, line 44: I would like to have the case names alphabetical.

Shared/ViewModels/AdminDashboard/DevicesViewModel.swift
```diff
@@ -12,136 +12,78 @@ import JellyfinAPI
...
+        self.devices = devices.sorted(using: \.dateLastActivity)
+            .reversed()
+    }
+
+    @Function(\Action.Cases.update)
+    private func _updateDevice(_ id: String, _ options: DeviceOptionsDto) async throws {
+        let request = Paths.updateDeviceOptions(id: id, options)
+        try await userSession.client.send(request)
+
+        if let index = devices.firstIndex(where: { $0.id == id }), let customName = options.customName {
```
LePips: Since this view model functions largely on `id`s, it may be more fitting to have the `devices` variable be an `IdentifiedArray`.

Swiftfin/Views/AdminDashboardView/ServerDevices/DeviceDetailsView/DeviceDetailsView.swift
JPKribs: Good call!

Shared/ViewModels/AdminDashboard/DevicesViewModel.swift
JPKribs: Apologies, this came from a previous version where the action was `updateDevice` which I shortened to `update`. I think `update` makes sense contextually and helps keep this more in line with the other migrated ViewModels so far.

Shared/ViewModels/AdminDashboard/DevicesViewModel.swift
JPKribs: I agree we should use an `IdentifiedArray` but unfortunately, device records aren't unique on the server's side. So it will build but I can force a crash by updating / installing the TestFlight which then gives me:

`Task 114: Fatal error: Duplicate key: 'Optional("iOS_A6FFD9EF-1DB6-406A-AB6D-A1C5DDE3A5C1")'`

I can alternatively create a composite ID that is unique. IMO, this seems like a Server oversight as well as deleting `iOS_A6FFD9EF-1DB6-406A-AB6D-A1C5DDE3A5C1` deletes both devices from this list as well.

JPKribs: This is all in my last commit. All else should be addressed. Let me know how you'd like to proceed for this! 

Aside from this, I believe this should otherwise be ready.

LePips: Ah, I see and now understand better what this endpoint actually is as a user-device connection list.

I could imagine an enhanced view where we group entries by device ID, and then the entries (user/timestamp) can be selected among that device. But, this only works since the devices/entries endpoint isn't paginated and we're given the full list and could process the results that way. I think a rework of this endpoint would be necessary on the server side anyways.

PR #1723, opened 2025-09-22
Change one instance of "ratings" string for "parentalRating"

rasko-dev:
Changes the string "Ratings" in the parental controls menu to instead use "Parental rating". Mostly for consistency for translated strings, because other places the "ratings" string is used are for the score ratings rather than age rating, which is a different word in some languages.


JPKribs (approved):
> because other places the "ratings" string is used are for the score ratings rather than age rating, which is a different word in some languages.

I didn't even consider this! Thanks for the PR!

PR #1724, opened 2025-09-22
Settings Fixes

### LINKED ISSUE #1722: [iOS] Next up days count does not match edit dialog count
### Describe the bug

1. tap profiel icon
2. accessibility -> customize (weird this is under accessibility header but whatever)
3. swipe down to 'home' section and note the days next to 'Days in next up'
4. tap it to edit
5. note the days are different, they seem off by 1)

Image

I'm pretty sure this showed up AFTER swapping to a new iPhone so going to iOS 26. It's probably related to that, as that version of tvOS also seems to have fixed the weird popup button text being wrong on tvOS.

### Application version

1.4 (13)

TestFlight

### Device information

iPhone 17 Pro

### OS version

iOS 26

### Jellyfin server version

10.10.7

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1722

Small cleanup. Use a duration label instead of time interval. The issue is, once getting past ~41, the `maxDaysNextUp` would get off by 1 day. This resolves this:

<details>

<summary>Now Working</summary>

https://github.com/user-attachments/assets/a301c4af-b541-4cb7-9c4e-ee911983e595

</details>

Fixed the ratings where both Community and Critic ratings were capped to 10.0. Community ratings should allow for 2 decimals (This is what we get from server) and Critic ratings are out of 100. This did not impact the values we got from server but this did prevent users from making updates to these values from Swiftfin.

This one was going to be part of the ViewModel migrations to `Stateful` but I just ran into this personally so it made the short list.

<details>

<summary>Now Working</summary>

https://github.com/user-attachments/assets/4f2d3a5b-eb4f-4e56-88e6-6c36064b92c3

</details>


Swiftfin/Views/ItemEditorView/ItemMetadata/EditMetadataView/Components/Sections/ReviewsSection.swift
```diff
@@ -27,7 +27,7 @@ extension EditMetadataView {
                 ChevronButton(
                     L10n.critics,
                     subtitle: item.criticRating.map { "\($0)" } ?? .emptyDash,
-                    description: L10n.ratingDescription(L10n.critics)
+                    description: L10n.ratingDescription(L10n.critics, 100)
```
LePips: I made a comment a long time ago about composed strings like this and how they aren't valid, these must have slipped through my eyes. To repeat that concern, we cannot have (outer) translated strings take in other (inner) translated strings since they may become invalid as the context of the outer string may determine what the inner string translates to.

These composed strings must have their own string entries.

LePips (changes requested):

Swiftfin/Views/ItemEditorView/ItemMetadata/EditMetadataView/Components/Sections/ReviewsSection.swift
JPKribs: Good call!

LePips (approved):
A correction in how the values should be formatted vs presented. The `TextField.format` will perform the formatting even on the initial value. So, if the value was `1.123` and the user selects this to edit, the value will automatically change to `1.12`. It doesn't really matter what the value is as it's fine that the user puts whatever but we should only display up to a certain precision regardless.

Issue #1502, opened 2025-04-16
Lost translations?


LePips:
We needed to fix the base english file encoding here: https://github.com/jellyfin/Swiftfin/pull/1491

I don't check the Weblate at all and don't scroll down on the README often. I just noticed at date of reporting (April 16, 2025) that the Weblate bar graph dipped from mostly translated (from recent memory) to mid-20% translated. The files themselves don't exhibit deletions in git history by Weblate.

This is just an investigatory issue to see if, actually: the keys for translations have long been incorrect and after changing the base file encoding to UTF16 that they get picked up by Weblate for translation.

JPKribs:
I *think* WebPlate wasn't updating but 🤷‍♂️. I've started seeing commits from WebPlates again which I don't remember seeing for a while?

Mostly, I'm just trying to evade responsibility for #1491 haha

PR #1730, opened 2025-09-23
Label Cleanup

JPKribs:
### Summary

This PR does the following:

1. Cleanup localization to sentence casing for various settings & admin dashboard items.
2. Cleanup footer and descriptions so we are consistent in our usage of punctuation.
3. Use `string.localizedCapitalized` for strings used in `NavigationTitle`s where there are 2+ words. This shouldn't be required for single word usages. Unless, I guess there could be 2+ words in a translated?
4. Creates specialized localizations for non-standard capitalization (TV not Tv, API not Api, etc.). A few of these are left capitalized by default where I believe a section or proper noun is apparent. Things like Quick Connect, Recently Added, & Up Next where it refers to a specific thing.
6. Consistently use plural cases in areas where other settings are plural (Dashboard -> Users -> Access -> Parental Rating***S***.
7. I ran the `PurgeUnusedStrings.swift` to clean up missing localization. I also found where https://github.com/jellyfin/Swiftfin/issues/1502 originated as `PurgeUnusedStrings.swift` converts our translation files to UTF8 which webplates cannot use. I've updated this to use the appropriate UTF16.
8. Split the action Bar and Menu buttons into 2 different router paths to include the `NavigationTitle`s at the top.

### Script Results

Screenshot 2025-09-23 at 13 45 19

### DIFF
--- a/Shared/Objects/MediaPlayerManager/MediaPlayerItem/MediaPlayerItem+Build.swift
+++ b/Shared/Objects/MediaPlayerManager/MediaPlayerItem/MediaPlayerItem+Build.swift
@@ -75,7 +75,10 @@ extension MediaPlayerItem {
         playbackInfo.liveStreamID = initialMediaSource.liveStreamID
         playbackInfo.maxStreamingBitrate = maxBitrate
         playbackInfo.userID = userSession.user.id
-        playbackInfo.mediaSourceID = initialMediaSource.id
+
+        if !item.isLiveStream {
+            playbackInfo.mediaSourceID = initialMediaSource.id
+        }
 
         let request = Paths.getPostedPlaybackInfo(
             itemID: itemID,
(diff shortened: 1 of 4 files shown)


Swiftfin/Views/SettingsView/VideoPlayerSettingsView/VideoPlayerSettingsView.swift
```diff
@@ -17,44 +17,58 @@ struct VideoPlayerSettingsView: View {
     private var jumpForwardLength
     @Default(.VideoPlayer.resumeOffset)
     private var resumeOffset
+    @Default(.VideoPlayer.videoPlayerType)
+    private var videoPlayerType
```
JPKribs: I'm drawing a blank. I was sure there was a way to call this inline like a variable where it's immutable but I couldn't find references to using Defaults this way only CoreStore.

JPKribs:
@LePips okay ready to go again. This is now purely labels, mostly capitalization and remove unused. A couple of new plurals.

JPKribs:
> I only changed the `PIN` to normal casing, as I don't consider us using the acronym here.

Makes sense. I was going based on "Personal Identification Number" but I think that's only applicable to banks? Regardless, I think that's a fair call!

Issue #1737, opened 2025-09-26
[iOS] Various UI elements read out by VoiceOver unneccessarily


samglt:
Some UI elements are are not hidden from voiceover and clutter up the interface.

The two examples I've found are:
* Poster images get OCRed and and read out after the accessibilityLabel text gets read out, leading to most items getting announced twice
* The backdrop and logo images in item view are focusable and have unnecessary labels that get read out

LePips:
The specific examples have been addressed. Create separate issues for other elements that are erroneously detected for accessibility.

PR #1740, opened 2025-09-26
[iOS] Hide elements in ItemView & PosterButton from voiceover

samglt:
#1737 

This hides the backdrop image in item views from VO, gives the logo images accessibility labels, and disables OCR on poster images.

For some reason the logo images sometimes still got their text read out even though I tried ignoring their contents with accessibilityElement(), so I've hidden them since the navigation title already contains the item's title anyway.

Demo:

https://github.com/user-attachments/assets/7420fc52-d459-49d4-8431-8b5a3b38a8ff

### DIFF
--- a/Swiftfin/Components/PosterButton.swift
+++ b/Swiftfin/Components/PosterButton.swift
@@ -63,6 +63,7 @@ struct PosterButton<Item: Poster>: View {
                 .backport
                 .matchedTransitionSource(id: "item", in: namespace)
                 .posterShadow()
+                .accessibilityElement(children: .ignore)
 
             label
                 .eraseToAnyView()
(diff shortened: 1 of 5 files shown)


Swiftfin/Components/PosterButton.swift
```diff
@@ -63,6 +63,7 @@ struct PosterButton<Item: Poster>: View {
                 .backport
                 .matchedTransitionSource(id: "item", in: namespace)
                 .posterShadow()
+                .accessibilityElement(children: .ignore)
```
LePips, line 66: We're gonna have to think more about how to properly set `PosterButton`, as there are a variety of states we have to consider where this direct ignore may not be correct.

- when `Show poster labels` is disabled, the title text is hidden from the label. Instead, an abbreviated title is shown on the poster in the placeholder/failure state.
- overlay elements like watched, progress, and favorited

LePips (changes requested):

LePips:
Closing due to changes to headers which no longer have image accessible to Voiceover. Logo now has accessibility title of the item's title.

PR #1741, opened 2025-09-26
[iOS] VoiceOver-only button to toggle player overlay

### LINKED ISSUE #1733: [iOS] Player overlay not accessible from VoiceOver
The overlay buttons in the video player can't be accessed by voiceover once they've faded out. This makes the player mostly unusable at best and usually impossible to exit. The way the native player handles it is by having a fullscreen "button" that toggles the overlay.

As an aside, this would also be helpful for automated screenshots/XCUITests since they behave similar to voiceover

samglt:
Fixes #1733 

This adds an invisible voiceover-only button to the VideoPlayer which toggles the overlay and keeps it visible rather than fading it out in a couple seconds. I also simplified `setPlaybackControlsVisibility` since I noticed some redundant ifs and stuff

### DIFF
--- a/Shared/Components/VideoPlayer.swift
+++ b/Shared/Components/VideoPlayer.swift
@@ -66,6 +66,17 @@ struct VideoPlayer: View {
         .onSceneDidEnterBackground {
             manager.set(playbackRequestStatus: .paused)
         }
+        .overlay {
+            Rectangle()
+                .hidden()
+                .accessibilityRepresentation {
+                    Button {
+                        containerState.accessibilityToggleOverlay()
+                    } label: {
+                        Text("Toggle Overlay")
+                    }
+                }
+        }
     }
 
     var body: some View {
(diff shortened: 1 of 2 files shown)


LePips (changes requested):
Since this is mostly the start of introducing more complex accessibility elements, I'm going to think about how to do it foundationally so that we don't have to refactor it all in the future.

We shouldn't be introducing new `accessibility_` functions just for it, but rather enhance the architectures (in this case, the video player). We may find that while working on accessibility features, it allows us to clean up the general case.

samglt:
In terms of general accessibility elements, if I come up with some thoughts worth posting I'll put them in the Help Wanted discussion.

zoli211:
Heads up before I open a PR. I ended up covering this along with the other VoiceOver issues in the player, and I kept your `accessibilityToggleOverlay` and credited you on the commit. Happy to step back if you'd rather finish this one.

PR #1751, opened 2025-09-30
Remove id guard

lostb1t:
So this is an edge case. But i have a plugin that uses an external search engine. And ids might not
match, breaking swiftfin search for me

I know this is an edge case but i don’t think removing the guard is in anyway harmful.

And if there's a mismatch i would say it would be intentional like my case

### DIFF
--- a/Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto.swift
+++ b/Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto.swift
@@ -511,10 +511,6 @@ extension BaseItemDto {
         let response = try await userSession.client.send(request)
         let newItem = response.value
 
-        guard newItem.id == id else {
-            throw JellyfinAPIError("Mismatching item IDs")
-        }
-
         return newItem
     }
 }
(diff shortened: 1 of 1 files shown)


LePips:
I understand the use case but it does raise a flag that we would ever get a different id here. Is there anything about the `BaseItemDTo.locationType` on these items where we can also allow returning the item?

lostb1t:
Not really, they can be both local and remote.

The plugin matches on providerids. But thats not bullet proof.

looking at the code i don't really understand what issue the guard tries to solve. It already request the item by ID. At this point if an id does not match you can safely assume its intentional

LePips (approved):
It's just a sanity check, as they never really hurt.

I do take issue with this operation not being invariant over the item id. Identification is used throughout the app to ensure proper item presentation and updating - regardless of the use case. I can never assume something like this is "intentional", the plugin is arguably breaking basic guarantees clients can make about server data. While there may be no harm in this function's immediate context, I can see issues elsewhere throughout the app I can't prevent.

I'm fine to relax this check, but I did add a comment when I inevitably forget why I did so.

lostb1t:
yeah sounds fair.

thanks!

PR #1759, opened 2025-10-12
Add session seed recovery flow for tvOS logout issue

davedean:
## Why

tvOS was purging our CoreStore cache whenever storage dipped, which wiped signed-in sessions. Users were tossed back to the login screen even though their tokens were still safe in the keychain. We needed to treat CoreData as disposable and rebuild it from a durable seed.

## What

- Added a SessionSeed model plus SessionSeedStore that serializes the minimal user/server details into the keychain and indexes them via Defaults.
- Hooked sign-in, server updates, local security changes, and deletions to keep those seeds in sync; cleanup wipes the corresponding keychain entries.
- Introduced a SessionRestorationService that kicks in when tvOS nukes the cache: it detects an empty store, shows a “Restoring Session…” overlay, re-fetches /Users/Me and /System/Info, repopulates CoreStore/StoredValues, and gracefully signs out if recovery fails.
- Added a lightweight SessionRestoreView and bootstrapped the flow through RootCoordinator.
- Dropped in unit coverage (SessionSeedStoreTests) with a mock keychain, plus the necessary Xcode target wiring.

## Testing

- Manual: sign in, delete Library/Caches/.../Swiftfin.sqlite*, relaunch, verify the session restores without re-authentication; run through server edits, PIN toggles, and sign-out to confirm seeds update/delete correctly.
- xcodebuild -scheme Swiftfin -destination "platform=iOS Simulator,name=iPhone 15" (fails here due to CoreSimulator access, but passes locally).

### DIFF
--- a/Shared/Coordinators/Root/RootCoordinator.swift
+++ b/Shared/Coordinators/Root/RootCoordinator.swift
@@ -24,20 +24,25 @@ final class RootCoordinator: ObservableObject {
             do {
                 try await SwiftfinStore.setupDataStack()
 
-                if Container.shared.currentUserSession() != nil, !Defaults[.signOutOnClose] {
-                    #if os(tvOS)
+                let restorationService = Container.shared.sessionRestorationService()
+                await restorationService.migrateExistingSeeds()
+
+                if await restorationService.needsRestoration() {
                     await MainActor.run {
-                        root(.mainTab)
+                        root(.sessionRestore)
                     }
-                    #else
-                    await MainActor.run {
-                        root(.serverCheck)
+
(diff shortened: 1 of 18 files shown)


davedean:
> Apologies, but this isn't the solution that we need for this problem. 

Perfect, thanks for the feedback! I knew I could be off target.

> What we need to do is make the backing store (UserDefaults, CoreData, maybe even Keychain) of `StoredValue` customizable so that for different values and platforms we can specify what backing store to use - UserDefaults in the tvOS case (500KB is enough for the platform with some constraints).

I'll see if I can get somewhere on this in a future PR.

Thanks for the response, apologies for the noise.

PR #1769, opened 2025-10-20
TV Shows watching indicators

hqueiroga:
Added Watched / Unwatched Indicators for TV shows, keeping consistency with the Library screen.

Simulator Screenshot - Apple TV 4K (3rd generation) (at 1080p) - 2025-10-20 at 10 47 34

Simulator Screenshot - iPhone 17 Pro - 2025-10-20 at 19 59 04

Simulator Screenshot - iPhone 17 Pro - 2025-10-20 at 19 59 14

### DIFF
--- a/Swiftfin/Views/ItemView/Components/EpisodeSelector/Components/EpisodeCard.swift
+++ b/Swiftfin/Views/ItemView/Components/EpisodeSelector/Components/EpisodeCard.swift
@@ -16,8 +16,6 @@ extension SeriesEpisodeSelector {
 
         @Default(.accentColor)
         private var accentColor
-        @Default(.Customization.Indicators.showUnplayed)
-        private var showUnplayed
         @Default(.Customization.Indicators.showPlayed)
         private var showPlayed
 
@@ -36,13 +34,8 @@ extension SeriesEpisodeSelector {
                     title: progressLabel,
                     progress: (episode.userData?.playedPercentage ?? 0) / 100
                 )
-            } else if episode.userData?.isPlayed ?? false {
+            } else if episode.userData?.isPlayed ?? false, showPlayed {
                 WatchedIndicator(size: 25)
-                    .isVisible(showPlayed)
-            } else if episode.canBePlayed && !episode.isLiveStream {
-                UnwatchedIndicator(size: 25)
-                    .foregroundColor(accentColor)
-                    .isVisible(showUnplayed)
             }
         }
 


JPKribs:
I like how this looks and give us some more color coming from the accent. IMO, reusing the watched indicator is smart and I'm in favor of this change! Can we ensure we're doing the same thing for iOS so we have consistency between platforms?

I haven't had a chance to check, but does this take the setting for watch indicator into account as well? I don't believe we'd need/want that but I wanted to check if toggling these on/off is expected behavior.

I'll defer to @LePips whether the unplayed indicator makes sense on the episode picker.

hqueiroga:
@JPKribs I have updated the PR with the iOS change. Kept same style from the Library screen :)

Swiftfin tvOS/Views/ItemView/Components/EpisodeSelector/Components/EpisodeCard.swift
```diff
@@ -30,16 +30,12 @@ extension SeriesEpisodeSelector {
...
-                        Color.clear
-
-                        Image(systemName: "checkmark.circle.fill")
-                            .resizable()
-                            .frame(width: 30, height: 30, alignment: .bottomTrailing)
-                            .symbolRenderingMode(.palette)
-                            .foregroundStyle(.white, .black)
-                            .padding()
-                    }
+                    // Use the same WatchedIndicator as library view for consistency
```
JPKribs: We should be fine to remove these comments. The code makes sense and is self documenting

Swiftfin tvOS/Views/ItemView/Components/EpisodeSelector/Components/EpisodeCard.swift
```diff
@@ -30,16 +30,12 @@ extension SeriesEpisodeSelector {
...
+                    WatchedIndicator(size: 45)
+                } else if episode.canBePlayed && !episode.isLiveStream {
+                    // Use the same UnwatchedIndicator as library view for consistency
+                    UnwatchedIndicator(size: 45)
+                        .foregroundColor(.jellyfinPurple)
```
JPKribs: Please change this to pull the `accentColor` instead of .jellyfinPurple. For now, on tvOS those are the always the same but in the future those might diverge so it'd be good to get this working now. Please see an example of using the `accentColor` here:

https://github.com/jellyfin/Swiftfin/blob/main/Swiftfin/Components/PosterButton.swift#L208

Swiftfin/Views/ItemView/Components/EpisodeSelector/Components/EpisodeCard.swift
```diff
@@ -29,16 +29,12 @@ extension SeriesEpisodeSelector {
...
-                        .symbolRenderingMode(.palette)
-                        .foregroundStyle(.white, .black)
-                        .padding()
-                }
+                // Use the same WatchedIndicator as library view for consistency
+                WatchedIndicator(size: 25)
+            } else if episode.canBePlayed && !episode.isLiveStream {
+                // Use the same UnwatchedIndicator as library view for consistency
+                UnwatchedIndicator(size: 25)
+                    .foregroundColor(.jellyfinPurple)
```
JPKribs: > Please change this to pull the `accentColor` instead of .jellyfinPurple. For now, on tvOS those are the always the same but in the future those might diverge so it'd be good to get this working now. Please see an example of using the `accentColor` here:
> 
> https://github.com/jellyfin/Swiftfin/blob/main/Swiftfin/Components/PosterButton.swift#L208

Same here as above please. The comments as well can be removed.

Swiftfin tvOS/Views/ItemView/Components/EpisodeSelector/Components/EpisodeCard.swift
hqueiroga: comments removed

Swiftfin tvOS/Views/ItemView/Components/EpisodeSelector/Components/EpisodeCard.swift
hqueiroga: Done! accentColor got from Defaults.

Swiftfin/Views/ItemView/Components/EpisodeSelector/Components/EpisodeCard.swift
hqueiroga: Done! accentColor got from Defaults.
Comments also removed.

hqueiroga:
Thanks @LePips for your feedback :) 
Think the watched episode indicator is nice to have, as you may see what you have already watched on the show... but unwatched one, it's not that important... let you to judge what's the best :)

Issue #1725, opened 2025-09-23
Getting error “The data couldn’t be read because it isn’t in the correct format”


TheEvilMonkey:
### Describe the bug

New install of latest Swiftfin from TestFlight, connecting to Jellyfin 10.11.0 rc5

Getting the error listed in the subject on Home tab

### Application version

1.4(13)

### Where did you install the app from?

TestFlight

### Device information

iPhone 15 Pro Max

### OS version

iOS 26

### Jellyfin server version

10.11.0 rc5

JPKribs:
Please provide a screenshot of this error.

Without seeing this, this sounds like 10.11 changes some APIs that we are using. If that is the case, we could not be able to resolve this until the final 10.11 release is out. But a screenshot would assist in troubleshooting

TheEvilMonkey:
<img width="645" height="1398" alt="Image" src="https://github.com/user-attachments/assets/570c5be4-6498-4954-857d-35a47e9965fa" />

JPKribs:
This is an API issue. Swiftfin is currently only supported up to 10.10. This issue occurs because the data your server is sending is not compliant with our structures generated from 10.10.

A previous example between 10.8 and 10.10:

https://github.com/jellyfin/Swiftfin/issues/1289

l2dy:
> Without seeing this, this sounds like 10.11 changes some APIs that we are using. If that is the case, we could not be able to resolve this until the final 10.11 release is out. But a screenshot would assist in troubleshooting

FYI, 10.11 is out!

JPKribs:
> FYI, 10.11 is out!

It's in the works: https://github.com/jellyfin/jellyfin-sdk-swift/pull/49

PR #1715, opened 2025-09-20
[tvOS] tvOS 26 Fixes for Navigation & Updated Linting

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1713

tvOS 26 has changed `.sheet` to mirror on iPadOS where it only fills the center of the screen. This means, our usage of `.sheet` for tvOS navigation is no longer viable in it's current form. There are a few routes that we can take to resolve this so so this will be WIP depending on the route we want to take.

See the current issue below:

https://github.com/user-attachments/assets/84dc7763-90cb-47e1-a36b-103cc81d1a6f

### Fix

https://github.com/user-attachments/assets/a932d972-861d-4bd7-bb3e-bb1acedd7129

PR #1772, opened 2025-10-20
Bump Jellyfin SDK: 10.10 -> 10.11

JPKribs:
### Summary

Companion PR: https://github.com/jellyfin/jellyfin-sdk-swift/pull/49

Resolves: https://github.com/jellyfin/Swiftfin/issues/1725

This bumps the Jellyfin SDK to use the 10.11 SDK. Currently pointed to my branch of the SDK but flipping this to 0.6.0 should work the same when that is available.

Overall, this is a small and fairly clean migration. Please find the outstanding items below. Most of the work will be on the [SDK side](https://github.com/jellyfin/jellyfin-sdk-swift/pull/49).

### TODO

- [x] Change from `JPKribs/10.11` to `Jellyfin/0.6.0` on official release.
- [x] Confirm that profile image removal of `maxWidth` is intentional (fix if reverted).

### DIFF
--- a/Shared/Components/Marquee.swift
+++ b/Shared/Components/Marquee.swift
@@ -41,7 +41,7 @@ struct Marquee<Content>: View where Content: View {
         delay: Double = 2.0,
         gap: CGFloat = 50.0,
         animateWhenFocused: Bool = false,
-        fade: CGFloat = 10.0,
+        fade: CGFloat = 10.0
     ) where Content == Text {
         self.speed = speed
         self.delay = delay
@@ -225,7 +225,7 @@ private struct _OffsetEffect: GeometryEffect {
         set { offset = CGSize(width: newValue.first, height: newValue.second) }
     }
 
-    public func effectValue(size _: CGSize) -> ProjectionTransform {
+    func effectValue(size _: CGSize) -> ProjectionTransform {
         ProjectionTransform(CGAffineTransform(translationX: offset.width, y: offset.height))
     }
 }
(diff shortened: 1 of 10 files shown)


JPKribs:
@LePips if we merge in https://github.com/jellyfin/Swiftfin/pull/1715 first, that has the linting fixes in it already. Or, I can add those to this manually since #1715 is temporary and should be fixed in 26.1. If we're not releasing tvOS before 26.1, I think we can skip that PR then it would be better to move the changes into here.

Shared/Extensions/JellyfinAPI/TranscodeReason.swift
```diff
@@ -66,13 +66,18 @@ extension TranscodeReason: Displayable, SystemImageable {
             return L10n.videoRangeTypeNotSupported
         case .videoCodecTagNotSupported:
             return L10n.videoCodecTagNotSupported
+        case .streamCountExceedsLimit:
+            return L10n.streamCountExceedsLimit
+        @unknown default:
```
LePips: We don't need `@unknown default` as it doesn't do anything. The SDK isn't a library in library evolution mode and extensible enums aren't in a language release yet.

LePips (changes requested):

Shared/Extensions/JellyfinAPI/TranscodeReason.swift
JPKribs: This is resolved in my last commit!

PR #1780, opened 2025-10-24
Update Feature Template

JPKribs:
### Summary

This updates our Feature Template in preparation of project dashboarding. The primary change of this is to appropriately assign new issues to our Issue Type: `Feature` instead of relying on labels.

While I am making these changes, I updated this template to collect additional information and provide additional resources to the requestor prior to the request. This points the user to other requests to see if their request exists prior to requesting. It asks for platform so we can get a split of "Downloads" - iOS only vs "Surround Sound" - tvOS only vs "HDR99++" - All platforms. 

I've added a checkbox for whether a user wants to contribute this feature themselves. The goal of this checkbox is to make the requestor aware that Swiftfin is FOSS and eligible for any contributors. Oftentimes, I hear people think that "only the core team can add features" so this toggle exists to keep users in the loop that they can become contributors. Additionally, this toggle signals to contributors to reach out in the comments to provide resources if someone wants to become a contributor.

Finally, I've removed the option for a blank template. As a contributor, I like having the ability to quickly make a feature/bug issue but with dashboarding it would means those need to be manually tagged, adding admin work. Keeping our options to *just* pre-defined templates helps ensure that our workflows are appropriately triggered and new issues are organized. Looking back over the last couple months, blank templates are used fairly often for bugs and features, leading to additional tagging.

### Questions

- [ ] Does the Platform selector make sense? The downside of this will be feature requests that *should* be multi-platform but the requestor only uses one platform so they only check their preferred platform.
- [ ] Should we leave blank templates in?

### Screenshot:

<details>

<summary>Full Feature Request</summary>

Screenshot 2025-10-24 at 14 29 38

</details>

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,4 +1,4 @@
-blank_issues_enabled: false
+blank_issues_enabled: true
 contact_links:
   - name: Jellyfin documentation
     url:  https://jellyfin.org/docs/general/getting-help.html


.github/ISSUE_TEMPLATE/feature-request.yaml
```diff
@@ -0,0 +1,49 @@
...
+    attributes:
+      label: "This feature request respects the following points:"
+      description: Please help us by ensuring that you meet all the following conditions.
+      options:
+        - label: This feature has **not** already been requested on [GitHub](https://github.com/jellyfin/Swiftfin/issues?q=is%3Aissue%20state%3Aopen%20type%3AFeature).
+          required: true
+        - label: I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
+          required: true
+
+  - type: checkboxes
```
LePips: I'm not worried about this checkbox, as I only consider a PR to be the ultimate signal of an intent to contribute.

LePips (approved):
After thinking about it a bit more, I would like to keep blank issues for now - mainly for myself...

However, we can still close requests that aren't in the feature template when we think they obviously should.

PR #1782, opened 2025-10-27
`@Stateful` - `ServerActivityDetailViewModel`

JPKribs:
### Summary

I was originally planning on doing `ServerActivityViewModel` as well but then I would need to refactor `PagingLibraryViewModel` which is a huge undertaking and also appears to be part of: https://github.com/jellyfin/Swiftfin/pull/1752

So, this is a small refactor to the new `@Stateful` and a TODO in `ServerActivityViewModel`


LePips (approved):
As a note I changed how libraries work in #1752 and have made the changes to `ServerActivityViewModel` already.

PR #1798, opened 2025-11-04
`ErrorView` Cleanup

JPKribs:
### Summary

This PR sets out to combine the `ErrorView` for iOS and tvOS into a single reused View since the only difference between them is sizing. This also migrates `JellyfinAPIError` to `ErrorMessage` per a TODO and changes `ErrorView.onRetry` to instead use the `Environment.refresh`. 

There is some additional random cleanup like removing some of the deprecated `String.text` usages in favor of `Text(String)` since we are getting warnings about that but that's only on the views where I am already making changes.

I’ve removed `DelayedProgressView` is favor of just using the native `ProgressView`. Please let me know if that is incorrect but we had split usage of it and I seem to remember it being on the chopping blocks a while back so I thought I would clean it up while I’m touching all these views anyway.

Finally, I took this as an opportunity to do a bit of a redesign for `ErrorView` to include some additionally information and try to match some of the other styling we have in Swiftfin. I put the logo with a background that matches our `.destructive` `ListRowButton` style. I added a `popOver` to display some server details to help with potential troubleshooting. I added a place for `recoverySuggestions` to go so, as those become available, they will be visible to assist in troubleshooting.

I think the rounded rectangle background to the icon is going to be the most controversial for this view. Feel free to let me know if we'd prefer not to have that. I think it looks nicer on iOS than tvOS.

### Videos

In testing, it should all work the same but now all instances where we were previously using `.onRetry` now use `.refreshable` instead. Here are some examples of some potential image types we could use for other errors. For now, this PR is purely just the `Generic` since we don't have errors types but I was looking at that for `NetworkError`. 

### iOS

**Generic**

https://github.com/user-attachments/assets/9ab277df-da66-40fe-b441-ff79a24678ec

**Network**

https://github.com/user-attachments/assets/08eda9ec-53d5-4568-b2df-936e58ff3df0

**Server**

https://github.com/user-attachments/assets/76fdb6fe-704e-48d9-bd08-4956849095de

### tvOS

**Generic**

https://github.com/user-attachments/assets/fb0e127f-e2a5-4c9f-a3d3-c7eb1f78b5fc

**Network**

https://github.com/user-attachments/assets/feb0a5b8-a372-4c7b-9e37-7183fbb5ede5

**Server**

https://github.com/user-attachments/assets/3ea269fe-981a-44f5-81ec-5f3dade53ce9

### Images

<details>

<summary>tvOS</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-11-05 at 10 36 57

</details>

<details>

<summary>iPadOS</summary>

Simulator Screenshot - iPad Pro 13-inch (M5) - 2025-11-05 at 10 46 20

</details>

<details>

<summary>iOS</summary>

Simulator Screenshot - iPhone 16 Pro - 2025-11-05 at 10 43 36

</details>

### `NetworkError`

Finally, I have decided to create the `NetworkError` object on another PR. I've left TODOs to update the `SystemImage` based on the error type and start using `RecoverySuggestions` which will be done with that `NetworkError`. The `NetworkError` will likely need a couple revisions while this work should be fairly non-controversial.

This can be seen [here](https://github.com/jellyfin/Swiftfin/pull/1800). I'm realizing this probably goes on the SDK level instead so I may be closing that out in favor of that.

### DIFF
--- a/Shared/Components/ErrorView.swift
+++ b/Shared/Components/ErrorView.swift
@@ -7,7 +7,6 @@
 //
 
 import Defaults
-import Factory
 import SwiftUI
 
 #if os(iOS)


JPKribs:
I am using `"xmark.circle"` instead of `"xmark.circle.fill"` because I would eventually like to use things like `"network.slash"` for network errors, etc. I'm game to revert this but that's just my reasoning for moving away from `.fill`. I should have all the types built out in a [`NetworkError`](https://github.com/jellyfin/Swiftfin/pull/1800) PR but I will want feedback to make sure it makes sense.

`"network.slash"` for reference:

Screenshot 2025-11-04 at 13 35 36

---

Eventually, I'd also love to catch unauthorized errors like 401 and 403 and have a secondary button for Switch users when those errors are present to better handle invalid tokens. But for now, I want to keep these changes standalone since I touch a lot of places.

JPKribs:
Eventually, I'd also love to catch unauthorized errors like 401 and 403 and have a secondary button for `Switch users` when those errors are present to better handle invalid tokens. But for now, I want to keep these changes standalone since I touch a lot of places.

JPKribs:
| Platform | Generic | Network | Server |
|:---:|:---:|:---:|:---:|
| iOS | https://github.com/user-attachments/assets/9ab277df-da66-40fe-b441-ff79a24678ec | https://github.com/user-attachments/assets/08eda9ec-53d5-4568-b2df-936e58ff3df0 | https://github.com/user-attachments/assets/76fdb6fe-704e-48d9-bd08-4956849095de |
| tvOS | https://github.com/user-attachments/assets/4b1d5ac3-7bf0-44ac-afb5-5b22b0ca3ec2 | https://github.com/user-attachments/assets/a7896bb3-8573-4928-a9ee-f6d2b39c4100 | https://github.com/user-attachments/assets/2d9bfd60-e0ae-4ad8-ad2f-8d7ef1aefda6 |

JPKribs:
Apologies, a bit of a rework. I re-found [where combining views was mentioned](https://github.com/jellyfin/Swiftfin/pull/1714) and realized I should be using the `PlatformView`. This is functionally still the same but a cleaner implementation without **as many** build flags. I still need those to remove non-iOS/tvOS components for building (`focusSection`, `popOver`, etc) but this greatly cleans this up.

LePips:
Please take a look at my changes. I don't find much value for previewing server information or for there to be design differences between the platforms here.

This finally prompted me to make the `PrimaryButtonStyle` and remove `ListRowButton/PrimaryButton` on each platform.

JPKribs:
> I don't find much value for previewing server information

Primarily, I wanted to use this as an opportunity for for using the the version check. This way, it was more visibly apparent for issues like [this](https://github.com/jellyfin/Swiftfin/issues/1725) that it's a version issue instead of a Swiftfin issue.

How would you feel about 2 changes:

1) Add something like this above the recovery suggestion when the Jellyfin Client version is off:

```swift
Label(
    L10n.serverVersionWarning(JellyfinClient.sdkVersion.majorMinor.description),
    systemImage: "exclamationmark.circle.fill"
)
.labelStyle(.sectionFooterWithImage(imageStyle: .orange))
```

2) Update the version check to to require a Server == Major/Minor opposed to Server >= Major/Minor to better handle people who get a version ahead of us as well since that seems to be our most common culprit for our Data Decode error.

Let me know what you think! I've got this open so I can make that change as needed

---

> prompted me to make the PrimaryButtonStyle and remove ListRowButton/PrimaryButton on each platform.

I'll make an issue to migrate those over as well. We have some new folks looking to help out and that kind of deprecation / find & replace is a great task for new people to get familiar with Swiftfin.

JPKribs:
Not necessary looking like this but functionally something like:

<details>

<summary>Example</summary>

Simulator Screenshot - iPhone 17 Pro - 2025-11-16 at 17 45 43

</details>

LePips:
Checking the server version is to warn the user that their server wouldn't support everything that Swiftfin uses, more as a hint to update their server and shouldn't be present for all errors. I don't want to do the comparison the other way around (SDK is out of date) and present that to the user. Issues like that issue will always be present and are a Swiftfin issue as we must update the generated SDK as necessary.

Something we could eventually add is our own error protocol to replace usage of `LocalizedError` and also provide custom actions for things like 401 errors.

JPKribs:
Makes sense. In hindsight, this would more likely go into `RecoverySuggestions`. If we have a Data decoding error, the suggestion should be to validate that your version of Jellyfin is compatible with Swiftfin. So this additional item is not required.

Testing looks/works as expected with your updates. I think this is good from my end!

PR #1800, opened 2025-11-04
[WIP] `NetworkError`

JPKribs:
### Summary

**This is a work in progress.**

Strongly types our *known* network errors that can come from Jellyfin. Has handling for unknown/new errors. Completes some of the currently unused features of the `LocalizedError` struct like `recoverySuggestion` for usage on the `ErrorView`.

This will come after https://github.com/jellyfin/Swiftfin/pull/1798 and I still need to localize this and configure the `JellyfinClient` to actually use this. Also, this might make more sense in the SDK? Which, means I should probably figure out how to generate this more dynamically opposed to my manual process.


JPKribs:
@LePips just confirming, does this belong on the SDK or is it fine here?

I was starting from the existing, commented out version but I'm realizing the best place for this is probably to catch from the SDK itself?

LePips:
These errors would live here. We should also make `NetworkError` `struct`s rather than enum cases.

JPKribs:
> These errors would live here. We should also make `NetworkError` `struct`s rather than enum cases.

I've made some updates & localized. I still don't think I'm done but I wanted to check if I'm on the right track. I did a protocol for `NetworkError` then built the errors out as `ClientError` (4XX Errors), `ServerError` (5XX Errors), `ConnectionError` (Essentially `NSError`), & `CodingError` (Encode & Decode Error).

Does that sound right for this?

I still want a way to more generically create something for this like `NetworkError(NSError)` -> `*Appropriate Network Error Here`. I've been looking around and the route I've seen is like a `NetworkErrorFactory` enum that returns the correct `NetworkError` type but I don't see anything like that in Swiftfin so I think that's probably incorrect..

JPKribs:
I'm going to close this one out. I remember I started on this but then I got in over my head. I'm not sure how much of this work is reusable for a final product. I'll leave the branch up if there's anything worth salvaging, but otherwise I might look at this at a later point in time from scratch.

Issue #1799, opened 2025-11-04
Replace `.text` Usages


JPKribs:
Find and replace all instances of `String.text` with `Text(String)` as the former is now deprecated.

PR #1806, opened 2025-11-11
#1799 Replace .text Usages

hqueiroga:
Text replacements for https://github.com/jellyfin/Swiftfin/issues/1799

### DIFF
--- a/Shared/Extensions/String.swift
+++ b/Shared/Extensions/String.swift
@@ -72,11 +72,6 @@ extension String {
         return padding + self
     }
 
-    @available(*, deprecated, message: "Use self in a Text instead")
-    var text: Text {
-        Text(self)
-    }
-
     var initials: String {
         split(separator: " ")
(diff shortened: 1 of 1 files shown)


hqueiroga:
@LePips sorry, I've missed a few of them :)
All fixed now.

JPKribs:
@hqueiroga He means lines 75 - 78 in:

/Shared/Extensions/String.swift

```swift
    @available(*, deprecated, message: "Use self in a Text instead")
    var text: Text {
        Text(self)
    }
```

I would recommend using:

Screenshot 2025-11-11 at 14 56 58

to find all instances. Or, if you delete that func, it will also give you build errors if you are missing any usages of it.

hqueiroga:
Got it! Removed the lines from the String.swift file.
Project building for both Swiftfin and Swiftfin tvOS, so think I didn't miss anything this time :)

LePips (approved):

Issue #1801, opened 2025-11-05
Replace `seconds.formatted(.hourMinute)` Usages


JPKribs:
Find and replace all instances of `seconds.formatted(.hourMinute)` with `Duration.seconds(X).formatted(.hourMinuteAbbreviated),` as the former is now deprecated.

---

<img width="330" height="33" alt="Image" src="https://github.com/user-attachments/assets/0f7219b6-f6ae-4370-802a-cba8945a7762" />

---

Once this is complete, please delete the following lines as they are no longer needed:

Lines 13-27 in:

`/Shared/Extensions/FormatStyle.swift`

```swift
struct HourMinuteFormatStyle: FormatStyle {

    func format(_ value: TimeInterval) -> String {
        let formatter = DateComponentsFormatter()
        formatter.unitsStyle = .abbreviated
        formatter.allowedUnits = [.hour, .minute]
        return formatter.string(from: value) ?? .emptyDash
    }
}

extension FormatStyle where Self == HourMinuteFormatStyle {

    @available(*, deprecated, message: "Use `Duration` instead.")
    static var hourMinute: HourMinuteFormatStyle { HourMinuteFormatStyle() }
}
```

PR #1807, opened 2025-11-11
#1801 Replace seconds.formatted(.hourMinute) Usages

hqueiroga:
Changes hour/minute formatting for: https://github.com/jellyfin/Swiftfin/issues/1801


JPKribs:
@hqueiroga Thanks for looking at this! Would you be able to resolve these linting issues?

Swiftfin tvOS/Views/AppSettingsView/Components/SignOutIntervalSection.swift
```diff
@@ -39,7 +39,7 @@ extension AppSettingsView {
                 if signOutOnBackground {
                     ChevronButton(
                         L10n.duration,
-                        subtitle: Text(backgroundSignOutInterval, format: .hourMinute)
+                        subtitle: Text(Duration.seconds(backgroundSignOutInterval).formatted(.hourMinuteAbbreviated))
```
LePips: Use the `Duration.ticks` function instead, along with the `Text(:format:)` init instead of `.formatted(...)`.

hqueiroga: @LePips the default value of backgroundSignOutInterval is **3600**. Just double-checking if we shouldn't be using Duration.seconds() in this case.

I think this would be the correct code:

`Text(Duration.seconds(backgroundSignOutInterval), format: .hourMinuteAbbreviated)`

LePips: Apologies, I only looked at the syntax and not the actual values while on mobile while also making a general statement for all of the changes.

Yes, use the `.seconds` function where applicable, update to use `Text(:format:)` instead where applicable, and don't use `ServerTicks`.

hqueiroga (commented):
@LePips all changed. Some instances we cannot use Text() because we need the actual String. On those cases the formatted(.hourMinuteAbbreviated) was used.

PR #1811, opened 2025-11-13
Removed UIScreen.main.bounds

jfredsilva:
Removed usage of UIDevice.main with CGSize from Geometry Reader.

Work requested on [help wanted](https://github.com/jellyfin/Swiftfin/discussions/1503) discussion


LePips:
Apologies, #1752 addresses this but as a review note this would have required an alternate implementation that doesn't introduce the environment value.

PR #1812, opened 2025-11-13
`@Stateful` - Server Tasks

JPKribs:
### Summary

Migrates `ServerTaskObserver` & `ServerTasksViewModel` to the `Stateful` Macro. Removes unused functions and cleans up the action naming to mirror the rest of Swiftfin.

### Random Cleanup

Changes the Action names to be mirrored for `ActiveSession` since they both operate in the same fashion. `.initial` before first load. `.content` after initial load. Utilize `background.state` for refreshes that are non-destructive / do not reset the contents.

I also found that the `stopObserving` action didn't do anything so I removed it.

Finally, there are some components in the Server Tasks views that were very small and never reused so I moved those back into the main file to declutter. I went a little overboard on creating new files the first go around so this is a bit of cleanup on some old work.

### Macro Issue

Since we are using a switch case on the view types for `.initial` to `.content`, the `ActiveSessionsView` on the TestFlight and now the `ServerTasksView` flicker on refresh. This can be resolved currently by removing `.to(.initial, then: .content)` in favor of purely using the background state. The issue with this is there is no initial state before the first call completes so there isn't good feedback if the first API call takes a while.

<details>

<summary>Example</summary>

https://github.com/user-attachments/assets/509086f3-2a93-4f97-aef4-141dbd76d621

</details>

I could misunderstand, but by using the below:

```swift
.onFirstAppear {
    viewModel.refresh()
}
.onReceive(timer) { _ in
    guard !isFiltersPresented else { return }
    viewModel.background.refresh()
}
```

My understanding is we should only trigger the `.whenBackground` but we are triggering the `.to(.initial, then: .content)` as well. If we can update the Macro to only call the `.whenBackground` when called from the background this should resolve the issue for both views.

Otherwise, I can change this to purely use `.background` instead.

### DIFF
--- a/Swiftfin tvOS/Views/ItemView/Components/ActionButtonHStack/ActionButtonHStack.swift
+++ b/Swiftfin tvOS/Views/ItemView/Components/ActionButtonHStack/ActionButtonHStack.swift
@@ -12,12 +12,6 @@ extension ItemView {
 
     struct ActionButtonHStack: View {
 
-        @StoredValue(.User.enableItemDeletion)
-        private var enableItemDeletion: Bool
-        @StoredValue(.User.enableItemEditing)
-        private var enableItemEditing: Bool
-        @StoredValue(.User.enableCollectionManagement)
-        private var enableCollectionManagement: Bool
         @StoredValue(.User.enabledTrailers)
         private var enabledTrailers: TrailerSelection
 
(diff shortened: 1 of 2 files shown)


JPKribs:
@LePips let me know if you believe that's on the Macro side or if I should remove the state changes.

Working on the last of the Admin Dashboard migration to the Stateful macro, I was finding we have some views with a cancel action.

Should canceling outstanding actions be included on the macro itself or is adding it ad hoc via an action correct?

PR #1816, opened 2025-11-16
Fix flickering issue when switching seasons on tvOS

svrem:
### Summary

Removes flickering issue for season switching in the `SeriesEpisodeSelector` by adding `PlaceholderHStack`, which maintains the height, and adding a transition to smooth over the change in the `EpisodeCard` stack.

### PlaceholderHStack

I added a new HStack in the `SeriesEpsideSelector` which the utilizes the pre-existing `EmptyCard` with the opacity set to zero and the interactions disabled. This ensures that the height of the `SeriesEpisodeSelector` will always remain constant, which removes the flickering. A new element could have created instead of using the `EmptyCard`, but that's just a lot of repeated code for minimal performance gains and it makes future changes in UI/UX harder to implement.

### Transition

Tried it, looked better.

### DIFF
--- a/Swiftfin tvOS/Views/ItemView/Components/EpisodeSelector/Components/HStacks/EpisodeHStack.swift
+++ b/Swiftfin tvOS/Views/ItemView/Components/EpisodeSelector/Components/HStacks/EpisodeHStack.swift
@@ -95,18 +95,22 @@ extension SeriesEpisodeSelector {
 
         var body: some View {
             ZStack {
-                switch viewModel.state {
-                case .content:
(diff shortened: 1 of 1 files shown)


LePips (approved):
This does seem to work around the actual issue, as technically there should always be a resolved `_HStack` to be presented. There must be some race or bug somewhere that is causing the `CollectionHStack`s to resolve into a 0 height for a moment and it's weird that `.hidden` doesn't work properly on them.

Thank you for your contribution and I'm good with this for now. I would want to actually reuse something like `EmptyHStack`, with some refactoring, when we consolidate this component.

Issue #1640, opened 2025-07-13
TvOS focus navigation issue in series details page


OlivierAlbertini:
## Issue summary

When navigating down from season buttons to episodes on the tvOS series details page, focus gets lost instead of moving to the first episode thumbnail. This creates a poor user experience where users have to manually search for focus or repeatedly attempt navigation.

### Devices
Apple Tv 4k A1842
tvOS 18.5[22L572]

Apple TV (just before the 4k one)
tvOS 18.5[22L572]

Note: It seems not to happen on the simulator Apple Tv 4k 3rd generation (18.4)

### Initial Symptoms
- Focus disappears when navigating from season selector to episodes section
- Sometimes focus navigation works, sometimes it doesn't
- User gets "stuck" around season buttons unable to proceed smoothly
- Issue does NOT occur in the old Swiftfin project

### The race condition
1. User navigates down from seasons to episodes
2. Focus guide `onContentFocus` fires → calls `getContentFocus()` → sets `focusedEpisodeID`
3. Season selection change fires → calls `getContentFocus()` again → overwrites `focusedEpisodeID`
4. Multiple rapid assignments cause SwiftUI focus system to fail
5. Result: `focusedEpisodeID` ends up as `nil` despite valid assignments

### Debug logs showing the issue
```
🎯 EpisodeHStack: Focus guide onContentFocus triggered
🎯 EpisodeHStack: getContentFocus() called, state: content
🎯 EpisodeHStack: Setting focus to lastFocused: 78d37e0779f1a4671f6ac0c26a04f36a
🎯 EpisodeHStack: About to set focusedEpisodeID = 78d37e0779f1a4671f6ac0c26a04f36a
🎯 EpisodeHStack: After assignment, focusedEpisodeID = nil  ← BUG HERE

🎯 SeasonHStack: Focus guide onContentFocus triggered
🎯 SeasonHStack: focusedSeason changed
🎯 EpisodeHStack: Focus guide onContentFocus triggered  ← SECOND CALL CONFLICTS
```

#### Simplified Natural Focus (Current Solution)
```swift
.focusGuide(focusGuide, tag: "episodes", onContentFocus: {
    if focusedEpisodeID == nil {
        focusedEpisodeID = viewModel.elements.first?.id
    }
})
```

### Application version

latest on master branch

### Where did you install the app from?

None

### Device information

Apple Tv 4k A1842

### OS version

18.5[22L572]

### Jellyfin server version

10.10.7

---
[...]

PR #1819, opened 2025-11-17
Fix tvOS episode focus (#1640)

3aa49ec6bfc910647fa1c5a013e48eef:
### Summary
  - stabilize `SeriesEpisodeSelector.EpisodeHStack` by introducing an `EpisodeFocusCoordinator` that queues pending IDs,
  defers FocusState writes via `withTransaction` + `DispatchQueue.main.async`, and ensures placeholders/episodes share a
  single focus path.
  - add `UIDevice.platformGeneration` (tvOS only) so `EpisodeHStack` can give older Apple TV hardware a slightly longer
  scroll delay when jumping to the play-button item.
  - keep focus logging scoped to `seriesEpisodeFocus` so we can trace season switches, pending focus consumption, and
  guard rails in Console.

  Addresses #1640.

  ### Testing
  - `carthage update --use-xcframeworks`
  - `swiftformat .`
  - `swiftgen config run`
  - `bundle exec fastlane buildLane scheme:"Swiftfin tvOS"` (DEVELOPER_DIR=/Applications/Xcode-16.4.0.app)

### Pending steps (have not yet done any hardware testing)
  - tvOS 26 hardware (Apple TV 4K gen3): season hopping, empty/error/loading states, “Next Up” mid-season
  - tvOS 18.5 hardware: reproduce prior focus loss and verify fix with same scenarios
  - tvOS simulators (18.4, 26): smoke focus tests (no regressions/crashes)


JPKribs:
I don't believe this is the correct approach. The ideal route for this would be to handle the scroll based on `viewModel.state` but track it so it doesn't scroll after the first scroll. 

Just increasing the delay will still result in issues if there is a delay loading or low/poor network conditions.

---

**Edit:** Please see the changes I have made to this and please confirm that this works for you. I've moved this logic to confirm that `viewModel.state` == `.content` to ensure that the content is loaded prior to scrolling. I also changed this to load based on the first appearance of the `EpisodeCard` instead of the `HStack`. Since, the `HStack` loads in prior to the cards and we cannot scroll the cards if they don't exist. Loading on the appearance of the cards ensures that content we are loading exists prior to scrolling. Ensuring we are on `.content` prevents us from scrolling during the `loadingCard`.

This should now always trigger in order regardless of OS/Jellyfin delays but I would appreciate you validating this from your end as well.

LePips:
Apologies for not doing anything with this earlier. I had thought about merging this anyways but don't think a tvOS release would be done until after #1752 anyways. However, I did take the change of waiting until the episodes are presented for scrolling and not on the stack's first appearance.

PR #1770, opened 2025-10-20
Libray Filters and Sorting

hqueiroga:
Added a Library header, with the name of the library and Filter / Sorting functionality

Simulator Screenshot - Apple TV 4K (3rd generation) (at 1080p) - 2025-10-20 at 10 43 37
<br/>
<br/>
Library has the filtering / sorting options menu:
<br/>
<br/>
Simulator Screenshot - Apple TV 4K (3rd generation) (at 1080p) - 2025-10-20 at 10 44 20
<br/>
Simulator Screenshot - Apple TV 4K (3rd generation) (at 1080p) - 2025-10-20 at 10 44 32
<br/>
Simulator Screenshot - Apple TV 4K (3rd generation) (at 1080p) - 2025-10-20 at 10 44 36
<br/>
<br/>
Sorting / Filtering options can be "remembered" by library at the settings screen:
<br/>
<br/>
Simulator Screenshot - Apple TV 4K (3rd generation) (at 1080p) - 2025-10-20 at 10 46 14

PR #1823, opened 2025-11-19
Unified & Composite `FilterView`

JPKribs:
## Summary

Reviewing https://github.com/jellyfin/Swiftfin/pull/1770 I was finding that there was a good opportunity to make a shared `FilterView` for both iOS and tvOS. Right now, on tvOS, there isn't a way to call this but this sets that up for success as we build out #1770 further so that item can be more manageable for review.

Please confirm that I am correct in my [assumptions here](https://github.com/jellyfin/Swiftfin/pull/1770#issuecomment-3549363427) first. **If I am incorrect there, this PR is not required and we can close this out.**

While working on the `FilterView`, I also addresses an outstanding TODO to allow multiple filters on the same `FilterView`. This changes the `init` to take `[ItemFilterType]` instead of a single `ItemFilterType`. This loops over each `ItemFilterType` to build it out in the same way this was built prior with the exception that we are using section headers when there are multiple filters (see screenshots). Additionally, the bindings are created inside the `selectorView(for:)` function.

Currently, I am not using multiple filters on the same view since I think that will require a bit of a rework of the `navigationBarFilterDrawer` and I wanted to keep this a little more focused than I normally do.

## Other Cleanup

1. The checkmark we were using on the `SelectorView` was the same as the `ListRowCheckbox`. I've just added an optional parameter to remove the O when the item was not selected. Functionally, there are no changes but this lets us re-use the sizing for the checkmark so we don't end up with any variations in that over time.
2. I added the `NavigationBarCloseButton` to tvOS to prevent the build errors when sharing a view with tvOS. This also enables this as an option moving forward **if we want it**. _If we don't want it_, I think this makes more sense to put a placeholder instead. Similar to...
3. I added a placeholder for tvOS and `navigationBarTitleDisplayMode` to prevent build issues for `PlatformViews`.
4. To get the icon for the filter on tvOS, I made `ItemFilterType` `systemImageable`. I also made it `Identifiable` to help with the `ForEach`.

## Outstanding Questions

- [ ] For a composite `FilterView`, I am currently making the title `types.map(\.displayTitle).joined(separator: " & ")` which works for 2 filters well _(see screenshots)_ but more filters this would be an issue. What would be the correct title for this composite `FilterView`?
- [ ] For a composite `FilterView` on tvOS, I am currently getting the `systemImage` from the first element with a fallback to `"line.3.horizontal.decrease"`. What would be the appropriate way to get the `SplitFormWindowView.descriptionView` value? I had an idea of making a grid of all elements' `systemImage` but that seemed like a lot so I thought I would ask first.
- [ ] If there are no filters in a `selectorView(for:)` in returns "None" _(as this does prior to this PR)_ but do we want this to show as "None" for a secondary filter on a composite `FilterView` as well? Or would we rather have the empty filters just not show on a composite `FilterView` so long as another filter has values?

### iOS

<details>

<summary>Single Filter</summary>

Simulator Screenshot - iPhone 17 Pro - 2025-11-18 at 21 22 24

</details>

<details>

<summary>Composite Filters</summary>

Simulator Screenshot - iPhone 17 Pro - 2025-11-18 at 21 22 35

</details>

### tvOS

<details>
<summary>Single Filter</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-11-18 at 20 46 39

</details>

<details>
<summary>Composite Filters</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) - 2025-11-18 at 20 46 59

</details>


JPKribs:
Will want to hold off on this until https://github.com/jellyfin/Swiftfin/pull/1838. Whatever route that takes should be used here.

JPKribs:
I think this is still "on track" but with some other items going on I think this would be better off like: https://github.com/jellyfin/Swiftfin/pull/1919#pullrequestreview-3873986221 where we take this out of the navigation stack for both iOS and tvOS.

I think the components here are fine but this might be an `_Alert` or at least like the ability to do multiple sheet sizes?

Would need to think how to best do that.

LePips (approved):
I brought over the `Stateful` migration of `FilterViewModel` from #1752 and had the filter types implement groupings, where sort by and ordering are grouped together. These are probably the only 2 filters to have this relationship, so everything else is pretty much just boilerplate.

Implementation now on tvOS should be pretty straightforward.

Issue #1821, opened 2025-11-17
Display remaining episodes count for TV shows


milanov:
### This feature request respects the following points:

- [x] This feature has **not** already been requested on [GitHub](https://github.com/jellyfin/Swiftfin/issues?q=is%3Aissue%20state%3Aopen%20type%3AFeature).
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).

### Platform

iOS

### Describe the feature you'd like to see

I am using Jellyfin to watch selected content from various video platforms (eg YouTube). Different shows upload with different frequencies, and I keep the shows only for some time. This leads to a rather dynamic episode count. In the jellyfin iOS app I could see the count of unwatched episodes with just a glance. In Swiftfin I need to click on each show with an unwached episode and count. I understand that my use case might not be too frequent, so perhaps this can be a view option, disabled by default.

Thanks for the amazing work, the app is very fast and a joy to use.

amtrevisan:
I would love to work on this as a first issue

PR #1825, opened 2025-11-23
Display remaining episodes count for TV shows #1821

amtrevisan:
## Summary

Adds a “Remaining” indicator toggle for TV shows in Swiftfin. When enabled, TV show posters show a small badge with the count of unwatched episodes.

### Key Changes

- **Settings:** New “Remaining” toggle in Poster Indicator Settings
- **Badge:** New `NumericBadge` component for dynamically sized circular badges
- **Logic:** `RemainingIndicator` now shows a badge for series with `unplayedItemCount > 0`

### Technical Notes

- Uses existing GeoServe default: `showRemainingUnwatched`
- Only appears when `type == .series`

### Design Notes 
- The `NumericBadge` is bigger on tvOS, if kept the same size as iOS it would be too small 
- The `NumericBadge` is placed in the **top-left** because the **top-right** is already crowded with other indicators. Moving it there would require shrinking it enough to fit, which hurts readability especially on iOS

## iOS:
Simulator Screenshot - iPhone 16 - 2025-11-22 at 21 49 51

## tvOS:
Simulator Screenshot - Apple TV 4K (3rd generation) (at 1080p) - 2025-11-22 at 21 47 34

### DIFF
--- a/Swiftfin tvOS/Views/ItemView/Components/EpisodeSelector/Components/EpisodeCard.swift
+++ b/Swiftfin tvOS/Views/ItemView/Components/EpisodeSelector/Components/EpisodeCard.swift
@@ -40,10 +40,6 @@ extension SeriesEpisodeSelector {
                 } else if episode.userData?.isPlayed ?? false {
                     WatchedIndicator(size: 45)
                         .isVisible(showPlayed)
-                } else if episode.canBePlayed && !episode.isLiveStream {
-                    UnwatchedIndicator(size: 45)
-                        .foregroundColor(accentColor)
-                        .isVisible(showUnplayed)
                 }
 
                 if isFocused {


JPKribs:
I think the correct route for this would be to use the existing `unplayed` indicator. What I think would be best would be changing the setting for the `unplayed` indicator to be:

`Disabled`
`Enabled`
`Count`

Those names aren't great but just updating that setting to be an enum instead of a bool. Then, use that existing to show unplayed count instead of just the triangle. Since there isn't really a reason to have both the unplayed indicator and the count. Plus, this reduces the number of settings we have while still accomplishing the same outcome.
 
I think something like this:
 
Screenshot 2025-11-22 at 21 01 46

So, disabled = nothing. Enabled = triangle. Count = number for **parent items** (Shows, Collections, etc.) but still uses a triangle for items that do not have a child count (Movies, Videos, etc.)

I think those names aren't very good so I'd defer to you on a better name for this haha.

Finally, as I alluded to above, I think we'd want to make this more generic. Right now, this only works for `.shows`. I would prefer we make this get the children regardless of type so this would work for collections as well and eventually albums, artists, etc.

amtrevisan:
I love it, thanks for getting back to me so fast, ill do that right away. What do you think about making the number white for light theme and black for dark theme, or maybe a transparent number so it's like a cutout?

JPKribs:
> What do you think about making the number white for light theme and black for dark theme, or maybe a transparent number so it's like a cutout?

Since the background is always the accent color, I would use:

`.accentColor.overlayColor`

I might try other fonts my screenshot I used `.body` with normal weight. Semi-bold is used a lot more in Swiftfin for weight. Font you might have to play around a bit to find something that fits.

For that alignment, though, I basically just commented out 3Q triangle (or whatever it was called) and then had the number `Text()` with a padding of 4 horizontal and 2 vertical and a background of `.primary` since we're passing in the accent color via `.primary`. So then they text color would be `.primary.overlayColor`

I hope that helps! You'll likely wanna play around with it a little bit. I think my screenshot looked a little off so for a final version, we might want to try out some options and see what looks best

amtrevisan:
@JPKribs What do you think? the problem is the count doesnt appear in the "Next Up" section since next up displays the episode not the series, i could make it so it displays the remaining episodes of the parent series

image

JPKribs:
> the problem is the count doesnt appear in the "Next Up" section since next up displays the episode not the series

That is intended functionality and I think that is correct behavior. What you've done to calculate this is exactly correct. For single episodes, we should not treat it as the show so this is perfect as is!

---

I've had a couple of changes. In the settings, I moved the `unPlayed` indicator to the top of the view. I did this since this is a different option than a toggle so moving the one different selector makes the view look cleaner. As a result, I moved the `played` indicator setting right below it to keep flow. Having `played` and `unPlayed` next to each other makes sense.

For the indicator itself, the current indicator and the count indicator will exist next to each other. So I've ensured that the height for both are the same as to not appear to dissimilar. The indicator now goes to the edge of the poster, mirroring the non-count version.

I have ensured that the height of both indicators is the same and the minimum width of the count should mirror the width of the existing indicator. This has the ability to grow further as needed. I've finally rounded the exposed edge of the indicator so it should roughly mirror the rounding of the poster as well. Rounding for tvOS is a little greater since rounding is more prominent on larger items.

Please find my screenshots below and I've committed this change to your branch.

**iOS**

Screenshot 2025-11-22 at 23 21 46

Screenshot 2025-11-22 at 23 21 05

**tvOS**

Screenshot 2025-11-22 at 23 54 16


---

Example of what I mean by the indicators will live next to each other. So, you can see that all of the indicators are the same height regardless of type. Additionally, the minimum width is the same width of the standard indicator so there will not be any items that are out of place for the height of the rest of the indicators.

**iOS**

Screenshot 2025-11-22 at 23 42 00

**tvOS**

Screenshot 2025-11-22 at 23 53 47

---

Please let me know what you think!

amtrevisan:
looks great man, especially the settings change, was thinking of doing something similar

JPKribs:
@LePips this one looks good from my end! I added some changes in here so I'll let you be the sign off for approval in case there was something I missed.

Issue #1525, opened 2025-04-29
Skip button for media segments (intro/outro)


rasko-dev:
**Describe the feature you'd like**
With media segments being added to the Jellyfin API a while ago, I suggest implementing buttons for skipping different segments. Related settings could be implemented for choosing what to do with each type of segment, for example ignoring, showing a skip button or even skipping automatically without user input.

I also suggest that instead of a "skip" button for outros specifically, the button should be a "play next episode" button. 

**Additional context**

- [Jellyfin-meta discussion](https://github.com/jellyfin/jellyfin-meta/discussions/30)
- [API implementation PR](https://github.com/jellyfin/jellyfin/pull/12345)

JPKribs:
Since #1203 was started back when we were on SDK 10.8 (pre-media segments), I wouldn't expect this to be included in that PR especially with all of the existing logic that needs to be validated around just regular media playback. So, this will likely need to be looked at after #1203. Most likely after 1.4 would be my best guess but these things are always subject to change!

PR #1826, opened 2025-11-26
Add support for Media Segment API and implement Skip Intro button

noranraskin:
Implements #1525 

This is a work in progress, but I wanted to open this PR already for people to comment on how to do things better.
To test this (and to use it) the Jellyfin server requires a plugin that scans the library for segments (intros, ads, etc.)
https://jellyfin.org/docs/general/server/metadata/media-segments/#plugin-support

This code uses types from JellyfinAPI but implement `getMediaSegments` since that doesn't exist there.

**Goals**:
- [x] Fetch Media Segments from Jellyfin server after starting playback
- [x] Store segments in MediaPlayerItem
- [x] Update currentSegment in MediaPlayerManager
- [x] Implement skipCurrentSegment function
- [x] Create SkipSegmentButton view
- [x] Display SkipSegmentButton in PlaybackControls
- [x] Get SkipSegmentButton to work for Native player
- [x] Add option in settings to define behaviour for different types of segments
- [x] Add localisation
- [x] Fix TvOS views
- [x] Fix autoskip for Swiftfin player

If possible I would really like to get this to work for the native player too, I tried a few different things. In its current state there's just an overlay on the NativeVideoPlayerView, however this is not working (and I don't understand why not).

I unfortunately don't have access to an AppleTV at the moment so I'd very gladly accept some help polishing the views for TvOS especially since it seems to not be working in general atm. This is what it looks like for me:
Simulator Screenshot - Apple TV 4K (3rd generation) (at 1080p) - 2025-11-26 at 01 05 14

On an iPad simulator it looks like this:

Simulator Screenshot - iPad Air 11-inch (M3) - 2025-11-26 at 01 42 45
Simulator Screenshot - iPad Air 11-inch (M3) - 2025-11-26 at 01 42 49


JPKribs:
> This code uses types from JellyfinAPI but implement `getMediaSegments` since that doesn't exist there.

Please use the SDK for this item and please call this via a `ViewModel`. We can likely call this from `ItemViewModel` but please note that this will likely need to come after https://github.com/jellyfin/Swiftfin/pull/1752 as there are some extensive reworks to `ItemViewModel`. Alternatively, this can be called when the MediaPlayer is generated which might make more sense. 

Please find the correct SDK usage below:

```swift
Paths.getItemSegments(itemID: String, includeSegmentTypes: [MediaSegmentType]?)
```

For reference:
https://github.com/jellyfin/jellyfin-sdk-swift/blob/main/Sources/Paths/GetItemSegmentsAPI.swift

noranraskin:
Ahh thanks was looking for getMediaSegments so I didn't find the appropriate function. Will change that. However the way I understand it is that `MediaPlayerManager` keeps state of playback und should therefore handle segment skips and updates. The `MediaPlayerManager` only has a reference to a `MediaPlayerItem`. I agree it would be preferable to have the fetching logic in `ItemViewModel` but I don't see an elegant way to pass the mediaSegments. Any idea or should I just wait for  #1752 ?

JPKribs:
I think the manager should be fine? Full disclosure, I haven't touched a lot of playback since the rework. I think it makes sense to proceed with it in the manager. Taking a quick look, that appears correct but I'd have to look at this more after Thanksgiving to confirm. I think you're good to proceed in the manager since that other PR is more ItemViewModel.

For tvOS, no need to touch the player View since that's being redone. I've personally taken a look at the tvOS player recently and SwiftUI differences on tvOS make building a player UI absolutely miserable IMO haha. That will all come in a later PR but I imagine I'm going to have to lean on someone with more UIKit experience since remote control functionality appears to be severely limited in SwiftUI. For this PR, I would make sure the shared components and settings exist but you can leave a TODO in the existing player view to add this for now.

I think for this PR:

- Settings/Config on iOS and tvOS
- Swiftfin Player UI element
- Make the skip generic to allow for auto-skipping

You can skip NativePlayer UI. If you are feeling adventurous, this would be a cool time for automatic skipping as well as a setting.

Last note is this button itself right now relies on padding for placement. I would prefer we figure out a cleaner was to do this as to prevent issue with screen sizing. Even just placing this in the bottom right with insets but moving it up when the standard overlay exists?

I might also ask we use the TintedMaterial button for styling to stay more consistent with other elements.

Let me know if there is anything I can help answer or assist with!

JPKribs:
I think we'd want, for settings we'd want a Default for:

MediaSegments: Bool

Then, if enabled, we'd have a secondary setting under for SkipMediaSegment which contains the types using:

https://github.com/jellyfin/Swiftfin/blob/main/Swiftfin/Components/OrderedSectionSelectorView.swift

You can see how I did customizing enabled filters for that usage. Default that setting to []

In practice, show media segments of enabled. Automatically skip if the bool is true and the media segment type exists in that list.

noranraskin:
I went ahead and implemented the settings how they would be most logical to me. Feel free to take a look and give me an opinion. I didn't quite get your explanation about customised enable filters, but I think my solution does that.

Here's what it looks like atm:

On iPad:

Screenshot 2025-11-27 at 12 57 33 PM
Screenshot 2025-11-27 at 12 57 48 PM


On apple TV:

Screenshot 2025-11-27 at 12 59 06 PM
Screenshot 2025-11-27 at 1 16 17 PM

In both cases the **configure media segments** is only visible when the **media segments** toggle is enabled.

All strings are localised but with english translations.

A few kinks to work out still:

- Skipping via button (when setting is *ask*) works well for TV and iOS under Swiftfin player
- Autoskipping does not work for Swiftfin player (log shows VLC state change to buffering and it doesn't continue playing) I don't really know what the problem could be since it should be doing the same as a a button press (I tried calling `skipSegment` from main thread but that didn't change anything). Autoskipping does however work for the native player.

I could use some help for the autoskip not working properly.

Next I'm gonna try to work improving the styling/placement of the button like you mentioned.

JPKribs:
Made a quick change to the settings like this:

https://github.com/user-attachments/assets/8eae529f-db30-4f6f-8ce6-fef0683638fc

This should make this more dynamic so as these types are added in the future this will automatically be added to our settings

noranraskin:
Ahh okay now I understand. I was more following the functionality of jellyfin web with my design

Screenshot 2025-11-27 at 10 31 18 PM

With your change how would you make it that ads are always auto skipped, intros and outros make a "Skip Intro/Outro" button appear and recaps and previews don't make a button appear?

JPKribs:
That's definitely a fair point! Let me think about that and get back to you. I agree, it would be nice to set these on a per segment type basis, but, where possible, we want to try and keep the settings manageable.

I think the correct solution might be a ForEach loop over MediaSegmentType.allCases. Then, three settings in total. Enable media segments, skip media segments, and ask media segments.

The first one would be a Boolean, and the other two would be arrays of items.

But, for now, definitely don't worry about this part of it! I can make a commit with what I'm thinking about when I have time this weekend. If you wanna focus on the functional part of auto skip, that would be perfect! I don't personally have media segments set up on my server so I will need to do some set up before I can help test this out, but I can help you test probably in a couple days after I've had a chance to set up that feature

JPKribs:
Okay, reworked the settings. This should be dynamic, in the sense that any future additions will be accounted for without needing to change the settings. This also accounts for the Off/Ask/Skip logic to allow more granular settings for this. Let me know what you think!

https://github.com/user-attachments/assets/938149ea-2915-45d7-b3c2-9b66e06d7569

noranraskin:
This looks very good! Will be trying to work out the remaining problems now

JPKribs:
I found a plugin for media segments so I'm starting that scan now. It might be a bit but I can assist with testing this whenever this wraps up!

Swiftfin/Views/SettingsView/VideoPlayerSettingsView/Components/Sections/MediaSegmentSection.swift
```diff
@@ -0,0 +1,65 @@
...
+        private var askMediaSegments
+
+        var body: some View {
+            Section(L10n.mediaSegments) {
+                Toggle(L10n.enableMediaSegments, isOn: $enableMediaSegments)
+            }
+
+            Section {
+                if enableMediaSegments {
+                    ForEach(MediaSegmentType.allCases.sorted(by: { $0.displayTitle < $1.displayTitle }), id: \.self) { segment in
```
LePips: Don't sort based on the display title. We should also conform `MediaSegmentDto` to `SupportedCaseIterable`, so we don't use `unknown`.

Swiftfin/Views/SettingsView/VideoPlayerSettingsView/Components/Sections/MediaSegmentSection.swift
```diff
@@ -0,0 +1,65 @@
...
+import SwiftUI
+
+extension VideoPlayerSettingsView {
+    struct MediaSegmentSection: View {
+
+        @Default(.VideoPlayer.enableMediaSegments)
+        private var enableMediaSegments
+
+        @Default(.VideoPlayer.skipMediaSegments)
+        private var skipMediaSegments
```
LePips: I would prefer a `[MediaSegmentType: MediaSegmentBehavior]` dictionary instead of 2 arrays to try and keep mutually exclusive.

LePips (changes requested):

For the segment handling, this should be implemented as a `MediaPlayerObserver` conforming object instead of having logic directly in the `MediaPlayerManager`. I understand this was probably following the `PreviewImageProvider` implementation, but that should eventually be changed as well.

Here is an example of what that object would look like, quickly spun up for testing:

```swift
class ItemSegmentObserver: ViewModel, MediaPlayerObserver {

    @Published
    private(set) var currentSegments: [MediaSegmentDto] = []

    private(set) var skippedSegments: Set<String> = []

    private var segmentTask: Task<[MediaSegmentDto]?, Never>!

    weak var manager: MediaPlayerManager? {
        willSet {
            guard let newValue else { return }
            setup(with: newValue)
        }
    }

    init(itemID: String) {
        super.init()

        self.segmentTask = Task { [weak self] in
            let request = Paths.getItemSegments(itemID: itemID)
            let response = try? await self?.userSession.client.send(request)
//            return response?.value.items

            return [
                .init(
                    endTicks: 100_000_000,
                    id: "test",
                    itemID: nil,
                    startTicks: 100,
                    type: .intro
                ),
            ]
        }
    }

    private func setup(with manager: MediaPlayerManager) {
        manager.secondsBox.$value
            .sink { [weak self] newValue in Task { await self?.secondsDidChange(newValue) } }
            .store(in: &cancellables)
    }

    private func secondsDidChange(_ newSeconds: Duration) async {
        guard let segments = await segmentTask.value else { return }

        currentSegments = segments.filter { segment in
            guard let startSeconds = segment.startSeconds, let endSeconds = segment.endSeconds else { return false }
            return startSeconds ..< endSeconds ~= newSeconds
        }

        if let introSegment = currentSegments.first(where: { $0.type == .intro }), !skippedSegments.contains(introSegment.id!),
           let endSeconds = introSegment.endSeconds
        {
            skippedSegments.insert(introSegment.id!)
            manager?.proxy?.setSeconds(endSeconds)
        }
    }
}

extension MediaSegmentDto {

    var endSeconds: Duration? {
        guard let endTicks else { return nil }
        return .ticks(endTicks)
    }

    var startSeconds: Duration? {
        guard let startTicks else { return nil }
        return .ticks(startTicks)
    }
}
```

For now, this is good enough to be initialized and added to the `MediaPlayerItem.observers` array, similar to `MediaProgressObserver`.

For the button, you've already put it in `PlaybackControls` like it should, but here's how it would be made with the provider above:

```swift
NavigationBar()

Spacer()

if let itemSegmentProvider = manager.playbackItem?.observers
    .first(where: { $0 is ItemSegmentObserver }) as? ItemSegmentObserver
{
    SegmentButton(segmentProvider: itemSegmentProvider)
        .edgePadding()
}

PlaybackProgress()
```

`SegmentButton` can just be a plain white button. We should extend `MediaSegmentDto` to have a display message for each skippable type: `Skip Intro`, `Skip Recap`, `Skip Outro`, etc. We can later take a look at implementing a `Play Next Item/Episode` later, where the outro segment would look at its end ticks, the media's runtime, and the queue to determine what message to present.

You'll notice that my example object has a set on what segments have already been skipped. If the segment is to be automatically skipped, we should still provide functionality to scrub back through media without skipping - and should instead present the skip button during that segment.

You'll also notice the example segment I have doesn't start at 0 ticks. Right now, the `VLCUI` wrapper will first play the media and _then_ set the correct starting seconds. This could be problematic if there are intro segments that start at or are close to 0 ticks. (Edit: don't implement logic to consider this. I should have a change in `VLCUI` to this behavior)

We should also not worry about the native player, as it will be removed. Also don't worry about tvOS for now, as its implementation will require more focus considerations.

Swiftfin/Views/SettingsView/VideoPlayerSettingsView/Components/Sections/MediaSegmentSection.swift
noranraskin: Thanks for the feedback and the code examples! Will try to update everything accordingly. Free time is a bit limited atm. Is the native player going to be removed before the next TvOS release? I rely on that since I'm using airplay speakers. Any updates on the audio issues with VLC or is mpv going to replace native? I don't want to capture this pr but I would like to learn more

LePips: Unsure if it would be removed on next release, but the usage of AVPlayer won't ago away. Instead, the video player is able to swap between VLC and AVPlayer, but the AVPlayer layer isn't complete yet along with proper PiP support in the video player. I can't say anything about what mpv would replace for now.

Swiftfin/Views/SettingsView/VideoPlayerSettingsView/Components/Sections/MediaSegmentSection.swift
JPKribs: Sorry that was me. For unknown, that's part of the DTO and essentially anything that isn't intro, outro, commercial, preview, or recap.

When I was messing around with this, things like "Trialer" where I typo'd would be in there. For our usage, maybe this is `L10n.other` instead of `unknown` and we have that string be "Skip "+"name". So, for my example, it's "Skip Trialer" instead of "Skip Unknown" and managed via an "Other" toggle.

All else we'd keep localized but `unknown` we'd just pull from server.

That's IF we want to build in a way to capture non-standard configs.

JPKribs (changes requested):
Some items to address from first glance:

1. Please clean up these comments. Your comments should roughly mirror what we use in other parts of the application. That is, so long as your code is not breaking any patterns or norms. There's no need to comment. Comments should be reserved for items where additional explanation is needed. Claude tends to over-comment which appears to be what is going on here.
2. Only implement the changes for the video player and ignore native player. AVPlayer will be rolled into that view. So, completing the regular overlay for the video player, and this should translate onto AVPlayer down the road: https://github.com/jellyfin/Swiftfin/issues/1853
3. The changes to the tvOS video player have been completed and your changes should be included in that supplement as well.

noranraskin:
@LePips Finally had some time to work on this. The requested change is made. Media segments are now handled by a MediaSegmentObserver.
Skip button works on iPad for VLCKit and AVKit player and for TvOS only for the Swiftfin player.
Same for auto-skip, since the native player that isn't going to be supported long term I didn't want to spend too much time trying to get it to work.

I removed the "enable media segments" setting and set the default to display the skip button for intros. The thinking being that media segments require setup on the jellyfin server. Users that went through the trouble of installing a plugin to get media segments likely want to use them and this would be a simple way of showing Swiftfin now supports media segments. People that don't care, don't have the plugin so the skip button will never be shown.

Here's how it works now:
The Skip button shows up for 8 seconds then it fades away, it's still there in the playback control overlay for the duration of the segment.

When autoskip is turned on the playhead automatically jumps to the end of the segment, but only if the beginning of the segment is reached via normal playback. Scrubbing into a segment or rewinding won't autoskip (again) but will show a skip button.

The same logic applies for the Skip button on TvOS when it comes to focus. Naturally entering a segment will focus the skip button, when scrubbing or rewinding into the segment the button appears, but isn't focused.

Here's what the settings look like now:
Screenshot 2026-06-11 at 6 25 15 PM
Screenshot 2026-06-11 at 6 25 38 PM
Screenshot 2026-06-11 at 6 26 49 PM

And here's what the button looks like:

Simulator Screenshot - Apple TV 4K (3rd generation) (at 1080p) - 2026-06-11 at 18 27 09
Simulator Screenshot - iPad Air 13-inch (M3) - 2026-06-11 at 18 31 22
Simulator Screenshot - iPad Air 13-inch (M3) - 2026-06-11 at 18 31 51

noranraskin:
@JPKribs I don't know why I can't just reply to your comment.
About the first point: I removed the comments.
What is the timeline for the change to AVMediaPlayerProxy? 
Can you explain the last point? What changes for the TvOS player are you talking about? I rebased my branch on yesterday's main. Are the changes already in main?

JPKribs:
> What is the timeline for the change to AVMediaPlayerProxy?

I have a [branch](https://github.com/JPKribs/Swiftfin/tree/avplayer) I am looking at now. Should be really simple now that we have the tvOS player complete. That being said, Ethan's input is invaluable so I would want to make sure they have time to review everything & provide feedback. So, when time allows for everyone involved.

> Can you explain the last point? What changes for the TvOS player are you talking about? I rebased my branch on yesterday's main. Are the changes already in main?

Elements like:

```swift
#if !os(tvOS)
        .overlay(alignment: .bottomTrailing) {
            if let segmentObserver = manager.segmentObserver {
                SkipSegmentButton(
                    observer: segmentObserver,
                    isPresentingOverlay: false
                )
                .padding(.trailing, 80)
                .padding(.bottom, 120)
            }
        }
#endif
```

Falls a bit under the AVPlayer so consider that point mute haha

JPKribs:
Here: https://github.com/jellyfin/Swiftfin/pull/2046.

This is an old branch so I need to bring everything up to speed. This will also require another [PR](https://github.com/jellyfin/Swiftfin/pull/1882) to work fully so it's a draft until then.

noranraskin:
@JPKribs Is there a chance this PR would be merged before #2046 ? Of course someone other than me should test it, but all scenarios I set out to support work now (basically the behaviour of web jellyfin). And it should be easy enough to remove the native player specific code when the Proxy gets merged. No worries if you'd like to hold off merging this

JPKribs:
> @JPKribs Is there a chance this PR would be merged before #2046?

That PR isn't blocking this one. Sorry, I think I'm explaining this poorly. I'm just saying we shouldn't add anything on top of the `NativePlayerView` as this just increases our surface area for this change and, by targeting the VLC UI, this will be inherited by `AVPlayer` shortly. If I had to guess, #2046 isn't going to be long before it's in.

Shared/Components/NativeVideoPlayer.swift
LePips, line 1: As already stated, please don't add this to the native video player.

Shared/Views/SettingsView/VideoPlayerSettingsView.swift
```diff
@@ -224,6 +231,30 @@ struct VideoPlayerSettingsView: View {
...
+    @ViewBuilder
+    private var mediaSegmentSettings: some View {
+        Section(L10n.mediaSegments) {
+            #if os(tvOS)
+            ChevronButton(L10n.mediaSegments) {
+                router.route(to: .mediaSegmentSettings)
+            }
+            #else
+            ForEach(MediaSegmentType.supportedCases, id: \.self) { segment in
+                PlatformPicker(segment.displayTitle, selection: mediaSegmentBinding(segment))
```
LePips, line 245: There should be a way to use `Binding.map(getter:setter:)` instead.

Shared/Extensions/JellyfinAPI/MediaSegmentDto.swift
```diff
@@ -0,0 +1,67 @@
...
+        return .ticks(startTicks)
+    }
+
+    var endSeconds: Duration? {
+        guard let endTicks else { return nil }
+        return .ticks(endTicks)
+    }
+}
+
+extension MediaSegmentType: Displayable, @retroactive Defaults.Serializable {
```
LePips, line 26: Conform to `Storable` instead of `Defaults.Serializable`.

Shared/Extensions/JellyfinAPI/MediaSegmentDto.swift
```diff
@@ -0,0 +1,67 @@
...
+
+extension MediaSegmentType: SupportedCaseIterable {
+
+    /// Same order as jellyfin web.
+    static var supportedCases: [MediaSegmentType] {
+        [.intro, .preview, .recap, .commercial, .outro]
+    }
+}
+
+extension MediaSegmentType: @retroactive LosslessStringConvertible {
```
LePips, line 58: This conformance shouldn't be necessary.

Shared/Views/SettingsView/MediaSegmentSettingsView.swift
```diff
@@ -0,0 +1,41 @@
...
+    #endif
+
+    @Default(.VideoPlayer.mediaSegmentBehaviors)
+    private var mediaSegmentBehaviors
+
+    var body: some View {
+        Form(systemImage: "forward.end") {
+            Section {
+                ForEach(MediaSegmentType.supportedCases, id: \.self) { segment in
+                    PlatformPicker(segment.displayTitle, selection: mediaSegmentBinding(segment))
```
LePips, line 28: Same using `Binding.map(getter:setter)` here.

LePips (changes requested):
Just an initial review, yet to look at the `MediaSegmentObserver`.

oskarkasprzyk:
Hey, are you still working on this? I’d be happy to help finish the remaining review comments and test tvOS.

Issue #770, opened 2023-05-09
Does not report new state event on tvOS


zuohaot:
### Describe the bug

I am a Home Assistant user. I wish to turn on Movie scene when Jellyfin is opened. When playing media via apps like Prime Video and YouTube. I got this
`{"app_id":"com.google.ios.youtube","friendly_name":"Living Room Apple TV","supported_features":450487},"last_changed":"2023-05-09T05:52:50.186095+00:00","last_updated":"2023-05-09T05:52:50.186095+00:00"}`

It seems that I can only get this type of event message when I turn on "native player" option in Swiftfin. I can not see whether my Apple TV is playing or paused when playing media through Swiftfin. Is that something can be improved? TIA!

### Application version

1.0.1 (70)

### Where did you install the app from?

App Store

### Device information

Apple TV 4K

### OS version

tvOS 16.4.1

### Jellyfin server version

10.8.10

vilhalmer:
It seems that the player doesn't register metadata or the fact that it's playing at all with the OS, so none of that state can make it to any integrations either. Even attempting to send play/pause has no effect. This also affects the remote speaker/TV control widget on other iOS devices, so fixing it would be more widely beneficial than just Home Assistant.

yanbreu:
It looks like this is still a problem with latest TestFlight build 1.1(78).
If I play a movie/tv show and put the add in background it still plays the audio but the OS (iOS and iPadOS) don’t recognize the playback.

eigenphase:
Chiming in here that I’d love to see a solution for this too. Are there any updates?

For me, media player doesn’t work, many items don’t have audio, others stay black….

Very surprised that not more people using this

PR #1830, opened 2025-12-02
fix: playing status on tvOS

fabiomatavelli:
Fixes the playing/pause reporting state on tvOS

# PoW

B57C34A1-67A0-448A-8A99-4829E588F911_4_5005_c
51444E63-538B-4524-8066-728DCFDCB1BF_4_5005_c

State reported by Home Assistant:

Screenshot 2025-12-02 at 17 28 54

Closes: #770


LePips:
I won't be able to test with a physical Apple TV for a while, but are you really sure that the fix is to not register with `AVAudioSession`? This is core functionality with registering with the now playing system and media playback properly at all.

fabiomatavelli:
Tbh I'm not 100% sure @LePips . I've tested it on my Apple TV and it worked, but it would be good if we could find someone else to test it also. I'll give it another try just in case.

LePips:
Apologies, but I'm closing this as I'm unable to replicate the issue: playing status is correctly registered with `AVAudioSession` as necessary.

PR #1831, opened 2025-12-03
[WIP] tvOS: YouTube Support (TubeArchivist)

JKamsker:
## Overview

This pull request introduces a dedicated **YouTube** media surface designed specifically for Jellyfin libraries backed by the [TubeArchivist](https://github.com/tubearchivist/tubearchivist) metadata plugin.

Currently, TubeArchivist libraries are identified as standard "TV Shows," forcing YouTube channels to behave like Seasons and Series. This creates a suboptimal playback experience and cluttered "Latest Media" feeds. This PR solves that friction by detecting these libraries and routing them to a specialized view optimized for flat video content, featuring a channel-filtering sidebar and landscape-forced grids .

Additionally, this PR includes specific layout refinements to the tvOS sidebar (channel rail) to improve focus margins and animation smoothness.

### 1\. Library Detection & Persistence

  * **TubeArchivistLibraryDetector:** Implemented a service that inspects the `MetadataFetchers` and `ImageFetchers` of user libraries. If the string "TubeArchivist" is detected in the options, the library is flagged as a candidate .
  * **Settings Persistence:** Added `Defaults.Customization.Library.youtubeLibraryIDs` to store user preferences regarding which detected libraries should be rendered using the new YouTube interface .
  * **Settings UI:** Introduced a "YouTube Libraries" menu in **Settings \> Libraries**. This list populates only when a compatible library is detected on the server, allowing users to toggle the feature on a per-library basis .

### 2\. Navigation & Data Architecture

  * **Surface Routing:**
      * Modified `MediaViewModel` to introduce a `.youtube(BaseItemDto)` media type. Libraries enabled in settings are now routed to `YouTubeLibraryView` instead of the standard Collection/Folder view .
  * **Feed Exclusion:**
      * Updated `HomeViewModel` to filter enabled YouTube library IDs *out* of the standard "Series" and "Latest" tabs to prevent content duplication across the UI .
  * **YouTubeLibraryViewModel:**
      * **Videos:** Fetches items with `includeItemTypes = [.episode]`, essentially flattening the library structure to show all videos sorted by "Newest" .
      * **Channels:** Fetches items with `includeItemTypes = [.series]` to populate the sidebar filter .

### 3\. Interface & Layout Refinements

  * **View Construction:** Created `YouTubeLibraryView`, utilizing a split layout:
      * **Sidebar:** Displays a scrollable list of Channels with circular avatars. Selection filters the main grid.
      * **Grid:** Renders a standard `PagingLibraryView` but forces the display mode to `.grid` and poster aspect ratio to `.landscape` to accommodate YouTube thumbnails .
  * **Channel Rail Layout:**
      * Defined `YouTubeChannelRailLayout` constants to strictly manage sidebar dimensions .
      * **Collapsed State:** Reduced width to **105pt** (previously wider) with centered icons to correct asymmetric margins on tvOS .
      * **Expanded State:** Maintained at **325pt** with leading alignment for text.
  * **Image Fallbacks:** Updated `BaseItemDto+Poster` logic for this view. If a video thumbnail is missing, it now falls back to the Channel/Series primary image or backdrop to ensure the grid is never populated with empty placeholders .

### 4\. Testing

  * **Unit Tests:** Added `TubeArchivistLibraryDetectorTests` to validate JSON parsing and provider string matching against mock data .
  * **Layout Tests:** Added `YouTubeChannelRailLayoutTests` to ensure the collapsed rail width maintains specific ratios relative to the expanded state, preventing regression during future UI tweaks .
  * **Visual Tests:** Added `YouTubePosterLayoutTests` to verify landscape card aspect ratios .

## Screenshot
TeamViewer_JohoAhxGj1
TeamViewer_op6jYOJwpK

## Verification Steps

1.  **Prerequisite:** Ensure access to a Jellyfin server with the TubeArchivist plugin installed and at least one library configured with that metadata provider.
2.  **Configuration:** Navigate to **Settings \> Libraries \> YouTube Libraries**. Ensure the new menu exists and your library is detected.
3.  **Activation:** Toggle the library to **ON**.
4.  **Home Feed:** Return to the Home screen. Verify the library is no longer visible in the "Series" tab.
5.  **Navigation:** Check the top media picker; a new "YouTube" tile should be present.
6.  **UX (tvOS):**
      * Enter the view. Navigate left to the sidebar.
      * Confirm the rail expands smoothly and collapses to a centered icon state.
      * Select specific channels to verify the grid filters content correctly.

### DIFF
--- a/Shared/Coordinators/Navigation/NavigationRoute/NavigationRoute+Media.swift
+++ b/Shared/Coordinators/Navigation/NavigationRoute/NavigationRoute+Media.swift
@@ -21,6 +21,15 @@ extension NavigationRoute {
         ChannelLibraryView()
     }
 
+    static func youtubeLibrary(viewModel: YouTubeLibraryViewModel) -> NavigationRoute {
+        NavigationRoute(
+            id: "youtube-library-(\(viewModel.parent?.id ?? "Unparented"))",
+            withNamespace: { .push(.zoom(sourceID: "item", namespace: $0)) }
+        ) {
+            YouTubeLibraryView(viewModel: viewModel)
+        }
+    }
+
     static let liveTV = NavigationRoute(
         id: "liveTV"
     ) {
--- a/Shared/Coordinators/Navigation/NavigationRoute/NavigationRoute+Settings.swift
+++ b/Shared/Coordinators/Navigation/NavigationRoute/NavigationRoute+Settings.swift
@@ -54,6 +54,12 @@ extension NavigationRoute {
         CustomizeViewsSettings()
     }
 
+    static let youtubeLibrariesSettings = NavigationRoute(
+        id: "youtubeLibrariesSettings"
+    ) {
+        YouTubeLibrariesSettingsView()
+    }
+
     #if DEBUG && !os(tvOS)
     static let debugSettings = NavigationRoute(
         id: "debugSettings"
--- a/Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto+Poster.swift
+++ b/Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto+Poster.swift
@@ -89,19 +89,33 @@ extension BaseItemDto: Poster {
     func landscapeImageSources(maxWidth: CGFloat? = nil, quality: Int? = nil) -> [ImageSource] {
         switch type {
         case .episode:
+            var sources: [ImageSource] = [
+                // TubeArchivist episodes carry the YouTube thumbnail in the thumb slot.
+                imageSource(.thumb, maxWidth: maxWidth, quality: quality),
+                imageSource(.primary, maxWidth: maxWidth, quality: quality),
+            ]
+
             if Defaults[.Customization.Episodes.useSeriesLandscapeBackdrop] {
-                [
+                sources.append(contentsOf: [
                     seriesImageSource(.thumb, maxWidth: maxWidth, quality: quality),
                     seriesImageSource(.backdrop, maxWidth: maxWidth, quality: quality),
-                    imageSource(.primary, maxWidth: maxWidth, quality: quality),
-                ]
-            } else {
-                [imageSource(.primary, maxWidth: maxWidth, quality: quality)]
+                ])
             }
+
+            // If episode images are missing, fall back to channel art/backdrop so cards are never blank.
+            sources.append(contentsOf: [
+                seriesImageSource(.primary, maxWidth: maxWidth, quality: quality),
+                seriesImageSource(.backdrop, maxWidth: maxWidth, quality: quality),
+            ])
+
+            return sources
         case .folder, .program, .musicVideo, .video:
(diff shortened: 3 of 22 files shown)


JPKribs:
Hi,

Thank you for the contribution. I'm a little confused by what you were trying to do here. This looks as though you are adding a feature that embeds YouTube channels into Jellyfin?

Just as two notes that may be dealbreaker for this PR:

This is a very large PR with, at the time of writing this, over 8,000 lines. I don't believe this is something we would be able to adequately review. I would lean towards breaking this up into smaller sections.

I don't believe this YouTube functionality falls under a standard Jellyfin installation. If this is based on a plug-in, we don't typically build around plug-in specific logic. While this looks as though it would work for TubeArchivist plug-in, this also looks as though this would cause a lot of navigation headaches pertaining to normal usage.

I believe the correct approach for this would be to use a Home Video / Video library type, where a nested folder structure is enabled. This should currently be live on 1.4 TestFlight and will be included in our next tvOS release.

JKamsker:
Hi @JPKribs,

Thanks for the *swift* response! You definitely caught me there-I submitted this at 3 AM and passed out immediately after, which explains the absolute mess in the file changes.

Regarding the **size of the PR**, that was entirely my bad. A lot of trash files got included by accident in my sleepy state. I’ve cleaned it up, so we’re down from 8,000 lines to just about **900 lines added** (and 32 deleted). Hopefully, that makes it much more reasonable to review!

To give you some background on my **motivation**: I recently cancelled YouTube Premium and moved to **TubeArchivist** to keep my library ad-free and organized. The problem is that Swiftfin currently treats these libraries like **standard TV Shows**. That forces a hierarchy where you have to dig through Creators and "Seasons" just to find a video, and the latest content gets buried at the end of sidescrollers (which I honestly find really frustrating for this type of content). I essentially wanted to create a **"YouTube Lite" experience** inside Swiftfin with a flattened feed and a proper channel sidebar.

I totally understand the hesitation around **plugin-specific logic**. To be fair, this implementation doesn't have any hard dependencies on TubeArchivist; it’s mostly just **string checks** in the metadata. If you prefer, we could make it more generic-perhaps by checking if a library is a "TV Show" type and contains **"YouTube" in the name**? That would allow users to organize it manually without the plugin. It's also **fully opt-in via Settings**, so standard users wouldn't see any difference.

I haven't tested the new **Home Video type** in the 1.4 TestFlight yet, but I do feel like the **specific layout tweaks** here are pretty necessary for a good YouTube experience.

Could you clarify what you meant regarding **"navigation headaches"**? 

Thanks again!

LePips:
Thank you for your contribution but I apologize, this is not functionality I would introduce to this project. While this isn't based on a plugin explicitly, it's still a special feature consideration based on a specific metadata provider - essentially the same thing. I would not be making special UI based on things like this.

However, reading this use case does spark the idea of making a similar UI in a more general sense: a [Miller columns](https://en.wikipedia.org/wiki/Miller_columns) design for library views - which could help achieve something similar. However, this should not be seriously sought out anytime soon.

Issue #1833, opened 2025-12-03
Improve Item Refresh Options


JPKribs:
### This feature request respects the following points:

- [x] This feature has **not** already been requested on [GitHub](https://github.com/jellyfin/Swiftfin/issues?q=is%3Aissue%20state%3Aopen%20type%3AFeature).
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).

### Platform

iOS

### Describe the feature you'd like to see

On Jellyfin-Web, there are more granular options when refreshing a `BaseItemDto`. This includes options for specifically refreshing images or trickplays. The goal should be to fully expose refresh options to the user instead of 4 preset options.

PR #1834, opened 2025-12-03
Complete `ItemMetadataRefresh` Logic

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1833

The `ItemMetadataRefresh` Logic was originally implemented on 10.8 prior to Trickplays being a available. The goal was to eventually update this to include Trickplay regeneration. I've matched how Jellyfin-Web does these refreshes. There are more variations that exist but, as far as I can tell, `.validationOnly` & `.none` don't do anything. So, I'm sticking the format that users are used to and just purely mirroring:

https://github.com/jellyfin/jellyfin-web/blob/master/src/components/refreshdialog/refreshdialog.js#L78

### Video

**iOS**

https://github.com/user-attachments/assets/96ca0a7d-db23-4ecf-ad5b-980d598d1857

**tvOS**

https://github.com/user-attachments/assets/27910579-a8d2-4ea6-8eff-6d55089ca489


Swiftfin/Views/ItemEditorView/ItemEditorView.swift
```diff
@@ -74,6 +76,7 @@ struct ItemEditorView: View {
             if canEditMetadata {
 
                 refreshButtonView
+                    .foregroundStyle(.primary, .secondary)
```
LePips: Please apply to the `Button` itself.

Swiftfin/Views/ItemEditorView/ItemEditorView.swift
```diff
@@ -98,26 +101,28 @@ struct ItemEditorView: View {
...
-                    value: L10n.replaceAllDescription
-                )
+            Button {
+                router.route(to: .itemMetadataRefresh(viewModel: metadataViewModel))
+            } label: {
+                HStack {
+                    Text(L10n.refreshMetadata)
+                        .foregroundStyle(.primary)
+
+                    Spacer()
```
LePips: The `Text` should have `frame(maxWidth: .infinity, alignment: .trailing)` instead of a `Spacer`.

Shared/Coordinators/Navigation/NavigationRoute/NavigationRoute+Item.swift
```diff
@@ -244,9 +244,18 @@ extension NavigationRoute {
...
-
     #endif
 
+    static func itemMetadataRefresh(viewModel: RefreshMetadataViewModel) -> NavigationRoute {
+        NavigationRoute(
+            id: "itemMetadataRefresh",
+            style: .sheet
+        ) {
+            ItemMetadataRefreshView(viewModel: viewModel)
+                .presentationDetents([.medium, .large])
```
LePips: I don't want to think about introducing additional detents quite yet, let's keep it as it would naturally be presented for now.

JPKribs:
I've made those changes and added tvOS. I still need some time to clean this up but I think this layout is locked in. I'm using a `PlatformView` to share this between the two but I'm struggling to figure out a good way to share the `LearnMore...` contents between `LearnMoreModel` and `LearnMoreButton` since we're using a `LabeledContentBuilder`.

https://github.com/user-attachments/assets/6cfa28ad-f8d0-4055-9457-3d90f389aa75

I'm also still looking to confirm what the `MetadataRefreshMode` actually mean since I'm not really able to find these documented anywhere. [What we have on the Enums themselves isn't super helpful...](https://github.com/jellyfin/jellyfin/blob/master/MediaBrowser.Controller/Providers/MetadataRefreshMode.cs). Once I have those cleared up, I'll localized this and mark this for review.

JPKribs:
This should be ready. I made some changes. In my testing, I found that some of the options just didn't do anything. We weren't using them on Web and I tried several combinations but nothing seemed to change. These where the `.validationOnly` & `.none` options.

I've pivoted this PR to purely just mirroring how Jellyfin-Web does this so the definitions that users get from Web exist in Swiftfin and vice versa. This also cleans up the need for our help text and those labels.

Finally, I lifted the following from https://github.com/jellyfin/Swiftfin/pull/1752 to help with building:

```
    /// - Important: This does nothing on tvOS.
    @ViewBuilder
    func navigationBarTitleDisplayMode(_ mode: NavigationBarItem.TitleDisplayMode) -> some View {
        self
    }
```

**Videos in the main post should now be accurate.**

JPKribs:
Will want to hold off on this until https://github.com/jellyfin/Swiftfin/pull/1838. Whatever route that takes should be used here.

LePips (approved):

Issue #1222, opened 2024-09-01
Reorganize Customization Settings


JPKribs:
**Describe the feature you'd like**
There are no issues with the current customization settings but I think there are rooms to split this out so it's easier to navigate. I am happy to make these changes. I wanted to have this issue to talk about it and build a consensus before I started moving things around. I think there are 3 approaches that all have pros and cons:

<details>
<summary>Move the Customization Settings into separate sections on the Main Settings View</summary>

It keeps the same "Flat" approach where most settings are still only a single layer down but it does make the home settings view longer.

![Simulator Screenshot - iPhone 15 Pro - 2024-09-01 at 16 27 00](https://github.com/user-attachments/assets/d095868a-5085-4d03-bb04-8b63ef8ca41a)

</details>

<details>
<summary>Layer the Customization Settings inside Customization</summary>

It moves to a less "Flat" design, which I understand people aren't a huge fan of, but it keeps the home settings view shorter while getting all the customize settings grouped.

![Simulator Screenshot - iPhone 15 Pro - 2024-09-01 at 16 33 51](https://github.com/user-attachments/assets/4bc22ac1-0a62-4e29-8993-afc874edcde6)

</details>

<details>

<summary>Retain the current layout but document the settings</summary>

I think the current layout is fine as well so long as we better group the sections. My best example of this is that there is currently a "Library" and "Libraries" section. If we want to keep the current layout, I'd love some guidance on what sections we want and I'd be happy to re-organize and write descriptions.

</details>

Regardless of which direction, I would like to add some descriptions to these settings so less technical users have a better understanding of what they are changing. I want to confirm that the best way to do this is using the 'footer' closure for the Section object?

JPKribs:
Nothing at this point! I think the primary item that comes up is if someone is taking this on, they'll need to figure out a good system for how to organize everything. I think we've got a couple things right now that are grouped together that shouldn't be, and a couple things on their own that could be grouped.

Other than that, this would be tooltips.

For consistently, unfortunately, I think we need to do a TV and iOS at the same time though.  Just to ensure that everything is comparable.

PR #1838, opened 2025-12-06
Deprecate `SplitFormWindowView` & `LearnMore` Types

JPKribs:
### Summary

Starts the process for: https://github.com/jellyfin/Swiftfin/issues/1222

`SplitFormWindowView` currently causes issues with building for both iOS and tvOS as it only exists in tvOS. Additionally, I understand that we are moving away from the builder pattern in favor of generics. My goal with this PR is to build the foundation so I can start migrating all of the settings over to a unified configuration.

I've created a `PlatformForm` that mirrors our existing `SplitFormWindowView` for tvOS and a standard form layout for iOS. I've added some logic for initializing the `SplitFormWindowView.descriptionView` since we have a pattern we re-used with the systemImage and 400 width but this builds this in as an init. I've also added our recurring logic in at this level so it's easier to maintain as we change things.

I was trying to figure out the right way of doing this. Whether I should do a `struct` where I basically retain `SplitFormWindowView` as the `tvOSView` of a `PlatformView` or if this should be a `protocol` like how I did this now. I preferred the `protocol` since I believe this is cleaner to just have a single `contentView` instead of `body` { `SplitFormWindowView` }.  Let me know if I went the wrong way with this!

I think I will want to move `CaseIterablePicker` to shared and remove `ListRowMenu` or vice versa to have a single type for this.

If this all looks good, I would like to start moving form views over to a shared state when applicable in iOS and tvOS. This should resolve https://github.com/jellyfin/Swiftfin/issues/1222 as part of this as well.

No screenshots as this is visually the same on both platforms.

### DIFF
--- a/Shared/Objects/ItemFilter/ItemFilterCollection.swift
+++ b/Shared/Objects/ItemFilter/ItemFilterCollection.swift
@@ -46,6 +46,10 @@ struct ItemFilterCollection: Hashable, Storable {
         self != Self.default
     }
 
+    var hasQueryableFilters: Bool {
+        genres.isNotEmpty || itemTypes.isNotEmpty || letter.isNotEmpty || tags.isNotEmpty || traits.isNotEmpty || years.isNotEmpty
+    }
+
     var activeFilterCount: Int {
         var count = 0
 
--- a/Shared/ViewModels/SearchViewModel.swift
+++ b/Shared/ViewModels/SearchViewModel.swift
@@ -54,6 +54,10 @@ final class SearchViewModel: ViewModel {
         items.values.allSatisfy(\.isEmpty)
     }
 
+    var canSearch: Bool {
+        searchQuery.value.isNotEmpty || filterViewModel.currentFilters.hasQueryableFilters
+    }
+
     // MARK: init
 
     init(filterViewModel: FilterViewModel = .init()) {
@@ -64,7 +68,7 @@ final class SearchViewModel: ViewModel {
             .debounce(for: 0.5, scheduler: RunLoop.main)
             .sink { [weak self] query in
                 guard let self else { return }
-                guard query.isNotEmpty else { return }
+
                 actuallySearch(query: query)
             }
             .store(in: &cancellables)
@@ -73,8 +77,8 @@ final class SearchViewModel: ViewModel {
             .debounce(for: 0.5, scheduler: RunLoop.main)
             .sink { [weak self] _ in
                 guard let self else { return }
-                guard searchQuery.value.isNotEmpty else { return }
-                search(query: searchQuery.value)
+
+                actuallySearch(query: searchQuery.value)
             }
             .store(in: &cancellables)
     }
@@ -84,13 +88,13 @@ final class SearchViewModel: ViewModel {
         searchQuery.value = query
 
         await cancel()
-        items.removeAll()
     }
 
     @Function(\Action.Cases.actuallySearch)
     private func _actuallySearch(_ query: String) async throws {
 
-        guard query.isNotEmpty else {
+        guard self.canSearch else {
+            items.removeAll()
             return
         }
 
--- a/Swiftfin tvOS/Views/SearchView.swift
+++ b/Swiftfin tvOS/Views/SearchView.swift
@@ -169,10 +169,10 @@ struct SearchView: View {
                 }
             case .initial:
                 if viewModel.hasNoResults {
-                    if searchQuery.isEmpty {
-                        suggestionsView
-                    } else {
+                    if viewModel.canSearch {
                         Text(L10n.noResults)
+                    } else {
+                        suggestionsView
                     }
                 } else {
                     resultsView
--- a/Swiftfin/Views/SearchView.swift
+++ b/Swiftfin/Views/SearchView.swift
@@ -191,10 +191,10 @@ struct SearchView: View {
                 }
             case .initial:
                 if viewModel.hasNoResults {
-                    if searchQuery.isEmpty {
-                        suggestionsView
-                    } else {
+                    if viewModel.canSearch {
                         Text(L10n.noResults)
+                    } else {
+                        suggestionsView
                     }
                 } else {
                     resultsView


JPKribs:
Let me know if I am off base with this. I just also added the `LearnMoreModal` since this is the only place we use this. I am trying to wrap my head around something where you pass in a single like `@LabeledContentBuilder` or maybe like `[Section: LabeledContent]` that then applies both the `LearnMoreButton` under the correct iOS section and applies the `LearnMoreModal` to the correct focused section on tvOS.

This might not be possible. For now, I have this version where the learn more context is an optional input as a `@LabeledContentBuilder` that applies to tvOS but iOS would still need to put it in the sections. This is an improvement but I feel I can do better.

My only other thought is like a `FormSection` struct that handles focus for elements for tvOS and just adds a `@LabeledContentBuilder`. Then, on iOS this would add a `LearnMoreButton` under the section footer. For this SettingsView, there currently is not a `LearnMore...` that is required but I think it would be valuable to add one to the `VideoPlayerType` that just explains Swiftfin v Native.

LePips:
You're not too far off of what I've thought about to consolidate these views. To share the `Learn More` functionality, we  would use `FocusedValue`. Here's a little sample I've whipped up for reference.

[SharedPlatformSettings.zip](https://github.com/user-attachments/files/24009506/SharedPlatformSettings.zip)

JPKribs:
> You're not too far off of what I've thought about to consolidate these views. To share the `Learn More` functionality, we  would use `FocusedValue`. Here's a little sample I've whipped up for reference.
> 
> 
> 
> [SharedPlatformSettings.zip](https://github.com/user-attachments/files/24009506/SharedPlatformSettings.zip)
> 
> 

Oh perfect! I'll take a look at this tonight

JPKribs:
Okay! This is all functioning how I would expect this to now! I'm happy with where this stands but I'm not sure about naming/deprecation. Some of the items that I moved around are basically just our existing `SplitFormWindowView`, `LearnMoreButton`, and `LearnMoreModal` but now in different places. I feel the functionality is there but the naming doesn't make a lot of sense.

Finally, I've added a `LearnMore` for the `VideoPlayerType`. This serves as a demo/confirmation that this works how we want and I think this is some valuable information to more casual users who might not look at our [documentation](https://github.com/jellyfin/Swiftfin/blob/main/Documentation/players.md). If we don't like this part, I am happy to remove it but I believe this serves as a good demo of this.

The description text for these `LearnMore` values is also up for debate. I tend to be a little wordy so feel free to lemme know if this should be reduced/changed.

Only other question. Do we like the section that leads to `CustomizeSettingsView` and, on iOS, `AccentColor` is called "Accessibility". IMO, this should be like "Customization" and the button that goes to `CustomizeSettingsView` should change from "Customize" to "Advanced" but I'm up for whatever. Just a good time to change this while we're changing it up.

---

**iOS**

https://github.com/user-attachments/assets/f74a08a5-a698-413f-9b09-bdcd31e8a56b

**tvOS**

https://github.com/user-attachments/assets/a786de65-f4c8-4e40-8b96-d8745c4def20

LePips:
I've changed the usage of `Learn More` sections to a `Section` overload, similar to how I've done the `SecureField` overload for mask toggling. We don't need to introduce a `PlatformForm` protocol, we can instead make a `Form` overload, that takes an image, and then based on the platform makes the relevant form view.

JPKribs:
Ohhh that's a much cleaner way of this! Good call, I'll work on that this afternoon. Thank you for the suggestion!

JPKribs:
I should have all of these changes included now. Videos should be unchanged. This should be good to go for a review now. 

I moved the overloads to extensions for `Form` and `Section`. Let me know if that isn't the right place for this.

---

I might also want to do something similar for `SplitLoginWindowView` in a later PR. Something similar but for `List`. I will want to focus with moving all the settings over first. Probably after https://github.com/jellyfin/Swiftfin/pull/1823 and https://github.com/jellyfin/Swiftfin/pull/1834 as well since those are done except for migrating to these new `Form` and `Section`s.

Shared/Extensions/Form.swift
```diff
@@ -0,0 +1,122 @@
...
+
+extension FocusedValues {
+
+    @Entry
+    var formLearnMore: AnyView? = nil
+}
+
+// MARK: - Form Overloads
+
+func Form<Content: View>(
```
LePips, line 19: We should not overload `Form { }`, this is causing a lot of unintended layouts throughout tvOS. Check out a media source on an item, or any setting view.

If this was to still use an underlying `PlatformForm` under the hood, it isn't much different if `Image == EmptyView` and I still want the proper `SwiftUI.Form` primitive.

Shared/Extensions/Form.swift
```diff
@@ -0,0 +1,122 @@
...
+    @ViewBuilder content: @escaping () -> Content
+) -> some View {
+    PlatformForm(content: content) {
+        EmptyView()
+    }
+}
+
+func Form<Content: View>(
+    systemImage: String,
```
LePips, line 20:
```suggestion
    systemName: String,
```

Shared/Extensions/Form.swift
```diff
@@ -0,0 +1,122 @@
...
+    PlatformForm(content: content) {
+        Image(systemName: systemImage)
+            .resizable()
+            .aspectRatio(contentMode: .fit)
+            .frame(maxWidth: 400)
+    }
+}
+
+func Form<Content: View>(
+    imageResource: ImageResource,
```
LePips:
```suggestion
    image: ImageResource,
```

Shared/Extensions/Form.swift
JPKribs: Sorry about that. This is resolved on the last commit. Along with Merge with Main.

Shared/Views/SettingsView/SettingsView.swift
```diff
@@ -0,0 +1,174 @@
...
+                },
+                icon: { EmptyView() },
+                subtitle: {
+                    Label {
+                        Text(viewModel.userSession.server.name)
+                    } icon: {
+                        if !viewModel.userSession.server.isVersionCompatible {
+                            Image(systemName: "exclamationmark.circle.fill")
+                            #if os(iOS)
+                                .labelStyle(.sectionFooterWithImage(imageStyle: .orange))
```
LePips: The `labelStyle` needs to be applied on the `Label`, or else it will only show the title due to `ChevronButton`'s internal label handling. Additionally, we can still use this label style but would just need to adjust its `HStack` spacing to be a bit smaller on tvOS.

Swiftfin tvOS/Views/LearnMoreModal.swift
```diff
@@ -8,6 +8,7 @@
 
 import SwiftUI
 
+@available(*, deprecated, message: "Use the PlatformForm protocol instead")
```
LePips:
```suggestion
@available(*, deprecated, message: "Use `Section(:content:learnMore:)` and a `Form` with an image instead")
```

Swiftfin/Components/LearnMoreButton.swift
```diff
@@ -8,6 +8,7 @@
 
 import SwiftUI
 
+@available(*, deprecated, message: "Use the PlatformForm protocol instead")
```
LePips:
```suggestion
@available(*, deprecated, message: "Use `Section(:content:learnMore:)` instead")
```

Shared/Views/SettingsView/SettingsView.swift
JPKribs: For this specific usage, I think an `HStack` makes more sense. However, I will likely migrate the `labelStyle` as part of other settings.

JPKribs:
Good call. I'll make an issue for that otherwise I'll forget. Lots to do haha

LePips (approved):

PR #1845, opened 2025-12-10
Allow search for empty string when tags are set

magshee:
# Summary

Additionally to the current behaviour when using a string to search, this PR allows to search across all libraries when at least one tag is set.

## Behaviour

- After navigating to the search tab, search suggestions are displayed as before
- If a tab is selected, a search across all libraries is performed that returns a result matching on the chose tags
- If no tab is selected and the search string is empty the suggestions will be displayed again

The entire behaviour is shown in the video below.

## Key Changes

- the `.search` transistion received an additional parameter that specifies whether tags are set or not
- guard statements that prevent the search from executing when the search string is empty were changed to allow a search for an empty string if at least one tag is set.

## iOS

https://github.com/user-attachments/assets/37db3526-36d4-4bee-bf19-b7cbe66a2b72

## tvOS

- There is no possibility to select a tag when searching in the tvOS app


magshee:
I just saw that the tvOS build had failed. I'm going to check what caused that later

Shared/ViewModels/SearchViewModel.swift
```diff
@@ -64,7 +68,11 @@ final class SearchViewModel: ViewModel {
             .debounce(for: 0.5, scheduler: RunLoop.main)
             .sink { [weak self] query in
                 guard let self else { return }
-                guard query.isNotEmpty else { return }
+
+                if query.isEmpty && !self.filterViewModel.currentFilters.hasFilters {
```
JPKribs: Please change this back to a guard or even combine this with the guard above:

```
guard let self else, query.isEmpty && !self.filterViewModel.currentFilters.hasFilters {
    return
}
```

Shared/ViewModels/SearchViewModel.swift
```diff
@@ -73,14 +81,13 @@ final class SearchViewModel: ViewModel {
             .debounce(for: 0.5, scheduler: RunLoop.main)
             .sink { [weak self] _ in
                 guard let self else { return }
-                guard searchQuery.value.isNotEmpty else { return }
-                search(query: searchQuery.value)
+                search(query: searchQuery.value, hasFilters: self.filterViewModel.currentFilters.hasFilters)
```
JPKribs: Please mirror the same guard from above. This isn't an issue in my testing but I would prefer that we keep this consistent:

```
guard let self else, searchQuery.value.isEmpty && !self.filterViewModel.currentFilters.hasFilters {
    return
}
```

JPKribs (changes requested):
Couple of guard changes. Also, we are bouncing between our `hasFilters` parameter and `self.` / `filterViewModel.currentFilters.hasFilters`. I would lean more towards the latter that way we aren't passing this bool in as a parameter. That being said, I haven't looked into when that changes or if we can kick off a search from that event.

Let's please only use of of those two locations for tracking `hasFilters` to make this more consistent. This is only for inside the `ViewModel`.

Shared/ViewModels/SearchViewModel.swift
magshee: I combined it with guard above, similar to you example. I made a small modification, as a guard statement expects an `else`, and therefore the condition must be different.

Shared/ViewModels/SearchViewModel.swift
magshee: If i understood @LePips comment below correctly,  there is no need for a guard statement here. If I am wrong, please correct me.

magshee:
Thanks to both of you for your feedback!

My initial solution was similar to what you described, it also did not had an additional parameter in the `search` function.

During testing, I noticed the following things: 

1. After selecting a filter, a search is correctly triggered. However, when the filter is deselected, the same items remain displayed as search results. Instead I would expect the app the return to the initial search screen with the suggestions shown, as this would be consistent with the behaviour when a search string is cleared after a successful search has happened.

2. If I search using a search string and a filter, then remove the search string, the app shows the initial search screen with the suggestions. I would expect it to instead show search results based on the still selected filter.


To address both issues, I decided to pass an additional parameter to the search function. Initial, this parameter consisted of the selected tags. However, since we only need to know whether any tags are set, I chose to pass a boolean value instead.


I’ve updated the code according the @LePips suggestions. If the behaviour described in the two points above is acceptable to you, I’ll leave the implementation as it is. Otherwise I’ll be happy to find a different solution.

Shared/ViewModels/SearchViewModel.swift
JPKribs: This was going off of the assumption that we would need to pass in the Bool. Referencing the FilterVM internally is better and means we are handling this check/empty string there.

JPKribs:
> Thanks to both of you for your feedback!
> 
> My initial solution was similar to what you described, it also did not had an additional parameter in the `search` function.
> 
> During testing, I noticed the following things: 
> 
>     1. After selecting a filter, a search is correctly triggered. However, when the filter is deselected, the same items remain displayed as search results. Instead I would expect the app the return to the initial search screen with the suggestions shown, as this would be consistent with the behaviour when a search string is cleared after a successful search has happened.
> 
>     2. If I search using a search string and a filter, then remove the search string, the app shows the initial search screen with the suggestions. I would expect it to instead show search results based on the still selected filter.
> 
> 
> To address both issues, I decided to pass an additional parameter to the search function. Initial, this parameter consisted of the selected tags. However, since we only need to know whether any tags are set, I chose to pass a boolean value instead.
> 
> I’ve updated the code according the @LePips suggestions. If the behaviour described in the two points above is acceptable to you, I’ll leave the implementation as it is. Otherwise I’ll be happy to find a different solution.

This is my fault. Sorry, I should've removed my review when LePips made their's as I was going a different, incompatible direction. Please find what I believe to be the correct changes below:

<details>

<summary>Full Code</summary>

```swift
//
//
//

import Combine
import Foundation
import JellyfinAPI
import OrderedCollections
import SwiftUI

@MainActor
@Stateful
final class SearchViewModel: ViewModel {

    @CasePathable
    enum Action {
        case getSuggestions
        case search(query: String)
        case actuallySearch(query: String)

        var transition: Transition {
            switch self {
            case .getSuggestions:
                .none
            case let .search(query):
                query.isEmpty ? .to(.initial) : .to(.searching)
            case .actuallySearch:
                .to(.searching, then: .initial)
                    .onRepeat(.cancel)
            }
        }
    }

    enum State {
        case error
        case initial
        case searching
    }

    @Published
    private(set) var items: [BaseItemKind: [BaseItemDto]] = [:]
    @Published
    private(set) var suggestions: [BaseItemDto] = []

    private var searchQuery: CurrentValueSubject<String, Never> = .init("")

    let filterViewModel: FilterViewModel

    var hasNoResults: Bool {
        items.values.allSatisfy(\.isEmpty)
    }

    // MARK: init

    init(filterViewModel: FilterViewModel = .init()) {
        self.filterViewModel = filterViewModel
        super.init()

        searchQuery
            .debounce(for: 0.5, scheduler: RunLoop.main)
            .sink { [weak self] query in
                guard let self else { return }

                actuallySearch(query: query)
            }
            .store(in: &cancellables)

        filterViewModel.$currentFilters
            .debounce(for: 0.5, scheduler: RunLoop.main)
            .sink { [weak self] _ in
                guard let self else { return }

                actuallySearch(query: searchQuery.value)
            }
            .store(in: &cancellables)
    }

    @Function(\Action.Cases.search)
    private func _search(_ query: String) async throws {
        searchQuery.value = query

        await cancel()
        items.removeAll()
    }

    @Function(\Action.Cases.actuallySearch)
    private func _actuallySearch(_ query: String) async throws {

        guard query.isNotEmpty || filterViewModel.currentFilters.hasFilters else {
            self.items = [:]
            return
        }

        let newItems = try await withThrowingTaskGroup(
            of: (BaseItemKind, [BaseItemDto]).self,
            returning: [BaseItemKind: [BaseItemDto]].self
        ) { group in

            // Base items
            let retrievingItemTypes: [BaseItemKind] = [
                .boxSet,
                .episode,
                .movie,
                .musicArtist,
                .musicVideo,
                .liveTvProgram,
                .series,
                .tvChannel,
                .video,
            ]

            for type in retrievingItemTypes {
                group.addTask {
                    let items = try await self._getItems(query: query, itemType: type)
                    return (type, items)
                }
            }

            // People
            group.addTask {
                let items = try await self._getPeople(query: query)
                return (BaseItemKind.person, items)
            }

            var result: [BaseItemKind: [BaseItemDto]] = [:]

            while let items = try await group.next() {
                if items.1.isNotEmpty {
                    result[items.0] = items.1
                }
            }

            return result
        }

        guard !Task.isCancelled else { return }
        self.items = newItems
    }

    private func _getItems(query: String, itemType: BaseItemKind) async throws -> [BaseItemDto] {

        var parameters = Paths.GetItemsByUserIDParameters()
        parameters.enableUserData = true
        parameters.fields = .MinimumFields
        parameters.includeItemTypes = [itemType]
        parameters.isRecursive = true
        parameters.limit = 20
        parameters.searchTerm = query

        // Filters
        let filters = filterViewModel.currentFilters
        parameters.filters = filters.traits
        parameters.genres = filters.genres.map(\.value)
        parameters.sortBy = filters.sortBy.map(\.rawValue)
        parameters.sortOrder = filters.sortOrder
        parameters.tags = filters.tags.map(\.value)
        parameters.years = filters.years.map(\.intValue)

        if filters.letter.first?.value == "#" {
            parameters.nameLessThan = "A"
        } else {
            parameters.nameStartsWith = filters.letter
                .map(\.value)
                .filter { $0 != "#" }
                .first
        }

        let request = Paths.getItemsByUserID(userID: userSession.user.id, parameters: parameters)

        return response.value.items ?? []
    }

    private func _getPeople(query: String) async throws -> [BaseItemDto] {

        var parameters = Paths.GetPersonsParameters()
        parameters.limit = 20
        parameters.searchTerm = query

        let request = Paths.getPersons(parameters: parameters)

        return response.value.items ?? []
    }

    // MARK: suggestions

    @Function(\Action.Cases.getSuggestions)
    private func _getSuggestions() async throws {

        filterViewModel.send(.getQueryFilters)

        var parameters = Paths.GetItemsByUserIDParameters()
        parameters.includeItemTypes = [.movie, .series]
        parameters.isRecursive = true
        parameters.limit = 10
        parameters.sortBy = [ItemSortBy.random.rawValue]

        let request = Paths.getItemsByUserID(userID: userSession.user.id, parameters: parameters)

        self.suggestions = response.value.items ?? []
    }
}
```

</details>

Essentially, remove the guard I told you to add (Sorry) and add a return from the `actuallySearch` for when we reset the filters. This should resolve the issues that you are seeing and bring this back to the intended behaviorr:

```
        guard query.isNotEmpty || filterViewModel.currentFilters.hasFilters else {
            self.items = [:]
            return
        }
```

---

I pushed these changes to the branch. Please let me know if this appears correct now with testing.

JPKribs:
Unrelated to this PR, I think we should add a `searchCases` or something to `ItemFilterCollection` to filter the filters down to search specific filters. Tags, Order, & Sort don't really belong there IMO. Tags will never populate and Order & Sort create weird edge cases for this where I can trigger a search based on`Order: Descending` but NOT by `Order: Ascending`. While not providing a real valid use case for searches IMO.

magshee:
No worries! Thank you very much for pointing fixing the two issues that I addressed!
I agree with you, that a selection of sort and order should not trigger a search. 
In `ItemFilterCollection` i added the variable `hasQueryableFilters` that checks if any of the other filters are selected. 

I also moved the guard condition into a function in the `SearchViewModel`, as this condition is used multiple times in the `SearchViewModel`, as well as in the `SearchView`. This would also make the `SearchView` a little bit cleaner and nicer to read in my opinion. 

I also removed the clearance of the items at the bottom of the `_search(...)` function, as this causes a minor glitch, when i have a filter selected and clear the search string. For a brief second, the "no results" text is shown, followed by the search.

edit: I kept tags in `hasQueryableFilters` as it made sense in my opinion to query for kinds like _action comedy_ movies.

Swiftfin tvOS/Views/SearchView.swift
```diff
@@ -181,6 +181,7 @@ struct SearchView: View {
                 ProgressView()
             }
         }
+        .animation(.linear(duration: 0.1), value: viewModel.filterViewModel.currentFilters.hasFilters)
```
LePips: `animation` isn't necessary.

Swiftfin/Views/SearchView.swift
```diff
@@ -203,6 +203,7 @@ struct SearchView: View {
                 ProgressView()
             }
         }
+        .animation(.linear(duration: 0.2), value: viewModel.filterViewModel.currentFilters.hasQueryableFilters)
```
LePips: `animation` isn't necessary.

Shared/ViewModels/SearchViewModel.swift
```diff
@@ -73,24 +73,28 @@ final class SearchViewModel: ViewModel {
...
                 guard let self else { return }
-                guard searchQuery.value.isNotEmpty else { return }
-                search(query: searchQuery.value)
+
+                actuallySearch(query: searchQuery.value)
             }
             .store(in: &cancellables)
     }
 
+    func hasSearchConstraints() -> Bool {
```
LePips: Please make a computed variable instead and this name sounds inverted. Should probably be something like `canSearch` instead.

LePips (changes requested):
I'm good with this for now, although what should actually happen is a proper `initial -> content` state loop.

magshee:
A proper `initial -> content` state loop would require a cancellation of pending searches, just like in the `_search` function of the `SearchViewModel` does it, when the search string is modified, right? 

My idea would be to add `viewModel.send(.cancel)` here:
```swift
let selectionBinding: Binding<[AnyItemFilter]> = Binding {
    viewModel.currentFilters[keyPath: type.collectionAnyKeyPath]
} set: { newValue in
    viewModel.send(.cancel)
    viewModel.send(.update(type, newValue))
}
```

However, I'm unsure whether I understood you correctly and even then if that is the right place to do, or if there are any other changes necessary then.

LePips (approved):
Apologies, my comment was a bit more high level and just would allow cleanup in the views.

It would show the suggestions on the `initial` state and then the results/empty state on `content`. However, that isn't important for this immediate work.

JPKribs (approved):
This looks good to me. Ran over testing below. Filters and text search the way I would expect it. Appropriately resets back to default when filters and text search are reset.

I think we've got it!

https://github.com/user-attachments/assets/6b386a25-671e-4012-9064-cdeac99ee6f4

magshee:
Great!
Thanks to the both of you for your guidance throughout this PR! :)

Issue #1846, opened 2025-12-10
Customizable Jump Forward Amounts


JPKribs:
### This feature request respects the following points:

- [x] This feature has **not** already been requested on [GitHub](https://github.com/jellyfin/Swiftfin/issues?q=is%3Aissue%20state%3Aopen%20type%3AFeature).
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).

### Platform

iOS

### Describe the feature you'd like to see

Setting to change the player's jump forward amount from 15 to another value.

PR #1849, opened 2025-12-13
Customizable Jump Intervals

JPKribs:
## Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1846

Adds an option to customize Jump Forward & Backwards intervals.

### tvOS

https://github.com/user-attachments/assets/0cef5511-52b4-4117-b5ee-796c3fcfea39

### iOS

https://github.com/user-attachments/assets/7868bd3d-b783-44cf-bdbb-af8219f903c7

## In Action

https://github.com/user-attachments/assets/294d1bb5-451d-49c3-941e-3787e0e914ea


JPKribs:
Some explanations to get ahead of potential questions:

1. The enum cannot be `CaseIterable` or `SupportedCaseIterable` because the `.custom(interval)` would require a valid interval for that.
2. I am not using a `Picker` because you cannot trigger an alert from a `Picker` only from a `Menu`.

JPKribs:
> CaseIterable is just a protocol with just a static var allCases requirement. When an enum has a case with an associated value, this just prevents the automatic synthesis that the compiler provides. Implementers would just have to implement allCases manually in scenarios where it cannot be synthesized, which is often.

Okay, I just assumed we wanted to avoid `allCases*` _`*Not ALL cases`_ haha.

I have the label for iOS 18+ now as well. I was struggling to do this in a way that didn't make this huge. I ended up doing this twice. Once the correct way for iOS 18+ and once in pieces to make this smaller and less duplicated for the current version of 16+. I have the 18+ version commented out with a TODO to swap this out when the time comes.

Here is what this looks like now:

https://github.com/user-attachments/assets/5ed9855d-3e1e-4559-8223-8d4ad28268af

https://github.com/user-attachments/assets/54cea31c-bc6e-4df0-9c43-6a399d22ef95

JPKribs:
Also, should be good on preventing 0 or negatives:

https://github.com/user-attachments/assets/ee6fdfd6-68a3-472c-bd87-7d9b93e5bd1f

JPKribs:
Once this is wrapped up, I can go back and do the same for these Max Failed Login Attempts & Maximum Sessions pickers that should use this as well. Should this be a `Picker` overload if I'm going to use this there as well? Similar to what you recommended: https://github.com/jellyfin/Swiftfin/pull/1838

<details>

<summary>The place in question</summary>

Simulator Screenshot - iPhone - 2025-12-14 at 01 10 32

</details>

---

Edit:

I overloaded the `Picker`. I think this works well / makes this easier to use. This also consolidates our `currentValueLabel` to make this easier to changeover when we move to iOS 18. I like how this turned out but please let me know if you would prefer I revert this to the more manual version of this instead of overloading.

JPKribs:
I have a version / change tested and working for migrating the MaxLogins and MaxSessions to this as well. Lemme know if we want this in this PR or another. The overall changes are the same so they kind of make sense where but they are in another part of the app so they also don't make sense here:

```swift
//
//
//

import JellyfinAPI
import SwiftUI

extension ServerUserPermissionsView {

    struct SessionsSection: View {

        @Binding
        var policy: UserPolicy

        private var isAdministrator: Bool {
            policy.isAdministrator ?? false
        }

        // MARK: - Body

        var body: some View {
            FailedLoginsSection
            MaxSessionsSection
        }

        // MARK: - Failed Logins Section

        @ViewBuilder
        private var FailedLoginsSection: some View {
            Section {
                Picker(
                    L10n.maximumFailedLoginPolicy,
                    selection: $policy.loginAttemptsBeforeLockout
                        .coalesce(0)
                        .map(
                            getter: { LoginFailurePolicy(rawValue: $0, isAdministrator: isAdministrator) },
                            setter: { $0.rawValue }
                        ),
                    currentValue: LoginFailurePolicy(rawValue: policy.loginAttemptsBeforeLockout ?? 0, isAdministrator: isAdministrator)
                        .displayTitle,
                    isCustom: { if case .custom = $0 { return true } else { return false } },
                    customTag: .custom(logins: 1),
                    customDefault: { _ in .custom(logins: LoginFailurePolicy.defaultLogins(isAdministrator: isAdministrator)) },
                    customDescription: L10n.enterCustomFailedLogins
                ) {
                    ForEach(LoginFailurePolicy.allCases, id: \.self) { policy in
                        Text(policy.displayTitle)
                            .tag(policy)
                    }
                } customInput: { $value in
                    TextField(
                        L10n.failedLogins,
                        value: $value.map(
                            getter: { if case let .custom(logins) = $0 { return logins } else { return 1 } },
                            setter: { .custom(logins: max(1, min($0, 1000))) }
                        ),
                        format: .number
                    )
                    .keyboardType(.numberPad)
                }
            } header: {
                Text(L10n.sessions)
            } footer: {
                VStack(alignment: .leading) {
                    Text(L10n.maximumFailedLoginPolicyDescription)

                    LearnMoreButton(L10n.maximumFailedLoginPolicy) {
                        LabeledContent(
                            L10n.lockedUsers,
                            value: L10n.maximumFailedLoginPolicyReenable
                        )
                        LabeledContent(
                            L10n.unlimited,
                            value: L10n.unlimitedFailedLoginDescription
                        )
                        LabeledContent(
                            L10n.default,
                            value: L10n.defaultFailedLoginDescription
                        )
                        LabeledContent(
                            L10n.custom,
                            value: L10n.customFailedLoginDescription
                        )
                    }
                }
            }
        }

        // MARK: - Max Sessions Section

        @ViewBuilder
        private var MaxSessionsSection: some View {
            Section {
                Picker(
                    L10n.maximumSessionsPolicy,
                    selection: $policy.maxActiveSessions
                        .coalesce(0)
                        .map(
                            getter: { ActiveSessionsPolicy(rawValue: $0) },
                            setter: { $0.rawValue }
                        ),
                    currentValue: ActiveSessionsPolicy(rawValue: policy.maxActiveSessions ?? 0).displayTitle,
                    isCustom: { if case .custom = $0 { return true } else { return false } },
                    customTag: .custom(sessions: 1),
                    customDefault: { _ in .custom(sessions: 1) },
                    customDescription: L10n.enterCustomMaxSessions
                ) {
                    ForEach(ActiveSessionsPolicy.allCases, id: \.self) { policy in
                        Text(policy.displayTitle)
                            .tag(policy)
                    }
                } customInput: { $value in
                    TextField(
                        L10n.maximumSessions,
                        value: $value.map(
                            getter: { if case let .custom(sessions) = $0 { return sessions } else { return 1 } },
                            setter: { .custom(sessions: max(1, min($0, 1000))) }
                        ),
                        format: .number
                    )
                    .keyboardType(.numberPad)
                }
            } footer: {
                VStack(alignment: .leading) {
                    Text(L10n.maximumConnectionsDescription)

                    LearnMoreButton(L10n.maximumSessionsPolicy) {
                        LabeledContent(
                            L10n.unlimited,
                            value: L10n.unlimitedConnectionsDescription
                        )
                        LabeledContent(
                            L10n.custom,
                            value: L10n.customConnectionsDescription
                        )
                    }
                }
            }
        }
    }
}
```

Shared/Extensions/JellyfinAPI/Picker.swift
```diff
@@ -0,0 +1,205 @@
...
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2025 Jellyfin & Jellyfin Contributors
+//
+
+import SwiftUI
+
+// MARK: - Picker Overloads
+
+func Picker<SelectionValue, Content, InputContent>(
```
LePips: I understand what you're trying to do (making a `Picker` API for "custom" values) but now this is going a bit too far.

I think I may have overemphasized making overloads of types. Good candidates for making overloads are if the feature itself can be self-encapsulated, don't require much overhead, and can be generalized in a way to cover many use-cases that doesn't require too much "assembly required" - like a protocol conformance vs a bunch of these parameters.

They can also just look nice when reading code. `CaseIterablePicker`, with some cleanup, would be a good next target.

I don't think there's a good way to generalize this pattern across use cases in a concise way and that this input component is important enough to justifies maintaining these overloads.

Actually, if I had remembered other areas where we have a similar pattern of static vs custom values I would have just had us mirror that instead for the time being as it also solves the current value label problem. If you want to, you can mirror that or revert back to something like I had suggested. I would be fine that on 16/17 the current value label would just be `Custom`.

JPKribs: No worries! I liked the override concept and I thought it fit here but I do agree. It's a pretty heavy implementation.

Since there are currently three locations (that I am aware of) that use this, do you think it makes sense to include that incorporate this into a struct of some kind or does it make more sense to do this ad hoc per location?

I can take a look at this later and do a bit of a reversion to the previous commit. Pending, what you think of the other two examples of this. I'd love to get this planned out around the time #1838 is ready.

From my end, for the rest of the year, I'd like to just focus on trying to get the settings all unified together and cleaned up where it makes sense. So, stuff like this I'd like to try and address/unify into a single solution _where it makes sense._

JPKribs: Latest commit reverts. If we want to go ad-hoc, this should be good to go. If we want to put this into a struct of some kind, this will need some additional work. Lemme know either way!

LePips (approved):
Each setting having their own component is just fine. Trying to come up with a general solution for "static + custom" seems like too much for now, unless we think of something later.

JPKribs:
@LePips sorry merge conflicts with strings from https://github.com/jellyfin/Swiftfin/pull/1834. Should be the only change. I ran the String Cleanup while I was fixing that:

Screenshot 2025-12-17 at 21 19 55

PR #1857, opened 2025-12-16
Unify `CustomizationSettingsView`

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1222

Consolidates all Customization settings into a single view for both iOS and tvOS. Also reorganizes the settings.

Customization is now in Customization -> Advanced instead of Accessibility -> Customize.

Removed all the sections that were their own files. I thought this made them cleaner and easier to use but the reality is this just meant searching more files to find things and often meant a lot of re-declarations that wasn't necessary. This is our largest single settings view and it's only 336 lines so I don't think this warrants the level of splitting I was doing prior and I never reuse the sections anyway so I think a single file is the most efficient and best way of doing this.

Settings are now ordered based on kind of the order you see this. Open to other suggestions but my thinking is:

1. `homeSettings` (First thing you see)
2. `mediaSettings` (First think you see in the media tab before going into a library)
3. `filterSettings` (First think you see in a library at the top)
4. `librarySettings` (Library settings as a whole)
5. `posterSettings` (Poster settings as a component of the library)
6. `itemSettings` (Item setting under the poster)
7. `itemViewSettings` (Item view when selecting the poster)
8. `itemManagementSettings` (Item management in the Item View)

Also migrates over the `IndicatorView` as that was 4 settings and a small add-on. I set this up for where we can add the poster for tvOS indicators to show as well. Left as a TODO for now.

### Videos

https://github.com/user-attachments/assets/6b6d0e9f-dfe9-4502-a587-6a2cf89cc81c

https://github.com/user-attachments/assets/ef9098ce-4ff6-4a02-b6ba-6d56de6f3ce6

### DIFF
--- a/Shared/Coordinators/Navigation/NavigationRoute/NavigationRoute+Settings.swift
+++ b/Shared/Coordinators/Navigation/NavigationRoute/NavigationRoute+Settings.swift
@@ -48,10 +48,10 @@ extension NavigationRoute {
         CustomDeviceProfilesView()
     }
 
-    static let customizationSettingsView = NavigationRoute(
-        id: "customizationSettingsView"
+    static let customizeSettingsView = NavigationRoute(
+        id: "customizeSettingsView"
     ) {
-        CustomizationSettingsView()
+        CustomizeSettingsView()
     }
 
     #if DEBUG
--- a/Shared/Views/SettingsView/CustomizeSettingsView.swift
+++ b/Shared/Views/SettingsView/CustomizeSettingsView.swift
@@ -12,7 +12,7 @@ import Factory
 import JellyfinAPI
 import SwiftUI
 
-struct CustomizationSettingsView: View {
+struct CustomizeSettingsView: View {
 
     #if os(tvOS)
     typealias PlatformPicker = ListRowMenu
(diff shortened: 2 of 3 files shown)


JPKribs:
Should be good to go. I think there is a better way to abstract `picker` and `listRowMenu` to eliminate the build flags but I am leaving that for a later PR in favor of just getting this settings merged into one. I've tried looking at this a few times in the past so I'm just excited to potentially get this all cleaned up finally!

JPKribs:
For the build error, the tvOS stepper needs a title for the _alert but there is no iOS init with a title. My workaround is an iOS Stepper extension where we take the title and just drop it for iOS:

```swift
extension Stepper {

    init<V: Strideable>(
        _ title: String,
        value: Binding<V>,
        in range: ClosedRange<V>,
        step: V.Stride = 1,
        @ViewBuilder label: @escaping () -> Label
    ) {
        self.init(value: value, in: range, step: step, label: label)
    }
}
```

The tvOS format was giving me trouble so I landed on `Strideable` instead of `where Format == VerbatimFormatStyle<Value>`. Testing this works for our usages at this point but I'll defer to you if this will cause us issues down the road.

The other route is build a build flag around each.

Issue #1818, opened 2025-11-17
Fix the `ListRowMenu` Styling


JPKribs:
When I first created `ListRowMenu`, this was on a much earlier version of tvOS and SwiftUI where some of the modifiers that we have today did not exist. As a result, there is a lot of trial and error focus logic in there that is no longer needed. 

https://github.com/user-attachments/assets/8e8fe348-2124-4b2b-bd7a-1a3186276167

Apple, in all their infinite wisdom, does not appear to allow us to mirror their default hover effect (see the `ChevronButton`s) so we will want to use `buttonStyle(.card)` or `.hoverEffect(.lift)` in their place. The goal should be to make this more in-line with the `PrimaryButtonStyle`. I'm currently leaning towards just overriding the `ChevronButton` `hoverEffect` to mirror our other buttons than trying to mirror the tvOS default behavior which is what I originally did with the `ListRowMenu`

I have a working version of this that gets us close but the goal should be to turn this into something that's a bit more polished and has more attention to detail than I am currently able to give this:

<details>

<summary>Working Version</summary>

```swift

//
// Swiftfin is subject to the terms of the Mozilla Public
// License, v2.0. If a copy of the MPL was not distributed with this
// file, you can obtain one at https://mozilla.org/MPL/2.0/.
//
// Copyright (c) 2025 Jellyfin & Jellyfin Contributors
//

import SwiftUI

struct ListRowMenu<Content: View, Subtitle: View>: View {

    private let title: Text
    private let subtitle: Subtitle?
    private let content: () -> Content

    var body: some View {
        Menu(content: content) {
            HStack {
                title
                    .foregroundStyle(.primary)
                    .padding(.leading, 4)

                Spacer()

                if let subtitle {
                    subtitle
                        .foregroundStyle(.secondary)
                }

                Image(systemName: "chevron.up.chevron.down")
                    .font(.body.weight(.regular))
                    .foregroundStyle(.secondary)
            }
            .padding(.horizontal)
            .frame(maxWidth: .infinity, maxHeight: .infinity, alignment: .leading)
            .hoverEffect(.lift)
        }
        .menuStyle(.borderlessButton)
        .listRowInsets(.zero)
        .listRowBackground(Color.clear)
    }
}

// MARK: - Initializers

// Base initializer
extension ListRowMenu where Subtitle == Text? {

    init(_ title: Text, @ViewBuilder content: @escaping () -> Content) {
        self.title = title
        self.subtitle = nil
[...]

PR #1858, opened 2025-12-16
[tvOS] `Stepper` & `ListRowMenu` for 26

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1818

Takes the iOS `BasicStepper` and makes this usable for tvOS. This should let us use the same API in both locations. Selecting Left decreases it. Selecting Right increases it. One it gets the upper or lower limit, you cannot increment it further in that direction.

If you click the button, it counts up and resets back to the lowest value from there.

This also creates a `.form` `ButtonStyle` for tvOS. This ensures that all Form buttons are styled in the same value and have the same hover animation & styling. In https://github.com/jellyfin/Swiftfin/pull/1857, I expand this to Pickers as well but for this PR this is just at the Form overload and the new `BasicStepper`. Videos below. If we choose to go this unified `.form` `ButtonStyle` route, this will need to be added to more locations. It will also automatically be added when `Form(systemImage: )` is used so we will get there eventually natively.

### Basic Stepper Video

https://github.com/user-attachments/assets/872ae065-09b7-48e2-a0e5-b0cb495f1256

### Form Button Style

Varying Button Styles. Apple has a default hover that we cannot replicate. We only have `.lift` & `.highlight` for `.hoverEffect`. This makes all form items use the same format so this should be more consistent.

<details>

<summary>Original Form Button Style</summary>

https://github.com/user-attachments/assets/b0084dd3-2285-4825-bbb0-9a5fc5875b16

</details>

<details>

<summary>Updated Form Button Style</summary>

https://github.com/user-attachments/assets/0c3fa6e1-4d9a-4e2e-a783-e72d40c719b2

</details>

### DIFF
--- a/Swiftfin tvOS/Extensions/View/Modifiers/_Alert.swift
+++ b/Swiftfin tvOS/Extensions/View/Modifiers/_Alert.swift
@@ -9,6 +9,8 @@
 import Engine
 import SwiftUI
 
+// TODO: tvOS 26: `.background(ultraThinMaterial)` to `.glassEffect`
+
 extension View {
 
     func _alert(


Swiftfin tvOS/Components/FormButtonStyle.swift
```diff
@@ -0,0 +1,35 @@
...
+// Swiftfin is subject to the terms of the Mozilla Public
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2025 Jellyfin & Jellyfin Contributors
+//
+
+import SwiftUI
+
+extension PrimitiveButtonStyle where Self == FormButtonStyle {
```
LePips: Let's please not make a button style like this. While it is very annoying that `Menu/Picker` don't have the same styling as `Button`, our workaround for it is just fine and I wouldn't instead change how everything else looks because of it. Instead, I would rather take a look at how we can make `ListRowMenu` more ergonomic for `Menu` or `Picker` use cases.

JPKribs: If there is a clean way to do this I am game but this was my like... 40+ hours of staring at buttons solution to this. I don't love we have to do this but I ended up here for 3 reasons:

1. The default `focusEffect` isn't accessible to us. I find this really annoying because like... Why doesn't Apple just include this?? See below the difference between `.hover` and whatever the default button is:

Screenshot 2025-12-17 at 20 42 48

2. The default `focusEffect` is very difficult for us to recreate. It looks like some slight increase & color change but it's like... Turn the button white with black text then place a different shade of white as a slightly vertically increase button shape but nothing horizontal. See color weirdness below:

Screenshot 2025-12-17 at 20 41 44



3. Finally, in tvOS 18, I got the `ListRowMenu` to very closely mirroring Apple's default but then they changed the default. The only way to prevent getting out of sync is for us to own the style.

Again, I'm not opposed to another solution for this but I really hate this button haha. This is kind of my nuclear option for this. I thought by isolating this to purely just the form that would make this more acceptable.

I'll go whatever direction you think is best for this. I think it might also make more sense to pull this part from this PR and tackle this separately too if we have a cleaner way for this

LePips: I'm very much familiar the frustrations that `Menu` and `Picker` are bringing us here but having something like `ListRowMenu` mimic it as close as we can is seriously just fine. Yes, that focus effect changed in 18 and is back in 26 with compatibility mode disabled, but is just such a small thing I could hardly see it.

How everything is styled goes a bit beyond SwiftUI. If you do think that owning the style entirely is the way to go, then I would prefer that it still matches the default style. If that isn't achievable just in the `ButtonStyle` itself, then we shouldn't pursue it.

JPKribs: tvOS `ButtonStyle` has been a giant gripe of mine with the platform... It's starting to feel personal haha. I'll move this to a draft and take crack at this without the `ButtonStyle` later in the week!

JPKribs: Okay. How does this look? I think the animation is a tiny bit off but I don't know if I can get much close than this:

https://github.com/user-attachments/assets/f0227fca-8a56-494e-862f-3513f9e3046e

The color is also a tiny bit off on the .sheet but that's because we are using the workaround for:

https://developer.apple.com/documentation/tvos-release-notes/tvos-26_1-release-notes

_Allegedly_, this is resolved in 26.2 but I tried reverting back to our previous usage but this still seems to be the smaller, iPad-esque modal. But this is what this looks like via .sheet:

Simulator Screenshot - Apple TV - 2025-12-18 at 12 03 31

LePips: I think that looks great! Let's just version it to the relevant OSs that have the focus effect. I'm fine with that `sheet` presentation for now, I have some changes in mind that would eliminate usage of `sheet` anyways.

JPKribs:
I think this is ready. I was finding that any `ButtonStyle` was interfering with the left/right remote actions. So I reverted the `BasicStepper` for tvOS back to something similar to where we started. My big change was using the Router for opening the `StepperView` instead of a `.sheet`. Additionally, I changed the current value in the stepper to be a `Textbox` to allow more manual entry if needed.

The Save button feels a bit weird. I feel it's more likely I'll accidentally forget to save and have to redo it than I will make a change and wish I could revert. So up to remove that if others agree.

Both iOS and tvOS should share an API for `BasicStepper` which should make sharing these settings either. Although, the tvOS version is a .sheet to another view.

Finally, I got the color to match for all `ListRowMenu` regardless of background. The coloring was because I forgot a `listRowBackground(.clear)` where I was testing so removing this and using the default background was the correct route where, for a bit there, I was trying to figure out how to match the default background manually...

So, colors all match, ignore that comment about trying to color match that was a mistake we're fine haha.

https://github.com/user-attachments/assets/6a325a6d-ffe1-494e-baf4-67f0920cd4ab

Example with formatting for the subtitle on the `ChevronButton`. Also a good example of where I believe a `TextField` is valuable as 0-30 means 15 button presses to get to the mid-point:

https://github.com/user-attachments/assets/19c64166-e6fb-49b0-bcc3-d6ead3fcab4f

LePips:
> Finally, in tvOS 18

Something I missed from that comment is that since we currently tvOS 17 minimum and no devices are dropped for tvOS 18, it could possibly be a bit better to just update to tvOS 18 minimum here.

- https://github.com/jellyfin/Swiftfin/pull/1076
- https://github.com/jellyfin/Swiftfin/pull/1455

JPKribs:
It looks like no one is missing 26 either:

https://en.wikipedia.org/wiki/TvOS#Supported_OS_releases

The only downside would be missing people who haven't updated.

Most of what I'm seeing for 26 is Liquid Glass specific. Google's AI summary claims "tvOS 26 brings a new focus engine" but I see no evidence of that in documentation...

That would be iOS 16 and tvOS 18. Would we want to do iOS 17 at the same time? 

My only thought for 18 is I am a fan of:

https://developer.apple.com/documentation/swiftui/tabviewstyle/sidebaradaptable

LePips:
I'm reluctant for a while on 26 minimum anytime soon, let's just do tvOS 18. For iOS, I probably want a few more things before bumping to iOS 18.

JPKribs:
Sounds good. For this PR, I have a check for 26+ for this newer button styling and the rest falls back to the previous. I did a second pass on button sizing since I think they were too wide. I think this PR should be good for a review:

Simulator Screenshot - Apple TV - 2025-12-22 at 11 25 31

I can also move the `ListRowMenu` changes to their own PR now since that style is no longer related to the `BasicStepper` on tvOS at all.

LePips:
Now coming back to this, I now want to just remove `BasicStepper` altogether. For iOS, just use a `Stepper` with a `LabeledContent` for the label directly.

For tvOS, we can make a `Stepper` function with the same signature that then presents the stepper overlay. I would also want us to rather make the stepper present a view that mimics the native `alert` view, which can be done somewhat cleanly. We would also do the same for some pickers, like the `HourMinutePicker` for the platform.

These changes don't need to happen in this PR.

JPKribs:
> Now coming back to this, I now want to just remove `BasicStepper` altogether. For iOS, just use a `Stepper` with a `LabeledContent` for the label directly.

I've got this knocked out and migrated in my last commit.

> For tvOS, we can make a `Stepper` function with the same signature that then presents the stepper overlay. I would also want us to rather make the stepper present a view that mimics the native `alert` view, which can be done somewhat cleanly. We would also do the same for some pickers, like the `HourMinutePicker` for the platform.

Let me know if I went too overkill on this. I ended up with:

`InputSheet{}` which is a `ViewModifier` to call `InputSheet` the tvOS struct. Later, `HourMinutePicker` can use this as well. I've tried using `.alert` for these but unfortunately, you can get the buttons to work but they also close so you can only iterate a single value at a time. `InputSheet` was just the best name I could think of happy to change that up.

---

### Regular Alert:

Simulator Screenshot - Apple TV - 2026-02-15 at 16 03 17

### Input Sheet Stepper:

Simulator Screenshot - Apple TV - 2026-02-15 at 16 03 11

JPKribs:
Also, I am using a lot of `geometry.size.width` and `geometry.size.height`. For the reason that testing this on hardware for a smaller screen it was hard to pinpoint a good size in straight up height and width values so I used a more dynamic percentage. Works better in testing but it also comes with its own host of trouble just since it's less static. Happy to swap this out if we'd prefer but I'm up for whatever is best.

Also, I found there is a difference in the Alert where the content has large horizontal padding while the buttons have very narrow padding. This resulted in a design I don't like where the buttons are a different closure than the content. IMO, this is an unpleasant way of doing this but I couldn't figure out a cleaner way of doing this.

Also, this fully removes the `BasicStepper`. I've included iOS in video below. Mirrors exactly what you were suggesting and the parameters to call it are mirrored for tvOS so the views should be sharable now!

---

## Videos

### tvOS

https://github.com/user-attachments/assets/e1706062-e54a-43b5-a480-31255ac8282b

### iOS

https://github.com/user-attachments/assets/2929782f-44e2-482e-bc8e-0267300ab3b8

JPKribs:
Okay, sorry just took a final pass of cleanup with a fresh set of eyes. I am done messing this as of [a0c805b](https://github.com/jellyfin/Swiftfin/pull/1858/commits/a0c805b1f7b1ba8a142cb42c6cf56fcb51effc7f).

JPKribs:
@LePips can I ask why `._alert` over `.inputSheet`? Just for future reference, I'm trying to get a better understanding for when `_name` usage is used

LePips:
There's no set-in-stone rule. Here, we're just making some container view to mimic the native `alert` so matching it but underscored just for consistency. There's nothing that forces it to be for value modification or even have user input, views that use `_alert` would just provide their respective bodies.

Issue #1852, opened 2025-12-15
Dolby Vision broken when remuxing - seems server is not informed about Swiftfin client DV compatibility


DrHurt:
### This issue respects the following points:

- [x] This is a **bug**, not a question or a configuration issue; Please visit our [forum or chat rooms](https://jellyfin.org/contact/) first to troubleshoot with volunteers, before creating a report.
- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/Swiftfin/issues?q=is%3Aopen+is%3Aissue).
- [x] I'm using an up to date version of Swiftfin; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).

### Description of the bug

It seems Swiftfin is not reporting to the server Dolby Vision compatibility. 
My current understanding is that Jellyfin 10.11 provides the API for that.

When watching mkv files from server 10.11.5 on macOS to Swiftfin 1.4 on iPhone, the server correctly remuxes the mkv to mp4 HLS however it seems to not use the ffmpeg -strict unofficial flag thus Dolby Vision data is dropped.

The ffmpeg log shows: [mp4 @ 0x7f87f380aa80] Not writing 'dvcC'/'dvvC' box. Requires -strict unofficial.

**_Impact of the bug:_**
DV Profile 5: green/purple as no HDR10 layer to fall back to
DV Profile 8: falls back to HDR10
DV Profile 7: appears to fall back to HDR

This will probably be a bigger problem on TV where native DV is more important (untested by me)

testing playback in Safari, the server correctly uses ffmpeg -tag:v dvh1 -strict -2

### Reproduction steps

1- start server 10.11.5 on macOS
2- start Swiftfin 1.4 on iPhone or iPad - (use native player for proper DV support)
3- play an mkv DV P5 file -> green/pink
4- play an mkv DV P8 file -> falls back to HDR

### Platform

iOS

### Application version

1.4

### Where did you install the app from?

None

### Device information

iPhone 15

### OS version

iOS 26.2

### Jellyfin server version

10.11.5

### Relevant log output

```shell

```

JPKribs:
I believe this is the same issue as https://github.com/jellyfin/Swiftfin/issues/1854.

DrHurt:
[...]

Issue #1848, opened 2025-12-13
Native AVKit player stops playing E-AC3 multi-channel audio after couple seconds


karolpeszek:
### This issue respects the following points:

- [x] This is a **bug**, not a question or a configuration issue; Please visit our [forum or chat rooms](https://jellyfin.org/contact/) first to troubleshoot with volunteers, before creating a report.
- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/Swiftfin/issues?q=is%3Aopen+is%3Aissue).
- [x] I'm using an up to date version of Swiftfin; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).

### Description of the bug

Swiftfin version: 1.3 and 1.4
iPadOS version: 26.0.1, 26.1 and 26.2
Audio codec: E-AC3 5.1

# Brief description
While playing the movie using the native AVKit after couple of seconds the audio stops. I've tested both 5.1 audio and 2.0. The latter plays without any issues.

# Expected behaviour
The native player should play the audio without any issues.

# Media information
```
Input #0, mov,mp4,m4a,3gp,3g2,mj2, from '/jellyfin/Movies/Django Unchained.mp4':
  Metadata:
    major_brand     : isom
    minor_version   : 512
    compatible_brands: isomdby1iso2mp41
    encoder         : Lavf59.27.100
  Duration: 02:45:22.92, start: 0.000000, bitrate: 24238 kb/s
  Stream #0:0[0x1](und): Video: hevc (Main 10) (hvc1 / 0x31637668), yuv420p10le(tv, bt2020nc/bt2020/smpte2084, progressive), 3840x1600 [SAR 1:1 DAR 12:5], 23781 kb/s, 23.98 fps, 23.98 tbr, 16k tbn (default)
      Metadata:
        handler_name    : VideoHandler
        vendor_id       : [0][0][0][0]
        encoder         : Lavc59.37.100 libx265
  Stream #0:1[0x2](eng): Audio: eac3 (ec-3 / 0x332D6365), 48000 Hz, 5.1(side), fltp, 448 kb/s (default)
      Metadata:
        handler_name    : SoundHandler
        vendor_id       : [0][0][0][0]
      Side data:
        audio service type: main
[...]

PR #1861, opened 2025-12-17
[10.11] Enable HDR/DV Device Profiles Conditions

JPKribs:
## Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1852

Adds `VideoRangeType`s to the Device Profiles for the Swiftin & Native Players. I don't have a ton of DV for testing but what I do have tests appropriately.

VLCKit:
- `.sdr`
- `.hdr10`
- `.hdr10Plus`
- `.doviWithSDR`
- `.doviWithHDR10`
- `.doviWithHDR10Plus`

AVKit:          
- `.sdr`
- `.hdr10`
- `.hdr10Plus`
- `.dovi`
- `.doviWithSDR`
- `.doviWithHDR10`
- `.doviWithHDR10Plus`

If any of these are incorrect, please let me know. I've been able to test all of these I believe. Some of what is tricky is there are some VLCKit are invalid formats but VLC is able to handle tone mapping inside the player. Primarily, it looks like DV without a fallback is where issues originate.

In my searching for AVKit, I believe these are all correct but If anyone is more familiar with these types let me know. I tend to avoid these as I don't have a TV for it and, even if I did, I cannot personally tell the difference.

---

I also updated the `ProfileCondition.value` to be a closure for ease of input and mapping. This required me to create a `PipeStringBuilder` which is just the `ComaStringBuilder` but with Pipes. If there is a good way to combine those, that might be best.

---

### Before

https://github.com/user-attachments/assets/fb3f2f70-afb7-4a37-8dfc-f3f605d69906

### After

https://github.com/user-attachments/assets/ae1bcf47-ca95-41f3-8838-b99a51171127

### DIFF
--- a/Shared/Objects/VideoPlayerType/VideoPlayerType+Shared.swift
+++ b/Shared/Objects/VideoPlayerType/VideoPlayerType+Shared.swift
@@ -25,67 +25,65 @@ extension VideoPlayerType {
 
     // MARK: - Shared Codec Profile Conditions
 
+    @ArrayBuilder<ProfileCondition>
     static var _h264BaseConditions: [ProfileCondition] {
-        [
-            ProfileCondition(
-                condition: .notEquals,
-                isRequired: false,
-                property: .isAnamorphic,
-                value: "true"
-            ),
-            ProfileCondition(
-                condition: .equalsAny,
-                isRequired: false,
-                property: .videoProfile
-            ) {
-                H264Profile.high
-                H264Profile.main
-                H264Profile.baseline
-                H264Profile.constrainedBaseline
-            },
-            ProfileCondition(
-                condition: .lessThanEqual,
-                isRequired: false,
-                property: .videoLevel,
-                value: "80"
-            ),
-            ProfileCondition(
-                condition: .notEquals,
-                isRequired: false,
-                property: .isInterlaced,
-                value: "true"
-            ),
-        ]
+        ProfileCondition(
+            condition: .notEquals,
+            isRequired: false,
+            property: .isAnamorphic,
+            value: "true"
+        )
+        ProfileCondition(
+            condition: .equalsAny,
+            isRequired: false,
+            property: .videoProfile
+        ) {
+            H264Profile.high
+            H264Profile.main
+            H264Profile.baseline
+            H264Profile.constrainedBaseline
+        }
+        ProfileCondition(
+            condition: .lessThanEqual,
+            isRequired: false,
+            property: .videoLevel,
+            value: "80"
+        )
+        ProfileCondition(
+            condition: .notEquals,
+            isRequired: false,
+            property: .isInterlaced,
+            value: "true"
+        )
     }
 
+    @ArrayBuilder<ProfileCondition>
     static var _hevcBaseConditions: [ProfileCondition] {
-        [
-            ProfileCondition(
-                condition: .notEquals,
-                isRequired: false,
-                property: .isAnamorphic,
-                value: "true"
-            ),
-            ProfileCondition(
-                condition: .equalsAny,
-                isRequired: false,
-                property: .videoProfile
-            ) {
-                HEVCProfile.main
-                HEVCProfile.main10
-            },
-            ProfileCondition(
-                condition: .lessThanEqual,
-                isRequired: false,
-                property: .videoLevel,
-                value: "175"
-            ),
-            ProfileCondition(
-                condition: .notEquals,
-                isRequired: false,
-                property: .isInterlaced,
-                value: "true"
-            ),
-        ]
+        ProfileCondition(
+            condition: .notEquals,
+            isRequired: false,
+            property: .isAnamorphic,
+            value: "true"
+        )
+        ProfileCondition(
+            condition: .equalsAny,
+            isRequired: false,
+            property: .videoProfile
+        ) {
+            HEVCProfile.main
+            HEVCProfile.main10
+        }
+        ProfileCondition(
+            condition: .lessThanEqual,
+            isRequired: false,
+            property: .videoLevel,
+            value: "175"
+        )
+        ProfileCondition(
+            condition: .notEquals,
+            isRequired: false,
+            property: .isInterlaced,
+            value: "true"
+        )
     }
 }
--- a/Shared/Objects/VideoPlayerType/VideoPlayerType+Swiftfin.swift
+++ b/Shared/Objects/VideoPlayerType/VideoPlayerType+Swiftfin.swift
@@ -144,7 +144,9 @@ extension VideoPlayerType {
     static var _swiftfinCodecProfiles: [CodecProfile] {
         CodecProfile(
             codec: VideoCodec.h264.rawValue,
-            conditions: _h264BaseConditions.appending(
+            type: .video,
+            conditions: {
+                _h264BaseConditions
                 ProfileCondition(
                     condition: .equalsAny,
                     isRequired: true,
@@ -152,19 +154,19 @@ extension VideoPlayerType {
                 ) {
                     VideoRangeType.sdr
                 }
-            ),
-            type: .video
+            }
         )
 
         CodecProfile(
             codec: VideoCodec.hevc.rawValue,
-            conditions: [
+            type: .video,
+            conditions: {
                 ProfileCondition(
                     condition: .notEquals,
                     isRequired: false,
                     property: .isAnamorphic,
                     value: "true"
-                ),
+                )
                 ProfileCondition(
                     condition: .equalsAny,
                     isRequired: false,
@@ -172,13 +174,13 @@ extension VideoPlayerType {
                 ) {
                     HEVCProfile.main
                     HEVCProfile.main10
-                },
+                }
                 ProfileCondition(
                     condition: .notEquals,
                     isRequired: false,
                     property: .isInterlaced,
                     value: "true"
-                ),
+                )
                 ProfileCondition(
                     condition: .equalsAny,
                     isRequired: true,
@@ -190,26 +192,26 @@ extension VideoPlayerType {
                     VideoRangeType.doviWithSDR
                     VideoRangeType.doviWithHDR10
                     VideoRangeType.doviWithHDR10Plus
-                },
-            ],
-            type: .video
+                }
+            }
         )
 
         CodecProfile(
             codec: VideoCodec.av1.rawValue,
-            conditions: [
+            type: .video,
+            conditions: {
                 ProfileCondition(
                     condition: .notEquals,
                     isRequired: false,
                     property: .isAnamorphic,
                     value: "true"
-                ),
+                )
                 ProfileCondition(
                     condition: .notEquals,
                     isRequired: false,
                     property: .isInterlaced,
                     value: "true"
-                ),
+                )
                 ProfileCondition(
                     condition: .equalsAny,
                     isRequired: true,
@@ -221,26 +223,26 @@ extension VideoPlayerType {
                     VideoRangeType.doviWithSDR
                     VideoRangeType.doviWithHDR10
                     VideoRangeType.doviWithHDR10Plus
-                },
-            ],
-            type: .video
+                }
+            }
         )
 
         CodecProfile(
             codec: VideoCodec.vp9.rawValue,
-            conditions: [
+            type: .video,
+            conditions: {
                 ProfileCondition(
                     condition: .notEquals,
                     isRequired: false,
                     property: .isAnamorphic,
                     value: "true"
-                ),
+                )
                 ProfileCondition(
                     condition: .notEquals,
                     isRequired: false,
                     property: .isInterlaced,
                     value: "true"
-                ),
+                )
                 ProfileCondition(
                     condition: .equalsAny,
                     isRequired: true,
@@ -252,9 +254,8 @@ extension VideoPlayerType {
                     VideoRangeType.doviWithSDR
                     VideoRangeType.doviWithHDR10
                     VideoRangeType.doviWithHDR10Plus
-                },
-            ],
-            type: .video
+                }
+            }
         )
     }
 }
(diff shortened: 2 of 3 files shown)


JPKribs:
I did some cleanup for `VideoProfiles` as well so we could use them in the same closure as other `ProfileCondition`s. This isn't a requirement for this PR. Feel free to let me know if we want to revert this.

---

Also added `.eac3` to the Native Player. I'm not sure where that got dropped but we have it documented as a supported audio type and it's one of the few advertised AVKit codecs. I don't think this resolves https://github.com/jellyfin/Swiftfin/issues/1848 but I haven't been able to recreate that issue so this could be related?

Shared/Objects/VideoPlayerType/VideoPlayerType+Shared.swift
```diff
@@ -11,76 +11,81 @@ import JellyfinAPI
...
-                )
+    // MARK: - Shared Codec Profile Conditions
 
-                ProfileCondition(
-                    condition: .lessThanEqual,
-                    isRequired: false,
-                    property: .videoLevel,
-                    value: "175"
-                )
+    static var _h264BaseConditions: [ProfileCondition] {
```
LePips, line 29: Attach these with `@ArrayBuilder<ProfileCondition>`

Shared/Objects/VideoPlayerType/VideoPlayerType+Swiftfin.swift
```diff
@@ -137,4 +137,124 @@ extension VideoPlayerType {
             SubtitleFormat.xsub
         }
     }
+
+    // MARK: - Codec Profiles
+
+    @ArrayBuilder<CodecProfile>
+    static var _swiftfinCodecProfiles: [CodecProfile] {
+        CodecProfile(
```
LePips, line 145: Use the init as defined in `CodecProfile` with the `ArrayBuilder` parameters for the profiles in this file.

JPKribs:
> I trust that the values are correct as I also won't be able to test them myself anytime soon, just some housekeeping.

While I am working on this, do we want a toggle to force HDR transcoding? I know some other clients have this. This would just take these and change this to `.sdr` as `isRequired: true`

JPKribs:
Revisions are made and tested. On Server, in testing on my last commit, you can see this is reporting and tone mapping / transcoding appropriately. I've included a screenshot of Swiftfin player on Main.

Screenshot 2025-12-18 at 18 30 19

### Native Player - No Transcode

NativePlayer - No Transcode

### Swiftfin - This PR

Swiftfin - Transcode

### Swiftfin - Pre-PR

Swiftfin - Pre-PR

### Notice/Question

VLC actually handles HDR10 and the documentation should reflect this now. VLC is also able to locally handle HDR10+ by treating it as HDR10. This isn't a direct usage of HDR10+ so I dunno if this is better to transcode or handle as HDR10. Transcoding probably handles this more "color accurate" but IMO, I'd rather just take the HDR10 version of HDR10+ than force the transcoding?

I've updated our documentation for this. For now, this PR is allowing HDR10+ since VLC will just ignore the dynamic metadata and handle this as HDR10.

I don't really use HDR10+ or DV so like... I'm not positive what is optimal here. I've taken the "Prioritize Direct Play" route and, if someone has stronger opinions, I am happy to adjust that or, if it makes sense, adding a like HDR toggle or something.

Screenshots I went for something with skin tones since I think that is the most obvious issues for DV. Hopefully, that is a good example/test.

holow29:
IMO it doesn't make sense to transcode HDR10+ if HDR10 is supported. Just like DoVi with HDR10 fallback is better to fallback to HDR10 on devices that don't support DoVi. Maybe in the future, there could be an option for it, but I don't think tonemapping server-side is going to produce a better image currently anyway - I actually think the HDR10 base layer will be more color-accurate.

JPKribs:
> IMO it doesn't make sense to transcode HDR10+ if HDR10 is supported. Just like DoVi with HDR10 fallback is better to fallback to HDR10 on devices that don't support DoVi. Maybe in the future, there could be an option for it, but I don't think tonemapping server-side is going to produce a better image currently anyway - I actually think the HDR10 base layer will be more color-accurate.

Good to know! I assumed as much but I appreciate a more qualified opinion haha

LePips (approved):

DrHurt:
> IMO it doesn't make sense to transcode HDR10+ if HDR10 is supported. Just like DoVi with HDR10 fallback is better to fallback to HDR10 on devices that don't support DoVi. Maybe in the future, there could be an option for it, but I don't think tonemapping server-side is going to produce a better image currently anyway - I actually think the HDR10 base layer will be more color-accurate.

+1

in my experience avplayer does an excellent job automatically falling back or tone mapping as required (as long as HLS manifest is compliant with Apple requirements which JellyFin servers seem to do very well anyway so no problem there)

DrHurt:
@JPKribs @LePips 

AVKit also supports Dolby Vision Profile 8.4:
`doviWithHLG`

Direct play profiles seems to be missing:
mp4 + `VideoCodec.h264`
m4v + `VideoCodec.hevc`

May I suggest those also get added to the device profile.

Thanks

JPKribs:
> AVKit also supports Dolby Vision Profile 8.4:
> doviWithHLG

Only for iPhone 12 and above:

https://professionalsupport.dolby.com/s/article/A-Guide-to-Dolby-Vision-on-iOS-From-Capture-to-Playback?language=en_US

Swiftfin supports iOS 16 which includes devices that do not have the appropriate SoC for this:

https://support.apple.com/en-us/103267

This was originally left out since I don't have a way to test this personally and this would create issues for a lot of supported devices. It might be worth seeing if we can add/remove based on SoC but for now I'm not personally comfortable adding what I cannot test.

JPKribs:
Actually, there is a way to pull supported HDR from AVPlayer that would be better for us to use. I can make a PR for that.

---

Nevermind. There is an `AVPlayer.availableHDRModes` and `AVPlayer.eligibleForHDRPlayback` but these are purely for whether the device is **connected to an HDR display** but does not tell us if the device is capable of it.

DrHurt:
I’m just brain storming here.

iPhone 11 support ends this year. It’s a very old device now. 

All iPhones 12 to 17 record in DV 8.4 **by default** so it’s an important profile to support for playing home videos.

Also, it looks like AVPlayer supports pure HLG as well which I can’t see in Swiftfin’s profile.

JPKribs:
You are welcome to create a PR with the changes to discuss this further. I cannot personally test these formats so I will not be making a PR for an item I cannot test.

I spent the weekend looking at and my only real takeaway is the profile for MP4 and M4V should be merged. It also might be prudent to have a toggle for tvOS to transcode HDR content if `AVPlayer.eligibleForHDRPlayback` is false.

DrHurt:
> You are welcome to create a PR with the changes to discuss this further. I cannot personally test these formats so I will not be making a PR for an item I cannot test.
> 
> I spent the weekend looking at and my only real takeaway is the profile for MP4 and M4V should be merged. It also might be prudent to have a toggle for tvOS to transcode HDR content if `AVPlayer.eligibleForHDRPlayback` is false.

I don’t currently have the setup to provide PRs so I’m just reporting as an end user.

Sounds sensible to merge mp4 and m4v though as they’re practically the same.

AVPlayer is capable of client side tone mapping when the display isn’t HDR compatible (but that depends on the HLS manifest). 

To get AVPlayer to tone map HDR/DV to SDR, you need to pass on the VOD playlist (the one with mp4 segments list) to AVPlayer not the master.m3u8 (tested on macOS not tvOS)

In my humble opinion, that is the more colour accurate and resource efficient way to do it.

JPKribs:
> I don’t currently have the setup to provide PRs so I’m just reporting as an end user.

Gotcha. I can try and look into something today. 

~It's been on my list for a while to just **make** some test files for these types of things. Just something with various audio tracks and some variations with DV, HDR10+ etc. that are very basic like "this color should be orange".~

~Do you happen to know about anything like this for testing DV/HDR?~

Edit: Turns out some folks on the Server team just have these ready to go! I can test this more fully today and get a PR out there.

DrHurt:
> > I don’t currently have the setup to provide PRs so I’m just reporting as an end user.
> 
> Gotcha. I can try and look into something today.
> 
> ~It's been on my list for a while to just **make** some test files for these types of things. Just something with various audio tracks and some variations with DV, HDR10+ etc. that are very basic like "this color should be orange".~
> 
> ~Do you happen to know about anything like this for testing DV/HDR?~
> 
> Edit: Turns out some folks on the Server team just have these ready to go! I can test this more fully today and get a PR out there.

some samples can be found here.
https://fra1.mirror.jellyfin.org/test-videos/

Bear in mind that AVPlayer behaves differently when playing HDR/DV mp4 (direct play) vs mkv -> HLS (remux & direct stream from server).

I suggest testing with DoVi P5 in MKV (because it has no HDR10 or SDR fall back so turns green/purple if not implemented & decoded correctly). Or just fails to play altogether.

HDR/DV mp4 are automatically tone mapped by AVPlayer if display is SDR (via direct play) 

HLS will NOT tone map on an SDR display if you try to play the server-provided master.m3u8 (because it tries to force mode switch to HDR which doesn’t exist so I get a black screen)

If you pass the VOD playlist (instead of master.m3u8), AVPlayer will tone map to SDR (regardless of display being HDR or SDR).

in short, you could use AVPlayer.eligibleForHDRPlayback to determine whether to play server provided master.m3u8 on HDR capable displays or pass on the VOD playlist instead on unsupported displays.

DrHurt:
That said, I don’t know how reliable AVPlayer.eligibleForHDRPlayback is for determining whether TV is HDR ready (meaning hardware is capable and tvOS is set to HDR/DV or SDR with match range enabled).

Someone will need to test that.

ps. Upstream server team are aware of this master vs VOD playlist issue affecting Apple tone mapping and are thinking of a reliable way to implement it server side.

JPKribs:
> https://fra1.mirror.jellyfin.org/test-videos/

I just found that as well! That helps a ton. Do you think this accurately captures our edge cases? Is there any value in 4K as well? I only bought this mac to work on Swiftfin so I cheaped out on storage so I'm shooting to keep this as small as possible haha

Screenshot 2026-01-05 at 14 14 37

DrHurt:
no need for 4k

Only thing missing here would be DV AV1 aka P10 (which is supported by A17 (pro) / M3 & newer). 
Current Apple TV chipset is too old for AV1 but next ATV will almost certainly support it.

test file: https://mega.nz/file/aPBiACoT#A3BqP2CXQrTguj-K82qWO4jrTft1w1UDFn2oEfES-FE (untested by me - but needs audio transcoding) 

I would keep testing to mkv files as they are the tricky ones that require server streaming via HLS (without transcoding if everything is setup correctly).

DrHurt:
You can use Subler (free OSS macOS app) to remux that AV1 DV P10 mkv to mp4 (and it can also convert the DTS audio to AAC).

The server should however stream the video directly & only transcode the audio if everything is setup correctly (will only work with A17 pro & regular A18 and newer) if you play the mkv.

JPKribs:
> will only work with A17 pro & regular A18 and newer

Hm. That gets me back into trouble. I only have an iPhone 13 Pro so I can see that the color is incorrect but I am not able to test if we are able to `DirectPlay` on a newer device. The good news is I found a way to get the [SoC from the Metal APIs](https://developer.apple.com/documentation/metal/detecting-gpu-features-and-metal-software-versions) but now I'm realizing I can only really cater the resolves to the hardware I own haha:

Screenshot 2026-01-05 at 15 43 16

DrHurt:
Swiftfin AVPlayer profile already seems to support AV1 including HDR/DV.

mp4 should direct play, and mkv should direct stream without transcoding on newer devices.

on older devices, it would need to transcode or fallback to vlc

———————————-

I think you should also look into how AVPlayer.eligibleForHDRPlayback behaves on tvOS. Is impacted by setting ATV output to SDR with match content on / off, HDR & DV?

I say that because setting the ATV to output SDR with match content off breaks HDR via HLS (master.m3u8) in my testing. File completely fails to play. MP4 files tone map & play though.

skipping master.m3u8 and playing the vod playlist directly fixes HDR via HLS on SDR mode by tone mapping just like direct playing the mp4.

this isn’t the case with iPhones as their internal display identifies as HDR since at least iPhone 12.

I did my testing a while back so might be worth rechecking on tvOS 26 to see if this behaviour persists.

JPKribs:
Does this look right?

```swift
    var supportsHEVCDecode: Bool {
        (isAppleSilicon && self >= .apple3) || (isIntelMac && self >= .mac2)
    }

    var supportsAV1Decode: Bool {
        isAppleSilicon && self >= .apple9
    }

    var supportsHDR10Decode: Bool {
        isAppleSilicon && self >= .apple4
    }

    var supportsHLGDecode: Bool {
        isAppleSilicon && self >= .apple4
    }

    var supportsDolbyVisionDecode: Bool {
        isAppleSilicon && self >= .apple5
    }
```

For reference:

GPU Family | Chips
-- | --
apple1 | A7, A8
apple2 | A8
apple3 | A9, A10
apple4 | A11
apple5 | A12
apple6 | A13
apple7 | A14, M1
apple8 | A15, A16, M2
apple9 | A17 Pro, M3
apple10 | A18, A18 Pro, A19, A19 Pro, M4, M5
mac1 & mac2 | Intel Macs

JPKribs:
> I think you should also look into how AVPlayer.eligibleForHDRPlayback behaves on tvOS. Is impacted by setting ATV output to SDR with match content on / off, HDR & DV?

https://developer.apple.com/documentation/avfoundation/avplayer/eligibleforhdrplayback

I will make sure to test this prior to any PR but this looks like this is purely for a display match. I don't know if that is the setting on the AppleTV but the simulator is appropriately returning false for my non-HDR mac. I'll do more digging/testing into this but IMO this looks promising:

> Indicates whether HDR content can be played to an appropriate display. This property is YES if an HDR display is available and the device is capable of playing HDR content from an appropriate AVAsset, NO otherwise.

I think the goal would be a `serverMappingOnIncompatible` setting or something to enable transcoding from the server if desirable over on device. Up for debate I'll need to figure out what this all looks like on the testing side. I'm still bogged down in Metal APIs so I haven't had a chance to do anything except see if we are returning the device name/family.

---

PR is still a work in progress I need to test more but:

https://github.com/jellyfin/Swiftfin/pull/1891

JPKribs:
Testing on iPhone 13 Pro, 8.4 always fails on Native Player. iOS 26.2, I do not believe 8.4 / HLG is supported on iOS.

PR #1868, opened 2025-12-20
Cleanup `[Culture|Country|ParentalRating]Picker`

LePips:
Cleans up these pickers, motivated by work in #1752.


JPKribs (approved):
Testing, I was getting the 500 error but that looks to be unrelated. Pickers themselves look good!

Issue #1068, opened 2024-05-24
Add option to set default audio/subtitle tracks


ranile:
**Describe the feature you'd like**
I would like to default to specific language audio and subtitle. There is espesically useful for anime where many want to select `jp` audio and `eng` (dialog) subtitles.

I don't see any setting for this in the app settings

**Additional context**
See how mpv does this:
https://github.com/mpv-player/mpv/blob/2fa66b850df1e0ecd653e9eea1e69fccc73afcee/etc/mpv.conf#L119-L123

JPKribs:
As of https://github.com/jellyfin/Swiftfin/pull/1581, we are now pulling default tracks using the same setting as Jellyfin-Web:

**Settings > Playback >  Audio Settings**
<img width="974" height="388" alt="Image" src="https://github.com/user-attachments/assets/563592ac-95a1-48da-932f-81edf39f257a" />

**Settings > Subtitles > Subtitles**
<img width="865" height="285" alt="Image" src="https://github.com/user-attachments/assets/be0857d5-e289-420d-b6d8-f1add5a35b88" />

So, setting this value there will impact the Default Subtitle and Audio track selection used by Swiftfin.

I have a PR in the works to update this value from Swiftfin but, for the time being, you can update this from Jellyfin-Web.

ggl-wolfy:
Currently Jellyfin only allows setting one default language, I would rather if I can set multiple, the first taking precedence if found.

I would rather if it allows setting language per series, but using a fallback system would be good enough.

Like OP, I could do jp,en and I would get jp audio for my anime. I can still get English in all my other tv series as long as it doesn’t include a jp audio track.

JPKribs:
We use the track selection logic provided by Jellyfin-Server. Any changes to how the track selection logic works would be a request in Jellyfin-Server's repo. When Swiftfin gets media for playback, the server indicates the track(s) we should default to based on the conditions set by the user in the `UserConfiguration` settings so none of the selection logic is maintained here.

hesshoss:
Installed jellyfin with a lot of enthusiasm but instantly deleted it after I saw that I can't change default language by specific show/anime, sadly plex have many year of success when basic thing like that are not available here.

PR #1882, opened 2025-12-31
Index/Track Fixes

### LINKED ISSUE #676: TRUEHD7.1 track has no sound

### LINKED ISSUE #771: Playing the incorrect audio track despite the option selected being different.
### Describe the bug

I have the Office on my JF server and I watch it across jmp web, jmp windows, android tv app (chromecast), and swiftfin. For only certain episodes (maybe 20-30%) the audio is selected as the default audio, but the audio playing is instead the extras or interview recording. I go and play it on anything other than swiftfin and I encounter zero issues. This has been pretty annoying because it means I can't watch on my phone anymore.

### Application version

1.0.1(78)

App Store

### Device information

iPhone 10 somethin

### OS version

iOS 16.1

### Jellyfin server version

10.8.10

### LINKED ISSUE #900: Wrong audio track is beeing played, both on appletv and iphone
### Describe the bug

When playing a movie on apple tv or iphone, directos commentary track are beeing played. 
Changing track manually does nothing, and the its not playing the one marked as default.

### Application version

1.0.1(78)

App Store

### Device information

Iphone 15 pro, apple tv 4k (newest)

### OS version

ios 17.1.. dont have the apple tv os infront of me right now.

### Jellyfin server version

10.8.11

### LINKED ISSUE #926: [VLCKit/Swiftfin Player] Audio stream index (?) incorrect during transcode
### Describe the bug

During transcoding/DirectStreaming using VLCKit/Swiftfin player, selected audio track index will be incorrect, causing no audio to play by default. To get audio to play, it appears that the next track in the index needs to be selected. The server is only sending the one track to begin with.

For example, say a file has two audio tracks: x, y.
The track selection will list: none, x, y.
Currently, 'none' will be selected by default graphically while selected track index sent to the server will indicate 'x'. Server will transcode 'x' and send that back to client. Client appears to have something other than 'x' selected, though, causing no audio to play. To get audio, 'y' needs to be selected graphically and this will actually play the transcoded 'x'.

A workaround for now is here: https://github.com/sy6sy2/Swiftfin/commit/029af180781bfe63bc7b172437257cfd5e2bccfb VideoPlayerViewModel.swift changes seem to allow audio to play by default, but I am not sure the implications for DirectPlay.

Might be related to #771 

### Application version

main eb17300

None

### Device information

Multiple simulators and devices

### OS version

iOS 16/17, tvOS 16/17

### Jellyfin server version

380c367

### LINKED ISSUE #1117: Embedded subtitles not appearing iOS
### Describe the bug

1. Go to Swiftfin iOS
2. Play media
3. Observe embedded subtitles not shown as option, only external. Moreover, external sub track is shown twice.

### Application version

1.1

App Store

### Device information

iPhone 14 Pro

### OS version

iOS 18 developer beta

### Jellyfin server version

10.9.6

### LINKED ISSUE #1531: Incorrect subtitles are played for media when embedded subs are disabled on server
### Describe the bug

When the server library setting "Disable different types of embedded subtitles" is set to "Allow None", and I try to play an external subtitle on an episode in Swiftfin, it will instead play an embedded subtitle (seems to be whichever embedded is set to default, at least for 1 item i checked). This embedded subtitle can also be a different language than what I am trying to play. The media is an MKV container. 

It works as expected when I allow all embedded subtitles and then select the external track, it plays the external subtitle as it should. Issue only seems to occur when all the embedded subs are hidden (or supposed to be).

This happens on both iOS testflight and tvOS (old, app store version)

I'm not sure if #1203 touches/fixes this, however I'll still open this issue to track this bug.

### Application version

1.3(9)

TestFlight

### Device information

iPhone 15 pro

### OS version

iOS 18.4.1

### Jellyfin server version

10.10.7

### LINKED ISSUE #1554: Swiftfin Player iOS - No audio E-AC3 Atmos
### Describe the bug

Working on the native player (but missing Atmos layer), working on Infuse (with Atmos layer). But no audio when attempting to use the swiftfin player.

### Application version

1.2

App Store

### Device information

iPhone 16 Pro

### OS version

iOS 18.5

### Jellyfin server version

10.10.7

### LINKED ISSUE #1728: Strange audio track selection behaviour with different device profiles
### Describe the bug
There appear to be some quirks with the device profiles I'm encountering specifically with TrueHD audio tracks. Some profiles hide all tracks other than the TrueHD tracks from being selected, and there are some issues where the server claims to be transcoding the audio, but Swiftfin still doesn't play any audio.

# Read update in comment for potential cause

Here are my observations with the different profiles:
**Automatic profile:**
- IF the playback starts with TrueHD selected, then TrueHD is the only availible track in the player, despite other compatible tracks being availible in the video file. If another track has been selected (which is possible in the direct profile, or probably if another is marked as default) then all tracks are shown and can be chosen. Selecting TrueHD and then quitting playback and starting it again will again lock it to only displaying TrueHD.
- The Jellyfin server seems to be "Direct streaming", a.k.a. transcoding the audio, however no audio is being played.

**Most compatible:**
- TrueHD again is the only availible to select audio, and the server claims to transcode yet no audio is played.

**Direct profile:**
- Now I can finally choose the other availible tracks, although TrueHD is still selected as default (this might be intentional though as this profile is not supposed to transcode at all?), however choosing one of the other compatible tracks by default would make sense IMO, even though the TrueHD is selected as default in the actual  video file, due to Swiftfin not actually being able to directly play it.

Overall it seems like the selected track is remembered across changing profiles, even though it isn't quite compatible, and also there are some issues where Swiftfin doesn't properly use the transcoded audio from Jellyfin server and instead doesn't play any audio at all. Tracks are also hidden from the playback selection in the Auto and Most compatible mode, specifically if a _TrueHD_ track is selected. There might also be some nessecary consideration for whether or not a TrueHD track should be selected by default across different profiles

### Application version

1.4(13)

TestFlight

### Device information

iPad Air M2

### OS version

iPadOS 26

### Jellyfin server version

10.10.7

### LINKED ISSUE #1854: SwiftFin Video Player doesn't trigger transcoding

### Description of the bug

I'ce started using Swiftfin on iOS more often and with Swiftfin video player it doesn't trigger transcoding at all. I've set bandwidth-limits from outside my network and on Web, Jellyfin.app and Swiftfin.app with Native video player it respects that and triggers transcoding but not with Swiftfin video player.

### Reproduction steps

1. Open Swiftfin
2. Go to settings
3. Set Video player type to Swiftfin
4. Start playing (high bandwidth) media
5. Nothing is playing, Jellyfin admin panel says the file is compatible and therefore must not be transcoded
6. Steps 1-2
7. Set Video player type to Native
8. Step 4
9. Playback starts after 1-3 seconds, Jellyfin admin panel says it's transcoding due to bandwidth limit

### Platform

iOS

### Application version

1.4

App Store

### Device information

iPhone 13 Pro

### OS version

iOS 26.1

### Jellyfin server version

10.11.5

### Screenshots or videos

_No response_

### Relevant log output

```shell

```

### LINKED ISSUE #1889: No audio selected when opening a movie

### Description of the bug

When starting up a movie or an episode, no audio is selected. I have to click the audio button, select the one (and only) english audio track to start hearing something.

### Reproduction steps

1. Open Swiftfin
2. Start playing something
3. Notice how no audio is being played
4. Open the audio selector and select the audio track

### Platform

iOS

### Application version

1.4

App Store

### Device information

iPad Pro M1

### OS version

ipadOS 26.2

### Jellyfin server version

10.11.5

### Screenshots or videos

_No response_

### Relevant log output

```shell

```

### LINKED ISSUE #1904: Incorrect Audio Track Selection on Playback

### Description of the bug

There is an issue with the automatic audio track selection during media playback. The player consistently skips the first available audio track.
• Scenario A: If only one audio track is available, the media plays with no audio.
• Scenario B: If two or more audio tracks are available, the player automatically starts with the second track in the list instead of the first (default).

### Reproduction steps

Steps to Reproduce
1. Open Swiftfin and navigate to any media item.
2. Start playback of a file that has at least one audio track.
3. Observe the audio output (or check the audio selector menu).
4. Note that for single-track files, audio is muted/missing. For multi-track files, the second track is active by default.

### Platform

iOS

### Application version

1.4.1

App Store

### Device information

Iphone 16 Pro

### OS version

iOS 26.2

### Jellyfin server version

10.11.5

### Screenshots or videos

_No response_

### Relevant log output

```shell

```

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/676
Resolves: https://github.com/jellyfin/Swiftfin/issues/900
Resolves: https://github.com/jellyfin/Swiftfin/issues/926
Resolves: https://github.com/jellyfin/Swiftfin/issues/1531
Resolves: https://github.com/jellyfin/Swiftfin/issues/1554
Resolves: https://github.com/jellyfin/Swiftfin/issues/1728
Resolves: https://github.com/jellyfin/Swiftfin/issues/1854
Resolves: https://github.com/jellyfin/Swiftfin/issues/1889
Resolves: https://github.com/jellyfin/Swiftfin/issues/1904

---

This PR is very dense but I think this solves a lot of our playback quirks! The crux of the issue is that we have two sources of truth for track selection. Jellyfin and the actual files read via VLCKit.

- Jellyfin handles this as External Audio > External Subtitles > Video > Audio > Subtitles.
- VLC handles this as File Track Order > External Audio > External Subtitles.

Previously, we were resolving this by just moving tracks around at the beginning. This ended up causing issues for us with the video player rewrite as we have a kind of 'sidecar' system for external files now. So, when remapping indexes, we didn't know the order of these files but just a guess. On top of this, we were not accounting for files types that Jellyfin was hiding in the `MediaSource` but exist in the file. Particularly, when Text or Image subtitles types are disabled. This then causes issues as our index we are using in Swiftfin doesn't match anything in Jellyfin anymore. To work around this, I build an indexMap with local and server indexes that are built first based on the Jellyfin provided `MediaSources` that are then revised as the 'sidecar' subtitles and other information load in. Round 1 is used to get the default / selected track for playback and Round 2 sures these up in case there are issues like Image or Text subtitles being hidden or things like external audio existing in the `MediaSources` but being unusable to us. Transcoding also causes issues as all internal subtitles because external and move down the row.

The other issue with this is that we can request stream based on default tracks. So, if the default track is AAC and the user switches to MLP / Atmos, a transcode will not be requested because we were given the raw file and we are changing tracks locally. This ave the impression that Swiftfin was not transcoding at all. I've updates this so, as tracks change, audio always requests a new stream to trigger a transcode if needed. Subtitles only triggers a new stream if we are requesting a PGS / Image Subtitle in an external format as those cannot be handled locally and need to be encoded into the video. Documentation and Device Profiles have been updated to reflect this!

Finally, we cannot play external PGS or image subtitles saves as `.sup` file. These need to be encoded into the file or embedded into the container. For this reason, if I have compatibility mode set to DirectPlay only, I hide these options from the drop down to prevent issues. 

I've fully tested every variation and have ZERO issues or mis-selections. The only issue is external audio tracks do not play but the have never played in Swiftfin. I left a TODO to figure this out but basically, these external audio tracks can DirectPlay so Swiftfin doesn't request a transcode but Jellyfin doesn't provide a delivery URL so we need to encode these into the stream to play this.

Testing files with subtitles can be found below. 25mb was the GitHub limit so I had to break this into 2 batches:

[Tests 1-3.zip](https://github.com/user-attachments/files/25435179/Tests.1-3.zip)
[Tests 4-7.zip](https://github.com/user-attachments/files/25435182/Tests.4-7.zip)

Happy to discuss any of what I've done here! It's all functional and clean (in my eyes) but I've been looking at this too long so any cleanup I can do to make this more universally understandable is welcome feedback!

---

## Videos:

Hopefully this shows what I mean but this is transcodes & directplays with various subtitles counts and audio counts all working without issue. All from a test file with the exception of the MLP / TrueHD which is LOTR since that's like my only file I left that in.

### MLP / TrueHD Transcoding (AC3 -> TrueHD)

1.4.1: 

If my starting track does not require transcoding, and then I switched to a track that does require transcoding, VLC just switches the track inside of the file, resulting in no playback.

If my starting track does require transcoding, or transcoding occurs at any point, I am only shown the audio track that is included with the new stream. Additionally, the subtitle indexes are incorrect, so you cannot select subtitles while transcoding.

https://github.com/user-attachments/assets/a9f390c5-fe71-4185-bb55-6abc2b6d2eca

All tracks are always shown to you and are always selectable. If you select a track that requires transcoding a transcode occurs, but if you select a track that does not require transcoding, it will just switch within the file itself. The position of playback is preserved.

https://github.com/user-attachments/assets/009e8465-da89-4cc1-a9f2-834ab05244b5

### External PGS File Transcoding (Text Embedded Subtitles Only)

https://github.com/user-attachments/assets/c4b34741-700f-406d-9773-88f57339351d

### Transcoding File Test (Text Embedded Subtitles Only)

https://github.com/user-attachments/assets/25c54853-6c0d-45b4-a628-150f15613688

### Transcoding File Test (Text Embedded Subtitles Only & DirectPlay Only)

https://github.com/user-attachments/assets/249b30e1-a6ed-4297-8618-affd78302218

### Transcoding File Test (Image Embedded Subtitles Only & DirectPlay Only)

https://github.com/user-attachments/assets/0580e43e-78b4-493b-b9d5-5f04bf6d203c


JPKribs:
@LePips I'm not sure how helpful I can be on this. I was troubleshooting this and found a fix for this but I don't think this is the correct implementation. I'm just hoping the items below help!

This PR resolves https://github.com/jellyfin/Swiftfin/issues/1854 but I don't think it does that correctly. I rebuild the stream from the manager since that made more sense but it means we're duplicating work between the manager and the proxy? I'd appreciate any insight into the correct way to do this!

https://github.com/jellyfin/Swiftfin/issues/1531 is essentially resolved in this PR but we would need to either pass in the `LibraryOptions` or we would need a route to go from `BaseItemDto` to `VirtualFolder` to `LibraryOptions` which I'm not sure how we'd want to do that cleanly. 

The subtitle issue is just completely beyond me. On transcode, we just don't seem to have subtitles at all. I am like 99% sure it's related to `VLCVideoPlayer.PlaybackChild` but I think you might have a better idea of how to use that appropriately.

JPKribs:
I'm scope creeping. I was also looking at https://github.com/jellyfin/Swiftfin/issues/1068 by:

```swift
extension [MediaStream] {

    func defaultTrack(for streamType: MediaStreamType, languagePreference: String? = nil) -> MediaStream? {
        let streams = self.filter { $0.type == streamType }
        
        guard streams.isNotEmpty else { return nil }
        
        if streamType == .audio || streamType == .subtitle {
            if let language = languagePreference?.lowercased() {
                let languageMatches = streams.filter { $0.language?.lowercased() == language }
                
                if let match = languageMatches.first(where: { $0.isForced ?? false }) {
                    return match
                }
                if let match = languageMatches.first(where: { $0.isDefault ?? false }) {
                    return match
                }
                if let match = languageMatches.first {
                    return match
                }
            }
        }
        
        if let forced = streams.first(where: { $0.isForced ?? false }) {
            return forced
        }
        if let defaultTrack = streams.first(where: { $0.isDefault ?? false }) {
            return defaultTrack
        }
        return streams.first
    }
```

Then we can get language preference via: `userSession.user.data.configuration?.audioLanguagePreference` and `userSession.user.data.configuration?.subtitleLanguagePreference`. In settings we'd probably want this to set that and pull that from the server so changes from web come to us and vice versa.

Then we'd just use `defaultTrack()` index instead as the starting point.

JPKribs:
@LePips sorry for starting this then putting this down for a while. This issue has honestly been the most frustrating Swiftfin issue I've worked on. Solving one interaction would break another. This solution is very robust and works in all circumstances except one*. The external audio part of this is something I would be happy to l look at at a later date but for now, this should solve 9 outstanding track selection issues.

_*If I start a non-MLP stream, then switch to an MLP stream but have DirectPlay forced, it will result in no audio because the transcode cannot trigger._

JPKribs:
@LePips if you have a chance I'd appreciate a second set of eyes on this before the next release. This resolves a lot of outstanding issues that come form VLC being frustrating but, to be completely frank, I was having trouble understanding the best way to do this in relation to the media player manager. All of the components are there and I have spent a ton of time and scenarios testing (per the main post) so I know the solution is good but I feel as though my solution is just kind of tacked onto the player as opposed to being integrated with it. The changes can be grouped into 2 categories:

1. Recreate the HLS stream on Track change. Some users have reported Swiftfin not transcoding. This is because if a file can be played natively, we get the full file with all tracks. If I switch to a track that cannot be natively played, we do not request a transcode because we are doing a local track change. This PR moves that to request the new stream which forces transcoding if the track doesn't work. Every audio track forces a new HLS stream but only external image subtitles force one as internal anything or external text subtitles will work as switches locally. External images need to be burned in and VLC does not work with them in my testing.

I moved this to the manager since HLS changes seemed more like a manager item than a proxy item? I was thinking maybe this is both? Like, manager creates the new stream then the proxy selects it? What I have now on the manager *works* but honestly, I'm feeling a bit lost. Sorry to toss this over the fence to you like this. I have it working in this PR but there is a layer of abstraction here I understand is required for VLC v AVPlayer v etc that I'm not positive if I'm breaking with this PR. Also, calling a new HLS will break downloaded playback so I'm not sure how to approach that.

I believe this HLS route will also work for bitrate changing so the structure is there for that but no real work.

---

2. This is a continuation/rework of https://github.com/jellyfin/Swiftfin/pull/1445. The issue at the end of the day is that VLC indexes based on the location of the track (internal v external) and also based on type/container while Jellyfin is just a straight global list. The reason this kept getting us into trouble was we were just overriding the old indexes for mapping. If the indexes every varied from what Jellyfin has/provides we could get off track. Some examples of where this can occur:

- Library is set up to exclude certain subtitles (Image, Text, All Internal) which meant we get the file with all subtitles but we're supposed to ignore some.
- The file is being transcoded so some of the subtitles that were IN the file are now external which changes the order. Also, some internal subtitles are not suitable as external subtitles and have to be dropped altogether. 

The way I got this to work is by creating an index map based on the original file in Jellyfin. This is based on the information that Jellyfin shows (Tracks we are supposed to have access to and where they are located). Once we receive the information about the file provided to us (Actual tracks), we have to re-map to help prevent issues with indexes on tracks we're not supposed to use. I think the index map is the correct move as it gives us an internal index for local track changes but also an external track for recreating the HLS stream (part of why these are bundled together). The reason this is two processes is this lets us request the stream first then this lets us remap based on what is actually provided to us. So like, we are requesting index 4 but it's 6 in the file because there are two internal image subtitles we are not allowed to see per library settings. As a result, without first mapping using information provided via API, we will start playback on the wrong track or with no subtitles/audio at all.

I'd be happy to help field questions or rework any of this! I think this is a big improvement/fix on the original #1445 solution

---

As a final note, this method for changing audio tracks would enable us to change tracks in AVPlayer as well as we are no longer reliant on the single audio track provided in HLS and instead would re-request HLS for track changes.

IMO, I think this track switching is the appropriate solution for what is being done here: https://github.com/jellyfin/Swiftfin/pull/1562

dlepaux:
+1 this would be so much appreciated

JPKribs:
I have a branch here I've been testing this change with AVPlayerProxy using the existing UI:

https://github.com/JPKribs/Swiftfin/tree/avplayer

I can confirm that these changes work for AVPayer to enable subtitles and track selection even in PiP!

https://github.com/user-attachments/assets/5be495aa-d461-4235-a01e-9dc8f3001693

Ignore the contents of the commentary lol I purely pulled this as a test of everything in a hurry while I wrapped up today. I'll pull a better video of this next time I'm home.

Shared/Objects/MediaPlayerManager/MediaPlayerManager.swift
```diff
@@ -297,6 +307,43 @@ final class MediaPlayerManager: ViewModel {
...
+        }
+
+        switch type {
+        case .audio:
+            guard playbackItem.audioStreams.contains(where: { $0.index == oldIndex }) else {
+                logger.warning("MediaPlayerManager.SetTrack call with an invalid audio track index")
+                return
+            }
+
+            try await updateMediaPlayerItem(
```
LePips: Just noting that I have started to take a look at this.

This will rebuild the stream every time the audio wants to change which might not be necessary, so I'm taking a look at some additional work around the index mappings.

JPKribs: Totally agree. Similar to subtitles, I wanted to use the local version of this where possible but I had 2 finds:

1. AVPlayer/Native player - Because of the HLS for AVPlayer only has a single of each track, we do need to request a new stream every time for that player.
2. VLCKit, we should be able to use tracks if available but for transcodes we have the same single track issue of 1) but also we have the the issue that some audio tracks require transcoding. So we would need to check audio tracks against the device profiles. Biggest offender is MLP/TrueHD tracks. The issue is, if you can play the default track Jellyfin sends you the raw file even if you cannot play any of the other tracks.

JPKribs: I was looking at this today. I've been trying to figure out how to phrase the issue correctly. I think this is 2 steps:

1. Does the track we're switching to exist locally.
2. Is the local track a valid playable track for our player.

The root cause is, if my video has an AAC primary track and a MLP secondary track, Jellyfin Direct Plays and provides us the file directly. If we switch to the MLP track locally, nothing will play as neither player can handle MLP. Therefore, we need to check the profile and rebuild the HLS if the new track would cause issues.

Right now, I am hard coding this and this is the incorrect approach. Not only does this not account for all circumstances but is also very fragile. I am testing a version that does a local device profile check for changing to a local track that I think should both prevent unnecessary rebuilds but also should be more robust moving forward.

Please find these changes here: https://github.com/jellyfin/Swiftfin/pull/1882/changes/aff67d8a5991905ec846f879f940419371ac0ab6

---

In that commit, I am checking against the device profiles for whether a local track will work. I am parsing this as a csv because, it kind of is a csv when we store and send this. Unfortunately, we handle this as a `String` instead of a strongly typed object. I think that would be work on the SDK but might be worth while in the long term?

Testing and working as expected in my test files. Additionally, this no longer rebuilds between AAC and AC3 in my test file (see below) but still appropriately requests a new stream for MLP/TrueHD to transcode. Then, unselecting it, it requests a new DirectPlay since we're back to valid tracks:

https://github.com/user-attachments/assets/ff94d7b6-052e-473f-a44d-7dfec3b77f45

---

This is a much much more robust way of doing local switching but it also adds local device profile parsing which IMO is undesirable but I think it's a necessary evil to prevent the "Swiftfin isn't transcoding" issues where it's just secondary tracks in the DirectPlay that's unplayable.

JPKribs: Also, I was doing this whole `guard proxy as VideoPlayerProxy` which I felt was wrong since audio will potentially need track switching but not subtitle switching. So I added a protocol for audio track switching and subtitle switching that `VideoPlayerProxy` conforms to. That way, when audio playback is used, this will not need updating except for lyrics which I left stubs for.

Lastly, I over-commented the `MediaPlayerItem` I've been staring at this for weeks so I've put reasoning in there since this isn't all intuitive. Feel free to remove this if unwanted!

JPKribs:
SDK changes the transcode URL is incorrect now. We will need to address:

https://github.com/jellyfin/Swiftfin/issues/2030

However, with this fix, this still works as expected. For #2030, I have that fix but I think this is better suited from the SDK instead. I just don't know where and I don't think I have the time for a bit to look into this.

LePips (approved):
I've had to fix a couple of bugs. The device profile wasn't being passed, instead an empty one was always used. Was unable to set a subtitle track from none.

Lyrics won't be a "track". They will be more like chapter segments, where we observe the current seconds against the lyric's intervals. I would expect that the lyrics UI component would have its own lyrics selection if there are multiple lyrics.

I'll approve to get this in, I expect some fixes after some time.

PR #1888, opened 2026-01-02
Default Track Settings & Unify `VideoPlayerSettingsView`

### LINKED ISSUE #1068: Add option to set default audio/subtitle tracks

### LINKED ISSUE #1726: Allow subtitle accessibility overrides on Apple TV
Allow Swiftfin to override subtitle fonts to use Apple TV's built in accessibility features.

**Additional context**
Apple TV allows for built in accessibility features to define subtitle font, color, and style. Would like to have Swiftfin allow this to be used in place of the fonts included in content. Plex and some streaming services have similar behavior to allow this, but I've not seen this on other video clients, including Swiftfin.

Specifically, these would be to allow the subtitles from:

Apple TV Settings -> Accessibility -> Subtitles and Captioning -> Style

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1068

Because of how https://github.com/jellyfin/Swiftfin/pull/1581 configured it, we are already pulling the default track selection based on the `UserDto.UserConfiguration` settings that we get from the server. This PR just adds a new setting to update these values from Swiftfin instead of relying on folks changing this from Jellyfin-Web:

**Settings > Playback >  Audio Settings**
Image

**Settings > Subtitles > Subtitles**
Image

This PR also fully refactors `VideoPlayerSettingsView` to be platform agnostic.

This PR also migrates `ServerUserAdminViewModel` to stateful and cleans up the views that use this. Most of these are a straight change but some of these there was some additional cleanup. The 3 big cleanups were reducing the comments that I originally put in there, combining `var`s when they were needlessly split up (by me), and finally, removing duplicate deletion logic.

Finally, this PR uses the same `ServerUserAdminViewModel` to track other server settings as well:

- `isHidePlayedInLatest` | Home View
- `isDisplayMissingEpisodes` | Item View for Seasons
- `enableNextEpisodeAutoPlay` | Player UI + Video Player Settings

These are items we had defaults / stored values for but most clients just pull this from the user configuration of the user session. This migrates these to use the user session > user configuration and updates the callers for these defaults / stored values to instead using the user session > user configuration.

### Videos

<details>

<summary>Swiftfin to Web Sync</summary>

https://github.com/user-attachments/assets/f9f1c0cb-e322-4e84-bc52-f897be537a50

</details>

<details>

<summary>tvOS</summary>

https://github.com/user-attachments/assets/38babb7e-bbeb-4d6a-90d3-ce2e1040435f

</details>

<details>

<summary>iOS</summary>

https://github.com/user-attachments/assets/290e7c1a-9cba-4700-b80a-06b2bfda0da4

</details>


JPKribs:
This PR is done but there is a lot of localizing. I also touch this same place via other outstanding PRs. When those are closed, I will finish up the localization and finalize this one. Videos are up-to-date!

---

Edit: Looks like I should do the same with the the `days in next up` and `Enable Rewatching in Next Up` to sync those with other clients / web. I will make that it's own PR. The changes here are *somewhere* more targeted but that change should be simple on top of this.

JPKribs:
Added an extension so tvOS Stepper could mirror iOS to stop the build issue. IMO, we can probably get away with that being the only stepper API since we only use it in tvOS in the contexts of the settings view. Primarily just wanted minimal impact so I didn't screw up anything you had offline.

LePips (approved):
I'm fine with this for now, but I anticipate change to the way user configuration settings are handled with some changes I've been thinking about for `UserSession` handling.

Issue #821, opened 2023-08-11
Long press to play faster


illusionaries:
**Describe the feature you'd like**
When user long presses the screen, set the playback speed to 2x until the user releases the finger.

**Additional context**
Some online streaming service providers (e.g. Bilibili in China) added this feature in recent years and I found it really helpful.

Long press is now used to toggle the "lock" feature, I suggest adding a menu to switch between the features as I personally do not use the "lock" feature a lot.

someone132s:
一年了好像还是没有支持啊

PR #1911, opened 2026-02-12
Gesture: Support long press for speed optimization

ShiSheng233:
This PR implements #821.

This implementation reuses [VideoPlayerContainerState.swift](https://github.com/ShiSheng233/Swiftfin/blob/c0cab9c37022c232309cb452bc8b77d53b03e401/Shared/Objects/VideoPlayerContainerState.swift) to define playback speed and adds a new long‑press option during playback, allowing both speed‑up and slow‑down.

When the player is long-pressed, a toast will pop up to inform the user. Meanwhile, lifting the finger from the screen will restore playback to its original speed.

This modification references Swiftgram's approach, which additionally allows adjusting playback speed finely by moving the finger left or right. I would like to gather feedback from users and other contributors to decide whether this feature should be included.

However, when I was debugging in the local environment, XCode throws error _"Function declares an opaque return type, but has no return statements in its body from which to infer an underlying type"_ in these two files. Since this issue is not related to my functional changes, I manually set it to `EmptyView()` locally to compile, but did not commit this modification.

- [Shared/Objects/MediaPlayerManager/Supplements/PlaybackRateMediaPlayerSupplement.swift#L84](https://github.com/ShiSheng233/Swiftfin/blob/c0cab9c37022c232309cb452bc8b77d53b03e401/Shared/Objects/MediaPlayerManager/Supplements/PlaybackRateMediaPlayerSupplement.swift#L84)
- [Shared/Objects/MediaPlayerManager/Supplements PlaybackInformationSupplement.swift#L57](https://github.com/ShiSheng233/Swiftfin/blob/c0cab9c37022c232309cb452bc8b77d53b03e401/Shared/Objects/MediaPlayerManager/Supplements/PlaybackInformationSupplement.swift#L57)

Additionally, I noticed that CI seems unable to build due to an issue with fastlane. However, this is beyond my ability, hope other developers can resolve this problem.

I'm new to iOS development, I welcome any feedback you may have.

---

<details>
<summary>Screenshots:</summary>
Snipaste_2026-02-12_15-29-05
Snipaste_2026-02-12_15-29-17

Snipaste_2026-02-11_18-32-06
Snipaste_2026-02-11_18-32-25
</details>


JPKribs:
Thank you for your contribution! An initial look at the code, I think this looks solid! That being said, I'll want to test it on real hardware which might take a sec for me to have the time.

In the meantime, would you be able to resolve the linting and the tvOS build issues? I think for tvOS, we can skip this feature for now and revisit it with the player.

ShiSheng233:
Hi @JPKribs , thanks for your feedback.

Regarding the tvOS build issue, I lack experience with Apple TV platform and am unsure if using `EmptyView()` is appropriate.

As for code linting, running the same command locally modified several files. Is this change acceptable?

JPKribs:
Nevermind, this builds fine for tvOS. I think there's some action updates required from our end. For now this is fine and we can let you know if there are any changes needed as we have a chance to test this!

ShiSheng233:
Hi, I just merged the changes from this repo, maybe it can build with CI now.

JPKribs:
Perfect sorry about that this looks all correct for the build. It's valentines weekend so I likely won't be able to test this on hardware until Monday but I'm not traveling for work next week. So I should be able to get this tested!

Swiftfin/Views/VideoPlayerContainerView/Gestures/VideoPlayerContainerView+TapGesture.swift
```diff
@@ -222,6 +222,48 @@ extension VideoPlayer.UIVideoPlayerContainerViewController {
                     systemName: VideoPlayerActionButton.gestureLock.systemImage
                 )
             }
+        case .speed:
+            guard checkGestureLock() else { return }
+            guard containerState.manager?.item.isLiveStream == false else { return }
+
+            switch state {
+            case .began:
+                // Save original playback rate
```
JPKribs: We can remove all of these comments

Swiftfin/Views/VideoPlayerContainerView/Gestures/VideoPlayerContainerView+TapGesture.swift
```diff
@@ -222,6 +222,48 @@ extension VideoPlayer.UIVideoPlayerContainerViewController {
...
+                containerState.originalPlaybackRate = containerState.manager?.rate
+
+                // Get configured speed multiplier
```
JPKribs: We can remove all of these comments

JPKribs (changes requested):
Sorry about the delay on testing this for you! This works well and I think this is good minus a bit of cleanup. Since there are some localization changes in Main, we need to resolve that prior to merging. I've actually taken care of this for you since it's a little convoluted since GitHub sees it as a binary it makes version control a giant pain.

Essentially, the process is:

1. Merge with main (removes all strings)
2. Copy all main strings into the localization files
3. Copy all of your strings int the localization files (duplicates do not matter)
4. Build and let the build scripts handle the duplicates and alphabetize everything
5. Push back to main with the updated localization

First, Let's change our usages of stuff like this:

`let rateText = originalRate == 1.0 ? "1x" : String(format: "%.2gx", originalRate)`

and instead let's make a `FormatStyle` which you can find here: `/Shared/Extensions/FormatStyle.swift`

Essentially, let's do what we did for `IntBitRateFormatStyle`. Something like:

```swift
struct DoubleRateFormatStyle: FormatStyle {
    func format(_ value: Int) -> String {
        let formattedValue = String(format: "%.2f", value)
        return "\(formattedValue)x"
    }
}
```

Then we can use that for the instances of this. It's basically what you already have but it centralizes this format so if we ever want to update this we have a spot to do this. Primarily I want this because, "2x" makes sense in English but I could see this making more sense in other locals as "2*something else*" or even for right-to-left languages being "x2". Centralizing this format just makes this easier to localize down the road.

Second, I appreciated these comments for review but these are unnecessary as your code is self-documenting enough in: `Swiftfin/Views/VideoPlayerContainerView/Gestures/VideoPlayerContainerView+TapGesture.swift`. I left 2 comments in there just for reference and something to check off for your side!

So let's just remove these comments as the code is self explanatory.

Finally, "Speed" works as the verbiage for this but I think we can do better. I don't personally have a good word for it but "Speed" is a bit too broad. Talking to a third party, they see "Gesture Lock" and that makes sense but they can't quite tell what "Speed" does from the setting. Their recommendation was to make the string "Playback speed" which I think captures it better but I'll let you determine what is best.

Other than these, I think this looks awesome! Again, really appreciate the contribution!

JPKribs:
As an extended note. I think it would be cool to change `PlaybackSpeed` to work more like `MediaJumpInterval` where there is also a custom implementation of this that can be set for this configuration. So:

```swift
enum MediaJumpInterval: CaseIterable, Displayable, Hashable, RawRepresentable, Storable, SystemImageable {

    typealias RawValue = Duration

    case five
    case ten
    case fifteen
    case thirty
    case custom(interval: Duration)

    init(rawValue: Duration) {
        switch rawValue {
        case .seconds(5):
            self = .five
        case .seconds(10):
            self = .ten
        case .seconds(15):
            self = .fifteen
        case .seconds(30):
            self = .thirty
        default:
            self = .custom(interval: rawValue)
        }
    }

    var rawValue: Duration {
        switch self {
        case .five:
            .seconds(5)
        case .ten:
            .seconds(10)
        case .fifteen:
            .seconds(15)
        case .thirty:
            .seconds(30)
        case let .custom(interval):
            interval
        }
    }

    var displayTitle: String {
        rawValue.formatted(.minuteSecondsNarrow)
    }

    var systemImage: String {
        switch self {
        case .thirty:
            "goforward.30"
        case .fifteen:
            "goforward.15"
        case .ten:
            "goforward.10"
        case .five:
            "goforward.5"
        case .custom:
            "goforward"
        }
    }

    var secondarySystemImage: String {
        switch self {
        case .thirty:
            "gobackward.30"
        case .fifteen:
            "gobackward.15"
        case .ten:
            "gobackward.10"
        case .five:
            "gobackward.5"
        case .custom:
            "gobackward"
        }
    }

    static var allCases: [MediaJumpInterval] {
        [.five, .ten, .fifteen, .thirty]
    }
}
```

This could require a more extensive rework here:

```swift
enum PlaybackSpeed: Double, CaseIterable, Defaults.Serializable, Displayable {

    case quarter = 0.25
    case half = 0.5
    case threeQuarter = 0.75
    case one = 1.0
    case oneQuarter = 1.25
    case oneHalf = 1.5
    case oneThreeQuarter = 1.75
    case two = 2.0
    case custom

    var displayTitle: String {
        switch self {
        case .quarter:
            "0.25x"
        case .half:
            "0.5x"
        case .threeQuarter:
            "0.75x"
        case .one:
            "1x"
        case .oneQuarter:
            "1.25x"
        case .oneHalf:
            "1.5x"
        case .oneThreeQuarter:
            "1.75x"
        case .two:
            "2x"
        case .custom(Interval):
            Interval.format(...etc)
        }
    }
}
```

But this would also change this to no longer conform to `Double` so this would likely be its own PR since it would touch a lot more places... More just putting this out there as a future state for this but I wanted to put this out there while I was thinking about this.

ShiSheng233:
Thank you for your suggestion. 
I have added a `DoublePlaybackRateFormatStyle` based on your advice to handle the formatting, which references the existing `PlaybackRateStyle`. (I'm not sure how to handle the i18n for the playback speed feature in other languages, so I just referred to the existing implementation.)
I've also removed the code comments and revised the i18n expressions. 
Currently, it looks like this in the settings: 
Snipaste_2026-02-16_13-44-50

JPKribs:
Couple of changes. 2 main changes:

1. When reviewing this I realized that the `PlaybackSpeed` enum was a `Double` but every instance of the speed of playback in VLC or the `VideoPlayerManager` was a `Float`. This means we can change this enum to be a `Float` which means we can re-use the existing logic for the `Float` version of this style instead of my suggestion for a `Double` version of this. While I was changing that, I added the custom option to allow for greater than 2.0 as needed.
2. Some string cleanup. We already had a "Playback speed" string so I just reused that instead.

Please test this out from your end and please make sure that everything is still working as expected. Since I made some corrections to this, I will want to run this by @LePips just to double check my work. Unless he sees something or you see something while testing again, I think this is good to go!

LePips (dismissed):
Simple enough, thank you for your contribution! I just performed some cleanup. I would also recommend using branches for any future contributions, as you'll now have to hard reset your `main`.

LePips (dismissed):

ShiSheng233:
Thank you for the fix. The current modifications are running as expected on my end. 
I will continue learning Swift and contribute to large projects.

JPKribs (approved):

PR #1912, opened 2026-02-13
Update Fastlane Runner and SwiftFormat

LePips:
- Use Xcode 26.2
- Update SwiftFormat version

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,3 @@
 source "https://rubygems.org"
 
-gem "fastlane"
+gem "fastlane", '2.232.0'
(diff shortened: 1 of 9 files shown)


chickdan:
@LePips it looks like the FastlaneSwiftRunner needs update like in https://github.com/jellyfin/Swiftfin/pull/1899. IIRC to update all I had to do was run `fastlane init swift` to update the Swift files. Surely there is an automated way to handle this rather than waiting for builds to start failing, but I've not had time to take a look into it.

An aside, should we have SwiftFormat ignore the Fastlane folder?

LePips:
I had updated the fastlane version to latest `2.232.1` and rebuilt the runner. However, GH's images have `2.232.0` atow and `2.232.1` seemingly changed something between those versions _and_ the runner will always prompt to rebuild on any version mismatch. I had tried a lot long ago to see if there was a CI/auto-`y` flag but couldn't get it to work.

The `fastlane` directory is ignored, any Swift file updates within it come from rebuilding the runner.

PR #1914, opened 2026-02-16
Fix ListRowMenu styling with modern tvOS focus effect

tobiasbischoff:
## Summary

- Removes manual `@FocusState` and custom background/scale effects from `ListRowMenu`
- Uses `.hoverEffect(.lift)` for native tvOS focus handling
- Simplifies styling to use `.foregroundStyle(.primary)` and `.secondary`
- Makes `ListRowMenu` consistent with `PrimaryButtonStyle`

This resolves issue #1818 which requested updating the legacy focus logic to use modern tvOS modifiers.

## Changes

- `Swiftfin tvOS/Components/ListRowMenu.swift`: Replaced manual focus handling with `.hoverEffect(.lift)`

## Testing

Built successfully with Swift syntax check. Should be tested on actual tvOS device/simulator to verify focus behavior matches other buttons in the app.

### DIFF
--- a/Swiftfin tvOS/Components/ListRowMenu.swift
+++ b/Swiftfin tvOS/Components/ListRowMenu.swift
@@ -10,51 +10,35 @@ import SwiftUI
 
 struct ListRowMenu<Content: View, Subtitle: View>: View {
 
-    // MARK: - Focus State
-
-    @FocusState
-    private var isFocused: Bool
-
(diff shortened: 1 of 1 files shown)


JPKribs:
Hi, apologies but this item is already started and tagged to the Issue here: https://github.com/jellyfin/Swiftfin/pull/1858

PR #1917, opened 2026-02-24
Unify `DeviceProfile` Views

JPKribs:
## Summary

Merges the views for iOS and tvOS for Custom Device Profiles & Edit Custom Device Profile. 

- Adds a check to tvOS to prompt for `Save` if exiting the Edit Custom Device Profile without saving
- Creates a `.labeledContentStyle(.deviceProfile)` instead of manually formatting this with `VStacks`
- Cleans up some of the naming between `DeviceProfile` & `CustomDeviceProfile` to make it all more consistent and less wordy
- When adding custom profiles on top of the default profiles, default profiles are now shown below them
- To go from defaults profiles into the format we are using, I added a generic string? utility to go from a list,or,strings to [Type] where `Raw Representable`

### iOS

https://github.com/user-attachments/assets/e5bf3f1d-59a0-4757-ac2e-b6d0ed868b08

### tvOS

https://github.com/user-attachments/assets/f0289efb-bb80-428b-a325-1cf67e9c9ca6


JPKribs:
There are two parts of this that might not be wanted:

1. Showing the default profiles. I added this because I think it's useful information to have when adding profiles on top of the existing. This way you can see what's already in the device profiles so you don't accidentally recreate an existing profile. It's also just good information where someone might say "I want X codec support" and then they can see it already exists in the profiles. IMO, it was simple enough to add and I believe the context is valuable but I can also understand the argument that it's extra clutter. 

2. For default profiles, I am using `ListRow`. This is an older implementation where we're are using modifiers. Since I am not using the leading content, this is: `ListRow {} content {...}`. I would normally change this but I can see this is changed here: https://github.com/LePips/SwiftFin/blob/f465963433b4bf98ca3765d56dca3681a76cd17e/Shared/Components/ListRow.swift. I can steal this change for this PR if that's the way we want to take this. Or, if we don't want 1) with the default profiles, this part doesn't matter and we can ignore this until it's part of #1752. The reason i am using the ListRow is that I wanted this to be focusable and consistent but these do not need to be buttons nor should they appear the same as buttons.

Let me know!

Otherwise, this should be a really simple copy from iOS and tvOS with some cleanup like moving to use the Chevron button instead of recreating this in our view and using labeled content styles instead of creating it for every usage manually.

PR #1902, opened 2026-01-16
[tvOS] Media Player

JPKribs:
### Summary

**Parent Issue**

Resolves: https://github.com/jellyfin/Swiftfin/issues/818

**Children Issues**

Resolves: #503
Resolves: #577
Resolves: #658
Resolves: #679
Resolves: #697
Resolves: #703
Resolves: #729
Resolves: #770
Resolves: #787
Resolves: #892
Resolves: #900
Resolves: #906
Resolves: #914
Resolves: #926
Resolves: #954
Resolves: #1005
Resolves: #1006
Resolves: #1102
Resolves: #1166
Resolves: #1378
Resolves: #1755
Resolves: #1962

---

Creates a tvOS version of the iOS Player UI. Attempts to mirror iOS as closely as possible but there are some limitations with UIKit Controllers and focus. iOS we are able to switch between the tabs using gestures and buttons but tvOS we need to account for buttons, swipes, and focus so my implementation of this is a bit different.

### Video

https://github.com/user-attachments/assets/6c37023f-2a42-4cc9-b7be-8d9747fd2d30

**Updated**: _2026-06-06_

Issue #818, opened 2023-08-08
Testflight - tvOS - Finish Swiftfin Video Player


LePips:
A big part of #593 was moving the video player layers from UIKit to SwiftUI. The tvOS overlay isn't complete yet and this is just to track that work.

JulianVennen:
Does this issue being closed mean that there's a testflight release for tvos now? The README still points to this issue

JPKribs:
There is not. Please follow this discussion for updates:

https://github.com/jellyfin/Swiftfin/discussions/1294

JulianVennen:
Thanks ❤️

JPKribs:
I am re-opening this to use as a parent to track tvOS Media Player issues that should be resolved by this. Currently using 1.4 as the milestone but that is subject to change and should mirror the final tvOS release.

---

I believe all attached items should be related. The only ones I am not positive of are the LiveTV ones.

PR #1918, opened 2026-02-28
Enable `RedundantViewBuilder` & Lint

JPKribs:
### Summary

Enables `redundantViewBuilder` in `.swiftformat` to remove where these are not needed. Then, lint the project to clean these up. Some of this I was doing [here](https://github.com/jellyfin/Swiftfin/pull/1902) but this separates this into its own PR.

**Let me know if this flag is intentionally disabled and I can close this branch out without merge!**

### Special Considerations

There are only 2 files where removing this caused issues. Neither of them are currently in use and are overridden by any [tvOS Player implementation](https://github.com/jellyfin/Swiftfin/pull/1902) we land on. 

1. `Shared/Objects/MediaPlayerManager/Supplements/PlaybackInformationSupplement.swift`
2. `Shared/Objects/MediaPlayerManager/Supplements/PlaybackRateMediaPlayerSupplement.swift`

For these, just replace Empty `{}` with what iOS is using for this to be overridden when https://github.com/jellyfin/Swiftfin/issues/818 is resolved or when we implement the `PlaybackRateMediaPlayerSupplement` which is WIP for iOS as well.

### DIFF
--- a/.swiftformat
+++ b/.swiftformat
@@ -42,7 +42,6 @@
     redundantFileprivate,   \
     redundantSelf,          \
     redundantType,          \
-    redundantViewBuilder,   \
     strongOutlets,          \
     trailingCommas,         \
     yodaConditions
--- a/Shared/Components/ListRowCheckbox.swift
+++ b/Shared/Components/ListRowCheckbox.swift
@@ -31,7 +31,6 @@ struct ListRowCheckbox: View {
 
     // MARK: - Body
 
-    @ViewBuilder
     var body: some View {
         if isEditing, isSelected {
             Image(systemName: "checkmark.circle.fill")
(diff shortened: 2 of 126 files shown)


LePips:
I had this intentionally disabled as I disagree the usage of `ViewBuilder`, even for a single `View`, is superfluous and find it helpful. I even wish that they weren't implicit for protocol members.

JPKribs:
> I had this intentionally disabled as I disagree the usage of `ViewBuilder`, even for a single `View`, is superfluous and find it helpful. I even wish that they weren't implicit for protocol members.

Makes sense! I think there was a brief period of time where it was enabled by default where I removed it from a branch I was working on so I wasn't sure which way we wanted to go with that, but your reasoning makes sense

PR #1919, opened 2026-02-28
Unify `OrderedSectionSelectorView`

JPKribs:
### Summary

From the device profile changes, there are two instances of `OrderedSectionSelectorView` that are almost identical. One for iOS and one for tvOS. This merges these into a single shared component.

I have also updated this to not have the modifiers anymore as I believe we are moving away from this design. To do this, I am using the standard `NavigationTitle` instead of the custom one I made in #1169. 

Additionally, I am changing the custom label logic which is only used in a single location (player buttons) to instead just always use the item if the element is `SystemImageable`. That one might be more controversial so let me know if we would prefer the label back.

I've removed the `ActionButtonSelectorView` as a result as this was a very thin wrapper where it was just the `OrderedSectionSelectorView` with a label modifier which is now in `SystemImageable`. This lets us use these routes for tvOS as well since this should be included in the same player. I plan on mapping this change to actual settings here: https://github.com/jellyfin/Swiftfin/pull/1888

Final add-on, iOS we were using text for this. This will be a `PillButton` style now to better match the rest of iOS toolbar buttons.

### Impacted Views

1. Device profiles for all 3 types
2. Player button customization stuff
3. Filters (iOS only for now)
4. `ItemView` attributes like HD or ratings and such

### Fixes

When I first did https://github.com/jellyfin/Swiftfin/pull/1169 I was less familiar with SwiftUI. One of the quirks of that implementation was that `OrderedSectionSelectorView`'s `move` function would also count as a select so it would get stuck whenever an item was moved where it was simultaneously in edit mode but unable to edit. To resolve this, I originally set this to just cancel the edit mode with each move. This prevented breakage but meant you could only move a single item at a time.

This version resolves that by just not registering this as a select during edit on tvOS so this works just like iOS now. It's a very minor change but should make this better to use.

### Conflicts?

This is changed very slightly in #1752 but it's just the binding box change so I don't think this will cause issues for that PR but let me know!

https://github.com/LePips/SwiftFin/blob/poster-library-home/Swiftfin%20tvOS/Components/OrderedSectionSelectorView.swift

### Videos

**tvOS**

As a note, none of the elements in iOS use the `SystemImageable` as those are only the `ActionButtons` for the player right.

https://github.com/user-attachments/assets/bec3b74f-3c5f-44bf-af4c-36d04bf35aca

https://github.com/user-attachments/assets/266b8144-cda9-4ebd-bbe6-7a2c35842a32

---

**iOS**

https://github.com/user-attachments/assets/f938e744-7fde-4eb7-b8e2-cf1fdd9978ad

https://github.com/user-attachments/assets/006fefc2-ac1d-479e-9e26-20af37c883b2

https://github.com/user-attachments/assets/1f4dfe25-ae7b-4b21-9d3c-70bd9e424c8e


JPKribs:
@LePips This is a fairly small one that shouldn't impact #1752. 4 line conflicts in total around `BindingBox` -> `PublishBox`.

Post should cover all of the changes. I wasn't sure if this should be a rework of what we have now or a shift over to `_Alert` for tvOS. I chose just to combine what we have now but let me know if the `_Alert` is what we should do for tvOS.

PR #1920, opened 2026-03-02
`LocalUser` View Unification

JPKribs:
### Summary

Unifies the `UserProfile` and `LocalUserSecurity` view for iOS and tvOS. I've renamed these to `LocalUser` to mirror `ServerUser` so it's local = Device and server = Server. `QuickConnect` and `ResetPassword` I left as iOS exclusive. I tried it out to allow those to be usable from tvOS and it's definitely a usable experience but I can see this being more of an annoyance *(my son reset my admin password from our TV!)* rather than a widely used feature. Same for QuickConnect, that authorizing the TV from my phone makes sense but authorizing my phone from my TV seems less valuable.

I've also left out the ability to set the profile image from tvOS as there is no real photo access from tvOS and it falls under the same boat as the password reset IMO.

So, migrate `UserProfile` and `LocalUserSecurity` to shared and move `QuickConnect` and `ResetPassword` into the root Swiftfin (iOS) > Views since they are used in both Settings and ServerUser so it made more sense for those to be generic views that Local User Settings specific. 

For `QuickConnect` and `ResetPassword` I cleaned up my excessive MARKing from last time (sorry!) but those should otherwise be untouched. Other than that, I updated the routes to reflect `localUser` instead of `userProfile`

Funcitonally, the biggest change from this is `.navigationTitle(L10n.user)` used to just be tvOS but now this is on both. We can wrap that in build flags if that looks off in iOS.

### Videos

**iOS**

https://github.com/user-attachments/assets/53c63500-b9a0-41de-8816-06c3de5fb868

**tvOS**

https://github.com/user-attachments/assets/f60be254-40f4-45b5-88eb-843f43e4ed2e

### DIFF
--- a/Shared/ViewModels/LocalUserSecurityViewModel.swift
+++ b/Shared/ViewModels/LocalUserSecurityViewModel.swift
@@ -10,7 +10,7 @@ import Combine
 import Foundation
 import KeychainSwift
 
-final class UserLocalSecurityViewModel: ViewModel, Eventful {
+final class LocalUserSecurityViewModel: ViewModel, Eventful {
 
     enum Event: Hashable {
         case error(ErrorMessage)
--- a/Shared/Views/SettingsView/LocalUserSettings/LocalUserSecurityView.swift
+++ b/Shared/Views/SettingsView/LocalUserSettings/LocalUserSecurityView.swift
@@ -26,7 +26,7 @@ struct LocalUserSecurityView: View {
     private var router
 
     @StateObject
-    private var viewModel = UserLocalSecurityViewModel()
+    private var viewModel = LocalUserSecurityViewModel()
 
     @State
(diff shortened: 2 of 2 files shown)


JPKribs:
As another note, the user view for tvOS is a bit bland. I was thinking of sending the user for tvOS directly to security but that changes up our navigation per-platform a bit. This also removes the profile section for other profile items down the road. Specifically, I think as we adopt the web socket, messages from server (see below) could cache in a `notifications` or `messages` section in the profile which would result in us adding this to tvOS anyway.

Screenshot 2026-03-02 at 12 36 31

So, I left this as is even though tvOS the `LocalUserSettingsView` is just:

- `Security`
- `Reset`

While iOS is a lot more "full" with:

- `Profile Image Edits`
- `QuickConnect`
- `Password Reset`
- `Security`
- `Reset`

PR #1923, opened 2026-03-04
Persist user selected playback speed

### LINKED ISSUE #711: playback speed control saved

jbenge1:
I added `playbackRate` to swiftfin defaults and write to it whenever there's a change to playback rate. Also added to vlc configurator when that's created to actually persist the playback rate.

Thif fixes #711


Shared/Objects/MediaPlayerManager/MediaPlayerManager.swift
```diff
@@ -123,7 +123,12 @@ final class MediaPlayerManager: ViewModel {
     @Published
     private(set) var playbackRequestStatus: PlaybackRequestStatus = .playing
     @Published
-    var rate: Float = 1.0
+    var rate: Float = Defaults[.VideoPlayer.Playback.playbackRate] {
+        didSet {
+            Defaults[.VideoPlayer.Playback.playbackRate] = rate
+        }
+    }
```
JPKribs, line 130: Since we're setting this at the `Default` I think we can just use:

```swift
@Default(.videoPlayer.Playback.playbackRate)
private var rate
```

But then we'd have to update all of the locations that call this from the published rate to do the same...

I'll defer to @LePips for this. Your solution works so I'll lean on them for whether or not it makes more sense to call this from the Default when we need it or keep the published rate.

LePips: No, as the `Default` property wrapper won't always work appropriately in non-View contexts (it's a `DynamicProperty`).

PR #1921, opened 2026-03-03
Unify `SelectUserView`

JPKribs:
### Summary

Combines the `SelectUserView` into a unified set of components for iOS and tvOS.

Some other goals of this process:

1. Use the environment `isEditing` rather than sharing a `Binding<Bool>`
2. tvOS ListView?

### Structure

Screenshot 2026-03-07 at 00 01 54

### Videos

**tvOS**

https://github.com/user-attachments/assets/22a6610a-219e-40eb-a59f-7227ded12064

**iPadOS**

https://github.com/user-attachments/assets/3fdcae45-a70e-46f9-a757-2567eab846d0

https://github.com/user-attachments/assets/17871215-51db-4ae5-b624-dfe7c3a74bd0

**iOS**

https://github.com/user-attachments/assets/0ae2f0a0-97fa-4b3d-b8c8-0f9cc9fabd3c

PR #1926, opened 2026-03-05
Cleanup `UserButtons`

JPKribs:
### Summary

This is a 2 parter to help with https://github.com/jellyfin/Swiftfin/pull/1921. 

1. The `PublicUserRowButton` is just a `ChevronButton`. We can (and #1921 does) use the same `ChevronButton` for the `SelectUserView` but this it's cleaner to just update it here.
2. The `UserGridButton` is almost identical for the `PublicUserGridButton` and the `SelectUserView.UserGridButton`. This aims to create a single, shared button that we can re-use for both views. For now, this is only used in the `UserSignInView` but my goal in #1921 is to use this there as well for the grid but I want to split this off to make this cleaner.

The end result is functionally no different for now but this should make this easier to maintain and, for the grid buttons, offer better re-usabe for when I finish up the `SelectUserView`

### Screenshots

**List:**
Screenshot 2026-03-04 at 22 29 24

**Grid:**

Unfocused:
Screenshot 2026-03-04 at 22 31 21

Focused:
Screenshot 2026-03-04 at 22 31 30

### DIFF
--- a/Shared/Components/UserButton.swift
+++ b/Shared/Components/UserButton.swift
@@ -10,7 +10,7 @@ import Defaults
 import JellyfinAPI
 import SwiftUI
 
-struct UserButton<Icon: View>: View {
+struct UserButton: View {
 
     @Default(.accentColor)
     private var accentColor
@@ -22,7 +22,7 @@ struct UserButton<Icon: View>: View {
     @Environment(\.isSelected)
     private var isSelected
 
-    private let icon: Icon
+    private let image: UserProfileImage
     private let title: String
     private let subtitle: String?
     private let action: () -> Void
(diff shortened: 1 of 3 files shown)


JPKribs:
So, for the time being, this is slightly more code than before but when also using these to replace their variants in the `SelectUserView` this nets out to less code overall. Some of the Init logic isn't used here as this will be used in `SelectUserView`. You can see the full context of it in https://github.com/jellyfin/Swiftfin/pull/1921/changes but it's still a giant PR so my goal was to split out this part as a standalone. I could link this to the existing `SelectUserView`s but my aim is to replace those so I didn't want to update those instance there only to change it in the next PR.

This also replaces the `SelectUserView` add user button as well.

LePips (approved):

PR #1927, opened 2026-03-06
Add WebSocket client for server DisplayMessage support (partial #408)

bones7456:
## Summary

Partially implements #408 (WebSocket Remote Control) — specifically the **"displaying a message onscreen"** feature. Playback control commands are not yet implemented.

When a Jellyfin server administrator sends a message from the dashboard to a Swiftfin session, it now appears as a non-intrusive pill overlay in the bottom-left corner of the screen. Video playback is never interrupted.

### What's included

**3 new files:**
- `Shared/Services/ServerWebSocketClient.swift` — Manages a persistent WebSocket connection to the server. Handles `GeneralCommand/DisplayMessage`, `ForceKeepAlive`, and `KeepAlive` messages. Reconnects automatically with a 5-second backoff on error.
- `Shared/Objects/ServerMessageProxy.swift` — Plain `ObservableObject` holding the current message state, registered as a Factory singleton. No UIKit presentation involved.
- `Swiftfin/Components/ServerMessageOverlayView.swift` — SwiftUI pill overlay that reads from `ServerMessageProxy`. Uses `.allowsHitTesting(false)` and a SwiftUI transition — never calls `UIViewController.present()`.

**3 modified files:**
- `RootCoordinator`: starts/stops the WebSocket on sign-in, sign-out, and server URL change.
- `RootView` (iOS): adds the overlay for non-video screens.
- `VideoPlayerContainerView` (iOS): adds the overlay inside the player's own view hierarchy so messages appear during playback without interrupting it.

### Dashboard "send message" button

The session's `SupportsMediaControl` flag (which controls whether the send-message button appears in the dashboard) requires both a reported capability **and** an active WebSocket controller. Capabilities (`DisplayMessage` + `SupportsMediaControl: true`) are reported after the server's first `ForceKeepAlive` — not immediately on connect — so the controller is already attached when `CapabilitiesChanged` fires, and the button appears reliably.

### What's not included

Playback control commands (play, pause, seek, etc.) are out of scope for this PR and can be addressed in a follow-up.

## Test plan

- [ ] Sign in to a Jellyfin server
- [ ] Open the Jellyfin web dashboard → Sessions → find the Swiftfin session → verify the send-message (chat bubble) button is visible
- [ ] Send a message — verify the pill overlay appears in the bottom-left corner and auto-dismisses
- [ ] Start video playback, send a message — verify the overlay appears inside the player without pausing or interrupting playback
- [ ] Sign out and sign back in — verify the WebSocket reconnects and messages still work


JPKribs:
Thank you for your contribution! However, we would want the web socket implemented from the SDK. There is already an outstanding PR for this as well:

https://github.com/jellyfin/jellyfin-sdk-swift/pull/46

Additionally, please see [Jellyfin's AI policy](https://jellyfin.org/docs/general/contributing/llm-policies/) as this PR is written by AI and would not be accepted on those ground as well. For future PRs, even if AI is used in the development of a solution, please ensure that the PR and discussion is written in your own words.

PR #1934, opened 2026-03-10
Unify `ItemOverviewView`, `MediaSourceInfoView`, & `MediaStreamInfoView`

JPKribs:
## Summary

This unifies`ItemOverviewView`, `MediaSourceInfoView`, & `MediaStreamInfoView`.

### `ItemOverviewView`

No real work here as they were always about the same. Just added the `FormItemSection` for iOS to make this prettier and better match the tvOS `Form` overload that has the poster in it.

### `MediaSourceInfoView`

tvOS and iOS this should be the same as prior. tvOS wraps the `MediaStreamInfoView` as trailing content based on the focused stream in the leading (same as prior). iOS opens `MediaStreamInfoView` as a `.sheet`. This could also be turned into a route as well. I used a `.sheet` with muliple `.presentationDetents`sizes. I'm not sold on this. I originally intended this as a way to keep it out of Navigation since tvOS wasn't using it but there is another `mediaStreamInfo` for subtitles editing so that's a but of a mute point. I'm game to revert this part if wanted, this would just be:

```swift
@ViewBuilder
private func streamRow(_ stream: MediaStream) -> some View {
    ChevronButton(stream.displayTitle ?? .emptyDash) {
        router.route(to: .mediaSourceInfo(source: source))
    }
    #if os(tvOS)
    .focused($focusedStream, equals: stream)
    #endif
}
```

Then delete all the `.sheet` logic here:

```swift
.sheet(
    isPresented: Binding(
        get: { selectedStream != nil },
        set: { if !$0 { selectedStream = nil } }
    ),
    onDismiss: { selectedStream = nil }
) {
    NavigationStack {
        if let selectedStream {
            MediaStreamInfoView(mediaStream: selectedStream)
                .navigationBarCloseButton {
                    self.selectedStream = nil
                }
        }
    }
    .presentationDetents([.medium, .large])
}
```

I can go either way and it's a quick fix. I think I'm just a sucker for `.presentationDetents([.medium, .large])`. My reason for using this here is it allows more a `peak` into a stream for those of us who go looking for a specific stream details but don't know which we're looking for.

### `MediaStreamInfoView`

Almost unchanged from original which the exception that this should be usable as both the trailing part of `MediaSourceInfoView` and a standalone iOS/iPadOS view.

### Generic Cleanup

- Instead of repeating `FormItemSection` & its tvOS replacement in multiple places, I added the tvOS replacement into the `FormItemSection` as part of a `PlatformView`.
- Made the `FormItemSection` text into `Marquee`s so the layout shouldn't shift between usages. This also helps for titles that extend past 2 lines _(e.g. Lord of the Rings: Fellowship of the Ring)_.
- Added the year to `FormItemSection` for Movies, etc. where there is no `subtitle` to flesh it out a little more for Movies or non-child items.
- Creates a `LabelContentStyle` for tvOS when you need the labels to essentially just be an empty button for focus. Centralizes all work for `LabelContent`s that need to be focusable for tvOS scrolling.
- Moves the `BaseItemDto`.`ParentLabel` out of the `ItemDeletionView` into the `BaseItemDto` extension.
- Remove the `BaseItemDto`.`SubtitleLabel` out of the `ItemDeletionView` in favor of `item.type?.displayTitle` since that's all I was getting from that.
- Remove the duplicate route for `itemOverviewView` in favor of `itemOverview`.

### Videos

**iOS**

https://github.com/user-attachments/assets/491f7587-a576-462f-bb7e-85dd401356f3

**tvOS**

https://github.com/user-attachments/assets/c809025a-16fa-4402-b856-3c920da6fe34

### DIFF
--- a/Shared/Views/MediaInformation/MediaSourceInfoView.swift
+++ b/Shared/Views/MediaInformation/MediaSourceInfoView.swift
@@ -14,11 +14,11 @@ struct MediaSourceInfoView: PlatformView {
     @FocusState
     private var focusedStream: MediaStream?
 
-    @State
-    private var selectedStream: MediaStream?
-
     @Router
     private var router
+
+    @State
+    private var selectedStream: MediaStream?
 
     let source: MediaSourceInfo
 
@@ -28,7 +28,8 @@ struct MediaSourceInfoView: PlatformView {
             router.route(to: .mediaStreamInfo(mediaStream: stream))
         }
         .focused($focusedStream, equals: stream)
-        .onChange(of: focusedStream) { newValue in
+        .backport
+        .onChange(of: focusedStream) { _, newValue in
             if let newValue {
                 selectedStream = newValue
             }


JPKribs:
Unrelated, but I was screwing around with a `NavigationTitle` overload in the `ViewExtensions`

```swift
func navigationTitle(
    _ marquee: String,
    speed: CGFloat = 60.0,
    delay: Double = 2.0,
    gap: CGFloat = 50.0,
    fade: CGFloat = 10.0
) -> some View {
    navigationTitle(marquee)
        .toolbar {
            ToolbarItem(placement: .principal) {
                Marquee(marquee, speed: speed, delay: delay, gap: gap, fade: fade)
                    .font(.largeTitle.weight(.bold))
            }
        }
}
```

This works fine in tvOS but iOS this does not display as the `.navigationTitle(marquee)` takes priority. However, we want that populated for accessibility as it's a defined value Apple bakes in.

If there was a way to make that work, I think that would be nice for these views as the titles are potentially huge as file names but not nice enough that I think it's worth spending too much time on 🤷‍♂️

JPKribs:
I've changed this to just use the route instead of the `.sheet`. Reviewing this later it just made more sense to preserve this as a route since this was used in subtitles already:

https://github.com/user-attachments/assets/c9c830e8-5005-4eaa-a970-18992fd0aae5

Shared/Components/FormItemSection.swift
```diff
@@ -27,13 +27,13 @@ struct FormItemSection: View {
                 VStack(alignment: .leading) {
 
                     if let parent = item.parentTitle {
-                        Text(parent)
+                        Marquee(parent)
```
LePips: Please don't propagate the usage of `Marquee` here. The line limits are enough for most titles of media.

Shared/Views/ItemEditorView/ItemDeletionView.swift
LePips, line 1: I would err to not really keep working on this view, as at the top of the file I would just move this to a confirmation dialog instead. Deleting items is somewhat already a problematic responsibility for the server and I don't see it as a feature requiring so much data of the item to be re-presented to the user as they would only be able to navigate to delete it via viewing the item first.

Shared/Views/MediaInformation/ItemOverviewView.swift
```diff
@@ -0,0 +1,58 @@
...
+// License, v2.0. If a copy of the MPL was not distributed with this
+// file, you can obtain one at https://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2026 Jellyfin & Jellyfin Contributors
+//
+
+import JellyfinAPI
+import SwiftUI
+
+struct ItemOverviewView: View {
```
LePips, line 12: Please revert the design of the overview, it's just supposed to be simple and not in a `List/Form` design.

You can use the same consolidation I've done in:
- https://github.com/jellyfin/Swiftfin/pull/1752/changes#diff-56a241a3d3400a0dc97b3370b7ad128ee5c5a7b698a1cb81426534ec623d79ab

LePips (changes requested):

Shared/Views/ItemEditorView/ItemDeletionView.swift
JPKribs: Part of why this view exists is because tvOS doesn't allow you to trigger an alert from a menu (where the delete button is located).

JPKribs:
I had to add back in the `focusState` & `state` pattern for tvOS. Since going to the trailing content changes the `focusState`, it was not possible to focus that content:

https://github.com/user-attachments/assets/08ffb6cf-258d-4a4c-b108-c86651305f67

Shared/Views/ItemEditorView/ItemDeletionView.swift
LePips: You would be able to wrap with a `StateAdapter` and have the delete button in the menu toggle the confirmation dialog, where then it actually deletes.

LePips:
Ah, my bad - thanks!

PR #1935, opened 2026-03-10
Unify `FontPickerView`

JPKribs:
### Summary

Straight forward combination. These are both built on top of the `SelectorView` which is already both an iOS and tvOS target. Some cleanup to `SelectorView` to use the same `ListRowCheckbox()` so the image sizing is consistent across the whole Swiftfin. Same reasoning & changes found here: https://github.com/jellyfin/Swiftfin/pull/1823

### Videos

**iOS**

https://github.com/user-attachments/assets/c952eeeb-2464-4d27-a84f-22c92f4382be

**tvOS**

https://github.com/user-attachments/assets/ce4610a3-3cb3-4d4b-8e30-27f54f4980f8

### DIFF
--- a/Shared/Components/ListRowCheckbox.swift
+++ b/Shared/Components/ListRowCheckbox.swift
@@ -9,6 +9,8 @@
 import Defaults
 import SwiftUI
 
+// TODO: remove isEditing observation
+
 struct ListRowCheckbox: View {
 
     @Default(.accentColor)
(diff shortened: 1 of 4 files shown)


JPKribs:
This is the second to last "loose" view that isn't combined. The following views are not in folders and not merged with into tvOS that don't have an open/closed PR for their merge:

Screenshot 2026-03-09 at 22 03 57

~At this point, I think only the `SearchView` needs to be merged in as the all of the other highlighted items are iOS only. I think that one is an easy enough `PlatformView` as iOS and tvOS are very different I don't think there is room to combine that in a more clever way.~

`SearchView` is resolved via https://github.com/jellyfin/Swiftfin/pull/1752!

`ServerCheckView` is iOS only right now but I'm a bit confused on whether we need it or now. So that might be another view but I'll get there when I get there. I haven't looked at this at all so this could be good to bring over but last time I looked at this it was something different with iOS routing vs tvOS that might not translate into anything from this unification stuff I've been up to.

I suppose also the `AppIconSelectorView` if that's coming over to tvOS as well but I don't have the talent to make icons so that seems like a later project.

JPKribs:
> In testing this, there's a bug where the changing of the user session isn't happening correctly. Signing into user A and then into user B, the accent color and settings values will still reflect user A's settings (and set new values to A's settings, rather than B's).

My initial thought, one of the items we're using for tvOS and not iOS is `userAccentColor` which is what we set in the settings while we use `accentColor` in all of our views.

https://github.com/jellyfin/Swiftfin/blob/main/Shared/Views/SettingsView/SettingsView.swift

Line 10-27, sorry don't know how to grab lines on mobile

Likely, the easiest route would be a global type alias (if that's possible) so tvOS is `userAccentColor` while iOS is `accentColor`? I can test if that distinction is actually the cause of this but that was something I noticed and didn't question/change from the `SettingsView` switch

---

Ignore the above that's just tvOS but this is applicable to iOS as well. I think we're just not updating the AccentColor on user switch.

LePips:
The issue isn't with the usage of the `userAccentColor` as that's the setting for the user. It is [observed elsewhere](https://github.com/jellyfin/Swiftfin/blob/main/Swiftfin/App/SwiftfinApp%2BValueObservation.swift#L58) and then sets the global accent color.

There is an issue somewhere that is re-using views. For example, if user A presents the user settings and then switches to user B, when user B presents the user settings the views will not be brand new - they will still be views that hold keys to A's settings. We just may need to stick an `id()` somewhere in/on `SettingsView`, along with checking some other views to see if this is happening elsewhere.

PR #1937, opened 2026-03-11
tvOS `Form` Clipping & Fading

JPKribs:
### Summary

tvOS Form is a bit annoying in that the buttons will clip against the side of the frame. Adding padding just means that the buttons clip against the padding instead. We resolved this a long while ago using `.scrollClipDisabled()`. The downside of this approach is that now the buttons don't appropriately clip/hide when the hit the top of the view. I've been screwing around for this for a while and I think I landed on a good solution:

1. Keep the `.scrollClipDisabled()`
2. AFTER the `.scrollClipDisabled()`, **re-clip** just the top.
3. Add mask to fade it so there isn't a clean line where this disappears.

Also utilize `.edgePadding()` instead of `.padding(.top, 100)` to be more dynamic and better handle various screen sizes.

The total change is:

```
.scrollClipDisabled() ← Old  |  New ↓
.mask {
    VStack(spacing: 0) {
        LinearGradient(colors: [.clear, .white], startPoint: .top, endPoint: .bottom)
            .frame(height: 30)

        Color.white
    }
    .scaleEffect(x: 1.5, y: 1)
}
```

#### This is mainly seen where the buttons would still be visible behind the title Text. Particularly noticeable when scrolling back up and the buttons re-appear.

**Before**

https://github.com/user-attachments/assets/a4a7f356-96fe-4ecd-ba64-77c5612d69e6

**After**

https://github.com/user-attachments/assets/139a6092-5ef9-40b7-b88d-f7d368258b22


JPKribs:
> My change will instead have the clip happen a bit outside of the `Form`'s bounds. Trying to have the gradient within the underlying `UITableView`'s margins seemed a bit tight, especially since the top section title will seemingly be in the margin and only increase it every so slightly, but I didn't measure everything while introspecting for the values.

Just tested, this looks very clean!

PR #1941, opened 2026-03-15
Playback Info

### LINKED ISSUE #339: Stream Stats During Playback
The ability to enable a debug feature that overlays information about the stream. Similar to how some video games expose what fps and other metrics are happening for the game.
The type of information that could be shown is: bit rate, resolution, AVC, etc

This can help narrow down where issues are happening. For example, when I play 4K streams on local connection with swiftfin, I get no audio. All 1080p video does have audio. (This will be a bug report when I do some more testing)

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/339

Adds Playback Info to the Supplement. Should roughly mirror what is covered in Jellyfin-Web. Makes the Supplments scrollable when too many entries exist. 

### Video

**Vertical**

https://github.com/user-attachments/assets/4e57f20f-40e1-4e8a-ad89-8d785439856f

**Horizontal**

https://github.com/user-attachments/assets/a6c2083a-aa9d-4b2f-a495-dc2064a92735


JPKribs:
Putting this down for a bit. Having some difficulty with the scrolling when needed where it doesn't either break the formatting or the Supplement containers. I'll revisit this with fresh eyes soon.

JPKribs:
Okay, I have this fixed. So, adding the `ScrollIfLargerThanContainer` works but I had 2 issues.

1. It filled all of the available vertical room so it pushed the buttons off the screen. To resolve this, I just set it to `fixedSize` then manually added my own vertical padding.
2. By adding the `ScrollIfLargerThanContainer` the alignment centered. So, I added a parameter to set the alignment which is then used in the underlying `AlternativeLayoutView`.

### Scrolling

https://github.com/user-attachments/assets/26ec426c-11b7-4893-9153-3d7acac458b2

### Stationary

https://github.com/user-attachments/assets/b9d79a2c-2e15-42ec-aa79-6537b9ffba20

LePips (approved):
I had to just change due to updates I had to VLCUI and changed the order. I think we'll have to move this to be owned by the `MediaPlayerItem` and have an option to disable it, but the whole supplement handling needs more attention.

Issue #1928, opened 2026-03-06
Implement Liquid Glass


JPKribs:
### This feature request respects the following points:

- [x] This feature has **not** already been requested on [GitHub](https://github.com/jellyfin/Swiftfin/issues?q=is%3Aissue%20state%3Aopen%20type%3AFeature).
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).

### Platform

iOS, tvOS

### Describe the feature you'd like to see

Implement Liquid Glass design elements into Swiftfin. I believe this falls into roughly 3 steps:

- [ ] Gating Liquid Glass vs standard design behind a os26 flag
- [ ] Determining the best alternative design for pre-os26 elements (or tvOS on older hardware)
- [ ] Actual implementation and testing of Liquid Glass throughout Swiftfin.

I do not believe this is a good issue for first time contributors as there will likely be some nuance or opinionated design decisions required.

_(The previous issue created for this was removed by the owner. I am creating this to track this task.)_

lavavex:
According to Apple Development guidelines [here](https://developer.apple.com/news/?id=ueeok6yw)
They will force all apps to be compiled on Xcode 26 or later, I noticed in the documentation for contributing that it still states Xcode 15, is that a doc typo or is development going to be forced to new sdk soon? (Commenting here cause xcode 26 is required for Liquid Glass anyways)

JPKribs:
> I noticed in the documentation for contributing that it still states Xcode 15, is that a doc typo or is development going to be forced to new sdk soon?

No, the documentation is just out of date. I've been using 26.4 but we build Swiftfin for release on 26.2.

yoveto:
https://github.com/jellyfin/Swiftfin/pull/2051

Implements a lot of the UI in liquid glass. This changes does not make any pre ios 26 changes.

PR #1946, opened 2026-03-21
Swizzle Liquid Glass

LePips:
Add experimental setting to swizzle liquid glass, testing with 26.2 (the key has changed over time).


JPKribs (approved):

JPKribs:
Would you say this resolves:

https://github.com/jellyfin/Swiftfin/issues/1928

LePips:
No, this is essentially just for development. I would resolve that after we've made all of our necessary changes.

PR #1949, opened 2026-03-23
`ItemImage` & `PhotoPicker` Rework

JPKribs:
### Replace `UIViewControllerRepresentable`'s `PhotoPickerView` with `.photoPicker` 

As of iOS 16, we now have access to utilize the SwiftUI modifier for [`.photoPicker`](https://developer.apple.com/documentation/photosui/photospicker) opposed to using our custom `UIViewControllerRepresentable` `PhotoPickerView`.

The current process is we go from `PhotoPickerView` to `PhotoCropView` to utilize Mantis to drop our images. This PR introduces overload for `.photoPicker` that directly routes the results into Mantis for cropping in a self contained `.sheet`. This takes both the `PhotoPickerView` and `PhotoCropView` out of the navigation stack. I took this as an opportunity to bump Mantis to 2.29 from 2.28.

### `PhotoCropView` Cleanup

I've updated cropping to utilize our toolbar buttons and added preset aspect ratios as most images for posters (our primary usage outside of user images) are set values like 16:9 or 2:3. This is kind of just extra fluff but it's nice for Item Images. This also the same `FilterDrawerButton` styling so I've turned that into a `ButtonStyle` and shared it between `PhotoCropView` and `FilterDrawerView`.

User Profile still only allows 1:1 so this only shows for Item Images.

### Unify Photo `ViewModel` Logic

Instead of a thin wrapper around `PhotoCropView` for User and Image, I've created a `ImageViewModel` which is extended by `UserImageViewModel` and `ItemImageViewModel`. The `PhotoCropView` looks for a generic `ImageViewModel` so the save functionalities are the same `ViewModel` `Action` but overriden based on `UserData` or `BaseItemDto`.

### `ItemImagesView` Rework

These have had their `ViewModels` updating (Minus the `PaginLibraryViewModel` one) so these were all updated to use the update `@Stateful` APIs. While there, I made some changes. I will go from smallest changes to largest:

#### `RemoteImageDetailView`

Renamed from `ItemImageDetailView`. This was renamed as I am no longer using it for both `ImageInfo` (existing images) and `RemoteImageInfo` (searched, remote images). Instead, I am purely using this for `RemoteImageInfo` to see search results in more detail. The `ImageInfo` (existing images) version of this was replaced by just a confirmation dialog for deletion (see `ItemImagesView`

Hide labels for details both if the values are nil OR if the values are empty. Most commonly, language was returning "" so it would just be an empty row.

This is otherwise unchanged.

#### `RemoteImageSearchView`

Similar renaming as `RemoteImageDetailView` to better match the other view.

Force 3 columns on iOS for square and portrait and 2 for landscape. iPadOS use dynamic sizing.

Remove all the custom Poster Button in favor of just using the `PosterButton` since this conforms to `Poster`.

Routing to `RemoteImageDetailView` now uses a push zoom instead of a .sheet. This was previously 3 stacked sheets so, IMO this works better and there are no longer circumstances where all 3 .sheet get dismissed at once. IMO, the zoom just looks nicer as well.

#### `ItemImagesView`

This was a complete overhaul! The previous view was functional but I believe the new view is much much better:

| **Before & After** |
| ------------------ |
| Screenshot 2026-03-25 at 23 50 25 |

1. Images descriptive text at the top with a link to images Jellyfin documentation. This is both more informative and also better matches our other settings in the dashboard.
2. Instead of displaying every image type in a row, utilize a dropdown menu, modeled after `SeriesEpisodePicker`, to only show selected items. Default to `.Primary` as it's the most used/important instead of having to scroll to it.
3. Provide an image description at the bottom. Current version, we have all the image sections but there isn't a clear way to tell what is the difference between a backdrop and a thumb. The description explains this and a warning is presented for images unused in clients (pulled from our documentation).

Screenshot 2026-03-25 at 23 58 43

4. Removed the unnecessary full screen sheet for deletions in favor of a confirmation dialog.

### Generic Cleanup

- `ServerLogView` had a hardcoded URL I moved to URLs while I was adding a new one. 
- Migrated everything in `LocalUserView` to use a `StateAdapter` while I was removing the `PhotoPickerView` routing in favor of the modifer.
- Better `Poster` conformance for `RemoteImageInfo` where we default the `CollectionVGrid` to use the `ThumbnailURL` in the image sources instead of the full image.

### User Profile

https://github.com/user-attachments/assets/4d87f21a-e383-46c4-b08f-d87a73740c83

#### Search

https://github.com/user-attachments/assets/93bfa809-af8d-45cb-b9d1-e2a1a98920db

#### Upload File

https://github.com/user-attachments/assets/42851674-bdf7-41c9-931a-3ed6b8595562

#### Upload Photo

https://github.com/user-attachments/assets/1cc99845-d7dd-4776-b10e-d2bdc8a99c34

#### Delete

https://github.com/user-attachments/assets/ac91a717-e8a9-4668-8da4-9961a1ed2ae9


JPKribs:
Realized I never tested this on iPadOS. Here is what that looks like:

https://github.com/user-attachments/assets/6041be1a-42a4-460d-8219-fbd36cb7a9fb

LePips (dismissed):

LePips (dismissed):

LePips (approved):

PR #1955, opened 2026-03-29
Fix marketing version

LePips:
Fix marketing version. If the TestFlight action still errors after this, I'll test in a private repo for a bit.


JPKribs (approved):

JPKribs:
I know you have access to the same logs I do but just in case it's giving:

https://developer.apple.com/library/archive/qa/qa1827/_index.html

I can also take a look at this as well I don't want to pull you away from other things. Just lmk

PR #1956, opened 2026-03-30
`AdminDashboard` Server Stats

JPKribs:
### Summary

Adds server stats to the Admin Dashboard. Right now, this is in the Admin Dashboard but it's a self contained section so we should be able to move it somewhere else if preferable. I'm not 100% sold on this being here because every other button takes you somewhere. It feels strange to mix action items and information like this into a single view.

Also, Apple's `DisclosureGroup` chevrons are slightly different than our's so having them side-by-side looks a tiny bit off if you stare at it too long like I have.

The content should roughly mirror Jellyfin-Web:

### Video

https://github.com/user-attachments/assets/9025e5f0-93b5-4819-9933-af13c6ca435b


LePips (changes requested):
For these I would differ from the web and have a `General` page with the server name, version, the server language setting (for now), QuickConnect toggle, and the item counts (kinda like iOS settings).

There should then be a separate `Storage` page for showing the storage of the folders like you've done so. This is where I would also have the web's `General/Paths` setting. We can also show the storage of the libraries, where they would then link to the settings of that library when implemented.

JPKribs:
> For these I would differ from the web and have a `General` page with the server name, version, the server language setting (for now), QuickConnect toggle, and the item counts (kinda like iOS settings).
> 
> There should then be a separate `Storage` page for showing the storage of the folders like you've done so. This is where I would also have the web's `General/Paths` setting. We can also show the storage of the libraries, where they would then link to the settings of that library when implemented.

Okay, I think I get what you are looking for! Here is what I am working on:

https://github.com/user-attachments/assets/1dddc482-52f6-43d7-85d1-c987a9cba28f

JPKribs:
Okay, current state:

https://github.com/user-attachments/assets/154188d1-79e4-4c44-bb9b-97ba0ad7da7a

JPKribs:
Not sure how to handle the paths view. Only two of them are update-able via API so I have them in their own section. Might make sense to use `DisclosureGroups` but then I think we have to use inset lists. Most of our more "complex" list rows we use plain list styles to take the full width. Insets might look strange or unnecessarily condense.

Updating file paths via a textfiedl in the alert is also not super intuitive. Also, it's very easy to get a 500 error when setting a folder to an already in-use folder or a 404 if the folder doesn't exist.

For now, I'm taking a break on this but I think this is a good start as a POC to kind of plan this out better. I've gone and highlighted our settings in the images below that we do not have in Swiftfin. I am not saying we SHOULD have all of these settings just that these are the missing ones. It might be easiest to draw designations on what goes where and what we drop now:

<details>

<summary>This PR Resolves...</summary>

Screenshot 2026-04-10 at 22-47-07 Dashboard
Screenshot 2026-04-10 at 22-46-48 General

</details>

<details>

<summary>Missing Settings</summary>

Screenshot 2026-04-10 at 22-46-07 Branding
Screenshot 2026-04-10 at 22 59 14
Screenshot 2026-04-10 at 22 59 04
Screenshot 2026-04-10 at 22-50-47 Backups
Screenshot 2026-04-10 at 22-50-35 Networking
Screenshot 2026-04-10 at 22-50-04 Trickplay
Screenshot 2026-04-10 at 22-49-46 Streaming
Screenshot 2026-04-10 at 22-49-36 Resume
Screenshot 2026-04-10 at 22-49-27 Playback
Screenshot 2026-04-10 at 22-48-48 NFO Settings
Screenshot 2026-04-10 at 22-48-39 Metadata
Screenshot 2026-04-10 at 22-48-31 Display
Screenshot 2026-04-10 at 22-47-42 Libraries

</details>

JPKribs:
Taking out of draft because I have other PRs I would rather focus on.

PR #1965, opened 2026-04-07
Bump Transmission to 2.8.3

### LINKED ISSUE #1709: [iOS] Rotation doesn't work in new media player

JPKribs:
### Summary

Fixes/Closes: https://github.com/jellyfin/Swiftfin/issues/1709

Now that the rotation issue is cleared up this the package, bumping this version to take advantage of the ~50 other releases between our pinned version and 2.8.3. Tested on iOS for various transitions and I see no issues on this version. Rotating the player works as expected as well without tearing. VLC still has the weird issue where rotating it loses the height for a second while turning but that exists on 2.4.X as well and appears to be related to VLC as I am not seeing it when testing on AVPlayer on a branch I am looking at using the [AVPlayerProxy](https://github.com/JPKribs/Swiftfin/tree/avplayer). This is not like a final or official branch I am working from but more just a way for me to test some track issues from another PR and also test rotation changes here.

Anecdotally, this feels a lot snappier as well and I can no longer create some weird interactions I can on Main where trying to press buttons mid-transitions causes weird animations/goes to an unexpected location. Very anecdotally seems like an overall improvement!

tvOS builds without issue now per the other change in 2.8.3!

### DIFF
--- a/Shared/Coordinators/Navigation/NavigationInjectionView.swift
+++ b/Shared/Coordinators/Navigation/NavigationInjectionView.swift
@@ -69,6 +69,7 @@ struct NavigationInjectionView: View {
                 NavigationInjectionView(coordinator: newCoordinator) {
                     route.destination
                 }
+                .environmentObject(rootCoordinator)
                 .background(.regularMaterial)
             }
         #else // <- Start: Use this for both OS when fixed
@@ -82,6 +83,7 @@ struct NavigationInjectionView: View {
                 NavigationInjectionView(coordinator: newCoordinator) {
                     route.destination
                 }
+                .environmentObject(rootCoordinator)
             }
         #endif // <- End
         #if os(tvOS)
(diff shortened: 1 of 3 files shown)


JPKribs:
As a note, this this bumps Engine as well, I did check all our `stateAdapters` on tvOS and their mirrors on iOS. These work as expected. I primarily checked `stateAdapters` since I believe those are our only usage of this package but please let me know if that is incorrect or if anything needs testing!

LePips (approved):

PR #1969, opened 2026-04-08
Filter Unusable Sort Options & Include More SortBy Options Based on Type

### LINKED ISSUE #1523: Enable More Sort By Options
As of 10.10, the enum for `ItemSortBy` offers a lot more options for ordering content. The issue is that not all `ItemSortBy` work with each `BaseItemKind` (I think). It would be nice to increase the number of `ItemSortBy` options based on `BaseItemKind` or whatever limiting factor exists. 

An example is sorting Movies based on `.similarityScore` results in a 500 Error. My guess is this `ItemSortBy` is only supposed to be used for "Recommended" sections.

Additionally might be good to move from defaulting from `.sortName` to `.default` but in testing I don't know what the difference is.

**Additional context**
 In `ItemSortBy` we have the following options:

        .default
        .airedEpisodeOrder
        .album
        .albumArtist
        .artist
        .dateCreated
        .officialRating
        .datePlayed
        .premiereDate
        .startDate
        .sortName
        .name
        .random
        .runtime
        .communityRating
        .productionYear
        .playCount
        .criticRating
        .isFolder
        .isUnplayed
        .isPlayed
        .seriesSortName
        .videoBitRate
        .airTime
        .studio
        .isFavoriteOrLiked
        .dateLastContentAdded
        .seriesDatePlayed
        .parentIndexNumber
        .indexNumber
        .similarityScore
        .searchScore

Of these, only the following are enabled Swiftfin:

        .premiereDate
        .name
        .sortName
        .dateLastContentAdded
        .random

### LINKED ISSUE #1967: Issues with the sorting method in the media library.

### Description of the bug

English:
In the media library, selecting different sorting methods and orders does not yield accurate results.

中文：
In the media library, selecting different sorting methods and orders does not yield accurate results.

### Reproduction steps

1. Open Swiftfin and navigate to your library.
2. Select sort by "Date added".
3. Order by “Ascending” or "Descending".
4. You will find that the results of both sorting methods are identical—they haven't actually changed based on the chronological order of when items were added.

### Platform

iOS

### Application version

1.5 (1)

None

### Device information

iPhone 12 Pro

### OS version

18.7.1

### Jellyfin server version

10.11.8

### Screenshots or videos

_No response_

### Relevant log output

```shell

```

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1967 & https://github.com/jellyfin/Swiftfin/issues/1523

To start, this PR creates a different variation of `SupportedCaseIterable` called `AssociatedCaseIterable`. This is like `SupportedCaseIterable` but is based on an external input. In our case, we only want filters that are applicable to the content being shown. So, I am passing in the `LibaryParent`.`supportedItemTypes` and using that to determine which filters should be used.

```swift
/// The `supportedCases` filtered against `filter` `AssociatedType`.
/// - `requiringAll` removes cases that only apply to some `AssociatedType`s but not all of them
static func associatedCases(for filter: [AssociatedType], requiringAll: Bool = false) -> [Self] {

    let filterTypes = Set(filter)

    return Self.supportedCases.filter { element in

        let applicableTypes = Set(element.applicableTypes)

        guard applicableTypes.isNotEmpty else {
            return true
        }

        return requiringAll ? filterTypes.isSubset(of: applicableTypes) : !applicableTypes.isDisjoint(with: filterTypes)
    }
}
```

The idea being, `SupportedCaseIterable` but for enums where the supported cases will vary based on an external variable. This still uses the Supported Cases as the master list of like, all possible usable elements. Then this is a subset of that list based on an external variable. Additionally, an optional flag of whether an element should be retunred if there is a single match or if ALL types need to match. 

I've tested all sort options in movies, shows, collections, and favorites and all work as expected. For composite libraries with multiple supported types (Studios, Favorites, etc.), this reduces the list to only sort options that are shared between the available types. This ensures we don't have sort buttons that behave erratically or not at all because there isn't valid content that can be sorted based on that type. See https://github.com/jellyfin/Swiftfin/issues/1967 as an example of this where the date added sort doesn't work for movies so it just doesn't appear to do anything.

Alphabetized the sorts just to be consistent then I grouped & sorted our `supportedCases` so that it kind of makes more sense how these are ordered. Since the `AssociatedCaseIterable` uses the `supportedCases` as the base list, the order of `supportedCases` is what the user gets in the `FilterView`

Finally, renamed "Date added" to a more accurate "Date of last added content".

### Weird Findings

`.seriesDatePlayed` doesn't do anything? Tried on series and episodes and movies and collections... This didn't change anything in ordering so I just removed it from the `supportedCases`. Just making a note of that for future reference.


LePips (changes requested):
I see the need and what you're going for but I don't think that a protocol like this is needed and could be even more restrictive in more general cases (what if one type needed to be "associated" with 2 types?).

Do the reverse: just have a variable on `BaseItemKind` that returns the intended `ItemSortBy`s and then in `FilterViewModel` it just performs a union against the supported cases.

JPKribs:
> Do the reverse: just have a variable on `BaseItemKind` that returns the intended `ItemSortBy`s and then in `FilterViewModel` it just performs a union against the supported cases.

Totally makes sense! I started that direction but I ended up switching thinking the `Associated` variation of this could be reused. Latest version of this works again as expected. See videos below:

| Movies | Shows | Multiple Types |
|-------|-------|----------------|
| <video src="https://github.com/user-attachments/assets/dec168d2-3804-494c-8034-e87669ca88b2" controls width="250"></video> | <video src="https://github.com/user-attachments/assets/c49972ba-0aca-4989-aab9-4eedc2e88b3d" controls width="250"></video> | <video src="https://github.com/user-attachments/assets/c1fd4d90-9b26-4af8-9de3-4767f6ee3f5a" controls width="250"></video> |

I've never used the `formIntersection` before so let me know if my union/join here is off:

```swift
let sortBySets = parent.supportedItemTypes.map {
    Set($0.supportedItemSortBy)
}

self.allFilters.sortBy = ItemSortBy.supportedCases
    .filter { sortBySets
        .dropFirst()
        .reduce(into: sortBySets.first ?? []) {
            $0.formIntersection($1)
        }
        .contains($0)
    }
```

I'm over-complicated this with the goal of only showing filters applicable to ALL types in the `LibraryParent` so we only want the shared elements between `BaseItemKind`'s `SupportedItemSortBy`.

LePips (approved):

PR #1979, opened 2026-04-22
Contributing Guide Cleanup

JPKribs:
### Summary

Updates the Contributing guide to update some [dated elements.](https://github.com/jellyfin/Swiftfin/issues/1928#issuecomment-4292608263) Primarily, Xcode 15 where it should be 15+ now.

Adds an Xcode version section. This is because I've been getting some questions about this lately so I think this would be best to document. Instead of hard coding this where we have to update this late, I instead just reference our CI workflow's Xcode version. The idea being, what we release on is our recommended version. This link also should reduce the need to manually updating this document every time that version updates.

Adds a section for testing. This is another question I've been seeing lately about how to run on real hardware for testing. This primarily just links https://developer.apple.com/documentation/xcode/running-your-app-in-simulator-or-on-a-device/ but I thought this would be good to add to our documentation to help provide some clarity for elements that should be tested on real hardware.

Also, Documentation was ### nested under Git Flow which didn't make sense so I just changes it to the same level at ##.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,11 +24,7 @@ jobs:
     runs-on: macos-15
 
     env:
-      XCODE_VERSION: "26.2"
-      IOS_VERSION: "18"
-      TVOS_VERSION: "18"
-      TVOS_SIMULATOR: "Apple TV 4K"
-      IPHONE_SIMULATOR: "iPhone 16 Pro"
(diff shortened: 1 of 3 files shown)


LePips (approved):
Just some adjustments and took this opportunity to also allow TestFlight pushes for tvOS.

PR #1452, opened 2025-03-15
[tvOS] Add accent colors to settings

Comet1903:
Hey there,

I’ve added the ability to select an accent color in the tvOS version of the app, similar to what’s already available in the iOS app. Since the `ColorPicker` component isn’t available on tvOS, I decided to implement a few preset colors for users to choose from. I also added the accent color to some more elements (like on iOS)

> I’m relatively new to Swift and SwiftUI, but I’m not new to programming in general. If there are any concerns with my idea (or code), I'll try my best to solve them.

PR #1980, opened 2026-04-23
Pin Fastlane & Ruby Versions

JPKribs:
### Summary

Updates Fastlane to get rid of the interactive error. Pins Ruby and Fastlane versions to ensure they run with the same version each time. Also updated to run on MacOS 13. I was getting a Swift Regex error that required 13+ to resolve.

I'm super new to Fastlane and am just following an update guide. Feel free to trash this if it's not the right direction!

Work originating to handle: https://github.com/jellyfin/Swiftfin/actions/runs/24813136606/job/72621951878#step:6:251 

Tested on my branch:

Screenshot 2026-04-22 at 21 26 43


LePips (changes requested):
I have actually reverted from using the local fastlane and instead opted into using the runner fastlane version, deciding to take on updating it as often as necessary. I've done the 2.233 update in #1981.

Sorry this was kinda hidden: https://github.com/jellyfin/Swiftfin/pull/1452/changes/52e63e3af16b76e817acd0bc006586acea57af65

JPKribs:
> I have actually reverted from using the local fastlane and instead opted into using the runner fastlane version, deciding to take on updating it as often as necessary. I've done the 2.233 update in #1981.
> 
> Sorry this was kinda hidden: https://github.com/jellyfin/Swiftfin/pull/1452/changes/52e63e3af16b76e817acd0bc006586acea57af65

No worries! IMO, makes sense to bundle into 1981 I'll defer to that PR

LePips (approved):

PR #1981, opened 2026-04-23
Cleanup Deprecated Code & Fastlane Updates

LePips:
- Remove ServerTicks
- Remove other deprecations I've made
- Fix TestFlight action scheme passing


JPKribs:
For the build errors, I think I have that in https://github.com/jellyfin/Swiftfin/pull/1980 if you need it

JPKribs (approved):
Jumped around the views touched and nothing is off. Media player works as well just since the manager had a single line change.

Looks good!

Edit: Changed the title to be more descriptive for the final 1.5 changelog

JPKribs:
@LePips still has:

# fastlane 2.233.0 is available. You are on 2.232.2.

Builds fine on device so workflow. My bad, I should've waited for that to conclude first...

LePips:
Well that's rather annoying. I had thought the erroneous version was 2.233.0, but it's currently in the process of being deployed.

I was gonna say we can wait this instance out and pin later, but this is too bothersome anymore. Reopened https://github.com/jellyfin/Swiftfin/pull/1980 for that.

PR #1982, opened 2026-04-23
Finish `ServerLogs`

JPKribs:
## Summary

This started by me just wanted to wrap up some TODOs during some downtime. There were 2 in the `ServerLogsView` which should now be knocked out by this PR:

```Swift
// TODO: download to device?

// TODO: super cool log parser?
//       - separate package
```

### FFMpeg & Non-System Logs

Since we don't know the format of client logs (varies by client that submits them) and I don't know how to parse the FFmpeg logs, these are only viewable in plaintext. I've added an option to open in browser to keep our legacy behavior and a new share option. This share sheet is just a wrapper around `UIActivityViewController`. But, this ineteracts with the locally downloaded log file and allows us to save this to a new location or like, do something else with the file like text it to someone.

<table>
  <tr>
    <th>Non-System Log Actions</th>
  </tr>
  <tr>
    <td>
      <video src="https://github.com/user-attachments/assets/f5ba1bb1-4c31-4903-ad74-6a038b8df27f" controls width="600"></video> 
    </td>
  </tr>
</table>

If there is a clean way to parse FFmpeg, I would be happy to add that as well. I'm 100% certain there is just not something I could find and frankly I'm not positive what people find valuable in those logs to order/group on.

### System Logs (Parsable)

Parses the system logs via regex. This also has a log entry view for each log line that allows copying the log itself.

<table>
  <tr>
    <th>System Log Actions</th>
  </tr>
  <tr>
    <td>
      <video src="https://github.com/user-attachments/assets/991ab638-b60c-4855-96f9-57ca762d4e34" controls width="600"></video> 
    </td>
  </tr>
</table>

### Empty Logs

The logs should really never be empty but if they are they will get:

<table>
  <tr>
    <th>No Logs</th>
  </tr>
  <tr>
    <td>
      Empty Log File
    </td>
  </tr>
</table>


JPKribs:
@LePips no rush at all this is a low priority PR but I have 2 questions about approach:

1. The smaller one, is there a reason we don't use [@ ScaledMetric](https://developer.apple.com/documentation/swiftui/scaledmetric)? I was struggling to get something formatting with larger accessibility text sizes and this solved it. Is there any reason I should avoid this macro?

2. For log parsing, I have this mapped out now. So, I download the log file and read the data in via [chunks as a psuedo-pagination](https://github.com/JPKribs/Swiftfin/blob/486ae8ff575522a784f69b9a6309f9551db4bfe4/Shared/ViewModels/AdminDashboard/ServerLogs/PagingLogViewModel.swift) to better handle massive logs. While it's being read in, I am parsing it using a parser that gets passed in. These conform to [`LogParser`](https://github.com/JPKribs/Swiftfin/blob/486ae8ff575522a784f69b9a6309f9551db4bfe4/Swiftfin/Objects/ServerLogs/LogParser/LogParser.swift) so you can either [parse system logs via their pattern](https://github.com/JPKribs/Swiftfin/blob/486ae8ff575522a784f69b9a6309f9551db4bfe4/Swiftfin/Objects/ServerLogs/LogParser/ServerLogParser.swift) or as [raw /n separated text](https://github.com/JPKribs/Swiftfin/blob/486ae8ff575522a784f69b9a6309f9551db4bfe4/Swiftfin/Objects/ServerLogs/LogParser/RawLogParser.swift). So we can swap between parsed logs and just the raw text like on the site.

Right now, what I have works but is this the right approach? It feels pretty overkill and I'm not sure I can parse the FFmpeg logs so I'm not sure if there was a cleaner way to do this.

Only outstanding issue is that `CollectionVGrid` doesn't let me do variable height buttons so the raw view gets crushed to 1 line per. I think other than that and trying to parse FFmpeg I have this? I'm also not positive on that paginated log reading I just stole that from a forum post so I need to do more digging.

Mostly, I just want to make sure I'm not completely off base here. Video below is current state:

https://github.com/user-attachments/assets/9fbac69b-87b0-452f-b41c-0dff87f39f0f

JPKribs:
This PR should be good! This being said, structurally, I took a bit of a swing with how this works with the nested type. The nested parsing inside the ViewModel, passed in, feels smart but I'll defer to you @LePips if you have any insight.

This is also, frankly, a fairly low impact PR. So I'm not urgently looking to pick this back up haha

Also realizing I still need to move this over to the new paging logic so that'll be part of this as well!

PR #1984, opened 2026-04-24
Supplement Customization, People Supplement, & VLCKit 3.7.2

JPKribs:
### Summary

Adds a setting to customize Supplements and change order using an `OrderedSectionSelectorView`. Looks to resolve the issue brought up here: https://github.com/jellyfin/Swiftfin/pull/1941#pullrequestreview-4166125266.

Super simple change I am pulling this from another branch I had this on:

| Settings | Result |
|---------|--------|
| Simulator Screenshot - iPhone - 2026-04-23 at 18 44 00 | Simulator Screenshot - iPhone - 2026-04-23 at 18 44 06 |

| Settings | Result |
|---------|--------|
| Simulator Screenshot - iPhone - 2026-04-23 at 18 38 40 | Simulator Screenshot - iPhone - 2026-04-23 at 18 43 36 |


LePips (changes requested):
I'll have to think about providing this granular of functionality. The supplements are a bit more important during media playback, especially is there is a present queue. Similar to how auto play won't work unless the button itself is ever shown, I wouldn't want to have queue progression if the supplement isn't present. But that is just one example of supplement behavior.

While supplements are currently only made in `setSupplements`, that should eventually change such that supplements can be provided from within the manager (queue) and from the item itself as well (info, playback information, chapters).

I know this was probably prompted by me wanting to have a toggle for the playback information. I may just want this to be a boolean for now, even if we implement what you've done in the future.

This also causes bad swipe up interactivity when there are no supplements.

JPKribs:
I had an idea for this. Optional second array of removable elements. The elements are still re-arrangable so you can customize it but we can prevent removing core functionality.

Video below shows what this looks like. I moved in the Person supplement I had in another branch since it looks kind of weird when there is only 1 item in there that can be removed:

https://github.com/user-attachments/assets/4950efef-d9eb-452e-acfd-13e11a5deb1c

Maybe leave the +/- icon but we gray it out instead? Or like a trailing `Required` text instead of the +/- icon?

---

This is the people supplement. I can move this back to another PR. It kind of hinges on this PR so it would come after but I think it makes sense to have more than one optional:

| Compact | Regular |
|---------|---------|
|  |  |

LePips (approved):
Great idea!

PR #1986, opened 2026-04-24
Localization Checks

### LINKED ISSUE #1985: Sorry, unable to translate some items.
In the last few test versions, I noticed that the "From Beginning" cannot be translated, as I am unable to locate the corresponding English entry.

I don't know if this is a bug, perhaps I wasn't looking in the right place. If anyone knows how to translate this, please let me know. 

Thanks.

Image

JPKribs:
## Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1985

I got detoured working on https://github.com/jellyfin/Swiftfin/pull/1956. This PR sets to resolve a couple of issues:

### `FindHardCodedStrings.swift`

Sometimes we have hard coded strings where we should not. Things like ["From beginning"](https://github.com/jellyfin/Swiftfin/issues/1985) where it just gets missed in the review process. I've created a script to add to our workflows that runs and finds missing strings. This is bit more complicated and will not be perfect but it should help find the bulk of offenders. This looks at the following functions and modifiers:
 
Screenshot 2026-04-24 at 16 06 14
 
This way, Text("String") will get flagged and Key["id"] does not. 

There will occasionally be items that are better suited as strings. For this, I have added a string modifier for `.doNotLocalize` that this script looks for. IMO, it's bit stupid to like have a modifier for this. The alternative was just:

```Swift
// Do not localize
Text("Hidden")
```

Which I can still revert us to but I used the modifier so it was strongly typed.

The final value of this, I have added a `--dry` & `--replace` to this script. This **creates** and **replaces** instances of hard coded strings. Running `--dry` will show all elements that will be updated. `--replace` creates the new strings in the English localization and updates all usages of the original hard coded value with the new `L10n` value. You will have to build after this is complete to generate the `Strings.swift` & `L10n` entries. If there is already an `L10n` instance for that particular string, this will just use that existing instance instead.

Screenshot 2026-04-24 at 16 29 19

So, in my testing, we should be able to just use "String" in development then run this script at the end to update everything and handle all of the localization. For me, personally, I think that's a big assist for dev work.

### `ProperNouns.swift`

Proper nouns do not need to be localized but they do need to be spelled correctly. I've added an extension on `L10n` for proper nouns. This way they are not picked up in the first script AND we can ensure that our usage of these are consistent. This idea came from the realization that I accidentally left THREE different spellings of Dolby with 2 misspellings. I know hard coding these is a bit of a chore but they should centralize spelling without localizing and, by extending `L10n`, this will come in using the same API as localized strings so it should be easy to use. 

I've moved over some of our media container items over and items like Dolby Vision (spelled correctly lol) and even Jellyfin & Swiftfin to ensure consistency throughout and avoid potential typos.

This is a lot less of a requirement but I wanted to do this to centralize spelling and also help limit the amount of `// Do not localize` or `.doNotLocalize` required throughout. If there is a better way to do this please let me know!

### `FindDuplicateStrings.swift`

This script is much much smaller. All this does is this finds localization where we have two keys that both share a value. I didn't think we had a lot of these but I found we had `auto` twice.

This also has a `--dry` and `--replace`. This takes the shortest key between these values and updates all instances to use the shorter of the two. In the example below, this found `L10n.auto` & `L10n.bitrateAuto` shared a string "Auto". Running `--replace` found all instances of `L10n.bitrateAuto` and replaced them with `L10n.auto` and removes `L10n.bitrateAuto` from the localizations file.

Screenshot 2026-04-24 at 15 53 19

### Other

We are using `\u{2026}` for `...` in several locations. Similar to `emptyDash`, I just added this to string. Changes nothing but just cleans this up a bit since `\u{2026}` is kind of a chore to remember for usage.


JPKribs:
This looks like this from the jobs now:

Screenshot 2026-04-24 at 16 46 47

Shared/Extensions/UIDevice.swift
```diff
@@ -33,12 +33,12 @@ extension UIDevice {
 
     static var platform: String {
         #if os(tvOS)
-        "tvOS"
+        L10n.tvOS
```
JPKribs, line 36: I know this is a bit overkill. IMO, this is just to enforce consistent casing.

Swiftfin/Views/ItemView/Components/CastAndCrewHStack.swift
```diff
@@ -32,7 +32,6 @@ extension ItemView {
                 SeeAllButton()
                     .onSelect {
                         router.route(to: .castAndCrew(people: people, itemID: nil))
```
JPKribs, line 34: This was unrelated. I just found the Cast & Crew casing was inconsistent. Then, I realized it routes you to Cast & Crew twice when you select it so I made the swap while I'm here.

Swiftfin/Views/ItemEditorView/ItemImages/ItemImageDetailsView/Components/ItemImageDetailsDetailsSection.swift
```diff
@@ -68,7 +68,7 @@ extension ItemImageDetailsView {
                 if let width, let height {
                     LabeledContent(
                         L10n.dimensions,
-                        value: "\(width) x \(height)"
+                        value: "\(width) \u{00D7} \(height)"
```
JPKribs: We use this for the x in other locations so this is a consistency change.

LePips: Like `.ellipsis`, make it a static string variable

LePips (changes requested):
I do like the intentions of this PR, but let's pivot to use SwiftLint instead for long term maintenance of the strings throughout the code. The following config file would warn on hard coded string usage within SwiftUI inits and `displayTitle` definitions. We can add our own views, like `PosterHStack`, as necessary.

We can then use `// swiftlint:disable:next hard_coded_display_string` and similar ways to disable the rule where necessary.

<details>
<summary>.swiftlint.yml</summary>

```yml
only_rules:
  - custom_rules

custom_rules:
  hard_coded_display_string:
    name: "Hard-coded Display String"
    regex: '(?:\b(?:Text|Button|Label|Toggle|Picker|Section|LabeledContent)\s*\(\s*"[^"\n]*(?:\\.[^"\n]*)*"|var\s+displayTitle\s*:\s*String\s*\{(?s:.*?)"[^"\n]*(?:\\.[^"\n]*)*"(?s:.*?)\})'
    message: "Avoid hard-coded user-facing strings. Use L10n instead."
    severity: warning
```
</details>

Add the script build phase to run SwiftLint to both targets and add SwiftLint as a brew dependency to install.

<details>
<summary>build script</summary>

```sh
export PATH="/opt/homebrew/bin:/usr/local/bin:$PATH"

if command -v swiftlint >/dev/null; then
  swiftlint lint
else
  echo "warning: SwiftLint not installed. Install with: brew install swiftlint"
fi
```
</details>

I may eventually migrate all of the scripts to a local `Tools` package and have us use [`SwiftLintPlugins`](https://github.com/SimplyDanny/SwiftLintPlugins) and the SwiftFormat package binary at some point in the future.

Documentation/contributing.md
```diff
@@ -57,6 +57,8 @@ If a Pull Request relates to an Issue, mention the issue correctly in the PR des
 
 [SwiftFormat](https://github.com/nicklockwood/SwiftFormat) is our linter. `swiftformat .` can be run in the project directory or install SwiftFormat's Xcode extension.
 
+[SwiftLint](https://github.com/realm/SwiftLint) enforces additional rules [Add better write up here]
```
JPKribs: TODO

JPKribs:
> I do like the intentions of this PR, but let's pivot to use SwiftLint instead for long term maintenance of the strings throughout the code. The following config file would warn on hard coded string usage within SwiftUI inits and `displayTitle` definitions. We can add our own views, like `PosterHStack`, as necessary.
> 
> We can then use `// swiftlint:disable:next hard_coded_display_string` and similar ways to disable the rule where necessary.
> .swiftlint.yml
> 
> Add the script build phase to run SwiftLint to both targets and add SwiftLint as a brew dependency to install.
> build script
> 
> I may eventually migrate all of the scripts to a local `Tools` package and have us use [`SwiftLintPlugins`](https://github.com/SimplyDanny/SwiftLintPlugins) and the SwiftFormat package binary at some point in the future.

I've added this and fixes up any remaining violating strings. I've added a test in validate PR that installs `SwiftLint` then tests against the project using --strict. I have confirmed this works [here](https://github.com/jellyfin/Swiftfin/actions/runs/25041227224/job/73344713873#step:5:524). I've also added references to install this in our documentation.

Let me know if this looks good to you! The only missing item is I need to make the SwiftLint section of the documention explain SwiftLint. I didn't have it top of mind of like SwiftLint vs SwiftFormat so I left it as a TODO if this looks otherwise the right direction: https://github.com/jellyfin/Swiftfin/pull/1986#discussion_r3152369644

LePips (changes requested):
I updated to using a brewfile and adjusted the regex, which caught a few more strings to either localize or ignore.

JPKribs:
I've cleaned up the remaining localization. I had one issue with the regex for `displayTitle` where it was catching both the individual line violations AND the `displayTitle` violation so I wasn't able to add an exception. This only apply to `displayTitle`s that had 2+ strings in them. See `VideoRangeType` or `DayOfWeek`. I just replaced:

`(?:^|\n)` with `(?<=\n)`

I'd appreciate another spot check on that haha

LePips (approved):

PR #1988, opened 2026-04-27
Finish `LetterPickerBar`

JPKribs:
## Summary

Going over old notes and TODOs. One that I _thought_ would be easy is finishing up the `LetterPickerBar`. The [fix in question](https://github.com/jellyfin/Swiftfin/pull/1326#pullrequestreview-2456826377) is letting the `LetterPickerBar` determine the height based on itself rather than a hard coded parameter. While I was there, I did some cleanup and go this ready for tvOS as well. I know the `PagingLibraryView` is reworked in https://github.com/jellyfin/Swiftfin/pull/1752 but this should be fairly portable to attach onto that or This isn't a huge PR but I think it comes down to 3 items:

### Determine Size from the Largest Letter

This is the reason I couldn't use the `AlternativeLayoutView` was because we are getting the dimensions of the letters in the bar rather than determining the layout once it exists. 

Screenshot 2026-04-26 at 21 58 47

So, on appear, we stack all letters on top of each other on the Z axis. This gives us a view with the height of the tallest letter and the width of the widest letter. Even if those are not the same letter. We need the largest overall so we can back the background square that fits ever letter in both direction. So loading this gets us the letter size:

```swift
.background {
    ZStack {
        ForEach(ItemLetter.allCases, id: \.hashValue) { letter in
            Text(letter.value)
                .font(LetterPickerBar.font)
        }
    }
    .hidden()
    .allowsHitTesting(false)
    .fixedSize()
    .trackingSize($letterSize)
}
```

Then I get the max of the height and width and use that as our square dimensions. I add 2 to this to add at least 1 around the edges so the largest letter doesn't poke out. Then we pass this into the `LetterPickerBar`.

### Turn the `LetterPickerBar` into a `ViewModifier`

Previously, for getting our spacing for this this, we padding the content then added the a ZStack layer with the letter picker over the now vacant space left by the padding. tvOS this was giving me a lot of trouble and I also felt like this might work better as a `.safeAreaInset` so this kind of just expands the existing safe area and places this in there. This keeps it on the same layer as the content and prevents weird focus issues.

This bundles up everything in the view modifer which is shared between platforms.

### Settings change

Changed the setting from `LetterPickerBar` on/off + orientation to just orientation then added disabled to the orientation. This is similar to what we did her and reduces the number of settings:

https://github.com/jellyfin/Swiftfin/pull/1825

### Other stuff

tvOS the letters are smaller but still comparable to our posters. I did this to better fit them all on the screen and make them take up less real estate. To use `.safeAreaInset`, `PagingLibraryView` now only ignores vertical safe area so we have horixontal available for the picker. The non-picker sider will be ignored as part of the modifier. If the picker is disabled, it will fully use both sides.

### iOS Examples

<table>
  <tr>
    <th>Scrolling</th>
    <th>Example 1</th>
    <th>Example 2</th>
    <th>Example 3</th>
  </tr>
  <tr>
    <td align="center">
      <a href="https://github.com/user-attachments/assets/f91d6032-97d6-44b4-a1af-94cf4aaeb905">
        
      </a>
      <sub>Click to play</sub>
    </td>
    <td>
      
    </td>
    <td>
      
    </td>
    <td>
      
    </td>
  </tr>
</table>

## tvOS Examples

<table>
  <tr>
    <th>Title</th>
    <th>Media</th>
  </tr>

  
  <tr>
    <td>Focus / Movement</td>
    <td align="center">
      <a href="https://github.com/user-attachments/assets/755e2ddb-93e4-4c15-9d38-f9c382d5ec1b">
        
      </a>
      <br/>
      <sub>Click to play</sub>
    </td>
  </tr>

  
  <tr>
    <td>Example 1</td>
    <td align="center">
      
    </td>
  </tr>

  
  <tr>
    <td>Example 2</td>
    <td align="center">
      
    </td>
  </tr>
</table>


LePips (approved):
Looks great! I think there are a few good ways to polish this button on each platform:

iOS: Instead of having there be a row of buttons, have the bar be a whole touch target and the letter underneath would be the selection. This would allow swiping up and down smoothly, somewhat similar to the native section element (Contacts app).

tvOS: Focus on top element on initial bar focus, focus selected letter if there is one.

JPKribs:
> iOS: Instead of having there be a row of buttons, have the bar be a whole touch target and the letter underneath would be the selection. This would allow swiping up and down smoothly, somewhat similar to the native section element (Contacts app).

I was looking at this! Some kind of gesture view and tracking indexes but I ran into trouble when the letter bar needed to scroll where I couldn't get a clean separation between scrolling and "hovering".

It's on the list of ideas if I can stop getting distracted with random PRs lol

PR #1992, opened 2026-04-29
Jump to End/Start

JPKribs:
### Summary

This is more of an issue on [tvOS](https://github.com/jellyfin/Swiftfin/pull/1902) but if I do the jump forward and the jump duration exceeds the time left in the file it just won't jump.

This PR moves the jump to the manager instead and uses the manager to calculate if the jump goes the full duration or a partial to just get to the end of the file. Same issue for jump backwards as well. This just says "If the jump is 15sec and there are 10 sec left then jump 10 sec instead to the end of the file."

### Before

https://github.com/user-attachments/assets/92a815ae-b010-4e65-864b-f144e331352f

### After

https://github.com/user-attachments/assets/0daee154-e08f-45ab-98da-29368e9b794a

### DIFF
--- a/Shared/Objects/MediaPlayerManager/MediaPlayerManager.swift
+++ b/Shared/Objects/MediaPlayerManager/MediaPlayerManager.swift
@@ -221,33 +221,6 @@ final class MediaPlayerManager: ViewModel {
         self.playbackItem = playbackItem
     }
 
-    func jumpForward(_ duration: Duration) {
-        guard duration > .zero else { return }
-
-        let target: Duration
-
-        if let runtime = item.runtime {
-            let remaining = max(.zero, runtime - seconds)
-            target = min(duration, remaining)
-        } else {
-            target = duration
-        }
-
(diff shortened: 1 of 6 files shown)


LePips (changes requested):
This stuff should be handled at the proxy level, the manager should not be concerned with runtime manipulation like this.

I've updated VLCUI to `0.8.1` to properly handle jumping at the edges, but there seems to be some oddities with remote streams vs local files like I was using for jumping forward.

So, implement the jump forward considerations in `VLCMediaPlayerProxy.jumpForward` instead.

JPKribs:
> So, implement the jump forward considerations in VLCMediaPlayerProxy.jumpForward instead.

Understood. Latest commit I've bumped to 8.0.1 & moved the check to only the `VLCMediaPlayerProxy.jumpForward`. Tested and this is now working forwards to 0:00 and within 1 sec at the end. I think that's just a rounding issue.

LePips (approved):

PR #1994, opened 2026-04-30
Focus `LetterPickerView` on Entry

JPKribs:
### Summary

I'm working on some focus elements. I've been digging into some of the default focusing tooling and found the [`defaultFocus`](https://developer.apple.com/documentation/swiftui/view/defaultfocus(_:_:priority:)). I took a detour to see if I could get this to work with the `LetterPickerView` to resolve half of [this comment](https://github.com/jellyfin/Swiftfin/pull/1988#pullrequestreview-4186218453).

The way this works, this will default to the first letter at the top of the bar on first focus. On future focuses, this will default to the selected letter. If there is no selected letter, this will return us to the previously focused letter.

```swift
.defaultFocus(
    $focusedLetter,
    viewModel.currentFilters.letter.first
        ?? lastFocusedLetter
        ?? ItemLetter.allCases.first
        ?? ItemLetter(stringLiteral: "#"),
    priority: focusedLetter == nil ? .userInitiated : .automatic // This prevents focus getting stuck for .leading
)
```

### Bug

Normally, `priority: .userInitiated` works. It works without issue on the `.trailing` side but on the `.leading` side you are trapped inside the `LetterPickerView` and cannot get out. I have no idea why this is but switching it from `.userInitiated` to `.automatic` once an item has been focused resolves this as a workaround.

### Video

Video is a sliced down to the relevant portions to get around the 100mb limit.

https://github.com/user-attachments/assets/c065e812-8de1-4f71-8e3c-edaab6549038

### DIFF
--- a/Shared/Components/LetterPickerBar/LetterPickerBar.swift
+++ b/Shared/Components/LetterPickerBar/LetterPickerBar.swift
@@ -18,8 +18,6 @@ struct LetterPickerBar: View {
     private var focusedLetter: ItemLetter?
 
     @State
-    private var lastFocusedLetter: ItemLetter?
-    @State
     private var letterSize: CGSize = .zero
 
     // Largest letter +1p on either side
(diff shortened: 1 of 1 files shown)


LePips (approved):
I haven't seen that modifier before, which changes some other things I've thought about regarding focus.

JPKribs:
> I haven't seen that modifier before, which changes some other things I've thought about regarding focus.

Haha I've gone super into the weeds trying to get the tvOS player to stop doing random janky things. The other one I found was `prefersDefaultFocus` that snags focus first on appear and you can always reset to it using the resetFocus environment variable.

But, I think I finally nailed this tvOS focus nonsense!

Issue #1809, opened 2025-11-11
Replace `UIScreen.main.scale` Usages


JPKribs:
`UIScreen.main.scale` is being deprecated because SwiftUI manages layout responsively and provides the screen’s pixel density via the `@Environment(\.displayScale)` property, eliminating the need for global screen queries. We need to find and replace all instances of `UIScreen.main.scale` with `@Environment(\.displayScale)`.

Below are all of the locations where `UIScreen.main.scale` exists:

---

<img width="618" height="57" alt="Image" src="https://github.com/user-attachments/assets/1e208ed2-5398-42b4-9bcf-f2a75b5f308e" />

---

Please be sure to test these changes thoroughly as these images are used in many locations in Swiftfin.

LePips:
Closing from #2075.

Issue #1808, opened 2025-11-11
Replace `UIScreen.main.bounds` Usages


JPKribs:
`UIScreen.main.bounds` is being deprecated because SwiftUI manages layout responsively and no longer relies on global screen metrics. We need to find and replace all instances of `UIScreen.main.bounds` with `GeometryReader` & `geometry.size`.

Below are all of the locations where `UIScreen.main.bounds` exists:

---

<img width="632" height="78" alt="Image" src="https://github.com/user-attachments/assets/37546b01-7259-425e-8f5e-3d5a7b8068dd" />

---

PR #1995, opened 2026-04-30
Fix deprecated UIScreen.main usage across iOS and tvOS

bnicholson2050:
UIScreen.main was deprecated in iOS 16. This replaces all usages across the codebase with the modern window scene API. On iOS, the active UIWindowScene is used to get the screen. On tvOS, UIScreen.main is kept as it doesn't have the same window scene limitations. Also fixes a bug where scaleHeight was accidentally using maxWidth instead of maxHeight when scaling image dimensions. This is a fix by Claude Code


JPKribs:
This is already being worked on as part of https://github.com/jellyfin/Swiftfin/pull/1752. That is why that PR was linked to the issues: #1808 & #1809.

PR #2001, opened 2026-05-01
[tvOS] Media Player Supplements

JPKribs:
### Summary

This PR is a slice of https://github.com/jellyfin/Swiftfin/pull/1902. My reason for splitting this out into it's own PR is a two parter:

1. I want to trim down the total footprint of #1902 and this was a logical group that can be reviewed independently.
2. Some of these changes are refactors to iOS as well so I want to address these on their own rather than accidentally change iOS as part of #1902. 

The primary changes are the tvOS version but this also accomplishes the following:

1. Mirate the Posters for Chapters and Episodes over to the `PosterButton`. This not only does this keep our sizing consistent but this also lets us inherit the overlays to knock out some of the TODO in there.
2. Backport some of the focus modifiers. iOS 17 gets most of the same focus modifiers as tvOS but these backports should help reduce some of the build flags required.
3. This should make the Episode Supplement stateful with error catching. This is somewhat mirroring the Episode Picker. I don't love how it all turned out, I think several copies of `CollectionHStack` that swap in based on state feels a bit clunky. That being said, I'm not sure what a better alternative is.

#### iOS

https://github.com/user-attachments/assets/b4dda9ba-c71b-498b-b803-ab0318b07a54

https://github.com/user-attachments/assets/8cb2fe34-1bf2-4c12-b0fc-df4bfec8912a

#### tvOS

Since this branch doesn't have the player changes to add the supplement containers, this is pulled out of #1902 but the state of the supplements is the same.

https://github.com/user-attachments/assets/e0308c9d-4014-4652-9e24-ff8b7f798800


JPKribs:
I'm closing this out because there are some performance considerations with tvOS. Without the rest of the context of the player changes, this isn't really feasible to test/sign off on.

PR #2002, opened 2026-05-01
Update TestFlight Deployment

LePips:
Finally just sat down and worked through everything necessary for properly shipping both iOS and tvOS.


RespectIsEverything:
Does the delay in pushing TestFlight builds pertain to the RC 12.0 versioning bump?

anthonylavado (approved):

PR #2003, opened 2026-05-02
Workaround for `TabView`

### LINKED ISSUE #1749: Swiftfin player: chapter button opened the episode list
### Describe the bug

While playing the video, when clicking the chapter button, it opened the episode list, NOT chapter list. 

**Step to reproduce** 
1. In setting, use Swiftfin player 
2. Play some video 
3. Tap screen to show on screen control 
4. Click chapter button 
5. See it open episode list 

Click the chapter button again (while the episode list is open) will correctly show chapter list 

**Screenshot** 

Image
Image

### Application version

1.4 (14)

TestFlight

### Device information

iPad mini 6

### OS version

iPadOS 26.0

### Jellyfin server version

10.10.7

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1749

Let me know if this isn't the right approach! My understanding of the `TabView` issue in #1749 is that the default `TabView` behavior is to scroll to a location and can over scroll ending up in the adjacent supplement when we are trying to do there programmatically instead of swiping.

This PR just creates a `TabView` but waits for the previous transition to settle before committing the next one.

This is not a permanent solution but hopefully this can hold us over until `TabView` can be triggered programmatically accurately.

As a heads up, this is **_totally unusable_** with SwiftUI focus for tvOS so this would need to be a split solution.

### Videos

https://github.com/user-attachments/assets/5770a7f1-ed66-4fce-9f45-584c51004da4

https://github.com/user-attachments/assets/d00e351e-2ce3-40c2-9c71-ed2dd3829a43


LePips (approved):
I had the tab view keep track of its current view controller, so that there isn't the forward/reverse animation on selecting/deselecting a supplement.

PR #2009, opened 2026-05-07
Bump SDK to `1.0.0`

JPKribs:
### Summary

Testing https://github.com/jellyfin/jellyfin-sdk-swift/pull/51

Super simple port of what we have now over. Locally, I'm pointing to the actual branch so I will just update this to 1.0.0 when that PR is merged. I just want to avoid all the like project HASH nonsense that comes with package updating until we actually have the final release.

### Changes

We were getting the version from `JellyfinClient.sdkVersion` where now we need an instance of the `client` for this. As a result, we have a version value in the `AboutAppView` where a `userSession` & `client` will never exist as the user will be logged out at this point. TBH, not really much of a lose but that's why that got removed.

Otherwise, simple ports and a single `TranscodeProfile` change to reflect the new variable order.


LePips (approved):
We just no longer need to set the user id on requests.

I also missed that the placement for the info I do in the SDK didn't make the values static, but oh well.

PR #2010, opened 2026-05-09
Port over `QuickConnect`

JPKribs:
### Summary

Testing branch. Functional but I think I'm doing this completely wrong.


LePips:
During my work on the SDK I've already performed this work, don't worry about it.

Issue #1662, opened 2025-07-23
Autodiscovery not working on iOS 1.3.1 (15)


kunzese:
### Describe the bug

I just installed Swiftfin on my iPhone and noticed that the autodiscovery of local servers is not working. The Connect (Verbinden) screen just tells me that there are no local servers (Keine lokalen Server gefunden).

Using 'tcpdump -i any -s0 -A udp port 7359" on the server I was able to confirm that no udp pakets arrive on port 7359.

Local network permission has been granted to the app.

On my AppleTV the autodiscovery in the Swiftfin app works though.

### Application version

1.3.1 (15)

### Where did you install the app from?

App Store

### Device information

iPhone 16

### OS version

iOS 18.5 (22F76)

### Jellyfin server version

10.10.7

samglt:
In ServerDiscovery I see that the broadcast is failing with the error "UDP connection failed to send data: No route to host" which is mentioned in [this issue](https://github.com/gunterhager/UDPBroadcastConnection/issues/25) in the UDPBroadcastConnection package we're using as being possibly related to needing the multicast networking entitlement, which is needed even when debugging on a physical device. 

There may be another problem here as well though, because multicast works in the simulator but I can't discover my server there either.

JPKribs:
> needing the multicast networking entitlement

That _sounds_ right IMO.

I have a branch on the SDK to rework Server Discovery: https://github.com/jellyfin/jellyfin-sdk-swift/pull/47

This is a bit more robust and works in a standalone app but we appear to have the same issue using this one. Which makes me believe we're missing something.

On a somewhat related note, I am also unable to find servers on other iOS apps too. StreamyFin & Infuse both appear to have this issue as well for me. So could be device/network related.

samglt:
Doing a bit more testing, if I use the simple `receive_and_reply.py` server in UDPBroadcastConnection to manually reply with the discovery payload, Swiftfin will receive it and handle it properly (in the simulator). That means nothing's wrong with the app beyond needing the entitlement, and I just don't have my networking set up right.

JPKribs:
But, why does this work from the simulator? I guess that's my confusion the simulator works fine for me for 1.3 / 1.4 / Main but none of those work on real hardware.

samglt:
Because Apple's decided to limit it this way for some reason. From [an official article about multicast](https://developer.apple.com/news/?id=0oi77447):
[...]

PR #2011, opened 2026-05-10
Use `ServerDiscovery` from the SDK

### LINKED ISSUE #554: Replace UDPBroadcast

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/554

This is an in-place change to the `ConnectToServerViewModel` that replaces our old `ServerDiscovery` with the new one from the SDK. This is fully functional with a few caveats that will keep this in draft for a while.

1. Running on `Main` of the SDK. Will want to wait for a release for 1.0.1 or 1.1.0 or whatever numbering we land on.
2. `QuickConnect` is broken so this cannot build. I know this is covered in another PR so I will just merge that on top of this one when that is done. Since we are using a local implementation of `ServerDiscovery` using the old version of this isn't breaking. For testing, I have confirmed this is all working independently but `QuickConnect` changes should be done first just to avoid all the hassle.
3. For our next iOS/tvOS release this will not be usable until we address: https://github.com/jellyfin/Swiftfin/issues/1662. I believe this is just an entitlement request from Apple but I think this might need to be done by Anthony as he's the account holder. This PR isn't blocked by that work but this will only be usable in debug/simulator until that is resolved. 

- https://developer.apple.com/documentation/BundleResources/Entitlements/com.apple.developer.networking.multicast
- https://developer.apple.com/contact/request/networking-multicast

If we'd prefer to loop this into the other `QuickConnect` changes feel free to steal this and I can close this out! Primarily making this to have visibility.


LePips:
I've just done the other 2.0.0 SDK work. If you're good about everything else, I can approve and merge.

JPKribs:
> I've just done the other 2.0.0 SDK work. If you're good about everything else, I can approve and merge.

Sounds good! Tested and we should be good now. One tvOS build issue of `MediaURL` -> `NamedURL`. I was thinking about just unifying the `TrailerMenu` as part of this but I see that's part of [#1752](https://github.com/LePips/SwiftFin/blob/e7bae1d216677aa3915b12846333e574a3948278/Shared/Components/TrailerMenu.swift) so let's just handle this then haha

JPKribs:
Also, do you believe we're on the right track for: https://github.com/jellyfin/Swiftfin/issues/1662? That is, do you want me to ping Anthony about this?

LePips:
We'll worry about the provisioning elsewhere.

LePips (approved):

Issue #1784, opened 2025-10-28
Local Downloads - `DownloadManager` and Metadata Storage


JPKribs:
## Summary

_This is one part of the full [Swiftfin Local Downloads](https://github.com/jellyfin/Swiftfin/issues/57) feature that we would like to implement. The full feature does not require that all of the work be completed in a single PR, so I am creating individual Issues to track each part of the overall work that needs to be completed. This is to better allow contributors to divide and conquer rather than waiting on a single contributor to complete all of the work on their own._

Breaking this out from https://github.com/jellyfin/Swiftfin/discussions/364 into issues.

---

Finalize the existing `DownloadManager` & `DownloadTask` to handle queuing, downloading, and persisting content with full metadata and images. Builds on existing implementation of `DownloadManager` & `DownloadTask`. Where applicable, this solution should use the `@Stateful` Macro.

---

## Expectations

- Input from `BaseItemDto` to queue downloads
  - Handle single items and items with children (series/seasons/episodes or collections/children)
  - When downloading an item with parents (e.g., episode), capture **parent metadata and images** (season, series) along with the item itself
  - Ensure that all parent/children have key relationships to represent this hierarchy in code later
- Download media files with progress tracking
  - Support pause, resume, and delete operations
  - Maintain queue state across app restarts (SwiftfinStore), including paused/resumed state
- Store complete metadata in CoreStore
  - Save all item details for offline access
  - Metadata should fully populate offline views like `DownloadItemView` and support all `FilterViewModel` filtering in `DownloadPagingLibraryView`
- Download and store associated images alongside media files
  - posters, backdrops, and logos
  - The CoreStore record, on read, must conform to the `Poster` protocol
    - Like `BaseItemDto` but `DownloadItemDto`
- Organize downloaded files in a logical structure
  - Maintain hierarchy by content type (movies, series, seasons, episodes)
- Group downloads logically to support navigation in offline views
- **ItemView Download Button**
  - Add a download button to `ItemView`
  - Button should **only be visible** if `Experimental -> Enable Downloads` is `true`
  - Button behavior:
    - If not downloading → start download
    - If downloading → pause/resume
    - If errored→ retry download
    - If completed → delete download
  - Button should be minimal and basic and does not need to mirror existing Swiftfin design language (although it should if possible). UI/UX polish will be handled later. This is purely to prove that the queuing functionality works.

---

## Acceptance Criteria

1. Items queue correctly and download as expected.
[...]

PR #2012, opened 2026-05-12
Download Manager

JPKribs:
### Summary

**Resolves:** https://github.com/jellyfin/Swiftfin/issues/1784

This should meet the following items. I wrote these for the original post so it was also my gameplan for this as well:

- Input from `BaseItemDto` to queue downloads
  **- Handle single items and items with children (series/seasons/episodes or collections/children)** _<- This is missing as folder items are not set with `canDownload` from server._
  - Ensure that all parent/children have key relationships to represent this hierarchy in code later
- Download media files with progress tracking
  - Support pause, resume, and delete operations
  - Maintain queue state across app restarts (SwiftfinStore), including paused/resumed state
- Store complete metadata in CoreStore
  - Save all item details for offline access
  - Metadata should fully populate offline views like `DownloadItemView` and support all `FilterViewModel` filtering in `PagingLibraryView`
- Download and store associated images alongside media files
  - posters, backdrops, and logos
  - The CoreStore record, on read, must conform to the `Poster` protocol
    - Like `BaseItemDto` but `DownloadItemDto`
- Organize downloaded files in a logical structure
  - Maintain hierarchy by content type (movies, series, seasons, episodes)
- Group downloads logically to support navigation in offline views
- **ItemView Download Button**
  - Add a download button to `ItemView`
  - Button should **only be visible** if `Experimental -> Enable Downloads` is `true`
  - Button behavior:
    - If not downloading → start download
    - If downloading → pause/resume
    - If errored→ retry download
    - If completed → delete download

Outside of these, I can some stubs set for transcoded downloads although that's mostly just a flag for bitrate. For now this only takes `Maximum`.

### WIP

This is still WIP while I test and try to break this more. I just nailed down the manager/task pattern so I'm likely to find something that doesn't work as expected. I also need to figure out where the downloads end up. There is still a bit I need to figure out so mostly making this as a draft to avoid duplicate work.

### Videos

https://github.com/user-attachments/assets/2551fa0a-31dc-4e09-afee-7c4e3aa64a29

https://github.com/user-attachments/assets/9a8dc2a4-976b-4332-89f5-12b42c5e37d7

https://github.com/user-attachments/assets/d5a5aba6-e000-4603-9760-aa0476b7de56


LePips (commented):
I understand that working on this looks exciting, but I'm going to caution going too much further. I really need to take a look at the many other open PRs and I would have something like this be implemented after #1752 anyways, which would change the direction of some parts that I see are being worked on.

JPKribs:
> I understand that working on this looks exciting, but I'm going to caution going too much further. I really need to take a look at the many other open PRs and I would have something like this be implemented after #1752 anyways, which would change the direction of some parts that I see are being worked on.

Understood. This is kind of a thinking out loud branch.

For this part I'm primarily just looking to lock down the downloading/metadata side of this. I am expecting this to be a draft until #1752 is done. The paging part of this is really bare bones with the understanding this will need to be reworked post-1752.

I also think if "transcoded downloads" are a route in the future that's an SDK item... essentially, request HLS F-MP4 and then combine the results into a file. It's pretty easy to do but it introduces so many edge cases!

JPKribs:
This is merged on Main and functional but I am going to close this out. I'm going to be getting busy here soon and I don't believe it's feasible for me to get this over the finish line with the other items I want to do first before that point. Branch will keep existing for reference if anyone wants to pick this up but nothing here is that interesting / novel.

I think the primary smart (or lazy?) thing that I'm doing here is there is a `String?` field on `BaseItemDto` for `Status`, used for things like Live TV recording status, that will basically always be empty for a downloaded item. I am overriding this with `"Downloaded"` to indicate it's a downloaded item. I then use this in all the server calls to images/children to `if item.isDownloaded...` and using that to look at local files instead of server. This let me re-use the entirety of the existing `ItemLibrary` & `ItemView` based on type so like, the the shows would still use the same series `ItemView`.

This made sense to me but with `ContentGroups` I think there are other, potentially better, ways of doing this as well. The rest is pretty standard stuff tacked onto the existing download process. Again, branch available if needed and I'm happy to pick this up later but I don't want to hold this item up while I turn focus to some [work](https://www.reuters.com/business/grant-thornton-buy-professional-services-firm-cbiz-5-billion-cash-deal-2026-07-29/) items.

PR #2015, opened 2026-05-12
Fix autofill for sign in fields

xrvk:
The username and password fields on the sign in screen are missing `.textContentType` modifiers, so iOS never offers Keychain/password manager autofill. 

This PR adds `.textContentType(.username)` and `.textContentType(.password)` to the respective fields so iOS offers Keychain/password manager support

### DIFF
--- a/Shared/Views/UserSignInView.swift
+++ b/Shared/Views/UserSignInView.swift
@@ -122,6 +122,7 @@ struct UserSignInView: View {
         Section {
             TextField(L10n.username, text: $username)
                 .autocorrectionDisabled()
+                .textContentType(.username)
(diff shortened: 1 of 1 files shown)


JPKribs:
Thank you for your first contribution! TBH, I've never even considered that we don't have that auto fill so this is a nice QOL change.

~I'm out of pocket today. Looks like a very straight forward change but let me do a quick test this evening and get back to you!~

Scratch that, I have 15m I can look at this now. I just want to check this on real hardware since simulator is kind of funky with text input.

JPKribs (approved):
Tested and works as expected! Thank you again for your contribution!

PR #2017, opened 2026-05-16
General Cleanup | `ChevronButton` & Deprecations

LePips:
In preparation of progressively bringing work over from #1752.

- Cleans up `ChevronButton`
- Lots of other cleanup of dead code, my deprecations, and inits


Swiftfin tvOS/Views/HomeView/Components/CinematicRecentlyAddedView.swift
```diff
@@ -36,24 +36,23 @@ extension HomeView {
...
-                                .fontWeight(.semibold)
-                        }
-                        .edgePadding(.leading)
-                        .aspectRatio(contentMode: .fit)
-                        .frame(height: 200, alignment: .bottomLeading)
-                }
-                .onSelect { item in
-                    router.route(to: .item(item: item))
-                }
+            CinematicItemSelector(items: viewModel.elements.elements) { item in
```
JPKribs, line 39: This predates this PR but no background images are shows in the `HomeView` in either Recently Added or Next Up:

Image

More of an FYI than something I think needs to be covered here. Could also be my configuration but I see this with demo.jellyfin.org/stable as well.

Shared/Components/ChevronButton.swift
```diff
@@ -8,375 +8,235 @@
 
 import SwiftUI
 
-// TODO: Remove alert and just be a Button
-// TODO: change "subtitle" to "content"
-// TODO: observe editing/selection for checkbox/chevron switching?
+// TODO: possibly consolidate with ListRow
 
-struct ChevronButton<_Label: View, _Content: View>: View {
+struct ChevronButton<Label: View>: View {
```
JPKribs, line 13: What was formerly subtitle is now `.primary` so it looks a tad out of place. See "Days in Next Up":

Image

JPKribs (changes requested):
Looks great! Tested and works as expected trying to go through all the impacted areas. 1 minor visual change I wanted to bring up in case it wasn't intentional. 1 historical tvOS bug that doesn't need to be changes here but is a touched component so just FYI.

Swiftfin tvOS/Views/HomeView/Components/CinematicRecentlyAddedView.swift
LePips: I've known this happens, we'll figure that out elsewhere.

Shared/Components/ChevronButton.swift
LePips: Thanks for catching this, looks like `LabeledContent`s just need to set the style themselves.

JPKribs (approved):
Looks good! Renamed the PR just so it's a bit more distinct. Primarily to make the final release post more descriptive.

PR #2019, opened 2026-05-17
TVOS LiveTV Player Fixes

michaelmoskie:
## Normalize malformed media URLs for VLC playback.
- Normalize malformed URLs like .../stream%3F&DeviceId=... to .../stream?DeviceId=...
## Improve caching settings for live streams and network variability.
- Force a buffer for streaming content. 
## General player fixes/enhancements
- Update subtitle size computation to ensure it remains within a valid range.
-  Refactor playback controls to better manage live stream states and overlays.
- Hide overlay after live video playback starts.

### DIFF
--- a/Shared/Coordinators/Navigation/NavigationRoute/NavigationRoute+Media.swift
+++ b/Shared/Coordinators/Navigation/NavigationRoute/NavigationRoute+Media.swift
@@ -31,6 +31,14 @@ extension NavigationRoute {
         }
     }
 
+    static var liveTVGuide: NavigationRoute {
+        NavigationRoute(
+            id: "liveTVGuide"
+        ) {
+            LiveTVGuideView()
+        }
+    }
+
     static func mediaSourceInfo(source: MediaSourceInfo) -> NavigationRoute {
         NavigationRoute(
             id: "mediaSourceInfo",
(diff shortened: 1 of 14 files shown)


JPKribs:
Hi,

Many of these changes will need to come after a Player UI is finalized for tvOS. For that reason, I think these changes will need to wait until that is finalized as the current player is incomplete and will be replaced soon.

Many of these changes are manual changes to how we handle elements of the SDK. If you believe the SDK handles elements incorrectly, please submit an issue/pr on the jellyfin-swift-sdk itself as those changes should be handled there.

Based on the contents of this PR, this has a lot of the signs of AI code. Would you be able to confirm if AI was used in this PR in conformance with Jellyfin's AI Policy. https://jellyfin.org/docs/general/contributing/llm-policies/

JPKribs:
I am going to close this out. The core bug was resolved on the SDK via 2.1.0 & added to Swiftfin via https://github.com/jellyfin/Swiftfin/pull/2031. The rest of the changes in this will conflict with https://github.com/jellyfin/Swiftfin/pull/1902. I would ask that any work start after that PR or its equivalent is completed as that will override the most of the items in this branch.

LePips:
Well, some of the parts for the close and key presses could be done individually, if desired. (But this is just from me glancing at it)

PR #2023, opened 2026-05-19
Consolidate App Defaults Observation

LePips:
tvOS needed the defaults observation, also did cleanup while in the area.

### DIFF
--- a/Shared/App/SwiftfinApp+ValueObservation.swift
+++ b/Shared/App/SwiftfinApp+ValueObservation.swift
@@ -11,14 +11,10 @@ import Defaults
 import Factory
 import Foundation
 import SwiftUI
+import UIKit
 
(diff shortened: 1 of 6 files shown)


JPKribs (approved):
Changes make sense to combine. Testing this now, nothing appears to have changed and everything continues working as expected.

JPKribs (changes requested):
Sorry, caught this right at the end. On tvOS, toggles are no longer using `Color.primary` but instead using the `accentColor`.

LG
Screenshot 2026-05-19 at 17 43 49

iOS 16
Screenshot 2026-05-19 at 17 44 47

LePips:
Ah, thanks.

JPKribs (approved):
Looks good!

PR #2024, opened 2026-05-19
Add SwiftfinTests / SwiftfinTVOSTests targets

new-usemame:
`Swiftfin.xcodeproj` has no test target, so `xcodebuild test` is a no-op. This PR adds `SwiftfinTests` (iOS) and `SwiftfinTVOSTests` (tvOS), each with one test against `Duration.ticks`, and wires them into the shared schemes. The pbxproj/scheme diff is generated by `Scripts/add_test_target.rb` (uses the `xcodeproj` ruby gem).

Both schemes pass locally on iPhone 16e (iOS 26.2) and Apple TV 4K 3rd gen (tvOS 26.2).

Follow-up PRs will add coverage for shared logic.

### DIFF
--- a/Scripts/add_test_target.rb
+++ b/Scripts/add_test_target.rb
@@ -1,0 +1,154 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+#
+# Adds the SwiftfinTests (iOS) and SwiftfinTVOSTests (tvOS) unit-test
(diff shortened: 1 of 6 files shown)


LePips:
As stated before, please read Jellyfin's AI policy. Tests wouldn't be added in this fashion, nor would they use XCTest or have a script that generates the scheme - you're already providing the scheme in the xcproj.

PR #2028, opened 2026-05-20
Fix `ListRowMenu` when `isLiquidGlassEnabled`

JPKribs:
### Summary

Since tvOS and Liquid Glass is more rounded we need to adjust that for the custom `ListRowMenu` implementation. Super simple, just change the shape from rounded rectangle to capsule when `isLiquidGlassEnabled` is true and on tvOS 26.

Examples of before and after below. See `Letter picker`:

### Before

https://github.com/user-attachments/assets/a00b97bc-7484-477c-8a40-8b4df31bcde1

### After

https://github.com/user-attachments/assets/26d60e26-c386-4cf0-9a43-a06572f13709

### DIFF
--- a/Swiftfin tvOS/Components/ListRowMenu.swift
+++ b/Swiftfin tvOS/Components/ListRowMenu.swift
@@ -6,9 +6,13 @@
 // Copyright (c) 2026 Jellyfin & Jellyfin Contributors
 //
 
+import Defaults
 import SwiftUI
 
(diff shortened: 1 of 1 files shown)


JPKribs:
Here are the two buttons next to each other. This isn't perfect but it's pretty close while preserving the pre-Liquid Glass:

Screenshot 2026-05-19 at 21 14 15

LePips (approved):

PR #2037, opened 2026-05-25
No audio on iOS after alarm or timer goes off (issue #1975)

### LINKED ISSUE #1975: No audio on iOS after alarm or timer goes off.

### Description of the bug

When using VLCKit, audio playback stops after a timer or alarm goes off. The only way to restore audio seems to be to close the player and reopen it.

### Reproduction steps

1. Ensure Swiftfin is configured to use the Swiftfin video player type
2. Begin playing media with audio
3. Ask Siri to set a timer for 10 seconds (this does not disrupt audio playback)
4. After the timer goes off there will be no audio until the player is closed and reopened

### Platform

iOS

### Application version

Does Swiftfin show this somewhere? Anyway I can reproduce on the latest public or Testflight releases

TestFlight

### Device information

iPhone 15 Pro

### OS version

26.4.1 though it has been happening since before iOS 26

### Jellyfin server version

10.11.5

### Screenshots or videos

https://github.com/user-attachments/assets/b2c5b7e4-37bd-4862-a15a-e73a7e83cb33

[Session Logs.pulse.gz](https://github.com/user-attachments/files/26836710/Session.Logs.pulse.gz)

### Relevant log output

```shell
I've uploaded the session logs pulse file under 'screenshots or videos'; it doesn't seem to show anything relevant, but for what it's worth the issue I reproduced happened just a few seconds after 10:56:00 or so.
```

quangDecember:
fix issue #1975
the original `guard let` was too strict, remove so it accepts `AVAudioSessionInterruptionOptionKey`
resume playing on `.shouldResume`

### DIFF
--- a/Shared/Objects/MediaPlayerManager/NowPlayable/NowPlayableObserver.swift
+++ b/Shared/Objects/MediaPlayerManager/NowPlayable/NowPlayableObserver.swift
@@ -183,6 +183,7 @@ class NowPlayableObserver: ViewModel, MediaPlayerObserver {
                 if playbackRequestStateBeforeInterruption == .playing {
                     if options.contains(.shouldResume) {
                         manager?.setPlaybackRequestStatus(status: .playing)
+                        manager?.proxy?.play()
                     } else {
                         manager?.setPlaybackRequestStatus(status: .paused)
                     }
--- a/Shared/Services/Notifications.swift
+++ b/Shared/Services/Notifications.swift
@@ -230,12 +230,8 @@ extension Notifications.Key {
             else {
                 return nil
             }
-            guard let optionsUInt = userInfo[AVAudioSessionInterruptionOptionKey] as? UInt
-            else {
-                return nil
-            }
-
-            let options = AVAudioSession.InterruptionOptions(rawValue: optionsUInt)
+            let options = (userInfo[AVAudioSessionInterruptionOptionKey] as? UInt)
+                .map(AVAudioSession.InterruptionOptions.init(rawValue:)) ?? []
 
             return (type, options)
         }


Shared/Services/Notifications.swift
```diff
@@ -230,12 +230,8 @@ extension Notifications.Key {
             else {
                 return nil
             }
-            guard let optionsUInt = userInfo[AVAudioSessionInterruptionOptionKey] as? UInt
-            else {
-                return nil
```
quangDecember, line 235: FYI  @JPKribs  after a timer, userInfo here having 2 elements

```
[AnyHashable("AVAudioSessionInterruptionReasonKey"): 0, AnyHashable("AVAudioSessionInterruptionTypeKey"): 1]
```

a.k.a no `AVAudioSessionInterruptionOptionKey` here, so it wasn't processed

JPKribs: Thank you for the insight! To me, this all looks good but I'll defer to Ethan for a final review in case there is something I'm missing

Shared/Objects/MediaPlayerManager/NowPlayable/NowPlayableObserver.swift
```diff
@@ -183,6 +183,7 @@ class NowPlayableObserver: ViewModel, MediaPlayerObserver {
                 if playbackRequestStateBeforeInterruption == .playing {
                     if options.contains(.shouldResume) {
                         manager?.setPlaybackRequestStatus(status: .playing)
+                        manager?.proxy?.play()
```
LePips, line 186: The proxy already receives the play command through `setPlaybackRequestStatus`.

LePips (changes requested):

Shared/Objects/MediaPlayerManager/NowPlayable/NowPlayableObserver.swift
quangDecember: we need that to force the VLC player to play again. Delete that and the problem happens again unfortunately.

LePips: That's fine, I had not experienced the problem with this removed. I may just change `setPlaybackRequestStatus` to just always call play/pause anyways.

LePips (approved):

PR #2040, opened 2026-06-01
Add session message support

u61d:
## Summary
- Advertise Swiftfin iOS sessions as supporting Jellyfin `DisplayMessage` session commands over the websocket
- Show incoming session messages as Swiftfin toasts
- Add a Swiftfin admin active-session message sheet that calls Jellyfin's existing send message API

## Context
Jellyfin already supports session messages through the server API and websocket `GeneralCommand` flow. Jellyfin Web and Android TV use this capability-based flow today. This PR adds the missing Swiftfin side so active iOS sessions can appear as message-capable and receive/display those messages.

No Jellyfin Web or server change is expected for this PR: Web/server already provide the admin message UI/API path, and this change makes Swiftfin advertise and handle the command.

## Validation
- `swiftformat . --lint --config .swiftformat`
- `swiftlint lint --strict --config .swiftlint.yml`
- `git diff --check`

Note: I could not run the iOS/tvOS Xcode build locally because this was developed on Linux without Xcode.


JPKribs (changes requested):
I don't know how feasible this is PR is without you having the ability to test this on your end. It's almost impossible to develop something blind like this and not miss something. The end result is this means our team has to do a lot of the leg work to validate on your behalf. Most of the items below I believe would be easily caught if you could test yourself.

Some notes on first glance:

1. We'd want a socket manager. The socket is a stream which is a single consumer only. As a result, this implementation means we can only do display messages. If I try to add anything else to the socket this results in crashes. So the socket needs to be consumed more generically and holistically for the app as a whole.
2. You've subscribed for media playback controls. This is required for messages but now Swiftfin is advertised as controllable from web. This means I can add Swiftfin to a remote session and it appears to work as a remote but all commands are unusable. For this reason, I don't think we can add messages without doing all elements of playback controls. Otherwise, this UX is confusing from other devices.
3. Socket currently fails to reconnect if the app backgrounds too long, network times out, or the server restarts. Once the socket dies this can't be resumed without restarting the app.
4. Instead of a send session message view model the way I see this is a rework so active sessions view model contains N session view models based on active sessions. Each session view model contains the actions needed so messages, send remote commands, etc. This helps make this reusable for remote control of another device or, eventually, sync play. This version would be heavily reliant on a socket manager or observer that handles socket items in a centralized spot.
5. More just thinking out loud. I know some reverse proxy/network setups mean the socket is unavailable. Do we need to do any gating or handling if the socket never connects? Something like sending messages this doesn't matter but for other elements like playback commands we will want to unsubscribe if the socket is inaccessible.

u61d:
Thanks for the review, that all makes sense.

I didn't realize the socket stream was single-consumer, so I see why doing this as a standalone message observer is the wrong shape. The remote-control point also makes sense. If Swiftfin advertises media control support, then Web making it look remotely controllable while most commands don't work would be confusing.

I'm fine with changing direction here. Would you prefer that I convert this PR to a draft and try reworking it around a central socket manager, or should I close this and come back later with a smaller groundwork PR for socket handling first?

Also fair point about testing. I'm working from Linux without Xcode, so I understand that this puts more validation work on maintainers than ideal.

LePips:
At the very least within the project, I'll only accept PRs that are able to test themselves.

PR #2042, opened 2026-06-01
Implement a `SocketManager`, Migrate `ActiveSessions` over to the Socket, & set up some `SessionCommands`

JPKribs:
### Summary

I've been sitting on this one for a while as part of another branch but https://github.com/jellyfin/Swiftfin/pull/2040 motivated me to split this out now.

This creates a `SocketManager` that lives in the `UserSession`. In testing, I found there were some items that I wanted to address. Let me know if these belong on the SDK instead but my understanding is the SDK should be a super simple stream and any like re-connection logic or multi-consumer publishing should be handled at the app level.

While testing, I found a lot of things that could break our connection. As a result, this is what the `SocketManager` does to try and be robust:

1. Reconnect on failure
2. Reconnect when going from background to foreground
3. Reconnect when switching networks like wifi to cellular

This is something I pulled from a Stack Overflow post to track the network changing:

```swift
private func observeNetworkPath() async {
let monitor = NWPathMonitor()
defer { monitor.cancel() }

let stream = AsyncStream<NWPath> {
    continuation in monitor.pathUpdateHandler = { continuation.yield($0) }
}

monitor.start(queue: .global(qos: .utility))

var previous: NWPath.Status?

for await path in stream {
    if path.status == .satisfied, previous == .unsatisfied {
        logger.debug("Reconnecting the socket (Network Restored)")
        reconnect()
    }
    previous = path.status
}
}
```

In testing, this works but please let me know if there is a better way to handle this instead!

Some issues I ran into that I wanted to build out here:

1. A stream can only have a single consumer so I am consolidating the stream into a publisher to ensure that we can consume this from multiple views at the same time. This is important because, if I loaded and de-loaded the `ActiveSessionsView` really quickly I could create 2 streams that caused issues. The published event is robust and allows multiple readers and ensure that we only ever have a single socket instance.
2. I am using `OSAllocatedUnfairLock` as partially some Swift6 logic but mostly that I was getting some crashes socket data updated and tripped over itself. This locking resolves issues but please let me know if `OSAllocatedUnfairLock` was the incorrect way of doing this.

### Fluff

Some of what is in this PR is fluff that I want to use later. Specifically, the `GeneralCommandType` and `PlaystateCommand` conformance to `Displayable` and `SystemImageable` are only used briefly here but over fully built out with the idea that a Swiftfin [`RemoteView`](https://github.com/jellyfin/Swiftfin/issues/408) would use this. I think that would be really easy to do from here but I wanted to keep this PR from ballooning since getting a clean `SocketManager` first should make future socket stuff easier.

One change I made to `BaseItemDto+Images` is to use more generic parent and child logic instead of series. The reason I did this is because Active Sessions for music was weird as most album art is going to live at the album level and we were just looking at the song's primary art. We get the song's primary art first but roll up to the album if the song doesn't have art as primary. 

I also remove the insets around the playback progress to match the header section better. On the fence over if this is better or not so no offense taken if we don't want this:

Screenshot 2026-06-01 at 16 34 10

### Usage

To start, I wanted this to be used in something small. This ended up a lot bigger than I intended. This PR updates the `ActiveSessionViewModel` to utilize this socket. Since the socket doesn't allow filtering, that logic has been moved locally. I made the `SessionDto` `Comparable` to centralize that logic a bit. The box has been replaced by a `SessionViewModel`. This is very simple and still takes the parent for updates but this view model is used for session commands and will one day be home to the remote commands I mentioned above. For now, I've just used the pause/play, stop, and message commands to mirror Jellyfin-Web for session details actions.

### Video

Video cuts off a hair short to stay under the 100mb limit...

https://github.com/user-attachments/assets/c330e2a8-179e-49ad-9523-3b45f59d5cf4


Shared/Services/UserSession.swift
```diff
@@ -14,6 +14,7 @@ import Pulse
 final class UserSession {
 
     let client: JellyfinClient
+    let socket: SocketManager
```
LePips: After #2045, conform the `SocketManager` as a `UserSessionService`, and register it with the session's services.

LePips (changes requested):

Shared/Services/ServerSocketManager.swift
```diff
@@ -0,0 +1,280 @@
...
+        delay: Duration = .seconds(0),
+        interval: Duration = .seconds(5)
+    ) -> AnyPublisher<[TaskInfo], Never> {
+        publisher(for: .scheduledTasks, delay: delay, interval: interval) { event in
+            guard case let .message(.scheduledTasksInfoMessage(message)) = event else { return nil }
+            return message.data
+        }
+    }
+
+    private static func publisher<Payload>(
```
JPKribs: @LePips Everything is tested and working well. I am happy with where I ended up in the main part of this but I am a bit stuck here.

We live on the `UserSession` but since this is static I am calling `Container.shared.currentUserSession()?.serverSocketManager` which I'm pretty sure is very incorrect lol.

I'm a bit caught on how to do this correctly. The reason I want these subscriptions here is because we use `Sessions` on both the `MediaPlayerManager` AND `ActiveSessionsViewModel` so I wanted to centralize the logic for reconnect on Network / Foreground Change once instead of calling it on each instance. I also wanted to make this an easier "Subscribe to new `SessionInfoDto` updates only" kind of call.

I've tried keeping this on the `ServerSocketManager` but the lazy var causes some headaches where it can attempt to load before it's ready and the view models that use have to exist and re-enter for this to kick off.

Happy with everything down stream of this but I just want to make sure I'm not building on top of a bad foundation.

Any insights would be helpful!

JPKribs:
This is "Ready" but more in the sense that I'm done with what I think this should look like. That being said, any insights/critique into structure would be great! The socket is functional here but if this is completely off just let me know!

Swiftfin/Views/AdminDashboardView/ActiveSessions/ActiveSessionDetailsView/ActiveSessionDetailsView.swift
```diff
@@ -0,0 +1,216 @@
...
+struct ActiveSessionDetailsView: View {
+
+    @Router
+    private var router
+
+    @ObservedObject
+    var viewModel: SessionViewModel
+
+    private var isPaused: Bool {
+        viewModel.session.playState?.isPaused ?? false
```
LePips: `== true`

Shared/Services/Notifications.swift
```diff
@@ -113,6 +113,10 @@ extension Notifications.Key {
         Key("didChangeServerConnection")
     }
 
+    static var didChangeNetwork: Key<Void> {
```
LePips: We shouldn't have decoupling of the server connection and the current network path.

Shared/Services/ServerSocketManager.swift
```diff
@@ -0,0 +1,286 @@
...
+            } else {
+                logger.debug("Socket waiting for signal")
+                _ = await wakeIterator.next()
+            }
+        }
+    }
+
+    /// Reconnect when the app returns to the foreground.
+    /// Fixes the socket when it expires in the background.
+    private func observeForeground() async {
```
LePips: I believe that with the publishers also observing foreground entrance and connection changing that the reconnections are being called twice.

Shared/Services/ServerSocketManager.swift
LePips: > I've tried keeping this on the ServerSocketManager but the lazy var causes some headaches where it can attempt to load before it's ready and the view models that use have to exist and re-enter for this to kick off.

I'm not seeing any loading issues with moving these to instance functions rather than being static.

LePips (changes requested):

Shared/Services/ServerSocketManager.swift
JPKribs: I should've retested this with the factory changes. I think I was doing something weird pre-factory with the UserSession but post those changes this would've been fine without static.

Shared/Services/ServerSocketManager.swift
JPKribs: In testing, I believe I now have this covered using what you recommend. Please let me know if this doesn't look how you were expecting!

JPKribs:
Spent a lot of today testing. I believe this is in a good stage (minus the merge conflicts) but I think I'm going to sit on this for a week just to be certain I'm not missing any weird edge cases. It's hard to intentionally test all the weird network situations so some time will hopefully help me build some confidence that I'm not missing something.

Any insight into structure is still very welcome! I tried to mirror the existing UserSession and connection manager since I would like this to slot into that ecosystem but this is new territory to me so getting this right takes priority haha

LePips (approved):

PR #1870, opened 2025-12-21
`BaseItemDto` Editing Cleanup

JPKribs:
## Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1869

While trying to resolve #1869, I wanted to convert over the `ItemEditorViewModel` and it's extensions to use the new `@Stateful` macros. This led me to the views themselves which I created about a year ago. With what I know now, there is some room for improvement. The primary changes made:

1. Move all Item Permissions from the `UserSession` to the `BaseItemDto`. The current usage of this was pretty bulky and required getting the permissions from a separate location using the item as an input. Instead, this just puts the permissions directly on the `BaseItemDto` since that's the only place where these can be used anyway.
2. Hide the `PlayButton` if the user does not have permission to play content from Jellyfin. I don't imagine this comes up often. I am only aware of this because I have an admin account that doesn't have playback permission and this button throws a non-descriptive error from the server. Hiding the button seems like the cleaner solution.
3. Unified `ItemEditorMenu` struct. The menu dropdown items are the same between iOS and tvOS (kind of) so this lets us keep all the permission logic and routing in a single spot for later.
4. A more detailed `ItemDeletionView`. Especially for Series, it can be hard to determine what specifically you are deleting. This gives the full count of items deleted, and the type/details. This should help prevent any heartache where someone thinks they are deleting an episode but delete a series.
5. For the `EditItemElementViewModels`s, I've changef this from using a trie to instead just do a search and debounce. This is more performant for items where searching is available. Tags cannot do this so that will still use the trie.
6. `EditItemElementView`s, was broken where I could re-add the same genre, tag, studio, or person unlimited times. This blocks adding duplicates to the same item now.
7. Moved `AdminDashboardView/Components/MediaItemSection.swift` -> `/Components/ItemFormSection.swift` since this is used outside of the `AdminDashboard` now. It's a good way to display `BaseItemDto` in a form so this helps with re-use.
8. Removed the progress on `RefreshMetadataViewModel` since we don't have the progress anymore. Eventually, most of this will be deleted since the websocket will be able to get the refreshed item instead of use counting to 5 and manually refreshing.
9. The ConnecToServerView -> ConnectToServerView typo fix.
10. Instead of doing many permission checks throughout the `ItemEditorView` I'm just purely leaving that as admin edit actions and moving subtitles (non-admin) and refresh Metadata functionality to the dropdown. It makes the permissions a bit easier to handle and centralizes the tvOS and iOS permission checks.
[...]

PR #2045, opened 2026-06-09
UserSession, ServerConnection, Permissions, DeepLink

### LINKED ISSUE #508: Home network detection and remote URL switching

LePips:
- Migrates UserSession handling to new system to no longer do forced unwrapping, which is a cause for a lot of crashes
	- Playing around with deep linking to at least verify how we would handle auth and item launching
- ServerConnection to represent connections to a server
	- Move from single URL selection to allow fallbacks, ordering, WiFi vs cellular, specific WiFi connections
	- Keeping as Experimental for now, as the switching needs some more work
	- Closes #508
- Permissions page to better handle permissions used in the app
	- Location is required for the WiFi SSID
	- Link to Settings for permissions that can't be manually allowed (local network)


JPKribs:
@LePips is there anything in particular you are that you would like me to focus on testing for a sanity check?

LePips:
If you would like to you can try a bit with the auto switching between Wi-Fi and cellular or Wi-Fi SSIDs with location permission enabled/disabled. However, I did leave the option as experimental for now to gather more input from the public if there's other fixes needed.

JPKribs:
Working on testing now. Getting caught up on this:

Screenshot 2026-06-13 at 00 27 29

but I should be able to get this reviewed soon!

Screenshot 2026-06-13 at 00 45 13

JPKribs:
Nevermind.... I will get back to you *sometime* this weekend:

Screenshot 2026-06-13 at 00 48 55

JPKribs (approved):
We're live! This works as expected but IMO I think we can drop it from experimental in favor of a disclaimer under the setting in `Connections` view. I spent longer than I care to admit thinking this was broken because I enabled this under Experimental thinking it enabled it altogether. I think just using the orange ! disclaimer under it there might be a better experience for testing. That being said, this worked exactly as expected when I actually had the setting turned on...

Is it possible to handle this with multiple WiFi networks? I see that we can set up multiple records pointed at the same URL so this works as well but it may be preferably to have a single like `LocalURL` and `WiFi #1`, `WiFi #2` rather than making 2 records for it. Certainly not a dealbreaker at all but [WireGuard](https://apps.apple.com/us/app/wireguard/id1441195209) has multiple URLs per record.

The other item of note, are we able to pull the current SSID? That's another item I see in WireGuard that would be nice if we could do. Not a major item at all but a nice QOL change if it's possible. That way, if you're on the Wifi for that server, it's pre-populated. Again, nothing worth holding this PR up over but would be a nice win if it was readily available.

All else, I attempted to look at all the places this touched and I didn't find anything off! Looks like good changes and I really attempted to break this with permissions. I agree the previous permissions usage was fragile and clearly [that person](https://github.com/jellyfin/Swiftfin/pull/1870) wasn't handling the force unwrap correctly! Jokes aside, this seems much more robust and didn't break when I attempted warp the `UserSession` in weird ways which I am able to get to break on 1.4.1 and current Main so this is a huge stability improvement!

I think this is definitely good to merge in my testing. 2 QOL changes but those can move to other PRs as well.

JPKribs:
@LePips I'll defer to you on the merge. 2 QOL change ideas we can roll into this, put in another PR, or skip. I'll leave the judgement up to you!

JPKribs (changes requested):
Just wrapping up and I found this:

If I add a trailing / to a URL in `Connections` saving crashes. 

https://github.com/user-attachments/assets/c5683cfb-aaae-461d-9885-249a4c5f09ce

Internal, external, FQDN, and IP all have this issue.

Shared/Extensions/String.swift
JPKribs, line 126:
```suggestion
guard s.isNotEmpty else { break }
s.removeLast()
suffix.removeLast()
}
```

JPKribs: This resolved the trailing / issue:

```swift
    func trimmingSuffix(_ suffix: String) -> String {

        guard suffix.count <= count else { return self }

        var s = self
        var suffix = suffix

        while s.last == suffix.last {
            guard s.isNotEmpty else { break }
            s.removeLast()
            suffix.removeLast()
        }

        return s
    }
```

JPKribs (changes requested):

LePips:
> Is it possible to handle this with multiple WiFi networks?

Good idea, I've made the underlying Wi-Fi SSID storage an array to allow multiple networks but right now only surface adding for one for now.

> pull the current SSID

This is what happens with the `currentWifiSSID` network population. Just have to ensure your development profile has the correct permissions.

I'm still going to want to keep this as experimental for now. I still have to consider a more robust connection manager for things like offline checking/mode, how we would properly deal with media playback, and ensuring some better UI.

JPKribs:
> This is what happens with the `currentWifiSSID` network population. Just have to ensure your development profile has the correct permissions.

Ah yes you are right! I think the first time I didn't have locations enabled so that makes sense that didn't work. On a fresh install, this worked

JPKribs (approved):
Just re-tested for the network switching for the parts since the last review. All looks good!

Genuinely, this is such a cool addition and I'm really happy with how smoothly this switching works! Great work on this, this is a tremendous QOL change for Swiftfin!

JPKribs:
@LePips when you get a chance, do you think we could get a TestFlight? I think this change plus the iOS side of the player changes would be good to get out for testing. Plus, I think we're down to ~30d on the current build

LePips:
I want a few more things in the build, I'm watching the build expiration.

Swiftfin/Resources/Info.plist
```diff
@@ -23,7 +23,7 @@
 			<string>Editor</string>
 			<key>CFBundleURLSchemes</key>
 			<array>
-				<string>jellyfin</string>
+				<string>swiftfin</string>
```
Cyberbeni, line 26: Why was this changed? It broke my home dashboard.

I think the jellyfin scheme should be kept, so that it can be used universally across platforms and (3rd party) apps. (And the swiftfin scheme can be kept for people who specifically want to open Swiftfin from several apps)

@JPKribs @LePips

LePips: While the previous tvOS release had this deep linking support, I had not considered to be complete. I had changed to `swiftfin` because that's the app, but am open to re-adding `jellyfin`.

Cyberbeni: I'm only using it to open the app currently. Also, while nobody is using my dashboard with an Android device currently, it would be nice if the same url scheme worked for both platforms (even if it just opens the app on one of them).

Issue #1968, opened 2026-04-08
The search displays only 20 results.


tobylibo:
### This issue respects the following points:

- [x] This is a **bug**, not a question or a configuration issue; Please visit our [forum or chat rooms](https://jellyfin.org/contact/) first to troubleshoot with volunteers, before creating a report.
- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/Swiftfin/issues?q=is%3Aopen+is%3Aissue).
- [x] I'm using an up to date version of Swiftfin; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).

### Description of the bug

English:
1. When using the search function, only 20 search results are displayed.
2. When selecting an actor, only the 20 films starring that actor are displayed.

中文：
1. 使用搜索时，只显示20个搜索结果。
2. 选择演员时，也只显示该演员出演的20个影片。

### Reproduction steps

1. Open Swiftin and Search with keywords, "boys"
2. Only the 20 movie titles containing "boys" will be displayed.
3. Or...
4. Select a movie, view the cast list, and select an actor, “Leonardo DiCaprio”.
5. Only the 20 movie actors containing "Leonardo DiCaprio" will be displayed.

### Platform

iOS

### Application version

1.5 (1)

### Where did you install the app from?

TestFlight

### Device information

iPhone 12 Pro

### OS version

iOS 18.7.1

### Jellyfin server version

10.11.8

### Relevant log output

```shell

```

JPKribs:
This ties to the same root cause as https://github.com/jellyfin/Swiftfin/issues/1890. I believe this is resolved as part of https://github.com/jellyfin/Swiftfin/pull/1752.

LePips:
Closing from #2075.

PR #2047, opened 2026-06-13
Generic Paging Libraries

### LINKED ISSUE #1705: Library Environment
As POC'd in #1677, we should allow libraries to define values that can be defined on themselves (like a shows vs episodes grouping) which can then be passed when retrieving items. The final solution should also cleanup with filters and the setting should be stored.

POC commit: https://github.com/jellyfin/Swiftfin/compare/main...LePips:SwiftFin:library-grouping-poc

### LINKED ISSUE #1890: Collections doesn’t display more than 20 records

### Description of the bug

I have a custom collection with around 60 items but i can only see 20 items in swiftfin.

### Reproduction steps

1. Open swiftfin
2. open collections having more than 20 items

### Platform

iOS

### Application version

1.4.1 (2)

TestFlight

### Device information

iPhone 17

### OS version

Ios 26.2

### Jellyfin server version

10.11.5

### Screenshots or videos

_No response_

### Relevant log output

```shell

```

### LINKED ISSUE #1896: Use Folder view for Home Videos and Photos libraries

### Platform

iOS, tvOS

Currently, Swiftfin displays libraries of type "Home Videos and Photos" as a flat list. I think it would make more sense for these libraries to always (or at least by default) be displayed in a Folder view.

### Rationale

Libraries of this type are naturally less suited to metadata tagging. Users are, of course, free to tag their content however they wish, but naturally, this kind of content is not going to appear on any centralized metadata DBs, and I think it's a safe bet that the majority of users with this library type are not tagging the files in them manually, either. As such, the most likely scenario is that if users organize these libraries at all, it will be on the filesystem level.

The official web UI also does it this way—and while that in itself is not a strong argument, I would hazard a guess that _the reason_ it does it this way for the reason laid out above.

#### My ulterior motive

I abuse this library type for content which may or may not have been released commercially, but is obscure enough to not really qualify for tmdb, omdb, or any of the other metadata sources Jellyfin scrapes from. And I don't really want to manually tag all the content in this library, so I organize the media into folders and call it a day—except it's currently terrible to navigate in Swiftfin, which is the only roadblock to me switching from away from the Jellyfin app on iPhone.

### Related issues

* #733

LePips:
Break out of library work in #1752.

- Libraries are now generically defined by their data source, while the actual paging and element state is owned by the object over the library. The mutable "environment" that a library may want is defined by its own `Environment` object, allowing things like value groupings. These aren't stored for now.
- Not coming over from my PR is additional functionality I've thought about for library styling, where elements can filter what their supported styles are. For example, movies can support portrait and landscape while people can only support portrait. Also got carried away consolidating the item value editing (genres, people, etc.) into a similar `ItemComponentEditor` protocol style and all of the library views.
- Implemented a poc for searching within a library, not surfacing for now.
- Fixed cinematic background focusing with tvOS libraries, unsure when that was lost.
- Media and channels are now just implementations of `PagingLibrary`, no longer requiring their own views.
	- `RemoteImageSearchView` should also be migrated, but yet to figure out the multi-library handling.

---

- Closes #1896
	- I've at least verified proper subfolder navigation throughout the library views, regardless of media type
- Closes #1705
- Closes #1890


LePips:
@JPKribs For review, you can of course look at the code for anything you think I've missed. Overall, just ensure that library styling and functionality is preserved and the fixes and new functionality I've described works.

JPKribs:
It looks like this resolves:

- https://github.com/jellyfin/Swiftfin/issues/1890

<details>

<summary>Test Collection</summary>

https://github.com/user-attachments/assets/ea460edb-c077-4e22-86f5-494cafbe2bd8

</details>

~This one still is limited by 25. I didn't know if you wanted to wrap that one into this since I think that's just our search isn't including in the pagination. That being said, I don't think it needs to be included in this PR:~

- ~https://github.com/jellyfin/Swiftfin/issues/1968~

I see in the code this is intentionally set as a `StaticLibrary`. If this is intentional, I still believe we should close out 1968 as Not Planned instead.

LePips:
Good catch, this will close out the collections too.

> as a StaticLibrary. If this is intentional

I think I'll actually close this out when I migrate over my content group work, as the search view will be redone with that.

JPKribs (changes requested):
Sorry for the delay! I went through and tested all of the impacted views. I have 1 real quirk and then I have a couple of items just to clarify:

### Issue

1. Activity logs are loading profile pictures & names weirdly. Like, once the image loads we're fine but on first load it defaults everything to system but then it changes it once it's loaded in:

https://github.com/user-attachments/assets/70ea49d4-20c3-40b7-a756-972c19975e87

### Confirmation

These items are not issues or require fixes. More just FYI kind of stuff in case they're unintentional or unplanned:

**a)** On 1.4, you can save the poster type and list type per-library. Now, this saves it for all libraries. Not an issue, just wanted to confirm this was intentional.
~**b)** Per my other comment, is https://github.com/jellyfin/Swiftfin/issues/1968 intentional behavior? If so, we should close it out.~
~**c)** Same comment, this definitely resolves https://github.com/jellyfin/Swiftfin/issues/1890 unless you see a reason not to include this to the PR?~
**d)** tvOS library rows clip to rectangles now when focused. Also not an issue but most other items are rounded so it stands out a bit:

Image

---

I don't think there were issues made for these but just for kudos / record keeping:

You also fixed the tvOS cinematic issue where it would adjust when focusing items NOT in the cinematic row! So like, if I went to the second row I could see the cinematic background changing in the line above.

You also seem to have fixed the `CollectionVGrid` issue where posters without set sized would blow up. Most notably, at the end of an image search for an item:

Image

Left is this PR which retains the shape. Right is 1.4.1 which would randomly have the occasional item that explodes out like this. So that's a nice improvement as well!

---

Outside of the 1 real issue, this looks and works great! I love how much cleaner this all is for set up now!

LePips:
Thank you for your thorough review. The activity bug was coming from the change of `ServerActivityViewModel` to plain libraries and the users were fetched after the logs.

I couldn't figure out the tvOS list clipping locally, I'll spend some time later adjusting tvOS's library UI specifically with that included.

JPKribs (approved):
Looks great!

Issue #2050, opened 2026-06-14
CollectionHStack layout performance with PosterButton


yoveto:
I was profiling the home page and noticed there are some expensive layout loops related to the use of CollectionHStack. Do we have plans to address those performance issues longer term? In my local testing I found that if we directly set a single frame dimension on the inner cells the auto layout logic in swift/uikit does a much better job here.

JPKribs:
Could you please attach and screenshots/files related to your findings on this?

LePips:
I know the performance issues you're mentioning as I've been running through the profiler lately.

I do have some fixes to hopefully address them.

yoveto:
@LePips Would you mind sharing some details? (also I'd be happy to help!) I spent a few hours today debugging the collection library and would like to compare notes. 

I did wind up hacking together a standard swift ui only implementation here which address all the issues (and is much more friendly with the liquid glass layout as a bonus): https://github.com/yoveto/Swiftfin/tree/homepage-perf

LePips:
Apologies, I did not see your message. Some of the details included:

- CollectionHStack was too eager to re-evaluate view bodies as the context around it changed
- From that, reading global poster settings from Defaults (essentially disk) on every poster button made it worse

I have since introduced improvements for poster settings (now available in the environment) and performance improvements in CollectionHStack/CollectionVGrid.

As a note, the SwiftUI implementation is not correct for this project. We need to display many items and properly dequeue (not just lazily load) their views with a `UICollectionView`.

PR #2051, opened 2026-06-14
liquid glass supporting changes

yoveto:
## Summary

The PR adds the following changes to support liquid glass in the UI:

- iPadOS only: adaptable side bar
- all: liquid glass play/fav/mark watched buttons
- various layout fixes for liquid glass enabled components (esp when sidebar is enabled from the adaptive style)
- various nav bar buttons have been liquid glass-ified
- the item view page has been updated to fully support liquid glass

Unfortunately this did require a bit more branching than I'd like so if anyone has strong push back against the if avaliable usage lmk and we can follow up.

## Known issue:
- there is a small pop in effect under transparent surfaces in the homepage when the adaptive side bar is enabled. This is a side effect of the bug noted in https://github.com/jellyfin/Swiftfin/issues/2050
- on iphone when entering the PagingLibraryView when liquid glass is enabled, the collectionview is clipped to bounds above the bottom tab bar. This also relates to https://github.com/jellyfin/Swiftfin/issues/2050

Given this is all gated behind the debug menu I think this is reasonable. (and work is in progress to address the bug on a separate PR)

## Testing
Validated both with liquid glass enabled and disabled in debug menu on iPad Pro v26.5 and iPhone 12 on 26.3.1

## AI Usage
None. mistake are my own.


Shared/Coordinators/Tabs/MainTabView.swift
```diff
@@ -60,32 +64,60 @@ struct MainTabView: View {
...
+        if isLiquidGlassEnabled, #available(iOS 26.0, tvOS 26.0, *) {
+            TabView {
+                ForEach(tabCoordinator.tabs, id: \.item.id) { tabData in
+                    if tabData.item.id == "search" {
+                        Tab(role: .search) {
+                            NavigationInjectionView(
+                                coordinator: tabData.coordinator
+                            ) {
+                                tabData.item.content
+                            }.environmentObject(tabCoordinator)
```
LePips, line 76: Just upon an initial skim, please lint/reformat modifier usage on separate lines.

LePips (changes requested):

Shared/Coordinators/Tabs/MainTabView.swift
yoveto: ah my bad, I am to used to leaning on the formater. Should be fixed up now. Do you recommend following a specific style guide for this project? (hopefully can save some back and forth)

Sorry for the slow turn around time, busy start to the week

Swiftfin/Extensions/View/Modifiers/NavigationBarCloseButton.swift
LePips, line 1: iOS 26 button just needs to be:

```swift
Button(role: .close, action: action)
    .disabled(disabled)
```

h3pdesign:
Please get it done to implement this change for Liquid Glass.

Swiftfin/Extensions/View/Modifiers/NavigationBarCloseButton.swift
```diff
@@ -23,10 +26,15 @@ struct NavigationBarCloseButtonModifier: ViewModifier {
                 Button {
                     action()
                 } label: {
-                    Image(systemName: "xmark.circle.fill")
-                        .fontWeight(.bold)
+                    Image(systemName: isLiquidGlassEnabled ? "xmark" : "xmark.circle.fill")
+                        .fontWeight(isLiquidGlassEnabled ? .regular : .bold)
                         .symbolRenderingMode(.palette)
-                        .foregroundStyle(accentColor.overlayColor, accentColor)
+                        .if(!isLiquidGlassEnabled, transform: { img in
```
LePips, line 32: These 2 conditionals resolve the same conditional.

Swiftfin/Views/ItemView/Components/ActionButtonHStack/ActionButtonHStack.swift
```diff
@@ -51,59 +54,90 @@ extension ItemView {
...
-                    Button(L10n.played, systemImage: "checkmark") {
-                        viewModel.send(.toggleIsPlayed)
-                    }
-                    .buttonStyle(.tintedMaterial(tint: .jellyfinPurple, foregroundColor: .white))
-                    .isSelected(isCheckmarkSelected)
-                    .frame(maxWidth: .infinity)
-                    .if(!equalSpacing) { view in
-                        view.aspectRatio(1, contentMode: .fit)
+                    if isLiquidGlassEnabled, #available(iOS 26.0, *) {
+                        Toggle(L10n.played, systemImage: "checkmark", isOn: Binding(get: {
```
LePips, line 58: Why change to a `Toggle` than keeping as a `Button` but with glass style applied?

Swiftfin/Views/ItemView/Components/PlayButton.swift
```diff
@@ -101,6 +98,18 @@ extension ItemView {
             }
             .isSelected(true)
             .enabled(isEnabled)
+
+            if isLiquidGlassEnabled, #available(iOS 26.0, *) {
+                button.buttonStyle(.glassProminent).buttonSizing(.fitted)
```
LePips, line 103: Don't have `.fitted`

Swiftfin/Extensions/View/Modifiers/NavigationBarDrawerButtons/NavigationBarDrawerModifier.swift
```diff
@@ -18,8 +18,10 @@ struct NavigationBarDrawerModifier<Drawer: View>: ViewModifier {
 
     func body(content: Content) -> some View {
         NavigationBarDrawerView {
-            drawer()
-                .ignoresSafeArea()
```
LePips, line 22: Instead of offsetting by reading the geometry, use `.ignoresSafeArea(edges: .top)` instead. I would assume this should also be used elsewhere you make these changes.

Swiftfin/Components/PillHStack.swift
```diff
@@ -25,18 +29,29 @@ struct PillHStack<Item: Displayable>: View {
             ScrollView(.horizontal, showsIndicators: false) {
                 HStack {
                     ForEach(items, id: \.displayTitle) { item in
-                        Button {
```
LePips, line 28: I would probably recommend a full availability + defaults conditional based button resolution, rather than inline modifications.

LePips (changes requested):
For our liquid glass efforts, I would highly recommend starting small by focusing on single views or modifiers.

We might have to be more opinionated in the changes that liquid glass brings, especially with its many modifiers and UI changes. Ideally, I would like the support to be such that when iOS 26/27 is made the minimum we could just do simple deletions instead of having to unravel many inline conditional statements.

Swiftfin/Views/ItemView/Components/ActionButtonHStack/ActionButtonHStack.swift
yoveto: I figured since we are toggling between the state of played/not played making it a toggle to match that function made sense. (within swift ui its actually using a button element) I can swap it back if you would like

Swiftfin/Views/ItemView/Components/PlayButton.swift
yoveto: Whats the reasoning for this? Not opposed to swapping it back, however the liquid glass default behavior is fitted not stretched so I am curious as to why we would want to deviate

Swiftfin/Extensions/View/Modifiers/NavigationBarDrawerButtons/NavigationBarDrawerModifier.swift
yoveto: ooo good point. I'll swap this out. As for other areas, I did an audit on ios+ipados and could not find other cases where this fix was needed. Regardless I'll do another pass before I ping you for review

yoveto:
@LePips thanks for the review. I'll get this feedback implemented this week. Do you want me to break down this PR or is it fine as is and I'll break down future liquid glass PRs into smaller changes?

Swiftfin/Views/ItemView/Components/ActionButtonHStack/ActionButtonHStack.swift
LePips: Sure that's the semantic primitive to use, but really has no strong reason for usage since we have to make the remote call anyways.

Swiftfin/Views/ItemView/Components/PlayButton.swift
LePips: I would want glass effects to still have the same geometry. With this also applied, the button is absolutely tiny.

LePips:
I would probably recommend narrowing to some easy wins/transitions at first.

LePips:
I will close this as in #2075 I "backported" glass support which I find works in many places and should be used for glass efforts instead.

PR #2054, opened 2026-06-16
Playback Quality & Versions

### LINKED ISSUE #184: Quality Menu

### LINKED ISSUE #2055: Swiftfin doesn't play selected movie version

### Description of the bug

I have a movie hosted on my server with multiple versions,
```
/Movies/Infernal Affairs (2002)/Infernal Affairs (2002) - Commentary.m4v
/Movies/Infernal Affairs (2002)/Infernal Affairs (2002) - Feature.m4v
```
and Swiftfin shows both options in the version selection UI, but no matter which version I select it always plays the Commentary version.

Issue is present in a build from `main` as well as the most recent TestFlight build 1.5.0 (2)

### Reproduction steps

As described

### Platform

iOS

### Application version

1.5.0 (2) and main (2026-06-15)

TestFlight

### Device information

iPad Pro (11in 2nd gen)

### OS version

iOS 26.5

### Jellyfin server version

10.11.10

### Screenshots or videos

Image

### Relevant log output

```shell

```

JPKribs:
## Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/184
Resolves: https://github.com/jellyfin/Swiftfin/issues/2055

My other PRs are bumming me out so I wanted to do something quick and easy in the meantime. Now that https://github.com/jellyfin/Swiftfin/pull/1882 is in, this is really trivial to add. While I was in there I made a couple of cleanups:

1. I changed `playbackQuality` to `playbackSettings` since we're adding versions to it there was a TODO for it as an idea.
2. Instead of having 1 hard string & translation for "1080p - 8mbps" etc, I've just building this using the formatter for `.bitrate`
3. Moved the bitrate testing to the `MediaPlayerManager` per our TODO.
4. There is a TODO that we should allow the auto-test for bitrate to be selectable here. While not hard to add, it's a bit hard with UI and there isn't a lot of room to explain what this test means. My thinking is the test can be configured and we'll just make this a run/set opposed to configuration of this. Happy to change this but I was leaning towards skipping this item.

There are some changes that I made to make this all work smoothly:

1. `item.runTimeTicks` now comes from the selected `MediaSource` when one is provided. This ensures that changing the version updates the total runtime.
2. The `MediaPlayerManager` now stores `playbackBitrate`. This defaults to the default bitrate but updates and allows us to carry the bitrate into later items in the queue (see video 2 & the queue supplement)
3. Added a `PlaybackBitrate.min` to return the lowest bitrate as a kind of an opposite to `.max`. The idea is the lowest bitrate may change in the future so this way we don't need to update callers. TBH, I thought we used `kbps420` in more places so maybe this isn't necessary.
4. Added a `PlaybackBitrate.validBitrates()` that returns the valid bitrates to display for a `mediasourceinfo`. This is because, if I have something that's 15mbps there really isn't a reason for the selector to offer me 120mbps. (I think video 1 shows this off pretty well between these versions)

### Version Switching

I don't really have multiple versions. So, I just merged the movie 9 (2009) with 7 Days in Hell. The result is that it's easier to tell that I am switching but that's why it's 2 different movies:

https://github.com/user-attachments/assets/96bee193-d279-4f20-b49c-7a8b864317f6

**As a note, the Chapters and Trickplay are stored at the `BaseItemDto` NOT the `MediaSourceInfo` so the result is that both of those items will be off for any versions aside from the first. This impacts ALL versions even if started from outside the player.**

**This is known and reported here: https://github.com/jellyfin/jellyfin/issues/11967**

### Quality Switching

I am storing the bitrate on the quality manager so the bitrate selected for the session will persist down the queue:

https://github.com/user-attachments/assets/8eae25a5-3fe9-4b5f-95e5-0dc89081d464


LePips (approved):
While I was working with this I went back and forth with having the version selection as an option as a button on `MediaInfoSupplement` instead and just having the quality as a single button, but couldn't get a sufficient layout. I also split out quality and version as separate buttons, but then don't want so many action buttons.

Overall, I don't necessarily want so much functionality condensed into a single menu, but this is also making me think about how to properly layout/group multiple settings that might change the stream. For example, while there is a TODO for the compatibility picker that's quite a more advanced feature to perform and would probably want on `PlaybackInformationSupplement`. Just food for thought.

PR #2057, opened 2026-06-17
Cleanup Playback Sessions on Transcodes

JPKribs:
### Summary

While looking at AVPlayer, I was realizing that we don't clean up our transcodes when switching tracks. This mirros the logic found on [Jellyfin-Web](https://github.com/jellyfin/jellyfin-web/blob/af2ca3adfc6e2e1839e82879e34e4e4e0a3ee86e/src/components/playback/playbackmanager.js#L1775) to cleanup transcodes when the player stops. This wasn't easily-noticeable in testing because VLC does a good job direct playing. In testing AVPlayer, I was noticing 40+ gb of transcodes that never went away and I found Web had a cleanup step we didn't. The issue is most noticeable when switching streams when transcoding on both streams as this will double up the transcode folder.

This just adds that step to our stop logic to ensure items are cleaned up on server. 

### Video

On the left is my transcode folder. Right is video that has TrueHD that requires transcoding on playback.

https://github.com/user-attachments/assets/cd4def70-f860-4eed-945a-53dc7a9e4a8f


JPKribs:
As a note, `applicationWillTerminate` doesn't really run the way I was hoping it would. That is, we still have the issue that force closing the app while playing still results in a stranded session until the server cleans it up. Unless there is something I am missing I don't believe there is a way to do this client side nor do I think web does this client side either so we're in good company.

absidue:
> That is, we still have the issue that force closing the app while playing still results in a stranded session until the server cleans it up. Unless there is something I am missing I don't believe there is a way to do this client side nor do I think web does this client side either so we're in good company.

Have you already considered cleaning them up the next time the app is started?

E.g. When a transcode is started save a reference/its ID to the disk, remove the reference/its ID from the disk when the transcodes are stopped, then every time the app starts check if there are any trancode IDs on disk, if there are stop them and remove the IDs from the disk.

Sure it's not stopping them immediately but there is still a chance that the user might reopen the app before the server does its own cleanup.

JPKribs:
Server already cleans up sessions once the Device ID responds again.

LePips (approved):
> applicationWillTerminate doesn't really run the way I was hoping it would

That's correct, and while I'm fine with the code handling it for now we would have better handling with background execution and heartbeat signals from throughout the app.

> Have you already considered cleaning them up the next time the app is started?

We will actually probably need to implement jobs that would run throughout a user session for various cleanup or data updating, but will kick that to later.

JPKribs:
> We will actually probably need to implement jobs that would run throughout a user session for various cleanup or data updating, but will kick that to later.

I think once we have a socket session in the user session that shouldn't be too bad. Just by default, we get all user data updates.

Issue #1734, opened 2025-09-25
[iOS] CollectionView's filter bar difficult to access with VoiceOver


samglt:
The filter bar at the top of a collection can't be selected by tapping and is in a strange position in the voiceover heirarchy. The filter buttons are located with the letter bar in between the bottom of the list of items and the bottom tab bar, when you'd expect them to be at the top in or near the navigation bar.

The similar filter bar at the top of the search tab doesn't have this issue for some reason

LePips:
Closed by #2058

PR #2058, opened 2026-06-19
Collection view's filter bar difficult to access with voice over  

quangDecember:
Solving #1734 
I tried messing with `self.view.accessibilityElements` directly but `_UIHostingView` will make it inefficient. 
Therefore the solution is to convert it to a normal UIView and stack drawerButtonsView on top


quangDecember:
https://github.com/user-attachments/assets/41ef7963-3a32-4add-9e01-4f12e8662c4a

forgot to add the screen recording

LePips (approved):
Thank you for the fix! I would recommend using Engine.HostingController instead, but I can do that as I also try to version this feature using `safeAreaBar`.

PR #2059, opened 2026-06-23
App state consolidation, remove AppDelegate

LePips:
Initially started out as just fixing that the user session was trying to be restored prior to the data stack being initialized, but then wanted to clarify app state startup and ownership.

- Remove arbitrary root navigation with `RootCoordinator`
	- Consolidate app value observation
- Remove `AppDelegate`
	- Somehow don't need the orientation logic migrated to a `SceneDelegate`
	- I experimented with proper external display support with `SceneDelegate`s, will implement soon
- Server connection will now resolve on user sign in if enabled
- User and server data are retrieved from server more often, rather than just on app startup
- More DeepLink handling
- Fix scrolling bug on `UserSignInView` with splash screen


JPKribs (approved):

PR #2066, opened 2026-06-28
BlurHashKit update and Color Quantization

LePips:
- Updates implementation of the previous `RotateContentView` to `FadeContentTransitionView`
	- Subtly changes behavior of transitioning content
	- Had issues when wanting to implement in `ImageView`, will look again later
- Updates BlurHashKit to 2.0
- As I removed the color options from BlurHashKit, finally just implementing a basic color quantization based on median cut
	- I think gets better results than average color from the blur hash, may tweak over time
	- While loading the color will be empty, but will resolve this by forcing image resolution to be complete with content group transition.


JPKribs (approved):

Issue #842, opened 2023-09-14
Live TV Guide


TechBLT:
Live TV would be very useful with a guide and bonus point for the ability to record and setup a pass from the app as well.

extraslute:
I second this. Similar to IPTVX's interface for the guide. I wish we could also get art related to the shows listed in the guide if we have EPGs enabled.

vikgor:
Hello! I have started working on the Live TV enhancement, including the guide view (nobody is working on it atm afaik).
I'm gonna open a draft PR in a couple of days, although I think I'll need to wait for #1752 before it could go to main.

I have come up with a pretty decent-looking UI, although I'll definitely polish it up a little.
This is just an early draft of the tvOS Live TV Guide view:
<img width="90%"  alt="Image" src="https://github.com/user-attachments/assets/115a0ddd-8a27-4322-af9e-e0b0858cec9d" />

extraslute:
> Hello! I have started working on the Live TV enhancement, including the guide view (nobody is working on it atm afaik).
> I'm gonna open a draft PR in a couple of days, although I think I'll need to wait for #1752 before it could go to main.
> 
> I have come up with a pretty decent-looking UI, although I'll definitely polish it up a little.
> This is just an early draft of the tvOS Live TV Guide view:
> <img width="90%"  alt="Image" src="https://github.com/user-attachments/assets/115a0ddd-8a27-4322-af9e-e0b0858cec9d" />

Holy cow, about time! I'm willing to test if you need any guinea pigs. I've sense paid for Emby because of the live tv UI on ATV. Live tv on Android TVs is killer compared to ATVs UI.

PR #2067, opened 2026-06-29
Live TV Filtering

### LINKED ISSUE #433: Ability to show only favorited live tv channels
Would be great to be able to filter live tv channels to show only those added to the favorites list!

### LINKED ISSUE #2078: Live TV play button not selectable

### Description of the bug

After selection of an item from the program view in a Live TV  library, the item is not selectable for playing. 

### Reproduction steps

1. Open Swiftfin and go to the Live TV library
2. Select a program
3. In the item detail page for the program, the play button is unselectable. 

### Platform

iOS

### Application version

1.5

TestFlight

### Device information

iPhone 13 Pro

### OS version

iOS 18.7.8

### Jellyfin server version

12.0 RC2

### Screenshots or videos

Image

### Relevant log output

```shell

```

JPKribs:
## Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/433
Resolves: https://github.com/jellyfin/Swiftfin/issues/2078

This removes the program view in favor of just using a paging library. I move the `OnNow` to the `HomeView` instead:

Screenshot 2026-07-14 at 22 23 37

Screenshot 2026-07-14 at 22 25 44

This adds a guide type for libraries that is only enabled for LiveTV. This type is just a listView + a lazyHStack for the content going out to the right. Performance is alright but nothing exciting. This allows filtering and sorting as well.

We used to have logic that hid sorts that were not usable by that parent type. I added this back in since Live TV is pretty restricted on that.

Some quirks addressed:

- Channels can be favorited but programs cannot. Removed the button from programs. This looks weird now but eventually this will be a record button for recording a program.
- Depending on their origins, Programs, TVPrograms, and LiveTVPrograms are all the same object just from different callers. Same for Channels, TVChannels, and LiveTVChannels.
- Programs have different logic for progress calc than regular `BaseItemDto` so this updates that to change when airing.

## Video

https://github.com/user-attachments/assets/c9eb4595-ec65-4196-8649-390f5e4b8c55


JPKribs:
As a note, I'm probably looking at a [Live TV Guide](https://github.com/jellyfin/Swiftfin/issues/842) next while I'm in the Live TV mood. I could be off but I was assuming that `ProgramView` was going to borrow a lot from `HomeView` so I was going to leave that alone.

Shared/Objects/ItemFilter/ItemFilterCollection.swift
```diff
@@ -30,6 +31,10 @@ struct ItemFilterCollection: Hashable, Storable {
         sortBy: [ItemSortBy.dateCreated],
         sortOrder: [ItemSortOrder.descending]
     )
+    static let channels: ItemFilterCollection = .init(
```
LePips: Don't need, this is just the same as the default value.

Shared/Objects/ItemFilter/ItemFilterType.swift
```diff
@@ -23,9 +23,23 @@ enum ItemFilterType: String, CaseIterable, Displayable, Identifiable, Storable,
     case tags
     case traits
     case years
+    case category
+
+    static var allCases: [ItemFilterType] {
```
LePips: I know this might be to not have the `category` in search, but I think we're fine having it as a filter elsewhere.

Now `SearchViewModel` should filter the category and check if it properly filters the channels.

Shared/Objects/Libraries/ChannelLibrary.swift
```diff
@@ -6,24 +6,60 @@
...
+
+        let categories = environment.filters.categories
+
+        parameters.isMovie = categories.contains(.movies) ? true : nil
+        parameters.isSeries = categories.contains(.series) ? true : nil
+        parameters.isNews = categories.contains(.news) ? true : nil
+        parameters.isKids = categories.contains(.kids) ? true : nil
+        parameters.isSports = categories.contains(.sports) ? true : nil
 
         let request = Paths.getLiveTvChannels(parameters: parameters)
```
LePips, line 28: With https://github.com/jellyfin/Swiftfin/pull/2067/changes#r3492718195, I wonder if we'll be better off retrieving channels via `getItems` to allow usage of all filters.

Does not need to be done now.

LePips (changes requested):

Shared/Objects/Libraries/ChannelLibrary.swift
JPKribs: ~I think the tradeoff is that items gives us all of the regular filters but we'd lose categories. I'm not sure what people's preference is. IMO, those `isKids`, `isMovies`, etc filters seem less useful than the other filters but I also think that some of the `getItem` filters will always return nil for programs/channels like `isPlayed` and `genre` I don't *think* channels have. I'd need to screw around with it a bit to see how many of the regular filters actually return results.~

JPKribs: I am completely wrong. Yes, all those category `isX` filters exist on `getItems`. Good call, we should just move to that instead.

JPKribs: The scope is expanding but I think this is going to set us up better for the future. Standby...

JPKribs:
@LePips some changes. Moved this over to just use the existing `ItemLibrary` so we can take advantage of all that work and the filters there.

I also added a Channel ItemView and added Programs to the simple ItemView. Right now it's somewhat bare but the idea is that is where we can put the recording options for these as well. I left some stubs for that. I was going to roll that into this PR but I reeled it back in a bit haha.

Depending on WHERE you access it, tvChannel, channel, and liveTvChannel are all used for the same items. Same for programs 🤷‍♂️. Hence the switches grouping those together when used.

The potentially controversial item, I am passing the poster from the currently active program for the Channel ItemView. More controversial, if the channel doesn't have an overview I am pulling that from the active playing item. It's a small change in a few places but I can remove that if we don't like it.

Video below:

https://github.com/user-attachments/assets/77bf66f1-e74f-4bdc-a15c-2d9539b261cc

JPKribs:
Downside in testing. The letter filter for `NameStartsWith` is looking at the channel number so any letter filter returns nothing only the # returns anything. Probably best to just hide that for this library type?

Also, in some previous PR I added the sort types for `BaseItemKind` that wasn't pulled into the generic paging. I've added it back in so our sort options are more tailored to the LibraryParent

JPKribs:
@LePips okay, I'm sorry to do this I am going a little overkill here but I felt inspired. This is still WIP while I figure out some of the quirks but here is my idea:

1. Remove the Program view and move the `On Now` section to the `HomeView`.
2. Change the `Live TV` library to instead take you right into a paging `ItemLibrary` for channels.
3. Add a new layout, filtered to only live tv, for `guide` that is just the `list` format but the inner content on the right is the tv guide layout in a `scrollview`. So this lets us paginate the content vertically in the guide but all the guide scrolls move in unison like a single block

This is my rough version I have now:

https://github.com/user-attachments/assets/f98ff864-b5de-4c97-9ec4-c726121fa6a6

My other idea in this, my biggest gripe with Jellyfin-web is that if you have a bunch of small items next to each other the buttons start stacking over one another. So I've set this up where if a button is less than 15m it should combine into another button like this:

https://github.com/user-attachments/assets/8ec2eaa0-21fa-4897-99dc-bd27b436677e

LePips:
Looks great!

But now that this has expanded to building views in the item and home view, I would have this wait until I've broken out the content group stuff from #1752, which I am working on. At the same time, what you implement here would arguably be easier afterwards.

JPKribs:
This is migrated over to the new `ContentGroups`. It was a lot easier to do these with those changes! Let me know if you have any feedback or changes that you would like me to make!

All videos in the main post are up to date!

LePips (changes requested):
I don't think having the guide as a library style is the right direction, but I will have to think about it. I would be fine with just a new view for the guide instead of making new library styles that all elements would have to consider that is just used for live tv.

JPKribs:
> I don't think having the guide as a library style is the right direction, but I will have to think about it. I would be fine with just a new view for the guide instead of making new library styles that all elements would have to consider that is just used for live tv.

Makes sense! Sorry, I was trying to loop this into the existing FilterViewModel. I was also looking at how Android TV does this and their grid hooks into their player as well so you can both look at the grid and play at the same time.

What I think is the right move, I can do the paging library and filters for the grid and list view with the channel ItemView then push the guide into its own PR since I realize there is more considerations for this.

JPKribs:
Okay. This is getting messy. I am going to split this out over a couple of PRs since most of these are just small, independent changes

PR #2071, opened 2026-07-03
Update StatefulMacros

LePips:
Have debugged the failing archiving a bit more, needed to restructure the products a bit on StatefulMacros.


JPKribs (approved):
Thanks for working on this BTW! This type of troubleshooting is always very frustrating so I wanted to say that your time on this is really appreciated!

PR #2073, opened 2026-07-10
Add tvOS home hero carousel

Kief5555:
## Summary

- add a focused hero carousel at the start of the tvOS home screen
- use the existing Recently Added items instead of adding a new data source
- support automatic progress, reduced-motion behavior, and left/right remote navigation
- keep Continue Watching as a regular poster row beneath the hero

## Validation

- SwiftFormat lint
- SwiftLint strict
- tvOS Xcode analysis

PR #2074, opened 2026-07-10
Add tvOS home hero carousel

Kief5555:
Replaces #2073 with the intended tvOS-only scope.

## Summary

- add a focused hero carousel at the start of the tvOS home screen
- use the existing Recently Added items instead of adding a new data source
- support automatic progress, reduced-motion behavior, and left/right remote navigation
- keep Continue Watching as a regular poster row beneath the hero

## Validation

- SwiftFormat lint
- SwiftLint strict
- tvOS Xcode build and analysis


LePips:
First, apologies but this does come right when I'm finalizing other work in #2075 that would finally be able to introduce a hero carousel element.

Second, please take some more caution in bringing changes from another fork and branch management to prevent the issues we are seeing in your last and this PR.

Kief5555:
Sorry, I'm not sure whats going on with GH. I select the other fork and it creates it on here instead of showing it somewhere else.

PR #2103, opened 2026-07-17
More glass support, tvOS glass

LePips:
Part of #1928

- Remove `ToolbarPillButtonStyle`, opting instead to back ported glass styling
- Make backported glass button styles
	- Initially generated by Codex
- Unify some save/confirm logic to be toolbar items instead of within lists
- Have tvOS have liquid glass enabled.

Glass effects are applied in a variety of methods and styles. Still trying to figure out best practices to do so, but good enough for now. Still trying to converge on glass effects to not have special-cased styles (supplement title, capsule). 

There are some tinting issues, but am willing to allow them pass to be fixed during TestFlight track:
- Delete toolbar button takes accent color instead of red on back ported glass style
- Text color on glass buttons in list is primary text instead of overlay color
- "From beginning" button has odd styling
- Video player episode supplement button overlay clipped


JPKribs (approved):

PR #2106, opened 2026-07-18
Fix library style sourcing

### LINKED ISSUE #2093: The Layout settings for the Libraries are not working.
If you click the layout settings in the top-right corner of the media library, the function works correctly.
However, if you go to the app's settings, select "Advanced," then "Libraries", and choose "Layout", selecting a different layout style has no effect. Regardless of the Layout chosen, the final Layout of the Libraries is Grid-Poster.

iPhone 12 Pro, iPad Pro 2th

iOS, iPadOS, 18.7.1

1.5 (12)

10.11.11

LePips:
Closes #2093

Also some cleanup.


JPKribs (approved):

PR #2108, opened 2026-07-18
Engine `PlatformViewRepresentable`

LePips:
Motivated by macOS work, begin to use some of Engine's shared platform protocols.


JPKribs (approved):
Exciting stuff! Tested and (expectedly) nothing is changed (for now)

PR #2112, opened 2026-07-18
Consolidate `ItemView`

LePips:
Consolidates `ItemView`, headers, and scroll views.


JPKribs:
Testing this now. While you're touching the PlayButton, could we add a fix for: https://github.com/jellyfin/Swiftfin/issues/2078

The working idea I had for this was to add this to BaseItemDto:

```swift
private var airingWindow: (start: Date, end: Date)? {
    switch type {
    case .program, .liveTvProgram, .tvProgram:
        guard let startDate, let endDate else { return nil }
        return (startDate, endDate)
    case .channel, .liveTvChannel, .tvChannel:
        guard let start = currentProgram?.startDate, let end = currentProgram?.endDate else { return nil }
        return (start, end)
    default:
        return nil
    }
}

var isAiring: Bool {
    guard let airingWindow else { return false }
    return airingWindow.start <= .now && .now <= airingWindow.end
}
```

And update the `PlayButton` to:

`.disabled(provider.selectedMediaSource == nil && provider.playButtonItem?.isAiring != true)`

The issue with `isLive` being I believe that just indicates whether it's liveTV or regular content.

LePips:
You can definitely add that after this is in.

JPKribs (approved):

PR #2121, opened 2026-07-19
Non-Romantic Subtitle Fix

### LINKED ISSUE #2004: External Chinese subtitle does not render properly
When playing videos with external Chinese subtitles (subrip), the characters are not rendered properly and are just shown as squares.

image

Thought maybe it's due to the font does not support Chinese (which ideally should fall back to a font that supports), I changed the subtitle font to Pingfang (which is Apple's recommended Chinese font for iOS: https://support.apple.com/guide/pages-iphone/format-chinese-japanese-or-korean-text-tanfbd4156e/ios). This time nothing is rendered at all:

image

The same video is rendered properly in Streamyfin under default setting.

image

Swiftfin: 1.5(2)
Jellyfin: 10.11.8
iOS: 26.3.1

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/2004

This does two things. 

#### Firstly:

To fix 2004, this bundles `Noto Sans CJK` into Swiftfin which resolves the rendering issue at the cost that Swiftfin will be ~19.5 MB larger. A non-insignificant increase ~175 -> ~195 is a ~11% increase but this does resolve our issue. As part of this, I am setting `Noto Sans CJK SC` as the default font. In testing, this works for Chinese, Japanese, and Korean with English and other romantic languages working with it by default as well. I'm sure there is a larger debate about which of the  `Noto Sans CJK` variants should be the default but this was recommended to me on a forum so I am willing to die on this hill /s.

As a note, I read that `Noto Sans CJK` could not render Hebrew and Arabic but in my testing **I found this to be false. This continues to render in VLCKit using this font.**

#### Secondly:

There was an outstanding TODO to turn the various subtitle settings into a `SubtitleConfiguration` object. Saving and encode/decode works as before. I haven't made a migration for this since I would argue migrating everyone to `Noto Sans CJK` as the default (since we know it works) would be beneficial to us but I'm open to migrating.

#### Before

Please note the □ in these subtitles:

https://github.com/user-attachments/assets/2f8aa96f-df84-4d4c-a42f-342ba86ee988

#### After

Please note that the □ are now gone and render correctly:

https://github.com/user-attachments/assets/ea00d02c-6806-4aa5-905c-f8b1e20b2570

### AI Disclaimer

All work was me but the subtitles for testing were generated by Claude. Please note that I have no idea what they say so apologies if there is anything off there!

### DIFF
--- a/Shared/Resources/Fonts/NotoSansCJK-OFL.txt
+++ b/Shared/Resources/Fonts/NotoSansCJK-OFL.txt
@@ -1,0 +1,92 @@
+This Font Software is licensed under the SIL Open Font License,
+Version 1.1.
+
+This license is copied below, and is also available with a FAQ at:
+http://scripts.sil.org/OFL
+
+-----------------------------------------------------------
+SIL OPEN FONT LICENSE Version 1.1 - 26 February 2007
+-----------------------------------------------------------
+
+PREAMBLE
+The goals of the Open Font License (OFL) are to stimulate worldwide
+development of collaborative font projects, to support the font
+creation efforts of academic and linguistic communities, and to
+provide a free and open framework in which fonts may be shared and
+improved in partnership with others.
+
+The OFL allows the licensed fonts to be used, studied, modified and
+redistributed freely as long as they are not sold by themselves. The
+fonts, including any derivative works, can be bundled, embedded,
+redistributed and/or sold with any software provided that any reserved
+names are not used by derivative works. The fonts and derivatives,
+however, cannot be released under any other type of license. The
+requirement for fonts to remain under this license does not apply to
+any document created using the fonts or their derivatives.
+
+DEFINITIONS
+"Font Software" refers to the set of files released by the Copyright
+Holder(s) under this license and clearly marked as such. This may
+include source files, build scripts and documentation.
+
+"Reserved Font Name" refers to any names specified as such after the
+copyright statement(s).
+
+"Original Version" refers to the collection of Font Software
+components as distributed by the Copyright Holder(s).
+
+"Modified Version" refers to any derivative made by adding to,
+deleting, or substituting -- in part or in whole -- any of the
+components of the Original Version, by changing formats or by porting
+the Font Software to a new environment.
+
+"Author" refers to any designer, engineer, programmer, technical
+writer or other person who contributed to the Font Software.
+
+PERMISSION & CONDITIONS
+Permission is hereby granted, free of charge, to any person obtaining
+a copy of the Font Software, to use, study, copy, merge, embed,
+modify, redistribute, and sell modified and unmodified copies of the
+Font Software, subject to the following conditions:
+
+1) Neither the Font Software nor any of its individual components, in
+Original or Modified Versions, may be sold by itself.
+
+2) Original or Modified Versions of the Font Software may be bundled,
+redistributed and/or sold with any software, provided that each copy
+contains the above copyright notice and this license. These can be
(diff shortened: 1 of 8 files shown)


Shared/Resources/Fonts/NotoSansCJK-OFL.txt
```diff
@@ -0,0 +1,92 @@
+This Font Software is licensed under the SIL Open Font License,
```
JPKribs, line 1: I've included the license because I'm always unsure about attribution and wanted to err on the side of over-attributing. It's like 92/171 overall LOC added. I can remove this if we feel it's unnecessary

JPKribs:
Attached are subtitles used for testing. These are just 10m of something I attached to my files. Again, I trust that Claude didn't put anything nasty in there but please note that I have no idea what this says! Remove the .txt to use these, I just uploaded this as .txt as github didn't like me trying to upload .srt and uploading a .zip felt like asking for trouble:

[test-subtitles-ar.srt.txt](https://github.com/user-attachments/files/30161430/test-subtitles-ar.srt.txt)
[test-subtitles-he.srt.txt](https://github.com/user-attachments/files/30161431/test-subtitles-he.srt.txt)
[test-subtitles-ja.srt.txt](https://github.com/user-attachments/files/30161432/test-subtitles-ja.srt.txt)
[test-subtitles-ko.srt.txt](https://github.com/user-attachments/files/30161433/test-subtitles-ko.srt.txt)
[test-subtitles-zh.srt.txt](https://github.com/user-attachments/files/30161434/test-subtitles-zh.srt.txt)

zhen-huan-hu:
Regarding different variants, Adobe does provide a variant OTF font that packages all variants in one file and according to them the most compressed form. However, I am not sure if such OTF is supported by FreeType.

https://github.com/adobe-fonts/source-han-sans/releases/tag/2.005R (Source Han Sans and Noto Sans CJK are the same font)

Also want to point out, on their repo, they also have Source fonts for other languages: https://github.com/adobe-fonts The Noto versions of some of these are already shipped with iOS

zhen-huan-hu:
I am sure the subs you tested are sufficient but here is the one that caused issue initially for me

[example.zh-tw.srt.txt](https://github.com/user-attachments/files/30170485/example.zh-tw.srt.txt)

JPKribs:
> I am sure the subs you tested are sufficient but here is the one that caused issue initially for me
> 
> [example.zh-tw.srt.txt](https://github.com/user-attachments/files/30170485/example.zh-tw.srt.txt)

I appreciate it. I didn't go through the entire file but I believe I recreate the part in your issue without rendering issues on this branch so I believe this still looks good to review!

LePips (approved):
Thank you for doing the research on this. I don't really have any comment on this specific font being used, but I will just make the statement now that I won't want to package custom/requested fonts in the future as I believe the iOS font downloading feature fulfills that.

zhen-huan-hu:
> Thank you for doing the research on this. I don't really have any comment on this specific font being used, but I will just make the statement now that I won't want to package custom/requested fonts in the future as I believe the iOS font downloading feature fulfills that.

Thank you for approving this. You mentioned the iOS font downloading feature. Is the font list in Swiftfin pre-specified and fixed or is it dynamically updated based on the actual installed fonts? In the latest version, it does not seem to recognize fonts I installed under 'More Fonts'.

See below the additional font Apple Braille is installed on my phone but it does not appear in Swiftfin:

image

image

PR #2122, opened 2026-07-19
Episode - Landscape: Use Backdrop if Available

JPKribs:
### Summary

Some metadata providers provide a `.primary` which is a landscape poster with text on top and a `.backdrop` which is the same landscape poster without text on it. More generically, if a backdrop image type exists, use that first.

Mirrors how web handles the existence of `.backdrop` + `.primary` vs just `.primary` as well.


JPKribs:
Ignore, this was the wrong approach this replaces valid .primary usages too. I think this would need to be a split of item view enhanced background and regular landscape which feels like more harm than good.

PR #2124, opened 2026-07-19
Various 1.5 Cleanup

JPKribs:
### Summary

This is kind of a grab bag of smaller items. Please let me know if you would rather I split these out.

1. Restores the changes from https://github.com/jellyfin/Swiftfin/pull/1825. It's very slightly different to fit into the new structures laid out in 1.5 so the indicator setting is just an array of indicator types so that is unchanged but this adds another setting for type but this is consolidated into a single menu selector.
2. Trailers default to .all but there was no way to select .all after you changed it once. This just adds this to .allCases to allow this to be selectable again.
3. The tvOS Save ItemSubtitle button was tiny because the maxWidth was applied to the button instead of the text inside the button. This just changes that and cleans up the view a little to be less bloated.
4. Some change up for the ItemEditMenu to allow us to put a StateAdapter around this to enable deletion from tvOS. This also cleans up the monochrome issues we had elsewhere where the symbol and the text didn't match.

### Various Screenshots

<details>

<summary>In No Particular Order</summary>

Bad Subtitle Save

Count Indicator (tvOS)

Count Menu + Preview

Count Menu

</details>

### iOS Item Menu (Unchanged but Proof of it)

https://github.com/user-attachments/assets/1509fd50-f04c-4198-b4d6-311a766616fd

### tvOS Deletion

https://github.com/user-attachments/assets/d110ba3b-c26c-4290-bf9c-62a48e0f0ccb


zhen-huan-hu:
Not really a bug but the three-dot menu for subtitle is a bit misleading as the first time I clicked it I thought this is for selecting what subtitle to play. Instead, it is for searching and downloading subtitles.

LePips (approved):

PR #2134, opened 2026-07-21
Allow DOVIWithEL video to DirectPlay

### LINKED ISSUE #2131: DV profile 7 won't DirectPlay in Swiftfin player

### Description of the bug

Dolby vision profile 7 (used on 4K Blurays) can be encapuslated in .mkv files. It uses the VideoRangeType `DOVIWithEL`. Currently, this is not in the `swiftfinHDRProfiles`, so the video is not DirectPlayed. However, the server realizes that the video is supported with this range and remuxes the streams, delivering what I believe to be the same stream in a different container. Since Swiftfin player already supports .mkv, this is unnecessary.

I wasn't paying a ton of attention when #1861 was introduced, but I think VLCKit plays these fine. To note: VLCKit is likely not handling the EL and falling back to HDR10, but I don't see the server handling it either. It is somewhat confusing because any `DOVIWithEL` has a HDR10 base layer (BL), but there is also `DOVIWithHDR10` which I believe is DV profile 8 with DV dynamic metadata on top of an HDR10 stream.

### Reproduction steps

Play a movie on Swiftfin that has a supported DirectPlay audio codec (such as DTS) with a Dolby Profile 7 (MEL or FEL) stream.

### Platform

iOS

### Application version

1.5

App Store

### Device information

iPhone 12 Pro

### OS version

iOS 26

### Jellyfin server version

10.11.11

holow29:
Summary:

Resolves #2131. Adds doviWithEL to both swiftfinHDRProfiles and nativeHDRProfiles.

This allows videos that are single track dual layer Dolby Vision Profile 7.6 to play directly if all else is in order, utilizing the HDR10 BL. The MEL and FEL is not utilized in this situation, but it is also not utilized by the server, so DirectPlay is preferred. I'm not familiar with a situation where this could cause playback issues because DOVIWithEL should always have a HDR10 BL AFAIK.

If at some point the server starts handling the MEL or FEL, it might be worth revisiting, but that raises questions as well about the whole way client capability is indicated and acted upon by the server. Likely the nativeHDRProfiles logic should be reworked because some of the DOVI ranges like this I believe might actually require HDR10 capability - but it also likely doesn't really matter right now because anything that supports DOVI should support HDR10 as well. In other words, I think PlaybackCapabilities.supportsDolbyVision is a superset of PlaybackCapabilities.supportsHDR10. This means _I believe_ (but am not confident enough at the moment to change anything, and it won't affect playback anyway given the current logic) the only DOVI videoRangeType that requires Dolby Vision support is VideoRangeType.dovi since all the other DOVI ones have a base layer of either SDR, HDR10, or HLG.

Compiled and tested on iPhone X and Apple TV HD. Tested with DOVIWithEL single track dual layer dv profile 7 .mkv and .mp4. MKV DirectPlayed on Swiftfin and MP4 DirectPlayed on Native as expected.

### DIFF
--- a/Shared/Objects/VideoPlayerType/VideoPlayerType+Native.swift
+++ b/Shared/Objects/VideoPlayerType/VideoPlayerType+Native.swift
@@ -265,6 +265,7 @@ extension VideoPlayerType {
         if PlaybackCapabilities.supportsHDR10 || PlaybackCapabilities.supportsDolbyVision {
             VideoRangeType.doviWithHDR10
             VideoRangeType.doviWithHDR10Plus
+            VideoRangeType.doviWithEL
             VideoRangeType.doviWithELHDR10Plus
         }
 
--- a/Shared/Objects/VideoPlayerType/VideoPlayerType+Swiftfin.swift
+++ b/Shared/Objects/VideoPlayerType/VideoPlayerType+Swiftfin.swift
@@ -325,6 +325,7 @@ extension VideoPlayerType {
             VideoRangeType.doviWithHLG
             VideoRangeType.doviWithHDR10
             VideoRangeType.doviWithHDR10Plus
+            VideoRangeType.doviWithEL
             VideoRangeType.doviWithELHDR10Plus
         }
     }


JPKribs:
I don't have a file to test this so I will need to create one later in the week. In the interim, would you be able to ensure that https://github.com/jellyfin/Swiftfin/blob/main/Documentation/players.md has the correct versions as a result of this change?

holow29:
HDR matrix looks good to me (subtitle matrix is a bit wonky IMO but separate issue). This PR allows DV 7.6 DirectPlay with the HDR10 layer, matching what the matrix already has.

If you create a test file, ensure codec tag is hvc1 on the single video track, otherwise it will have to remux so server can alter it.

JPKribs:
> If you create a test file, ensure codec tag is hvc1 on the single video track, otherwise it will have to remux so server can alter it.

That is a very good note, thank you for that! I was about to drive myself crazy haha

holow29:
For future reference, here is my understanding of an oversimplified videoRangeTypes:

| videoRangeType | Dolby Vision Profile | Base | Enhancement layer | Metadata |
| ---- | ---- | ---- | ---- | ---- |
| sdr | | SDR | | |
| hdr10 | | HDR10 | | |
| hdr10Plus | | HDR10 | | HDR10+ metadata |
| hlg | | HLG | |
| dovi | Profile 5 | DV Profile 5 | | DV metadata |
| doviWithEL | Profile 7.6 | HDR10 | DV MEL or FEL | |
| doviWithELHDR10Plus | Profile 7.6 | HDR10 | DV MEL or FEL | HDR10+ metadata |
| doviWithHDR10 | Profile 8.1 | HDR10 | | DV metadata |
| doviWithHDR10Plus | Profile 8.1 | HDR10 | | HDR10+ metadata; DV metadata |
| doviWithSDR | Profile 8.2 | SDR | | DV metadata |
| doviWithHLG | Profile 8.4 | HLG | | DV metadata |

JPKribs (approved):
Tested and functioning as expected! Thank you for your contribution!

PR #2139, opened 2026-07-22
Display Live TV `MediaInfoSupplement`

JPKribs:
### Summary

This was part of the other LiveTV changes but I dropped it at one point and never added it to the final PR.

If a `currentProgram` exists, use that for the `MediaInfo` only if the `Channel` does not provide it.

### Screenshots

tvOS

Simulator Screenshot - Apple TV 4K (3rd generation) - 2026-07-22 at 23 21 25

<details> 

<summary>iOS Comparison</summary> 

| Orientation | Screenshot |
|---|---|
| Vertical | Vertical Screenshot |
| Horizontal | Horizontal Screenshot |

</details> 

### Transition Between Items

https://github.com/user-attachments/assets/0b2fe3a9-28fc-40d9-9044-4c4847a5e249


JPKribs:
This was in an earlier version of the other work. The main reason I moved into it's own PR was the usage of:

```swift
@State
private var item: BaseItemDto

init(item: BaseItemDto) {
    self._item = State(initialValue: item)
}
```

Shared/Objects/MediaPlayerManager/Supplements/MediaInfoSupplement.swift
```diff
@@ -99,6 +108,15 @@ extension MediaInfoSupplement {
             }
             .padding(.leading, safeAreaInsets.leading)
             .padding(.trailing, safeAreaInsets.trailing)
+            .task(id: item.currentProgram?.endDate) {
```
LePips, line 108: I see that this functionality is for updating the program as it's changing. I will have to think about a mechanism where this would be done as an `MediaPlayerObserver` instead in the future, but am fine with this for now.

This also needs to be applied to tvOS.

Shared/Objects/MediaPlayerManager/Supplements/MediaInfoSupplement.swift
```diff
@@ -29,6 +30,9 @@ extension MediaInfoSupplement {
 
     private struct InfoOverlay: PlatformView {
 
+        @Injected(\.currentUserSession)
+        private var userSession
```
LePips: Use `manager.userSession` instead.

Shared/Objects/MediaPlayerManager/Supplements/MediaInfoSupplement.swift
```diff
@@ -149,7 +175,15 @@ extension MediaInfoSupplement {
                         .lineLimit(2)
                         .multilineTextAlignment(.leading)
 
-                    if let overview = item.overview {
+                    if let currentProgram = item.currentProgram {
+                        Text(currentProgram.displayTitle)
+                            .font(.callout.weight(.semibold))
```
LePips: Use `.fontWeight(.semibold)` instead. Also fix the one above this.

LePips (changes requested):

Shared/Objects/MediaPlayerManager/Supplements/MediaInfoSupplement.swift
JPKribs: > This also needs to be applied to tvOS.

My bad 🤦‍♂️

Tested and working now on this version. I left a TODO here for moving this to an `MediaPlayerObserver` for later.

Updated main post with a screenshot of tvOS as well.

LePips (approved):

PR #2141, opened 2026-07-23
Live TV Guide

JPKribs:
### Summary

Starts: https://github.com/jellyfin/Swiftfin/issues/842 - Supplement still required!

Creates a Live TV or [EPG](https://en.wikipedia.org/wiki/Electronic_program_guide) for Live TV on Swiftfin.

I think the only unique item I am doing for this is that when I first ran this I found there were a ton of instances where smaller videos (Music Videos, 5m segments, etc) where the buttons were too small to press and performance dragged considerably drawing that many items. As a solution, smaller items are bundled into a single entry (see below).

<details>

<summary>Several Smaller Entries</summary>

https://github.com/user-attachments/assets/06dae929-afbf-487f-8963-60fa697e3b71

</details>

This also enabled the feature on Web where you can color code your grid items by type. To get ahead of the eventual requests about color selection, I've added a setting to save different colors by type. The checkbox to enable color for that type is saved to defaults as well so it carries into the supplement.

### New Settings

https://github.com/user-attachments/assets/e35a84e9-e99e-46f2-929d-5aaf91e07c1e

#### iOS

https://github.com/user-attachments/assets/59723fa2-1cb0-4f9a-bd8d-f97d3e02aff5

#### iPadOS

https://github.com/user-attachments/assets/83237826-6589-42a0-a879-c19902799a88

#### tvOS

https://github.com/user-attachments/assets/24a69303-581c-41c3-857d-f23b23964e77

https://github.com/user-attachments/assets/0fb14ed9-dec1-4aa0-88de-f27f721ac8f1

#### iOS

https://github.com/user-attachments/assets/70f06e67-1abc-433e-a95d-5826492b3544

#### iPadOS

https://github.com/user-attachments/assets/236549c4-a595-4b7b-a5f6-beb630b372aa

#### tvOS

https://github.com/user-attachments/assets/b10166ab-e7c5-4b2d-a1e8-5733de4980d7

https://github.com/user-attachments/assets/93d37567-c201-4184-8aa5-8bb907c1b731


JPKribs:
Performance is solid until a lot of items load in. With a full stack, it's not perfect but still definitely acceptable. The simulator seems to perform worth than real hardware. Tested on iPhone 13 Pro and was how this looked:

https://github.com/user-attachments/assets/f88c1c00-21a9-4567-bd8e-c483a7e117ea

https://github.com/user-attachments/assets/3278d426-5423-4cfd-85bb-ba1014f7060d

I think the alternative would be replacing the `CollectionVGrid` with a custom `UICollectionView` which starts to get pretty heavy in terms of maintenance. 

---

Going to continue to look for areas of improvement while I cleanup some of these files but I'm pleased with where this stands.

LePips:
> replacing the CollectionVGrid with a custom UICollectionView

Yes, that has always been my intention with the guide. I'm not concerned about maintenance.

The buttons should also not have glass effects.

JPKribs:
> Yes, that has always been my intention with the guide. I'm not concerned about maintenance.

Damn, switching to this is genuinely a world of difference. I'm still pretty green on `UICollectionView` but I leaned on projects like [EPGGrid](https://github.com/CoderXpert/EPGGrid) for reference. I *think* this is in a really good spot now! Will definitely appreciate your expertise on the `UICollectionView` in case there is something stupid I'm missing there.

Doing another cleanup pass this evening after I wrap up work. ~From there, I will update the videos and screenshots in the main post and should hopefully be ready for review.~

Edit: Screenshots are up to date!

JPKribs:
After this PR I have one more Live TV item I want to knock out then I intend on finishing the remaining draft PRs I have open. I am working on a `ProgramTimerViewModel` for starting, stopping, and canceling recording timers. Once we knock out this I think that's the last Live TV item we're missing in Swiftfin? My goal is I want to just knock out all the Live TV now while it's top of mind for me so if there is anything I'm forgetting about let me know and I can add that to the list!

LePips (commented):
Largely migrated some views to SwiftUI for usage in the collection view and would want this presented as a fullscreen supplement instead in the video player. The change of layout for supplements still needs some work and I very much still consider this a work in progress, so also had the video player accessibility as experimental.

Please test again for functionality.

JPKribs:
> Please test again for functionality.

iOS looks good and works as expected:

https://github.com/user-attachments/assets/13f6f0c5-97f4-4ef2-a0ab-ea9c783ee592

tvOS works but getting to the trailing edge it starts to behave weirdly for the cutoff:

https://github.com/user-attachments/assets/9cb11785-0e51-4b0b-a33a-2f23585e74a9

An earlier version of this I had was it would just paginate out `onTrailingEdge` if you think that helps here as a route.

LePips (approved):
I'm fine with that as this continues to be iterated upon.

Issue #2132, opened 2026-07-21
Prioritize resolution when transcoding


JustinHoMi:
### This feature request respects the following points:

- [x] This feature has **not** already been requested on [GitHub](https://github.com/jellyfin/Swiftfin/issues?q=is%3Aissue%20state%3Aopen%20type%3AFeature).
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).

### Platform

tvOS

### Describe the feature you'd like to see

Currently, the transcoding option prioritize bitrate over resolution. For example, the transcoding options look like this for an example video:

1080p - 10mbps
720p - 8mbps
720p - 6mbps
720p - 4mbps

On some tvs, it does not look good to drop the resolution to 720p. I would rather prioritize resolution. For example, this would look better on many sets: 

1080p - 10mbps
1080p - 8mbps
1080p - 5mbps
720p - 3mbps

This is closer to how cloud based streaming services do it, if my research is correct.

JPKribs:
These don't set resolutions, only bitrate. Any resolution changes are dictated by the server. These labels are purely based on Jellyfin web's for the corresponding bitrates which I believe are more "estimates" based on bitrate.

But, these only control bitrate.

PR #2150, opened 2026-07-27
Register Socket for Commands

### LINKED ISSUE #408: Support Websocket Remote Control

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/408

~This is a bit of a smaller one.~ <- sorry

This registers for playback commands via the socket & `UserSession`. This should cover the bulk of commands that are available to use with some exceptions that are not feasible at this time.

What we're missing:

#### Possible

1. **Queue Commands** - `SetRepeatMode`, `SetShuffleQueue`, `SetPlaybackOrder`, `PlayInstantMix`, `PlayShuffle`. Queue / Shuffle PR would come first.
2. **Live TV Commands** - `ChannelUp`, `ChannelDown`, `Guide`. Would only work for Live TV and would need some kind of lookup for this.
3. **Message Command** - `DisplayMessage`. Toast changes first.
4. **App Navigation Commands** - `GoHome`, `GoToSearch`, `GoToSettings`. Would be usable via `routePublisher` but do we need this?

#### Not Possible

1. **Directional Commands** - `MoveUp/Down/Left/Right`, `PageUp/Down`, `PreviousLetter`/`NextLetter`, `Select`, `Back`, `SendKey`, `SendString`, `ToggleContextMenu`, `TakeScreenshot`. I don't think there is a clean/good way for us to do this?
2. **Player UI Commands** - `ToggleOsd`, `ToggleOsdMenu`, `ToggleStats`, `ToggleFullscreen`. This is possible via `VideoPlayerContainerState` but isn't super useful without the Directional Commands since this would just open these but then there is no way to interact with them.
3. **Volume Commands** - `VolumeUp/Down`, `SetVolume`, `Mute`, `Unmute`, `ToggleMute`. I know there are ways to do this via iOS but tvOS I don't think we can control these. I imagine casting to a device is more from device to tvOS so rather than iPadOS/iOS so I don't know if it's worth the effort.

### Random Cleanup

This makes some changes to `PlaybackBitrate` since there is a `setBitrate` command. This started with making it so we can go from `Int` to `PlaybackBitrate` to keep everything in check but while I was there I added more bitrates for audio since movies 128kbps doesn't make sense but that's a reasonable music bitrate. Finally, I removed the text for bitrate since it wont make sense for audio (1080p doesn't mean anything for audio) and to [clear up confusion](https://github.com/jellyfin/Swiftfin/issues/2132).

Cleaned up instances of this so `MediaSource` will show bitrates that make sense based on the type so that way we don't have way too many bitrates for videos.

### Videos

https://github.com/user-attachments/assets/234b6601-ba60-4f71-af8b-f79d74f7d9c5

https://github.com/user-attachments/assets/d1d32e60-3750-4849-960a-288165d790fb

https://github.com/user-attachments/assets/f395cb8e-1a72-4013-a8ff-7925ee342f2b

https://github.com/user-attachments/assets/4b0e0f91-089b-4ad7-ae45-412e4c1f8b63

https://github.com/user-attachments/assets/1b36a247-ae75-4f42-a456-9093408e05c9


LePips (changes requested):
If the media manager is just going to have the subscription on the user session, then just have the subscription on the user session and call the relevant manager commands instead.

That does remove messaging though, which I'm fine with for now. I have a local revamp of the toasting architecture locally that would deal with that.

JPKribs:
> If the media manager is just going to have the subscription on the user session, then just have the subscription on the user session and call the relevant manager commands instead.

Actually, moving this here I think we can fully close out https://github.com/jellyfin/Swiftfin/issues/408!

JPKribs:
Okay, here is current state:


https://github.com/user-attachments/assets/4b0e0f91-089b-4ad7-ae45-412e4c1f8b63


https://github.com/user-attachments/assets/1b36a247-ae75-4f42-a456-9093408e05c9

tvOS:

https://github.com/user-attachments/assets/8c41156f-c168-4cd5-9f32-56053c18517e

tvOS videos get big fast so more just showing the same thing works there.

LePips (approved):
Great work!

Shared/Services/UserSession/UserSessionManager+SocketCommands.swift
```diff
@@ -0,0 +1,234 @@
...
+            guard let index = generalCommand.arguments?["Index"], let index = Int(index) else { return }
+            mediaPlayerManager?.playbackItem?.selectedAudioStreamIndex = index
+        case .setMaxStreamingBitrate:
+            guard let bitrate = generalCommand.arguments?["Bitrate"], let bitrate = Int(bitrate) else { return }
+            mediaPlayerManager?.setBitrate(bitrate: PlaybackBitrate(for: bitrate))
+        case .setSubtitleStreamIndex:
+            guard let index = generalCommand.arguments?["Index"], let index = Int(index) else { return }
+            mediaPlayerManager?.playbackItem?.selectedSubtitleStreamIndex = index
+        case .displayContent:
+            guard let itemID = generalCommand.arguments?["ItemId"] else { return }
```
LePips, line 129: It is possible to navigate to an item during playback which we should prevent.

PR #2160, opened 2026-07-31
Focus Play Button on Episode Details on tvOS  

### LINKED ISSUE #2129: Request: Focus Play Button on Episode Details on tvOS

### Platform

tvOS

Thanks so much for the huge version 1.5 update for tvOS, the app so much better in so many ways now! I've just got one tiny annoyance that I keep running into in the new version.

On the episode details screen (the screen you get to by selecting an episode on Next Up or by selecting the episode description from within the shows overview/episodes screen), the show name is the default focused button. When selecting an episode on "Next Up", it would be preferable for the default button to be the Play button so that I don't have to click Down and then click Play when I'm going into an episode with the intention to play it. Because of Apple's default UI style, it's also no super obvious at first which button is focused so I've accidentally just clicked an episode in Next Up and then instinctively clicked again to play and get brought to the shows overview screen instead of playing the episode. In the case of a Next Up episode I can just click the default focused Play button on the shows overview screen which plays the next episode instead of having to go back, but it's still an extra click and a bit of an unintuitive interaction between the screens and inconsistency between the default focused button on each of these screens.

quangDecember:
fixing #2129 
# Debug

`RegularSimpleHeaderContentGroup` keeps losing focus to `PosterGroup` underneath

# Cause

prefersDefaultFocus doesn't work inside ScrollView on tvOS — focus is still landing on posters because defaultFocus targets a non-focusable header container. 
defaultFocus was aimed at the non-focusable header container (itemView-header). The focus engine couldn’t land there, so the season PosterGroup won.

# Video

Simulator Screen Recording - Apple TV 4K (3rd generation) (at 1080p) - 2026-07-31 at 12 06 16

**AI Disclosure**: 
Cursor fixed it after Claude Code failed

P/s: the pull request template seems missing?


quangDecember:
seems like we don't need the changes for all the headers. Let me know if you need me to reverse that or limit the changes to `itemType` `episode`

LePips (approved):
I've hopefully abstracted away focus for `ContentGroup` views into `FocusCoordinator`, as I found the manual `FocusState/Binding` passing a bit annoying. We might be able to use this coordinator for general focus-related matters with `ContentGroup` views.

PR #2163, opened 2026-08-01
Configuration for Poster Quality

### LINKED ISSUE #2158: Episode preview images have a low resolution

### Description of the bug

On show pages, episode preview images are displayed in a low resolution. (This is especially noticeable with animated shows)

In the old Swiftfin tvOS version (1.0.1), the resolution is higher.

The preview image for the same episode on the Home screen, however, looks better.

### Platform

tvOS

### Application version

1.5

App Store

### Device information

Apple TV 4K (1st gen)

### OS version

tvOS 26.5

### Jellyfin server version

10.11.11

### Screenshots or videos

Episode preview images are low-resolution: (It's a little difficult to see in the pictures)
Image

The preview image for the same episode looks better on the home screen:
Image

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/2158

Back in https://github.com/jellyfin/Swiftfin/pull/1648 we limited poster sizes to `90` from the unconstrained `nil` / original size. On iOS, iPadOS, and a lot of tvOS this is nothing but upside. However, some folks with larger TVs have commented that the posters are a downgrade on larger/higher resolution TVs or just larger screens in general. 

This PR adds a flag to the posters to allow un-reduced images. Default is still `performance` for the 90 quality but allows removing this with `nil` if the user so chooses.

### Setting

Simulator Screenshot - Apple TV 4K (3rd generation) - 2026-08-01 at 02 50 42


LePips (changes requested):
Are you sure that it's the quality parameter rather than the size parameters? Quality at a setting of 90 is near imperceivable compared to original.

Regardless, settings like this would not be the way to go.

JPKribs:
> Are you sure that it's the quality parameter rather than the size parameters? Quality at a setting of 90 is near imperceivable compared to original.

In testing, it does make a difference BUT I think I figured it out it has more to do with the original image format via conversion weirdness. webp is handled weirdly while jpg / png is not. More likely server than us.

> Regardless, settings like this would not be the way to go.

Makes sense! I'm going through new issues to see what is a quick fix so no harm no foul

PR #2167, opened 2026-08-03
Recently Played

### LINKED ISSUE #1939: Recently Played section

### Platform

iOS, tvOS

Having a Recently Played section on the homepage of Swiftfin to list TV Shows and/or Movies that have been marked as”Played” within a user settable timeframe (e.g. user defines period in Settings a Last Played within 30 days — similar to Days in Next Up setting).

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1939

Adds a `Recently Played` section to the home view. Disabled by default but enableable from the the `Home` settings section. This is basically just a copy-paste of the `Recently Added` with a different `SortBy`. The primary difference is I am not registering this in the cinematic rows because I don't think it makes sense to use it.

Other than that, uses `StateAdapter` for the `.alert` setting for `Days in Next Up` while I'm touching that view. Also, when I added a `RecentlyPlayed` I noticed that `RecentlyAdded` was title case so I changed that and updated the called to use the `localizedCapitalized` since that's been the localization structure so far. Happy to revert that if we don't see a value.

<details>

<summary>Screenshot</summary>

Simulator Screenshot - iPhone 17 Pro - 2026-08-03 at 12 24 41

</details>


JPKribs:
This should be a small one. Not sure what the demand is for something like this so defaulting to off. I kept it series and movies because adding episodes make it get really cluttered and the portrait meant all episodes shared the same portrait. I can revert that to episodes as well if we think that makes sense to keep. This is what that looks like with episodes included:

Screenshot 2026-08-03 at 12 35 58

I supposed we could just remove the type as well for like home video & music videos as well. Again, feels like it may get cluttered but whatever makes the most sense!

LePips (approved):
Simple enough!

Issue #2149, opened 2026-07-26
1.5 regression: orientation locked to portrait when returning to video player from background or PiP


AYDJI:
## Describe the bug

**Version 1.5 introduced a regression** of the orientation lock fix from #1898.

When playing a video in landscape and backgrounding the app (or using PiP), returning to Swiftfin locks the video player into portrait mode. Rotating the device has no effect — the player stays stuck in portrait until the app is force-quit and reopened.

## Reproduction steps

1. Open Swiftfin and start playing a video
2. Rotate to landscape — video fills the screen correctly
3. Background the app (or trigger PiP) and rotate to portrait
4. Return to Swiftfin
5. Try rotating back to landscape — **orientation is locked to portrait**

## Root cause

PR #2059 (`App state consolidation, remove AppDelegate`, merged Jun 25, 2026) deleted `AppDelegate.swift`, which contained the `application(_:supportedInterfaceOrientationsFor:)` override from PR #1959 (the original fix for #1898).

Without the AppDelegate, UIKit has no way to determine the correct orientation mask when the app transitions to foreground. The Transmission SwiftUI swizzling alone does not cover this path.

## Platform

iOS

## Application version

1.5 (regression — was working in 1.4.1)

## Device information

iPhone

## Related

- PR #1959 — original fix (restored in PR #2148)
- PR #2059 — removed AppDelegate, causing the regression
- PR #2148 — proposed fix re-adding AppDelegate

LePips:
What OS version? In the future, please disclose AI usage.

AYDJI:
iOS 18.7.7 (from the original report on #1976 which I'm also experiencing).

Apologies for not disclosing sooner — this issue and the associated PR #2148 were created with assistance from an AI coding tool. I've now disclosed this in the PR body as well.

AYDJI:
Correction: the iOS version where this is reproduced is 26.5.2, not 18.7.7 (apologies for the incorrect guess).

PR #2169, opened 2026-08-03
Fix video player freezing when the window may rotate

### LINKED ISSUE #2165: Video player is locked to portrait on iPhone since #2148 (AppDelegate restore)
# Video player is locked to portrait on iPhone since #2148 (AppDelegate restore)

## Summary

On iPhone, the video player no longer rotates to landscape on `main`. Release 1.5 rotates correctly. The regression comes from the orientation handling added in #2148 (`c5153d67`), which takes precedence over `PreferencesView`'s existing mechanism and reports `.portrait` for the window in every case that matters.

## Environment

- Device: iPhone 13, iOS 26.5.2
- Build: stock `main` @ `90321fd7`, Debug, no local modifications
- Reproduced on `main`; **not** reproducible on tag `1.5`

## Steps to reproduce

1. Build stock `main` to a physical iPhone
2. Play any item
3. Rotate the device to landscape

**Expected:** player rotates to landscape (1.5 behavior)
**Actual:** player stays portrait; rotation never occurs

## Cause

At 1.5 there is no `AppDelegate` and no `application(_:supportedInterfaceOrientationsFor:)`. Orientation is resolved through the view-controller chain, where `PreferencesView`'s swizzle answers:

```swift
// PreferencesView/Sources/PreferencesView/UIViewController+Swizzling.swift
@objc
func swizzled_supportedInterfaceOrientations() -> UIInterfaceOrientationMask {
    search()?._orientations ?? .all
}
```

The player route sets `.supportedOrientations(.allButUpsideDown)`
(`Shared/Coordinators/Navigation/NavigationRoute/NavigationRoute+Media.swift`), which lands on
`UIPreferencesHostingController._orientations`, so the chain returns landscape and the player rotates.

#2148 added a window-level delegate callback, which **overrides** that chain:

```swift
// Swiftfin/App/AppDelegate.swift
guard UIDevice.isPhone else {
    return .allButUpsideDown
}

if let presentedViewController = window?.rootViewController?.presentedViewController,
   let preferencesHostingController = presentedViewController as? UIPreferencesHostingController
{
    return preferencesHostingController.supportedInterfaceOrientations
}

return .portrait
```

The player is not presented as a bare `UIPreferencesHostingController`. It is presented through
Transmission, so `window?.rootViewController?.presentedViewController` is:

```
PresentationHostingController<ModifiedContent<Optional<ViewControllerRepresentableAdapter<UIViewController>>, PresentationBridgeAdapter>>
```

The `as?` cast fails, control falls through to `return .portrait`, and the player is locked to portrait
on iPhone.

## Evidence

Instrumenting the delegate to walk to the topmost presented controller and log what it finds, during playback:

```
[orientation] topmost: PresentationHostingController<ModifiedContent<Optional<ViewControllerRepresentableAdapter<UIViewController>>, PresentationBridgeAdapter>> -> All But Upside Down
```

The correct mask is available on that controller — the delegate simply never consults it, because it only
matches one concrete type.

## Note for whoever fixes this

Returning the topmost presented controller's `supportedInterfaceOrientations` unconditionally is **not**
a sufficient fix. Tried locally, and while it produces the right mask, the value then alternates between
`.all` and `.allButUpsideDown` across the presentation lifecycle. On device that churn tore down and
rebuilt the player view during presentation, and `VLCVideoPlayer` frequently never got constructed —
black screen, no VLC state events, no stream request to the server. Playback stability appears to depend
on this callback returning a stable value while the player is being presented.

A fix likely needs to either resolve the mask through the presentation chain in a way that is stable
during presentation, or drive it explicitly via `setNeedsUpdateOfSupportedInterfaceOrientations` /
`UIWindowScene.requestGeometryUpdate` at the point the player is presented, rather than answering
per-query at the window level.

## Related

- #2147 (`77c5f59d`) removed the `AppDelegate`
- #2148 (`c5153d67`) restored it and introduced this behavior
- #2149 is the issue #2148 fixed (orientation lock when returning from background/PiP). This report is a
  *separate* regression introduced by that fix: the player never rotates at all on iPhone, background/PiP
  not required.
- Prior art in the same area: #1898, #1900, #1976, #1709

---

*Disclosure: this report was written with the assistance of an AI coding tool. The reproduction, the
`main` vs `1.5` comparison, and the logged controller type/mask were all observed on a physical device.*

dwertle:
Fixes #2165.

## The visible bug

On iPhone, the video player cannot rotate to landscape on `main`. It rotates correctly on 1.5.

## Why the current code locks it

`AppDelegate` reports `.portrait` for the window unless the root's immediate `presentedViewController` casts to `UIPreferencesHostingController`. The player is presented inside Transmission's hosting controller, so that cast never succeeds and the window is clamped to portrait.

The obvious fix — report the player's actual mask — **breaks playback**. I tried three variants (walk to the topmost presented controller, search descendants for the preferences controller, and a static `.allButUpsideDown`); all three left the player unable to start: black screen, no VLC state events, no stream requested from the server. Details and results table in #2165.

## The actual cause

The clamp was hiding a second defect.

`VideoPlayerContainerView` erases the player to `AnyView` and hosts it once in `makeUIViewController`, and `updateUIViewController` does nothing. After creation, the hosted view's only route to a new playback item is the `@EnvironmentObject` subscription captured at that moment. That subscription does not survive every geometry pass, and permitting rotation reliably triggers one.

Once it is lost the hosted view never re-evaluates. Instrumenting both sides shows the manager working perfectly while the child is frozen:

```
PARENT received playbackItem: true     <- publisher fires, parent observes it
Playing new item                       <- item built, URL valid
PARENT received state: playback        <- gate condition satisfied
(VLCPlayerView.body never re-runs)     <- player never constructed
```

`/Sessions/Playing` is reported to the server, so the session looks live, but no stream is ever requested. That also explains intermittent black screens in portrait: the same freeze, just triggered less often. In one user's logs, 8 of 9 playback attempts on device never constructed the player, all in portrait.

## The fix

1. Re-host the player content in `updateUIViewController`, giving it an input path that does not depend on an observation that can be dropped.
2. Let `AppDelegate` report the full mask, so `PreferencesView` resolves orientation per screen as it did before the delegate was reintroduced.

(1) is what makes (2) safe. With only (2), playback breaks as described above.

## Testing

iPhone 13, iOS 26.5.2, Swiftfin player, direct play and transcode:

- Movie and Live TV / IPTV playback
- Rotation to landscape and back, mid-playback
- Background and return mid-playback (the #2149 scenario the delegate was restored for)
- Seeking and scrubbing, which exercise the new `updateUIViewController` path repeatedly

Not tested: iPad, tvOS (unaffected — the delegate path is `UIDevice.isPhone` only), PiP (Native player only).

---

*Disclosure: this change was developed with the assistance of an AI coding tool. All diagnosis and testing was done on a physical device.*

### DIFF
--- a/Shared/Views/VideoPlayer/VideoPlayerContainerView/VideoPlayerContainerView.swift
+++ b/Shared/Views/VideoPlayer/VideoPlayerContainerView/VideoPlayerContainerView.swift
@@ -74,7 +74,9 @@ extension VideoPlayer {
         func updateUIViewController(
             _ uiViewController: UIVideoPlayerContainerViewController,
             context: Context
-        ) {}
+        ) {
+            uiViewController.refreshPlayerContent()
+        }
     }
 
     // MARK: - UIVideoPlayerContainerViewController
@@ -218,12 +220,30 @@ extension VideoPlayer {
             return view
         }()
 
+        private func makePlayerContent() -> AnyView {
+            PlayerContainerView(player: player)
+                .environmentObject(containerState)
+                .environmentObject(manager)
+                .eraseToAnyView()
+        }
+
+        /// Re-hosts the player content.
+        ///
+        /// The hosted view is created once, so its only route to a new playback
+        /// item is the `@EnvironmentObject` subscription baked in at creation.
+        /// That subscription does not survive every geometry pass — notably
+        /// when the window is allowed to rotate — and once it is lost the view
+        /// is frozen on stale state with no way to recover: the manager reaches
+        /// `.playback` with an item ready and the player is never built.
+        /// Pushing fresh content gives it an input path that does not depend on
+        /// observation.
+        func refreshPlayerContent() {
+            playerViewController.content = makePlayerContent()
+        }
+
         private lazy var playerViewController: HostingController<AnyView> = {
             let controller = HostingController(
-                content: PlayerContainerView(player: player)
-                    .environmentObject(containerState)
-                    .environmentObject(manager)
-                    .eraseToAnyView()
+                content: makePlayerContent()
             )
             controller.disableSafeArea = true
             controller.automaticallyAllowUIKitAnimationsForNextUpdate = true
--- a/Swiftfin/App/AppDelegate.swift
+++ b/Swiftfin/App/AppDelegate.swift
@@ -23,16 +23,11 @@ class AppDelegate: NSObject, UIApplicationDelegate {
         supportedInterfaceOrientationsFor window: UIWindow?
     ) -> UIInterfaceOrientationMask {
 
-        guard UIDevice.isPhone else {
-            return .allButUpsideDown
-        }
-
-        if let presentedViewController = window?.rootViewController?.presentedViewController,
-           let preferencesHostingController = presentedViewController as? UIPreferencesHostingController
-        {
-            return preferencesHostingController.supportedInterfaceOrientations
-        }
-
-        return .portrait
+        // UIKit intersects this window-level mask with the top view
+        // controller's own `supportedInterfaceOrientations`, which
+        // `PreferencesView` already resolves per screen. Reporting a narrower
+        // mask here clamped that intersection and locked the video player to
+        // portrait on iPhone.
+        .allButUpsideDown
     }
 }


LePips:
Frankly, I can't reproduce the original issue from #2149 with the `AppDelegate` removed, I had been fine with the re-patching as I would have expected the behavior to be unchanged/be an inconsequential fix. I can rotate just fine on an iOS 26+ on main with the `AppDelegate` removed and will revert it for now until there is something I can reproduce.

Lastly, thank you for disclosing AI usage but that does not excuse walls of text in issues, PR descriptions, git commits, or comments that really don't make sense. Same for these code changes. I will no longer accept AI changes like these.

dwertle:
Ok, sorry about that. Thanks for your comment, I'll be more careful next time - still learning things as I go. May I ask your process for developing? I'm prompting Claude or Codex within VS Code to get me through things, but I'm curious if there's a more professional approach.

PR #2173, opened 2026-08-04
Fix consecutive tvOS deep links reopening first route

potelux:
Currently tvOS uses .fullScreenCover instead of .sheet due to an Apple bug.  However, the .fullScreenCover has a bug where it only changes what is being displayed when the value goes from nil to some value.  In other words, setting some deeplink value and then setting a new deeplink value will not cause the .fullScreenCover to render something new.

The workaround is, if .fullScreenCover is non-nil, first clear it to nil.  Then, when the cover is dismissed, set the new deeplink value.

TLDR;

Problem: non-nil -> non-nil doesn't work
Solution: non-nil -> nil -> non-nil


JPKribs:
Hey! Apologies, Deep Links aren't something I've used before. Could you walk me through your steps for testing/working on this so I can vet from my end and see if I have any questions/issues?

potelux:
Sure thing!  I currently am testing it through Home Assistant.  If you have a Home Assistant setup, these are the steps to test:

0. Ensure that the Apple TV is configured as a media player in Home Assistant
1. Create a Script or Automation
2. In the `Actions` section, add a `Play specified media` action
3. Select the Apple TV as the target device
4. In the media section select `Manual entry`
5. For the `Media content type` enter "url"
6. For the `Media content ID*` enter "swiftfin://SERVER-ID/USER-ID/item/ITEM-ID"
7. Ensure the Apple TV is on, and Swiftfin has been opened at least once, then run the action

To retrieve the necessary IDs, open Jellyfin in the browser and select any piece of content.  The URL will end in `id=ITEM-ID&serverId=SERVER-ID`.  To find the `USER-ID`, select the user icon in the upper-right corner, then select `Profile`.  The URL should end in `userId=USER-ID`.

EDIT: Note that the user corresponding to the `USER-ID` must already be logged into Swiftfin on the TV.

potelux:
Before using my branch, you can recreate the bug by following the above steps twice.  Change the `ITEM-ID` only.  You will see the screen flicker as if loading a new deep link, but instead the first item will reload.  Run the second `ITEM-ID` deep link as many times as you'd like to confirm that it will always reload the first link instead.

potelux:
> Hey! Apologies, Deep Links aren't something I've used before. Could you walk me through your steps for testing/working on this so I can vet from my end and see if I have any questions/issues?

@JPKribs - Let me know if there is anything else I can do to help you test this.

JPKribs:
I can understand how you came to this solution but I don't think this is the right route. The core issue is that we create a `fullScreenCover` binded to the item that we then can't update because we try to update on top of the same modal. Your solution works but if I move quickly I can get it to break. Instead, could you test this by updating:

#### /Shared/Coordinators/Navigation/NavigationCoordinator.swift

```swift
//
//
// Copyright (c) 2026 Jellyfin & Jellyfin Contributors
//

import SwiftUI

@MainActor
final class NavigationCoordinator: ObservableObject {

    @Published
    var path: [NavigationRoute] = []

    @Published
    var presentedSheet: NavigationRoute?
    @Published
    var presentedFullScreen: NavigationRoute?

    func push(
        _ route: NavigationRoute
    ) {
        let style = route.transitionStyle

        switch style {
        case .push:
            path.append(route)
        case .sheet:
            presentedSheet = route
        case .fullscreen:
            withAnimation {
                presentedFullScreen = route
            }
        }
    }
}
```

This just removes the build flags and uses the same push for iOS and tvOS instead of presenting on the same sheet. This seems to work for me in testing but please confirm if this works for you as well. I will defer to @LePips on if this is the appropriate solution but I would lean more towards this than adding another mechanical part of the dismissal process.

potelux:
> This just removes the build flags and uses the same push for iOS and tvOS instead of presenting on the same sheet. This seems to work for me in testing but please confirm if this works for you as well. I will defer to @LePips on if this is the appropriate solution but I would lean more towards this than adding another mechanical part of the dismissal process.

I agree that this is the optimal solution, but I was under the impression that the `#if os(tvOS)` branch existed to deal with a bug in Apple tvOS.

EDIT: I see that [tvOS 26.1 release](https://developer.apple.com/documentation/tvos-release-notes/tvos-26_1-release-notes) specifically references the bug as fixed:
`Fixes: sheet uses a modal sheet presentation style instead of full-screen presentation style.`
This is above my pay-grade, but can all code related to that workaround be removed? 

EDIT 2:  I noticed that Swiftfin has its deployment target set to 26.0.  Should the deployment target be bumped up?  Or should this workaround be scoped to 26.0? @JPKribs

JPKribs:
This appears to no longer be a workaround, but Apple's intended implementation of .sheet. Changing the patch as the build target does not resolve the sheet changes.

LePips:
Yeah, the current SwiftUI tvOS behavior looks like it's how it will be going forward which should allow cleanup in `NavigationInjectionView`. I'm also fine with bumping the minimum OS, as I had just been doing the major versions for ease.

potelux:
If I am understanding correctly, we want to use the .sheet modal for tvOS now.  

First, I will bump the version: https://github.com/jellyfin/Swiftfin/pull/2203

Then I'll update this branch

potelux:
FYI, this workaround is also present in Shared/Coordinators/Navigation/NavigationInjectionView.swift

JPKribs:
Sorry, there's a bit more under the surface I think needs to be sorted out here. Testing some stuff out then I can report back

LePips (approved):
The issue is that tvOS has a different presentation stack. It is deliberate that pushes don't append the `NavigationStack`'s path, instead presenting on top of current content. For received deep links, the top-most coordinator would need to present the item instead.

This touches the "navigation issue" I've thought about for tvOS, where navigation can cause a deep tree on tvOS which can be annoying to navigate. I have some thoughts on how to remedy that. This also had to make me think about how macOS would present items, which may have its own stack. This may also get more complicated as non-item/library deep links are implemented.

Tested with `xcrun simctl openurl booted`.

PR #2182, opened 2026-08-06
Server Backups & Restore

JPKribs:
### Summary

This enables server backups and restoring. Straight forward with two changes:

~1. Disallow restoring from anything that is prior to the SDK Jellyfin version just to be safe.~
2. Sign the user out on restore. I found this to be a better experience than having Swiftfin work fine until it hits the next 503. Because this restarts the server, there isn't a great way to track unless we are able to pulse until the server stops sending a 503? My thinking is you back out to the sign in and if you immediately try again you are met by a 503. Additionally, there is a world where the backup you are restoring from doesn't contain your user account anymore so the login would give you a login error instead of 503ing until we start 401ing.

### Video

https://github.com/user-attachments/assets/bc15b4c1-0e67-4182-944d-b5baf26f3fc5


Shared/Extensions/JellyfinAPI/JellyfinClient.swift
```diff
@@ -11,6 +11,21 @@ import Get
 import JellyfinAPI
 import UIKit
 
+extension JellyfinClient.Version {
+
+    init(backupVersion: String) {
```
JPKribs: This is because the backup versions are like:

`10.11.12.0`

And our logic for parsing would read that in as `11.12.0` when it should be `10.11.12`

LePips: I don't believe so, as instead the version parsing is hardcoded to 3 components. I would be fine to relax that requirement, instead taking the first 3 components and defaulting to 0 in the event there are less than 3.

https://github.com/jellyfin/jellyfin-sdk-swift/blob/37a2f5028bd24689b772effb559f72ec5388f021/Sources/Extensions/Info.swift#L49

Shared/Strings/Strings.swift
```diff
@@ -186,6 +186,14 @@ internal enum L10n {
...
   internal static let backdropImageDescription = L10n.tr("Localizable", "backdropImageDescription", fallback: "A wide background image displayed behind the item details on its media page. Multiple backdrops can be added and will cycle or be selected randomly.")
+  /// Backup
+  internal static let backup = L10n.tr("Localizable", "backup", fallback: "Backup")
+  /// Backups
+  internal static let backups = L10n.tr("Localizable", "backups", fallback: "Backups")
+  /// Create a snapshot of your server which can include the database, metadata, subtitles and trickplay.
+  internal static let backupsDescription = L10n.tr("Localizable", "backupsDescription", fallback: "Create a snapshot of your server which can include the database, metadata, subtitles and trickplay.")
+  /// This backup was created on an older version of Jellyfin and will not restore correctly without first downgrading to that version.
+  internal static let backupVersionWarning = L10n.tr("Localizable", "backupVersionWarning", fallback: "This backup was created on an older version of Jellyfin and will not restore correctly without first downgrading to that version.")
```
LePips: I don't find this language to be that clear as written, but this is also used by just comparing against the SDK's generated version. Can remove this and just use the same string that is used for server mismatch if we really want a warning here.

LePips (changes requested):

Shared/Extensions/JellyfinAPI/JellyfinClient.swift
JPKribs: ~You are correct. I made the assumption without debugging when it didn't match but checking it, this just returns 0.0.0. I can leave a TODO for now then tie that into the next SDK change if that makes sense to you?~

I'm dumb ignore this. We don't even need to convert this because the only reason I was converting this was the version test. If I'm dropping the version test I don't need this.

Shared/Strings/Strings.swift
JPKribs: That's a good point that I'm assuming they're on a server that matches the SDK. TBH, I don't know if this is something that matters. I was trying to be overly cautious but it looks like restoring from version mismatches is intentional and server has their own catch here:

https://github.com/jellyfin/jellyfin/blob/34cc90f5e285a0b896e07d0bc195e4cff768cb56/Jellyfin.Server.Implementations/FullSystemBackup/BackupService.cs#L116

So I'm less worried about this than I originally was.

LePips (approved):

PR #2194, opened 2026-08-09
Maximum Resolution Setting

### LINKED ISSUE #1703: Set Max Bitrate for 1080p Devices

At the current moment neither Swiftfin or Infuse seem to offer this, my Apple TV is an older 1080p only version and whenever I go to stream any 4K content I end up with a black screen with choppy audio, being able to transcode 4K content down to 1080p would be a great feature as it would certainly save me from having to upgrade the Apple Tv boxes likely sending them to e-waste. 

Thanks for reading!

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/1703

My quest to resolve all the easy issues continues!

This PR adds a new setting for Max Resolution. When enabled, this creates a `CodecProfile` that limits the video dimensions on height. So, 1080p sets the maximum height to 1080. I originally had an idea that this would set both height and width maximum but I settled on just the height since I assume some people will have some oddly shaped media and I want to avoid the `1080x1920`, `1080x1770`, `1080x990`, etc. where height seems a lot more universal than width.

This also does a bit of clean up on strings in the session supplement since I know I changed the `a x b` in https://github.com/jellyfin/Swiftfin/pull/2184 I'd like to keep that consistent and some items there I use `"\(value)"` where `.description` works as well.

#### Swiftfin

https://github.com/user-attachments/assets/0a84560d-4920-446f-bf4b-ef656355707b

#### From Server

Screenshot 2026-08-08 at 21 49 39


JPKribs:
Tiny change. Feedback I got was "Maximum resolution: Maximum" might've been using the word `Maximum` too much. I've updated the label to be `Original` so that's the one difference vs the video

Shared/Views/SettingsView/PlaybackQualitySettingsView.swift
```diff
@@ -29,18 +37,24 @@ struct PlaybackQualitySettingsView: View {
...
+                PlatformPicker(
+                    L10n.maximumResolution,
+                    selection: $appMaximumResolution
                 )
-                #else
-                ListRowMenu(
+            } header: {
+                Text(L10n.resolutionDefault)
+            } footer: {
+                VStack(alignment: .leading) {
```
LePips: I believe having the `VStack` is a holdover from when the learn button still had to be manually added. Should no longer be necessary and can be removed from other sections if necessary.

Shared/Strings/Strings.swift
```diff
@@ -1366,6 +1370,10 @@ internal enum L10n {
   internal static let resetSettingsDescription = L10n.tr("Localizable", "resetSettingsDescription", fallback: "Reset Swiftfin user settings.")
   /// Are you sure you want to reset all user settings?
   internal static let resetSettingsMessage = L10n.tr("Localizable", "resetSettingsMessage", fallback: "Are you sure you want to reset all user settings?")
+  /// Default resolution
+  internal static let resolutionDefault = L10n.tr("Localizable", "resolutionDefault", fallback: "Default resolution")
```
LePips: Just have the section titles as `Resolution` and `Bitrate` instead of having "default".

LePips (changes requested):
My comment from #1703 was to instead restrict the existing values based on the device's support, and neither does this.

I'm fine with the separate settings for now, but at some time would want to think about how to have the overall quality settings be consolidated for improved UX. Especially as having a random splatting of bitrate values doesn't make sense.

Shared/Strings/Strings.swift
JPKribs: Simulator Screenshot - Apple TV 4K (3rd generation) (at 1080p) - 2026-08-10 at 14 58 19

This is where I landed for this. Combine the sections into a single quality section. My thinking is this is eventually:

Resolution:
Video Bitrate
Audio Bitrate

> I'm fine with the separate settings for now, but at some time would want to think about how to have the overall quality settings be consolidated for improved UX. Especially as having a random splatting of bitrate values doesn't make sense.

My reason for splitting this out is https://github.com/jellyfin/Swiftfin/issues/2132 where I worry combining bitrate and resolution into a single item opens a can of worms. 1080p @ 8mbps for H.264 vs AV1 are very different in terms of quality. *Or so I'm told I can't personally see it*, but I'd rather just let this be variable for those who care about this level of granularity and have unset defaults for the average user where this isn't a concern

LePips: I like the combining!

LePips (approved):

PR #2188, opened 2026-08-08
Common Issues Documentation

JPKribs:
### Summary

Creates a document for [common issues](https://github.com/jellyfin/Swiftfin/issues/2183#issuecomment-5219810046) that we receive in Swiftfin. The goal of this is to help new contributors who want to field new issues or Matrix troubleshooting to give them a central place for commonly seen items. This also works to alleviate new issues that come in by filtering out common issues through documentation.

This also updates the Jellyfin badge `10.11` -> `12.0` since I forgot to update that with the SDK again.

PR #2195, opened 2026-08-09
Long Press to Always Gesture Unlock

### LINKED ISSUE #2185: Gesture Lock Cannot be Disabled if the Long Press Gesture is Disabled
I'm not sure if this is a bug or if I just haven't figured out how to do it. After clicking 'Gesture lock', I can't seem to disable it.

I have been noticing this issue since at least version 1.5.

iPhone 12 Pro, iPad Pro 11-inch 4th  |  iOS, iPadOS, 18.7.8   |  TestFlight  1.6 (Build 15)  |  Jellyfin Server 10.11.11

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/2185

If you gesture locked while also using the long press for `None` or `Speed` you had no way out of the Gesture LOck. This changes some ordering around so:

1. If gesture locked, show the label "Press and hold to unlock" as a clear action
2. If gesture locked, have an immediate guard/action at the top where a long press unlocks so this is consistent regardless of the long press action

The outcome is that gesture lock is either initiated via the long press OR the button but it's always disengaged via a long press. The label *should* hopefully alleviate that question in the future but it should also be included in:

https://github.com/jellyfin/Swiftfin/pull/2188

#### Please note that the long press is consumed by the playback speed!

https://github.com/user-attachments/assets/35ecb1bd-4d18-45e2-80b9-e47b1630aa20


Swiftfin/Views/VideoPlayer/Gestures/VideoPlayerContainerView+TapGesture.swift
```diff
@@ -200,30 +200,32 @@ extension VideoPlayer.UIVideoPlayerContainerViewController {
...
-                    L10n.gesturesLocked,
-                    systemName: VideoPlayerActionButton.gestureLock.systemImage
-                )
-            }
+            containerState.toastProxy.present(
+                L10n.gesturesLocked,
+                systemName: VideoPlayerActionButton.gestureLock.systemImage
+            )
         case .playbackSpeed:
-            guard checkGestureLock() else { return }
```
JPKribs, line 226: Since this is guarded at the top, I didn't think we needed this here anymore. I am still calling this func for the double tap functions.

LePips (approved):

Issue #2186, opened 2026-08-08
Accent color fails to apply correctly to 'Continue' progress.


tobylibo:
I'm not sure if this is a bug. If it's not, and is actually by design, please feel free to ignore this issue.

Regardless of Accent Color the set for ②, the color of ① remains white.
Personally, I feel the style isn't consistent.

iPhone 12 Pro, iPad Pro 11-inch 4th | iOS, iPadOS, 18.7.8 | TestFlight 1.6 (Build 15) | Jellyfin Server 10.11.11

<img width="1146" height="1250" alt="Image" src="https://github.com/user-attachments/assets/94bbb8a6-dbb4-4a96-a877-0f1c22bfbc64" />

PR #2197, opened 2026-08-10
Use Accent Color for Landscape Posters Progress

chickdan:
Making the Landscape poster align with Portrait and Square by re-coloring the progress indicator with the accent color.

I've also noticed the Landscape progress indicator is styled quite differently. Should it look flat and square like it does on the Portrait poster type? If so I can tackle that in this PR as well.

### Screenshots
| Before | After |
|--------|--------|
| IMG_9042 | IMG_9041 |


tobylibo:
Resolved #2186

Thanks

LePips (approved):

PR #2202, opened 2026-08-11
Convert Item `ActionButtonHStack` to Mirror Video Player

### LINKED ISSUE #2090: Allow selecting video quality, audio and subtitle tracks before playing the media

### Platform

iOS

Currently, we can only select and change audio and subtitle tracks when playing the media. It would be nice if we can do so through the item page before playing the media as well, along with changing video quality. This applies to both iOS and tvOS.

Streamyfin interface as an example:

Image

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/2090

The original goal for this PR was to add the ability to select tracks & quality into our same Version menu:

https://github.com/user-attachments/assets/4c219a23-c325-44bd-817a-b4ec690220da

This led me down a rabbit hole...

Before, the version selector was on the same row as the `PlayButton` which looked fine pre-liquid glass but was now just looked off because 1) it was not a capsule and 2) it would often be a different size than the `ActionButtons` below it. So, I made the decision to move this to the `ActionButtonHStack` which led to its own host of issues about positions and sizing and getting everything to fit since we're accruing a lot of buttons here. So I had a genius solution! What if, I just stole all the same work that was already done for the `VideoPlayerActionButtons`!

This should roughly mirror `VideoPlayerActionButtons` now where, when there is not enough roome (iOS: 4, tvOS: 5) we should move the buttons into a menu. For tvOS, this menu is just at the end of the row (just like the Video Player's handling of this) but for iOS, these buttons go into the toolbar:

Settings

Anything that overflows from the bar buttons ends up at the top of the menu in the same fashion as the player:

Buttons

So this allows customization as always menu or menu on overflow.

### Other Cleanup

- Adds more spacing for tvOS simple view as right now the button focus covers the `MetadataHStack`
- Makes the Action Buttons a capsule to match the `PlayButton` so it looks more inline with eachother
- Remove the settings to turn off and on editing. The idea (to be) behind those was the ability to remove the accidental edit or deletion. Allowing customization to the action buttons that are present does this same thing and let's us just purely use the canDelete, canEdit flags on the user/item instead of using our own.
- Move the delete outside the `Metadata` editing view. Users can have permission to delete but not edit *(not sure who would do that but...)* so this moves it outside. This also unifies the iOS and tvOS implementation of deletion into an `ActionButton` so it's easier for maintenance/changes.
- Moves the toolbar menu into the `ActionBar` struct. Not positive if this is bad practices but this enables the menu/overflow to be managed from one place rather than passing thing between views/structures.

### tvOS

<details>

<summary>New Button Spacing</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) (at 1080p) - 2026-08-11 at 12 11 26

</details>

<details>

<summary>Version Selection</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) (at 1080p) - 2026-08-11 at 12 11 32

</details>

<details>

<summary>Menu - All Intentionally Menu</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) (at 1080p) - 2026-08-11 at 12 11 37

</details>

<details>

<summary>Enhanced View</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) (at 1080p) - 2026-08-11 at 12 28 29

</details>

<details>

<summary>Menu with 1 Playback as the Overflow</summary>

Simulator Screenshot - Apple TV 4K (3rd generation) (at 1080p) - 2026-08-11 at 12 28 40

</details>


JPKribs:
@LePips Currently in draft I've been screwing around with this for a bit so it's mostly ready but I wanted to get this out in front of you in case this unwanted. My motivation was the Video Player Action Buttons since adding/changing those is a lot easier than the `ItemView` onces. tvOS is like a dead copy but iOS I'm not sure if the toolbar menu for overflow is bad practice.

Let me know if this isn't wanted and I can close out!

LePips (commented):
I have long been thinking about what to do as we begin to add more actions, especially as different media might have different actions so it becomes very dynamic. Balancing the variety of customization we should allow vs putting the presentation on the user is what I've been thinking about since this PR has been made.

LePips (approved):
I've come around to be good with this. Although, I wouldn't want this to be too complicated with an abundance of actions depending on the media, like recording for example.

PR #2203, opened 2026-08-11
Bump tvOS deployment target to 26.1

potelux:


JPKribs:
As I mentioned in the other thread for this. 26.1 and above **does not resolve the original sheet issue**. The messaging from Apple around this was misleading. The current implementation of the sheet, where it is not a full screen modal, appears to be the intended functionality from Apple. Anything full screen would still need to use the full screen cover. If I had to guess, I believe Apple is trying to use a similar modal structure to iPadOS, where the sheet only populates the middle of the screen.

If you are able to test this and I am incorrect please let me know. Last week I tried building against every build versions for 26 and the only way to get a full screen was to either push or use the full screen cover.

If we otherwise don't have a reason for moving to 26.1 I would personally vote against it

JPKribs:
I'm going to close this out as I don't see any value in chasing the batch versions for build. In testing, the reason we originally sited for 26.1 is not applicable. Apple's explanation was misleading in that it sounded like the `.sheet` not being full screen was a bug but this appears to be intended behavior as it mirrors iPadOS.

I think any work related to the routing troubles will be done on https://github.com/jellyfin/Swiftfin/pull/2173. If we do see a benefit to bumping this build based on a patch, I am not opposed but if there isn't a direct benefit then I don't know if this is worth while on its own as it just adds other potential regressions.

LePips:
I had stated that I was fine with bumping the minimum version. As SwiftUI has API and changes with point releases, I want to be more aggressive with doing so if applicable.

- https://github.com/jellyfin/Swiftfin/pull/2173#issuecomment-5245968443

LePips (approved):

PR #2204, opened 2026-08-12
Show runtime for special features

kmishukov:
I've added an `extrasLabel` for `.video` items with a non-nil `extraType`. Runtime is now shown in that label, and `.unknown` is omitted from it. Tested the implementation in the iOS Simulator and verified that the tvOS build succeeds.

Screenshots: [before](https://github.com/user-attachments/assets/537e9b5b-c115-4891-9266-d3df0e5906d1) & [after](https://github.com/user-attachments/assets/8571d9a8-9857-4590-8dc6-9ab142c60582) (+ [web](https://github.com/user-attachments/assets/f40d7d19-1cba-4d3f-919b-0fb1b5e5653a) for reference)

This is my first contribution. The [development guide](https://jellyfin.org/docs/general/contributing/development#contributorsmd) suggests adding myself to the `CONTRIBUTORS.md` file, which I could not locate.

Fixes #1691

PR #2207, opened 2026-08-12
Align special feature labels

### LINKED ISSUE #2212: Setting to hide titles in library view

### Platform

tvOS

I'm not sure if this is a question or a feature request.

I just updated (thank you and congratulations to devs on the release), and found that titles are now showing below posters in library view. I think I had enabled a setting somewhere to hide these before, but I can't find it now. 

Was this feature eliminated? I really liked how clean and compact it looked before, and most titles are too long to display fully anyway, so you get ellipses... plus most posters have titles on them anyway.

Is there any workaround to hide titles now?

kmishukov:
Follow-up to #2204.

Special features with a known `extraType` use three label rows, while items
with `.unknown` use only two, causing inconsistent card heights.

This keeps the runtime directly below the title when the extra type is unknown,
while reserving the final row to keep neighboring cards aligned.

Screenshots: [before](https://github.com/user-attachments/assets/3affd165-3407-41ea-97be-4fcd619d9f2e) & [after](https://github.com/user-attachments/assets/3498ee22-c98b-4212-aebf-3ad4425d838e)

### DIFF
--- a/Shared/Components/BulletedList.swift
+++ b/Shared/Components/BulletedList.swift
@@ -15,8 +15,7 @@ func BulletedList(
     MarkedList(spacing: spacing) { _ in
         ZStack {
             // Capture local font line height
-            // swiftlint:disable:next hard_coded_display_string
-            Text(" ")
+            Text(String.space)
                 .hidden()
 
             Circle()
--- a/Shared/Components/PosterButton.swift
+++ b/Shared/Components/PosterButton.swift
@@ -86,6 +86,7 @@ struct PosterButton<Item: Poster>: View {
                 .trackingSize($posterSize)
             #endif
         }
+        .environment(\.posterDisplayType, displayType)
         .foregroundStyle(.primary, .secondary)
         .buttonStyle(.borderless)
         .buttonBorderShape(.roundedRectangle)
--- a/Shared/Extensions/EnvironmentValues.swift
+++ b/Shared/Extensions/EnvironmentValues.swift
@@ -38,6 +38,9 @@ extension EnvironmentValues {
     var playbackSpeed: Binding<Double> = .constant(1)
 
     @Entry
+    var posterDisplayType: PosterDisplayType = .portrait
+
+    @Entry
     var safeAreaInsets: EdgeInsets = UIApplication.shared.keyWindow?.safeAreaInsets.asEdgeInsets ?? .zero
 
     @Entry
--- a/Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto+Poster.swift
+++ b/Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto+Poster.swift
@@ -510,6 +510,14 @@ private struct BaseItemDtoPosterLabel: View {
                     .foregroundStyle(.secondary)
                     .lineLimit(1, reservesSpace: true)
             }
+
+            if item.extraType == .unknown {
+                Text(verbatim: " ")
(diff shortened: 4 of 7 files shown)


kmishukov:
@JPKribs

Hi, while the context is still fresh, could you please review this one as a follow-up? 
I noticed a minor UI alignment issue yesterday after the #2204 was merged and made a quick fix. Thanks!

JPKribs:
Seems like we should do a layout for this AND to resolve to outstanding TODO in there. 

That is, we have X components and 2/3 rows of space at all times based on the item type. Then use this to determine sizing to ensure that 1) all labels are the same height to avoid the issue you are encountering and 2) allow longer labels without a secondary element to utilize the second or third row of space. This would keep everything consistently sized moving forward.

I can look at this this evening US Mountain Time for this PR but I would want to loop in @LePips when we're done since I already missed this the first time around it would be best to have a third set of eyes haha.

JPKribs:
@LePips This is a bit bigger than I intended. Here is the pitch:

Always cap to 2 rows. In recommendations, any items can be placed next to each other. So, if some have 2 and some have 3 this will cause layout issues. This looks like this was your original plan for this but I took us off track with the programs so apologies for that. The primary dig against 3 rows is then we always have to reserve three rows which starts to consume a lot of room, especially on iOS Libraries.

First item is title, N+ are subtitle. 

1) If there is only one item, use it as the title but allow the title to use up to 2 rows. (Per TODO)

Screenshot 2026-08-14 at 10 47 25

2a) Portrait/Square, only use the first subtitle item.

Screenshot 2026-08-14 at 10 28 06

2b) Landscape, if there are 2+ subtitles, put them in a `DotHStack`.

Screenshot 2026-08-13 at 15 58 07

Net of this, related to the original PR, is the extras:

Screenshot 2026-08-14 at 10 28 42

Last two are `unknown` which doesn't show up intentionally including on web.

---

Work is on: https://github.com/jellyfin/Swiftfin/pull/2207/commits/76c3ced4acc08084eddba515c2d0783bcc12b492

Which is the label object and an environment object to track if this is landscape or portrait. Also added .space to the Strings. I know it's not really needed but so we're not having to add the `ignore hardcoded string` whenever we want to get a font height.

Other note, I use this:

``` swift
ForEach(details.indices, id: \.self) { index in
    details[index]
        .layoutPriority(index == 0 ? 1 : 0)
}
```

So that when we truncate text at the end of the subtitle it attempts to leave the first item in the `DotHStack` intact and only truncate the down-string item(s). So just like, why I'm trying to capture the index for this loop:

Screenshot 2026-08-14 at 10 56 44

See that the text is truncated at the end but the start/end times are not. Default behavior would be to truncate both.

LePips (commented):
I'll have to think about this a bit more and I do like the overall presentation consolidation by the number of views presented. This would also help if/when we want to have an option for displaying other poster metadata (rating, runtime, etc.).

> In recommendations, any items can be placed next to each other.

I believe we can make assumptions for what types of items are presented together as the similar items API will query/return similar items by the item's type. Meaning programs and extras could have 3 rows just fine as we can assume they would be grouped with one another. I don't think we'd have to be worried about this presentation in libraries either as the folder view won't present items with extras separate, but instead just as the media item.

LePips (approved):
This is great, but in the future we could still take into account how elements are presented alongside one another to allow more vertical elements as necessary. Also, the runtime labels should use the necessary runtime format style.

PR #2209, opened 2026-08-13
Remove iOS 16 & 17 Backports

JPKribs:
### Summary

Removes backports for iOS 16 and iOS 17. Additionally, clears out deprecated: `'onChange(of:perform:)' was deprecated in iOS 17.0: Use onChange with a two or zero parameter action closure instead.` where found.

Migrates `JumpIntervalPicker` & `PlaybackSpeedPicker` to use a `StateAdapter` while those are being changed as part of the backport removal.

Tested and I don't see any impact from these changes. Most files changed where purely just dropping the `.backport` or updating `onChange` but there are 2 bigger noes that I'd appreciate confirmation:

- `MainTabView`- Remove `legacyTabContent()`
- `TabItem`- Remove `labelStyle`

We're still using `NavigationTransition` so I wasn't sure if that should move out of the backport or if there was a grander task to unwind it's usages throughout. Either way, I landed on leaving it along for this pass.

`.toolbarTitleDisplayMode(router.isRootOfPath ? .inlineLarge : .inline)` had an iOS and tvOS branch that I've moved to our 2 callers.

Last item, I had a 17+ version of `TextTransferable` but found it added more options but dropped copying. Might look at that more later but for now just keeping with the existing version of this that include copy + file.


LePips (approved):
> Migrates JumpIntervalPicker & PlaybackSpeedPicker to use a StateAdapter

`StateAdapter` is to be used in smaller contexts. If it's wrapping the entire body, it's the same thing as just putting the state on the view.

> We're still using NavigationTransition so I wasn't sure if that should move out of the backport

I believe this was also created for the `Hashable` conformances used in the navigation stack, but I can't remember.

PR #2211, opened 2026-08-14
Jellyfin Remote

JPKribs:
### Summary

In https://github.com/jellyfin/Swiftfin/pull/2150 we intake most socket commands into Swiftfin. This PR creates the remote & `MediaPlayerProxy` to **send** those commands over.

My goal was purely to try and recreate Apple's built-in Apple TV Remote. The directional pad really only works on Web since I can't really find other clients that register for these commands. It is hidden for devices that do not have this available but still allows media controls. Swiping works to control or pressing the directional arrows. Pressing in the middle does a selection. Queue only shows/works if the client reports their queue to the socket which we do not do at this point in time from the Swiftfin side.

### WIP

I am working on this and a socket refactor for server tasks. Outside of those, and my open PRs, I think this will be my last new item for a bit while I shift focus to some personal items.

I have some cleanup that I need to do for this mostly on the `ViewModel` & `Proxy` side of this so I think the UI is just about done. I also would like to see what happens with https://github.com/jellyfin/Swiftfin/pull/2202 before calling this one done because that will create some changes to where this is located.

Drafting this a hair earlier than I usually would mostly because my mac is out of room so I am trying to clear videos and screenshots off of it and these are ones I don't really feel like recreating again when I'm done.

### Swiftfin to Swiftfin (Cast)

https://github.com/user-attachments/assets/359c4c81-68be-4f1d-a904-b78e5458d9c1

### Swiftfin to Web (Control)

https://github.com/user-attachments/assets/8520a3d0-86a6-470f-bc3f-f5fd5cfda1a7

#### These are offset by a hair. If I knew how to combine 2 videos I would left is Swiftfin controlling the Web for music playback and the left is what the looks like on web.

<table>
  <tr>
    <th>Phone</th>
    <th>Web</th>
  </tr>
  <tr>
    <td><video src="https://github.com/user-attachments/assets/7cfa7d98-a710-47c0-8b76-e094f9b5a1c3" controls></video></td>
    <td><video src="https://github.com/user-attachments/assets/3d13e1e6-2bea-401d-95dd-8b761b3a7aac" controls></video></td>
  </tr>
</table>


JPKribs:
### Some UI Cleanup:

#### 2026-08-21

| **Empty Selection** | **Controlling Web** | **Controlling Swiftfin** | **Controlling Music** | **Controlling Movie/Show** |
| :---: | :---: | :---: | :---: | :---: |
| The track pad area is disabled along with buttons too. | Track pad works, volume selection/mute, back, play, etc. Based on whatever the client reports back as available but this is the fullest example that works. | Only playback controls since that's all that Swiftfin reports. When nothing is playing, the trackpad area is replaced with a nothing playing indicator instead. | | |
| Empty Selection | Controlling Web | Controlling Swiftfin | Controlling Music | Controlling Movie/Show |

**Queue Management**

All the queue management that's exposed. I'm pretty proud of the equalizer thing I made that's meant to mirror Apple Music.

https://github.com/user-attachments/assets/dbd40afb-30ae-404a-b110-5612cc0c36e4

**Casting from Swiftfin**

https://github.com/user-attachments/assets/b53852d7-aa9c-41de-8bb5-b507f050813b

**Apple Music Equalizer**

https://github.com/user-attachments/assets/ed2d74f6-89d6-468f-9e5d-3c14a3cc5f63

PR #2227, opened 2026-08-17
Focus the Library not the `LetterPicker`

### LINKED ISSUE #2218: Movies/TV Library focuses on the letter picker by default

- [x] I have read the [Common Issues](https://github.com/jellyfin/Swiftfin/blob/main/Documentation/common_issues.md) documentation and this issue was not mentioned there.

### Description of the bug

When the Letter picker is enabled, the “#” in the Letter picker is selected by default in the Movies and TV Shows libraries when they are opened.

### Platform

tvOS

### Application version

1.6

App Store

### Device information

Apple TV 4K (1st gen)

### OS version

tvOS 26.5

### Jellyfin server version

10.11.11

### Screenshots or videos

Image

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/2218

Uses the same `focusCoodinator` as the `PlayButton` for the `ItemView`.

### Video

https://github.com/user-attachments/assets/b8caaec4-56ee-453d-a608-6de1bb7c0c99


LePips (changes requested):
As we want work for tvOS library filters, I would prefer this is done within the context of how we would manage focus with filters on this view as it may be more complex.

Focusing on a button shouldn't happen until the first load is complete and we may have to think about how focus works on filter selection while the library is loading. I overall want to prevent the behavior of snatching the focus, like how this does when the library takes a little bit longer to load.

JPKribs:
Is that through something like https://developer.apple.com/documentation/swiftui/view/focusscope(_:)? I was able to use that for some things while testing but I'm not sure how to pass a `namespace` between nested views.

LePips:
I'm unsure what you mean, as all I'm really advocating for is disabling focusing a button until first load.

JPKribs:
Sorry for the delay. I've been looking at this and have gone down the focus rabbit hole. I hate focus... So, here is what I have:

**SwiftUI Focus Tools**

`focusable` - This item CAN be focused
`focused` - Ties the current focus state of this item to a variable. When true, that item is focused.
`focusSection` - Wrapper around a group of `focusable` items.
`defaultFocus` - When this `focusSection` is focused, what child item takes the initial focus
`prefersDefaultFocus` - When the items all first appear in this `namespace`, which item takes initial `focus`. **This doesn't work inside of a `TabView` 😕**
`resetFocus` - Resets the focus of the `namespace` back to `prefersDefaultFocus`.

Here is the issue that I am having here. SwiftUI needs *something* to be focused. So, on initial load, it attempts to find and focus the first item available. In this case, I have tried a few options:

1. Disable the `LetterPickerBar` until `viewModel.state` == `.content` but the issue is that `.content` means the data exists NOT that the `CollectionVGrid` has rendered an item. 
2. On first item of the `CollectionVGrid`.`onAppear`, divert focus to this item. The issue is that creates that "Randomly pulls the user focus to something new" effect that's not great.
3. Create an empty, clear overlay that "holds" focus until an event occurs, like `CollectionVGrid`.`onAppear`, then removes itself while it diverts focus.

I've been legitimately working on this one focus thing for like 3 days...... I think the solution here is something that's a bit more of a system rather than just purely a fix for this one view. The existing `FocusCoordinator` does the job and works but I think we need a focus nil to hold focus so nothing* is focused then applying it after a condition.

I have a version of this working. Every. Single. Guide. I have found for tvOS focus seems to defer to `UIFocusGuide` instead of working on SwiftUI so I'm flying a bit blind. I'm working through some jank but I'll push my WIP version of this today just for visibility.

JPKribs:
I'm not going to be looking at this for a while. TBH, it seems like a UIKit vs SwiftUI not talking to each other well. The work on this branch is functional but probably not the way we want to go. Leaving this branch around but I'm closing this to reduce PR bloat.

PR #2233, opened 2026-08-18
Follow Apple's Recommended `Long Press` Gesture Duration

### LINKED ISSUE #2217: Long press for playback speed requires too long a hold; make duration shorter or configurable

### Platform

- [x] iOS

When "Long press" is mapped to "Playback speed" in the gesture settings, the playback speed only kicks in after holding for **1.2 seconds**, which feels sluggish when you just want a quick speed boost. Please make the required hold time shorter and/or configurable.

The duration is currently hardcoded in the long press gesture recognizer at `Swiftfin/Components/GestureView.swift`:

https://github.com/jellyfin/Swiftfin/blob/17c46d35f804164d3b136dcb76f6e0bd0205103d/Swiftfin/Components/GestureView.swift#L54-L61

The gesture action handling (including playback speed) lives in `Swiftfin/Views/VideoPlayer/Gestures/VideoPlayerContainerView+TapGesture.swift` (`handleLongPressGesture`), and the gesture settings UI is in `Swiftfin/Views/GestureSettingsView.swift`.

Ideally the hold time would be either:
- shortened to something closer to the iOS default (~0.5s), and/or
- user-configurable in the gesture settings.

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/2217

Apple's documentation for long press suggests 0.5s. Keeping in line with that, this changes this to 0.5 instead of the existing 1.2. There is an argument for customization, but IMO, I think it makes the most sense to just follow the Apple documentation for this to keep this consistent between apps instead of opening the door up to customization on every gesture.

https://developer.apple.com/documentation/uikit/handling-long-press-gestures

Image

Open for changes if we want to make this a customization but IMO, it makes sense to have that be the line as we risk the Gesture settings becoming overwhelming if we include things like `Minimum Swipe Distance`, etc.

If we prefer the 1.2s we have now, I think this is fine to document and close out as well. Just trying to clear out new smaller issues.

### DIFF
--- a/Swiftfin/Components/GestureView.swift
+++ b/Swiftfin/Components/GestureView.swift
@@ -56,7 +56,7 @@ struct GestureView: PlatformViewRepresentable {
                 target: self,
                 action: #selector(handleLongPress)
             )
-            recognizer.minimumPressDuration = 1.2
+            recognizer.minimumPressDuration = 0.5
             return recognizer
         }()
 


LePips (approved):
The suggestion is at least 0.5s to distinguish a long press gesture from a longish-single tap. I'm just gonna have to trust that this doesn't conflict with other gestures.

Issue #2234, opened 2026-08-18
Home screen navigation sidebar shows when loading user


wektec:
### This issue respects the following points:

- [x] This is a **bug**, not a question or a configuration issue; Please visit our [forum or chat rooms](https://jellyfin.org/contact/) first to troubleshoot with volunteers, before creating a report.
- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/Swiftfin/issues?q=is%3Aopen+is%3Aissue).
- [x] I have read the [Common Issues](https://github.com/jellyfin/Swiftfin/blob/main/Documentation/common_issues.md) documentation and this issue was not mentioned there.
- [x] I'm using an up to date version of Swiftfin; We generally do not support previous older versions. If possible, please update to the latest version before opening an issue.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).

### Description of the bug

When you select a user in the user selection and the home screen loads, the navigation sidebar appears.
Maybe this menu should be closed by default. (The navigation sidebar also appears when the home screen loads after an app restart)

In addition, a gray box appears in the upper left corner after the home screen loads for a new user.

<img width="2000" height="1125" alt="Image" src="https://github.com/user-attachments/assets/c7264338-47ca-40ee-9400-c908c269b2b4" />

### Platform

tvOS

### Application version

1.6

### Where did you install the app from?

App Store

### Device information

Apple TV 4K (1st gen)

### OS version

tvOS 26.5

### Jellyfin server version

10.11.11

PR #2236, opened 2026-08-18
`ProgressView` Renders into `TopLeading` Corner

### LINKED ISSUE #2235: Home screen loading indicator

- [x] I have read the [Common Issues](https://github.com/jellyfin/Swiftfin/blob/main/Documentation/common_issues.md) documentation and this issue was not mentioned there.

### Description of the bug

The loading indicator on the home screen is currently displayed in the upper left corner instead of in the center of the screen, as it is in other views.

In addition, this loading indicator is obscured by the open navigation sidebar when the home screen loads following a user selection or an app restart. See [#2234](https://github.com/jellyfin/Swiftfin/issues/2234).

Image

### Platform

tvOS

### Application version

1.6

App Store

### Device information

Apple TV 4K (1st gen)

### OS version

tvOS 26.5

### Jellyfin server version

10.11.11

JPKribs:
### Summary

Resolves: https://github.com/jellyfin/Swiftfin/issues/2235
Some Changes For: https://github.com/jellyfin/Swiftfin/issues/2234

`ProgressView` gets crammed into the top leading so you can see the material background in the top right and the loading progress circle thing pushed out by the tab. This just forces it to use all of the space when it exists. I think it's just an issue with the various layering for the content view 

#### Left is 1.6 and Right is this PR

Keep an eye on the top left.

https://github.com/user-attachments/assets/b2ad6e82-3de3-41fe-80cc-fc0313d6384e

Also visible by the placement of the loading indicator.

https://github.com/user-attachments/assets/e515307b-63c5-46bc-b39e-4439e2f57e31


LePips (changes requested):
> Two symptoms of the same issue.

This doesn't address #2234 directly, as the issue is instead that there isn't anything to focus on the view when it presents the loading indicator so it will default to the tab. To better work with these issues, please work by introducing a delay in the network calls to mimic what a non-development environment would see.

> removes the nested refreshable

The internal refreshable is so the content would instead refresh on the background instead of a full refresh. This is invalid, as refreshing on the error view already worked.

JPKribs:
Some cleanup. This fixes the loading icon and the dark material box in the top leading. For the `TabView` focusing, it looks like that would need to come from [`UITabBarController`](https://developer.apple.com/documentation/uikit/uitabbarcontroller). In testing, it looks like [`preferredFocusEnvironments`](https://developer.apple.com/documentation/uikit/uifocusenvironment/preferredfocusenvironments) resolves this but it creates a lot of headaches so I'm limiting this PR to just be the `ProgressView`.

Main post reflects the scope here. Just the rectangle when loading and the indicator being in the wrong place.

LePips (approved):

Issue #784, opened 2023-06-01
Library Update


jrnas:
TVOS: Please develop a feature or a settings option for a automatically library refresh, because it doesn’t refresh the library 

**Additional context**

dalehumby:
To expand on this feature request a bit:

Once the app is launched the Apple TV library is not updated when new shows/movies are added on Jellyfin server. The app needs to be restarted (Double-tap the TV button on the remote. Then swipe up on the app to terminate) and then relaunched for new content to show.

**To recreate:**
1. Launch Jellyfin on Apple TV. Exit Jellyfin app, launch another app. Or turn off the Apple TV (press hold middle button, Power Off)
2. Using radarr etc., a new movie is added and Jellyfin server imports the movie (it's available through the web player, with all metadata)
3. Turn on Apple TV, open Jellyfin
4. New movies/show is not availabe
5. Terminate Jellyfin app (Double-tap the TV button on the remote. Then swipe up on the app to terminate.)
6. Relaunch Jellyfin, the movie is now available

Ideally the content can be lazy-loaded, so when the app is open new content shows up after a few seconds. 

Personally I don't think this needs to be a setting, just something the app does in the background.

github-bryld:
Seconded.
Oddly, the library seems to refresh automatically on iOS/iPadOS, it only fails on TVOS.

kSapewpew:
This is still an issue. Any way one can help with on this one? It's making it rather annoying to use Swiftfin at the moment.

sIRwa2:
Is this issue acknowledged and on the road map to be fixed is one way or another? a "refresh library" button would already help until auto refresh is solved imo.

Nanorulez123:
Still an issue

KintsugiUwU:
> Seconded. Oddly, the library seems to refresh automatically on iOS/iPadOS, it only fails on TVOS.

It isn't working for me on iPad OS. Added a new library with videos from YouTube. Jellyfin mobile has the updated library but SwiftFin not

foorschtbar:
I can see the same behavior under tvOS - the app has to be restarted after every library update. Would be nice if someone could take a look at this. Automatically or, if necessary, with a button.

h4knet:
Hi,
I can confirm the issue on Apple TV.
The only workaround is to reboot the Apple TV 😓

JPKribs:
[...]

PR #2240, opened 2026-08-20
Refresh Home and item details after playback stops

Adavid114:
Swiftfin's player doesn't update the real playback state in real time. This is especially annoying because there's no (intuitive) way to refresh and update the watched state of each piece of content — unlike on iOS, where there is one.

This PR proposes a solution for that, without a manual refresh: everything happens automatically.

When you exit playback, didSendStopReport is posted (a notification that already existed in the code but was unused). It updates Home and other navigation menus, and it does so without the user noticing.

The content detail is updated more selectively. Instead of refreshing the whole page, it only refreshes the play button (the remaining playback time) and whether the content is marked as watched or not.

This PR was tested on a physical Apple TV 4K (tvOS) and on the iPhone 17 Pro simulator (iOS). I verified that the change feels fluid and improves the user experience.

I don't know whether there's an open issue about this, since it was something I found annoying and I decided to contribute a solution that benefits us all. Thanks.

### DIFF
--- a/Shared/Objects/ContentGroup/ContentGroupProvider.swift
+++ b/Shared/Objects/ContentGroup/ContentGroupProvider.swift
@@ -14,8 +14,16 @@ protocol ContentGroupProvider: Displayable, Identifiable {
     var environment: Environment { get set }
     var id: String { get }
 
+    var refreshesOnPlaybackStop: Bool { get }
+
     @ContentGroupBuilder
     func makeGroups(environment: Environment) async throws -> [any ContentGroup]
+}
+
+extension ContentGroupProvider {
+    var refreshesOnPlaybackStop: Bool {
+        true
+    }
 }
 
 extension ContentGroupProvider where Environment == Empty {
--- a/Shared/Objects/MediaPlayerManager/MediaProgressObserver.swift
+++ b/Shared/Objects/MediaPlayerManager/MediaProgressObserver.swift
@@ -152,6 +152,8 @@ class MediaProgressObserver: ViewModel, MediaPlayerObserver {
 
             let request = Paths.reportPlaybackStopped(info)
             try await send(request)
+
+            Notifications[.didSendStopReport].post()
         }
     }
 
--- a/Shared/ViewModels/ContentGroupViewModel/ContentGroupViewModel.swift
+++ b/Shared/ViewModels/ContentGroupViewModel/ContentGroupViewModel.swift
@@ -60,6 +60,15 @@ final class ContentGroupViewModel<Provider: ContentGroupProvider>: ViewModel {
             self?.lastRefreshSignalDate = Date.now
         }
         .store(in: &cancellables)
+
+        if provider.refreshesOnPlaybackStop {
+            Notifications[.didSendStopReport]
+                .publisher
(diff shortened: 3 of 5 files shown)


JPKribs:
Thank you for your contribution! Unfortunately, I don't think this is our best approach to this issue. What is happening here is we are re-pulling the entire home worth of data when, at most, a couple of items have changed. We should be able to get that information from the socket for free which would then also catch updates where playback ended from another device as well. 

I have a branch that I started working on this here you are welcome to pull from if you are interested:

https://github.com/JPKribs/Swiftfin/tree/userNotificationManager

The concept is pulling both socket updates and internal notifications into a centralized point then distributing the output. This should work for all instances of items in Swiftfin opposed to just home. This was mostly done but there are some edge cases like libraries where this logic updates items in the middle of the libraries correctly but I didn't quite know how to handle like... The library is open and a new item was added to that library. That one might actually need a full refresh to work but I think new items is really the last part I was struggling with here.

Again, you're totally welcome to take from this what you would like but I think that updating home fully at playback end only solves the issue at home while missing the `ItemView` and other locations.

I mention this approach here: https://github.com/jellyfin/Swiftfin/issues/784

Adavid114:
Hi JPKribs!
Thanks a lot for the feedback and for sharing your branch. You're totally right: a full home refresh is too heavy and only fixes home, while pulling changes through the socket would gets us updates for free and also covers playback ending on another device, which mine doesn't. (i didn't think of that initially)

I'd like to try contribute with your approach. From what I can see, the paging library views already handle didChangeItem and didDeleteItem, but the remaining gap is newly-added items. An open library doesn't pick up items added to it (the foldersAddedTo / global-refresh path). That matches the edge case you mentioned, so I'll start there.

I'll close this PR in favor of that approach. I'll follow up on #784 as I make progress. Thanks again for taking the time to review my PR!

Adavid114:
Hi JPKribs, Quick update. I built the open-library case on top of your userNotificationManager branch and verified it on my physical Apple TV. made one commit (+41/-5).

libraryChanges now carries the affected library IDs (foldersAddedTo + collectionFolders) on a small scoped notification, and PagingLibraryViewModel does a full refresh only when library.parent.id matches. So an open library picks up added/removed items in place, other open libraries aren't touched, and global refresh stays as the fallback. The missing piece was collectionFolders — the branch was dropping it and always going global. I verified it live by creating/removing a collection with the Collections library open (changes arrive ~30s late, but that's the server's LibraryUpdateDuration batching, not the client).

The downside is that the match is by library.parent.id, so it only covers folder/collection/top-level libraries. Filtered or recursive views (genre, studio, person, etc) have no matching folder ID, so those won't auto-refresh on new items. (These are used less frequently i think, but wanted to clarify that)

Since it's your architecture and your branch isn't a PR yet, do you want me to open a PR to jellyfin/Swiftfin:main crediting you, or to your branch?

Thank you!

JPKribs:
> Since it's your architecture and your branch isn't a PR yet, do you want me to open a PR to jellyfin/Swiftfin:main crediting you, or to your branch?

Feel free to open your own branch and PR when you feel you have what you need! I'm not particularly worried about attribution

PR #2244, opened 2026-08-22
Distinguish Focus from Selection in the Season Picker

nkt:
On tvOS the season picker draws focus and selection identically, so there is no way to tell that the row has been entered.

`SeasonButtonStyle` fills a capsule when `isFocused || (!isPickerFocused && isSelected)`, and both branches produce the same `.regular.selection(tint: .white)` appearance. Moving focus into the row only carries the fill from the chosen season across to the one under focus — and because the row is normally entered *on* the chosen season, the common case is that nothing changes on screen at all. From there, arrowing along the row looks like a single capsule that never clearly belongs to either "what is selected" or "where I am".

The focused season now lifts slightly and casts a shadow, and the chosen season drops to 72% opacity while focus is elsewhere, so the fill reads as position once the row is entered. This follows what `SupplementTitleButtonStyle` already does for the same problem in the player's supplement bar.

Focus is also passed down from the picker's own `FocusState` instead of being read from `\.isFocused` inside the `ButtonStyle`. Logging both values while moving through the row showed them reaching the same totals but disagreeing on individual passes — the environment value trails the focus state by a render, which let the style briefly draw a focus that had already moved on.

**Testing.** Verified in the tvOS 26.5 simulator on an Apple TV 4K (1st generation), i.e. `AppleTV6,2`, where `UIDevice.supportsLiquidGlass` is false and `BackportGlassEffectModifier` takes its legacy path. Entering the row, arrowing between seasons, and leaving the row all read correctly now. tvOS only; nothing on the iOS side of `SeasonSelector` is touched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Shared/Objects/ContentGroup/SeriesEpisodeContentGroup/SeriesEpisodeContentGroup+SeasonSelector.swift
+++ b/Shared/Objects/ContentGroup/SeriesEpisodeContentGroup/SeriesEpisodeContentGroup+SeasonSelector.swift
@@ -49,7 +49,12 @@ extension SeriesEpisodeContentGroup {
                             Button(season.library.parent.displayTitle) {
                                 selection = season.id
                             }
-                            .buttonStyle(SeasonButtonStyle(isPickerFocused: isPickerFocused))
+                            .buttonStyle(
+                                SeasonButtonStyle(
+                                    isFocused: focusedSeason == season.id,
+                                    isPickerFocused: isPickerFocused
+                                )
+                            )
                             .isSelected(isSelected)
                             .focused($focusedSeason, equals: season.id)
                             .accessibilityAddTraits(isSelected ? .isSelected : [])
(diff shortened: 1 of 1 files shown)


LePips (changes requested):
I'm fine with the intention of the changes but don't have the scale effect and have the animation curves be `.linear(duration: 0.1)`. Please remove all AI comments.

Please see Jellyfin's AI policy regarding direct communication for PRs.
- [jellyfin.org/docs/general/contributing/llm-policies](https://jellyfin.org/docs/general/contributing/llm-policies/)

JPKribs:
Should we just use this style with the the Supplement buttons from the player? They both have the same `Focus-To-Select` pattern for child data underneath. IMO, they look similar as well:

#### Current Season Buttons

Simulator Screenshot - Apple TV 4K (3rd generation) (at 1080p) - 2026-08-23 at 12 15 31

#### Supplement Buttons

Simulator Screenshot - Apple TV 4K (3rd generation) (at 1080p) - 2026-08-23 at 12 15 41

This is lazy just for demo:

```swift
Button(season.library.parent.displayTitle) {
    selection = season.id
}
.buttonStyle(VideoPlayer.UIVideoPlayerContainerViewController.SupplementContainerView
    .SupplementTitleButtonStyle())
...
```

I think, ideally, we would do some cleanup/polish before this but this is what that style would look like in this context (tvOS 26):

https://github.com/user-attachments/assets/9351789f-ada9-45f1-a776-47cfebf597fe

LePips:
Replaced by #2306

PR #2245, opened 2026-08-22
Remove mediaSources from listing MinimumFields

### LINKED ISSUE #1707: Remove MediaSources field from listings
Mediasources is set as an minimal field here: https://github.com/jellyfin/Swiftfin/blob/main/Shared/Extensions/JellyfinAPI/ItemFields.swift

This is fine for detail
pages but for listings this can slow down responses significantly, especially when having multiple versions. As far as i can tell its also not neded for
listings

mitre88:
## Summary

Home, Search, and library listings currently request `mediaSources` for every item via `MinimumFields`. That field carries codec, resolution, and version payloads. On libraries with multiple versions it makes those list calls slower for data the posters never use.

Playback already reloads the item in `MediaPlayerItem.build` (`getFullItem`). Listings only need `parentID`.

## Why this is not a repeat of #1950

#1950 was closed waiting on #1752. #1752 closed without merging, and `main` still ships `.mediaSources` in `MinimumFields`. This change is against current `main` and keeps playback working without assuming listing rows include sources.

Fixes #1707

## Changes

- Drop `.mediaSources` from `MinimumFields`.
- Keep `.mediaSources` on the single next-up / resume / first-episode fetch used by the series Play button, so the version menu still has data.
- Episode overlay no longer force-unwraps `mediaSources?.first`; it uses the same `MediaPlayerItem.build` path as next/previous episode.

## Test plan

- [ ] Home, Search, and a large library with multi-version items: posters still load; list requests no longer include `MediaSources`.
- [ ] Play from a poster / search result (item without preloaded sources).
- [ ] Series page Play button, including an episode with more than one version.
- [ ] Play an episode from the in-player episode overlay, then next/previous.
- [ ] Live TV / program play still starts.

### DIFF
--- a/Shared/Extensions/JellyfinAPI/ItemFields.swift
+++ b/Shared/Extensions/JellyfinAPI/ItemFields.swift
@@ -14,8 +14,11 @@ extension ItemFields {
     /// The minimum cases to use when retrieving an item or items
     /// for basic presentation. Depending on the context, using
     /// more fields and including user data may also be necessary.
+    ///
+    /// `mediaSources` is intentionally omitted: it inflates listing
+    /// responses when items have multiple versions. Playback paths
+    /// load sources via `getFullItem` / `MediaPlayerItem.build`.
     static let MinimumFields: [ItemFields] = [
-        .mediaSources,
         .parentID,
     ]
 }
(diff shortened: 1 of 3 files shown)


LePips:
This is practically just a repeat of #1950 as noted in the issue is that we ought to just be able to remove `MinimumFields` entirely.

Please see Jellyfin's AI policy regarding direct communication for PRs.
- https://jellyfin.org/docs/general/contributing/llm-policies/

PR #2248, opened 2026-08-23
Remove Minimum Fields

### LINKED ISSUE #1707: Remove MediaSources field from listings

LePips:
Closes #1707 

Removes `.MinimumFIelds` entirely. I've at least verified that for programs we aren't using `ItemLibrary` and that for similar items we will now check the type of the parent item. I believe we could remove `channelInfo` elsewhere, but didn't track down the endpoints in the server.

While doing the cleanup in `EpisodeMediaPlayerQueue` I also believe we could eliminate the full channel retrieval and just make the item with the channel ID, but we'd have to verify.


JPKribs (approved):
Tested and working. Made an extra point to test Live TV since that always seems to behave weirdly but everything checks out!

PR #2253, opened 2026-08-25
Refresh an open item view from LibraryChanged socket messages

freedreamer82:
### **_Summary_**

Swiftfin keeps a WebSocket open to the server but never listens for LibraryChanged, the message a server sends when items it already handed out have changed. Pressing Refresh metadata on an item is the clearest case: the request returns, the server does the work, and the open item view keeps showing the old data until you leave it and come back.

This PR listens for that message and updates the item view in place:

You press Refresh metadata on an item.
The server does the work and, when it lands, sends LibraryChanged with that item's id in ItemsUpdated.
ServerSocketManager now exposes those updates, UserSessionManager turns them into the existing itemShouldRefreshMetadata notification, and an open ItemView whose id matches re-reads its item and refreshes its groups — without a spinner over the whole screen and without leaving the view.
Everything it needs is already there: the message is in the Jellyfin API description, the SDK decodes it, and the notification exists and is already posted from elsewhere. Nothing observed it until now

### DIFF
--- a/Shared/Services/ServerSocketManager.swift
+++ b/Shared/Services/ServerSocketManager.swift
@@ -245,6 +245,13 @@ extension ServerSocketManager {
         }
     }
 
+    var libraryUpdates: AnyPublisher<LibraryUpdateInfo, Never> {
+        commands { event in
+            guard case let .message(.libraryChangedMessage(message)) = event else { return nil }
+            return message.data
+        }
+    }
+
     private func commands<Command>(
         extract: @escaping (JellyfinSocket.Session.Event) -> Command?
     ) -> AnyPublisher<Command, Never> {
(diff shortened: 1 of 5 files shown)


JPKribs:
I'll defer to @LePips but I would prefer we go the https://github.com/jellyfin/Swiftfin/pull/2265 route. Apologies, I had discussed with that contributor on an earlier PR but I didn't do a good job elevating that discussion so it was easily discoverable. 

My reason for my preference is that solution is more holistic, including updates for items like posters and home view as well and handles both internal notifications and socket messages. It also works on an item we've mentioned on https://github.com/jellyfin/Swiftfin/issues/784 for centralizing notifications. This item should resolve your `ItemView` issue as well if you are able to test and confirm this meets your expectations for that item as well.

PR #2254, opened 2026-08-26
Add setting to disable the Latest prefix on home library rows

freedreamer82:
Summary

Home library rows are titled with the `latestWithString` format ("Latest %@"), which produces awkward results in several languages — in Italian "Recenti Film" / "Recenti Serie TV" instead of a natural phrase. The section already only ever shows latest media, so the prefix carries no information the row's context doesn't.

Add `Customization.Home.showLatestPrefix`, defaulting to true, and a toggle for it in the Home section of Customize settings. With it off, rows are titled with the library name alone. Existing users see no change.

with Latest :

image

----------------------

without Latest:

image

----------------------

Config switch

image


LePips:
If the core issue is with a translation then offer a fix for the translation in Weblate. Smaller things like this are not to be customizable.

PR #2256, opened 2026-08-26
Add tvOS tab bar placement setting

freedreamer82:
## Description

On tvOS the main tab bar currently always uses `.sidebarAdaptable`, which renders
as a collapsed bar in the top-leading corner that expands into a sidebar on focus.
Some users might prefer the classic tvOS tab bar, centered along the top.

This adds a per-user setting to choose between the two. Default is `Sidebar`, so
existing behavior is unchanged.

Settings → Customize → **Tab bar** → **Layout**: `Sidebar` (default) / `Centered`

image

image


LePips (dismissed):
I'm fine with this and we'll have to introduce some more work for views to change their padding accommodations. If having to accommodate both becomes too much I would remove this.

LePips (approved):

PR #2257, opened 2026-08-26
Add poster orientation settings for Home sections

freedreamer82:
## Description

Lets you choose whether each Home row shows portrait or landscape
posters, from a new **Customize > Posters > Orientation** screen.

`Customization.nextUpPosterType`, `recentlyAddedPosterType` and
`latestInLibraryPosterType` already existed, but nothing read them any
more once Home moved to content groups — `DefaultContentGroupProvider`
hardcodes the display types instead. This reconnects them and adds the
two keys that were missing.

## Changes

- `SwiftfinDefaults`: add `resumePosterType` (default `.landscape`)
  and `recentlyPlayedPosterType` (default `.portrait`); change
  `latestInLibraryPosterType`'s default to `.landscape`
- `DefaultContentGroupProvider`: read the five keys instead of
  hardcoding `.landscape`/the `PosterGroup` default. The Continue row
  also drops to `.small` when portrait, since `.medium` resolves to the
  landscape width and would leave 2 columns instead of 3 on iPhone
- `PosterOrientationSection`: new screen, one picker per Home row
- `PosterSection`: `Orientation` chevron as the last section
- New `orientation` / `posterOrientationDescription` strings

## Known limitation

The change applies on the next full Home refresh, i.e. after an app
restart. `makeGroups` only runs in `ContentGroupViewModel.fullRefresh`,
while pull-to-refresh goes through the inner `.refreshable` on the
scroll view and calls `background.refresh()`, which reuses the existing
groups. This already affects the existing Recently added / Recently
played toggles in Customize, so it's pre-existing rather than
introduced here; happy to address it separately if you'd like it in
scope.

## Testing

Built and run on tvOS (Apple TV 4K simulator), navigated to the new
screen and confirmed the pickers render and persist. iOS builds clean.

image

image

image


LePips (changes requested):
Thank you for wanting to contribute to fix this but the actual fix should come through an alternate `poster-group-id: poster orientation` store instead of individual sections, which I have local work on for making custom views.

Issue #1738, opened 2025-09-26
[iOS] Duration texts read incorrectly by VoiceOver


samglt:
Progress labels and similar that express a duration like "1h 15m" are read out like "one h fifteen meters"

JPKribs:
Is this still the case on 1.4? We should no longer have custom strings for durations post-https://github.com/jellyfin/Swiftfin/pull/1807

samglt:
I can still replicate this in the app store release of 1.4, iOS 26. I haven't kept up with the repo in a while so I don't know about what's happened in main since then.

JPKribs:
> I can still replicate this in the app store release of 1.4, iOS 26. I haven't kept up with the repo in a while so I don't know about what's happened in main since then.

Apologies, this is likely still the  older TimeInterval usage as opposed to the Duration > `HourMinuteAbbreviated` which **should** just inherit Voice Over from Apple. I'm realizing there is another version of this we use for playback position that likely needs to be moved over per:

https://github.com/jellyfin/Swiftfin/blob/main/Shared/Extensions/FormatStyle.swift

I **believe** that should resolve this as a byproduct

PR #2258, opened 2026-08-27
[iOS] VoiceOver support in the video player

### LINKED ISSUE #1733: [iOS] Player overlay not accessible from VoiceOver

zoli211:
### Summary

The video player was close to unusable with VoiceOver. Once the playback controls faded there was no element left to focus, no way to bring them back, and usually no way out of the player. VoiceOver also read straight through to the item view behind it, the controls that were on screen didn't say what they were, and the progress bar couldn't be moved at all.

There is one commit per fix.

The player is presented with a custom presentation controller rather than a system one, so it doesn't inherit the isolation a modal presentation provides. The outermost presented view is now marked modal while the player is up.

An accessibility element stands in for the video, placed beneath the controls, and toggles them when activated. This is the approach from #1741, whose `accessibilityToggleOverlay` I kept and credited on the commit. While the controls are hidden the element collapses to a corner target rather than tracing a focus ring around the whole screen.

The decorative elements are hidden. The dimming, the gradient that covers the bottom 55% of the screen, `GestureView`, the scrubbing preview and the slow scrubbing indicator were all silent elements to swipe past. `isVisible` now hides from accessibility as well as setting opacity, so faded controls aren't focusable either.

The controls no longer disappear from under someone still reading them. With an assistive technology running they only fade with focus resting on the video, which is how they are dismissed there.

The jump buttons read "15s" without saying which way they jump, and the timestamps were digit patterns like 1:23:45, read as a run of separate numbers. Both are labelled now, and a `spokenRuntime` format style spells the durations out. Gestures that report only through a toast are announced.

The progress bar is a `ProgressView` driven by a drag gesture, so it carried no value and no way to move it. It is now a single adjustable element that seeks by the same interval as the jump buttons.

### Testing

- Built and ran on an iPhone 16 Pro, iOS 26.6, with VoiceOver on
- SwiftFormat, SwiftLint (`--strict`) and `FindUnusedStrings` pass

### Not covered

The tvOS half of #1738. The timestamps in the tvOS player are untouched, and the only change reaching tvOS at all is the shared `isVisible` modifier.


LePips (changes requested):
Thank you for wanting to contribute accessibility settings. However, I will have to request to remove the subtitle reading. I have been working on some exciting developments that would enable you to contribute that in the future.

zoli211:
Hey! If I submit subtitle reading as a separate feature PR, would it be possible to get it into the repo? Right now, there’s no way to consume any content with subtitles using VoiceOver, and even information from forced subtitles gets lost. It feels like a pretty basic accessibility feature to have. Thanks so much!

zoli211:
Removed — the last commit is gone, the PR is back to the player fixes only. Happy to wait for what you're working on and bring the subtitle reading separately once it lands.

PR #2261, opened 2026-08-30
Fix incorrect episode selection for series

### LINKED ISSUE #2143: Shows selected choose correct season, but wrong episode.

### Description of the bug

This may feel like a weird thing to care about, but after updating to v1.5 when selecting a season of a show in the previous versions it would set the episode list to the latest unwatched episode. After the update when selecting a season it starts at episode 1. Currently I have a few YouTube series’ that are over a hundred episodes and scrolling to the episode I’m on can be rather cumbersome. Not sure if the old functionality was a bug or the new functionality is.

### Reproduction steps

Open Swiftfin, navigate to media, select a subfolder (I leverage favorites), select the show, and it will show the correct season but the first episode despite what has been watched.

### Platform

iOS

### Application version

1.5

App Store

### Device information

iPad 10th gen and iPhone 16

### OS version

iPadOS 26.5 and iOS 26.5.2

### Jellyfin server version

10.11.8

### Screenshots or videos

_No response_

### Relevant log output

```shell

```

Cyrus-Krispin:
### Summary

Series and season pages now select the earliest unplayed episode when there is no next-up or resumable item. Once the episode cards appear, the carousel scrolls to the selected episode on both iOS and tvOS.

### Testing

- Built and ran Swiftfin using the iOS 26.5 Simulator
- Verified a partially watched episode remains selected
- Verified completing an episode selects the first unplayed episode
- iOS and tvOS builds passed

# BEFORE
The selected episode is outside the visible carousel.

swiftfin-2143-before

# AFTER
The carousel scrolls to the selected episode.

swiftfin-2143-after

### DIFF
--- a/Shared/Objects/ContentGroup/SeriesEpisodeContentGroup/SeriesEpisodeContentGroup+EpisodeCollection.swift
+++ b/Shared/Objects/ContentGroup/SeriesEpisodeContentGroup/SeriesEpisodeContentGroup+EpisodeCollection.swift
@@ -127,6 +127,11 @@ extension SeriesEpisodeContentGroup {
         @FocusState
         private var focusedElement: EpisodeElement.ID?
 
+        #if os(iOS)
+        @StateObject
+        private var proxy = CollectionHStackProxy()
+        #endif
+
         let elements: [EpisodeElement]
         let preferredElementID: EpisodeElement.ID?
         let header: Header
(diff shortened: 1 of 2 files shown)


JPKribs:
We've done this before with the 100 ms delay and it has historically led to issues. Which, I believe that's why this was removed.

The way that this would need to work instead, is we would need to wait for the first appearance of the actual poster itself. The reason this would break historically is because it would attempt to scroll to a poster that does not exist yet.

The reason the 100 ms works in your testing is because SwiftUI draws the poster it needs to scroll to inside that window so that it has something that can scroll to. However, if you have slower network conditions, or a much longer season, that starts to cause issues.

Ideally, we should not have solutions that have a fixed delay, if avoidable.

I would also prefer any solution work for both iOS and tvOS.

Cyrus-Krispin:
Thanks for the context! That makes sense. I’ll continue to investigate this issue.

Cyrus-Krispin:
I reworked this based on your feedback @JPKribs. 
The carousel now waits for the actual episode card to appear before scrolling. 
(Same approach on iOS and tvOS)

Cyrus-Krispin:
Hi @JPKribs  and @LePips, when you have a chance, could you please review the latest changes? I reworked the implementation based on the earlier feedback, and it should now be ready for another look. Thanks!

LePips:
Apologies I have not taken a look at this yet. Frankly, I don't think that having this as a part be done as the stack's content is being presented is the correct fix. I've instead been trying to adjust this from the `CollectionHStack` side, where as the data is set instead it will set the proper offset.

LePips:
I've done the work as above in https://github.com/jellyfin/Swiftfin/pull/2297.

However in this PR you've also added additional work in `firstAvailableItem` for "first unplayed". That should already be captured either by resume or up next. Is there an instance when the items don't return as expected?

PR #2262, opened 2026-09-02
Fix playback of in-progress live tv recordings

nintwentydo:
Fixes "Unable to load this item" playback error on live tv recordings that are still in progress.

The item's only listed source is a placeholder which the server filters out of playback sources, so sending it returns no sources.

This fix skips the media source id when the initial source is a placeholder. 

Tested on iOS and tvOS against Jellyfin 12 RC7.


JPKribs (approved):
Thank you for your contribution! I've tested this and this solution makes sense to me.

PR #2265, opened 2026-09-02
Live item, user and library updates over the socket (#784)

### LINKED ISSUE #784: Library Update

### LINKED ISSUE #1514: Media Page doesn’t refresh, when going back from an item
### Describe the bug

1. Go to Media and select a category (e.g. Movies)
2. click on a movie
3. press on favourite / watched
4. go back 
The tag doesn’t shown up
When you go back to All Media and the click on movies the page refresh and the Tag is showing up.

### Application version

1.3 (78)

None

### Device information

iPhone 13 Pro

### OS version

18.4.1

### Jellyfin server version

10.10.7

Adavid114:
This continues the work from @JPKribs on #784 (originally in JPKribs/Swiftfin#20). I took it over from his branch and finished it, so this is one PR with the whole feature.

What it does

Adds a UserNotificationManager that works as a central hub for real time updates. It listens to the WebSocket (userData changes, library changes, user updates/deletions, server restarts) and to internal app notifications, fetches the fresh data from the server and re-emits a small set of unified notifications (didChangeItem, didDeleteItem, didChangeServerUser, didRequestGlobalRefresh, didRequestLibraryRefresh). The views subscribe to those, so marking something as watched/favorite, deleting an item, or changing a library now updates the UI live and across devices.

A 1s debounce (waitForSocket) avoids doing the same fetch twice when a manual action and the socket message arrive together.

Changes on top of the original branch

- Restore the instant local update when toggling played/favorite from the item detail (it was waiting ~1s for the socket instead of using the response we already had).
- Dismiss the item detail when its item gets deleted, for any viewer (before it only dismissed for users who had the edit menu, so a non privileged viewer got a ghost page on a remote delete).
- Scope the library refresh to the libraries that actually changed instead of a global refresh.
- Remove dead/redundant code (an unused server shutdown notification chain, a redundant observeCandidateGroups).

Notes

- fetchItems requests ItemFields.allCases right now. It works but it over fetches, could be narrowed later.
- I looked into coalescing rapid socket fetches but the server already consolidates rapid changes into a single message, so it was not needed.

Tested on a physical Apple TV and iOS.


JPKribs:
@Adavid114 Thank you for your contribution! Please resolve our merge conflicts for us to proceed.

Adavid114:
Hi @JPKribs, conflicts resolved and pushed. Thank you!

orehmane:
@Adavid114 just wanted to say thank you. While I'm able to handle the quirk, this has been driving the less technical members of my family bananas.

Adavid114:
> @Adavid114 just wanted to say thank you. While I'm able to handle the quirk, this has been driving the less technical members of my family bananas.

Thanks @orehmane ! This was also the one thing my family kept bringing up while using the app. Glad I can help, and hopefully it makes the experience better for everyone!

orehmane:
@Adavid114 just to make sure I understand properly, this will fix the issue where you stop watching something and the main screen doesn't update (so like you watch 3 episodes of something and then it still shows the one you started on as "next up"), right?

Adavid114:
@orehmane Yes exactly, that same issue. This will help to add a 'live' status on the progress of the movies/series watched.

JPKribs:
Due to this being built on my original changes, I will defer to @LePips in case our starting point is incorrect. Obviously, **I** think this was the starting point haha but he is going to be the expert on this system and making sure this is correct

PR #2268, opened 2026-09-05
Add `jellyfin` scheme for deep links

LePips:
Requested from a comment I don't have the time to find.


JPKribs (approved):
I know the comment you're talking about but I couldn't find it either. I'll tag it here if it comes up again but this looks good to me!

PR #2269, opened 2026-09-05
Final Stateful macro implementations

LePips:
Final view models migrating to Stateful macro.


JPKribs (approved):
Looks great and works as expected! Didn't realize we were this close to the full replacement

PR #2276, opened 2026-09-08
Merge cast & crew posters for people with multiple crew roles

shbernal:
**Changes**

The server sends one `People` entry per credit, so a person credited for several crew jobs on the same item gets one poster per job — a director who also produced appears twice in Cast & Crew, same photo, one saying "Director" and another saying "Producer". This combines the director, writer and producer credits belonging to the same person into a single poster listing every role, e.g. "Director / Producer". Actors are left alone, so someone playing two characters keeps a poster per character.

It follows what jellyfin-web does since https://github.com/jellyfin/jellyfin-web/pull/8209 (shipped in 12.0), using the same merged kinds — director, writer, producer, which is `PersonKind.supportedCases` minus actor, and also what the server keeps in `TmdbUtils.WantedCrewKinds` — and the same `" / "` separator.

Merging lives in `mergingCrewRoles()` on `Collection where Element == BaseItemPerson`, applied where `item.people` is read for display: the Cast & Crew poster group in `ItemContentGroupProvider` and the people supplement in `MediaPlayerManager`.

One related change worth calling out: `BaseItemPerson.subtitle` used `firstRole`, which cuts a role at the first `"/"` so that an actor's many characters stay short. That would hide everything after the first job of a combined credit, so crew subtitles now use `role` directly. For an uncombined crew credit this is identical, since job names do not contain `"/"`.

**Testing**

Being upfront: **I develop on Linux and have no macOS machine, so this has not been compiled or run.** Please treat it as unverified and build it before merging — I would rather say so than imply a test I did not do.

What supports it:

- The identical change on Android TV, where I could run it, behaves as intended against a real library: on *Scott Pilgrim vs. the World*, Edgar Wright collapses from three cards (Director, Screenplay, Producer) into one reading "Director / Screenplay / Producer", and actors are unaffected.
- The same logic on jellyfin-vue, also verified against a live server.

I am happy to iterate on anything CI flags, or to close this if you would rather implement it yourselves.

**Code assistance**

Written with Claude Code using the Opus 5 model. It audited which clients were missing the behavior and wrote the extension and the call sites. I reviewed the change myself. As noted above, no build or on-device verification was possible on my machine.

I originally hit the duplicate cards on Android TV, but the same gap exists here, in jellyfin-roku and in jellyfin-vue, so I am opening the equivalent PR against each of those rather than leaving the fix partial across clients.

**Issues**

N/A — no existing issue. Follows up on https://github.com/jellyfin/jellyfin-web/pull/8209.

### DIFF
--- a/Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto+Poster.swift
+++ b/Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto+Poster.swift
@@ -45,7 +45,7 @@ extension BaseItemDto: Poster {
         case .episode:
             seasonEpisodeLabel
         case .person:
-            people?.first?.firstRole
+            people?.first?.displayRole
         case .video:
             extraType?.displayTitle
         default:
--- a/Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto.swift
+++ b/Shared/Extensions/JellyfinAPI/BaseItemDto/BaseItemDto.swift
@@ -137,6 +137,30 @@ extension BaseItemDto {
         return L10n.episodeNumber(episodeNo)
     }
 
+    /// Merges crew credits
+    var mergedPeople: [BaseItemPerson]? {
+        guard let people else { return nil }
+
+        let crew = Dictionary(grouping: people.filter(\.isCrew), by: \.id)
+        var seen: Set<String> = []
+
+        return people.compactMap { person in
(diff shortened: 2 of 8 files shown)


LePips (approved):
I'm fine with it, just did some cleanup. I'm still preserving behavior we had before where for actors with a multi-role string we only get the first role, which seems asymmetric. I'll just wait to see what happens with the server changes.

PR #2281, opened 2026-09-10
Add a native macOS target

### LINKED ISSUE #215: Swiftfin on M1 Macs
Swiftfin can currently run on M1 Macs via TestFlight for Mac. While this is a capability of the project, this **_will not_** be actively developed for and no issues will be accepted regarding the functionality of Swiftfin in this way.

The current goal for Swiftfin is to develop an iOS and tvOS client. After this goal is readily met, the development of Swiftfin towards a macOS client may be addressed. If anything, we may want the client to mirror the tvOS client instead of the iOS client.

ComicBit:
This adds a native macOS target to the app, with a Mac-native UI and a floating Picture in Picture window.

Closes #215

## Architecture

The macOS app reuses `Shared/` and `Swiftfin/` through Xcode's synchronized folder groups rather than forking them, so there is one copy of the source and the iOS/tvOS targets are untouched. A small platform layer under `Shared/Platform/` maps the UIKit names the shared code uses (`UIImage`, `UIColor`, `UIFont`, `UIDevice`) onto their AppKit equivalents so those files compile unchanged.

`CollectionHStack` and `CollectionVGrid` are reimplemented natively for macOS because the upstream packages declare iOS and tvOS only. Everything else links the same packages as iOS and tvOS, so this adds no new dependencies.

It builds universal (arm64 and x86_64) with a macOS 15.0 deployment target, but I have only tested it on Apple Silicon.

`Swiftfin macOS` is also added to the CI build matrix.

### Library

Screenshot 2026-09-10 at 12 14 45

### Playback

The touch overlay does not translate well to a window, so macOS gets its own control bar: transport, scrubber, timestamps, audio and subtitle menus, aspect fill, Picture in Picture and full screen. Controls appear on pointer movement and hide again once the pointer goes idle.

Screenshot 2026-09-10 at 12 15 05

### Picture in Picture

This is a floating always-on-top window, not AVKit's system Picture in Picture. VLC and MPV render into an `NSView` rather than an `AVPlayerLayer` or a sample buffer layer, and AVKit cannot host those.

Screenshot 2026-09-10 at 12 19 10

## One fix outside macOS

Commit `777e307c` changes behaviour on all platforms, so it is worth reviewing separately. Resetting filters restored `ItemFilterCollection.default` instead of the filters the library was created with, so resetting on the Movies tab stopped restricting results to movies. The same comparison also made a freshly opened library look filtered, which is why the reset control was always visible. The macOS filter bar is what surfaced it, but the bug is not macOS specific.

## Testing

The macOS debug build is clean, SwiftLint `--strict` passes and `FindUnusedStrings` reports nothing unused. I have used the app on Apple Silicon for browsing, search, settings and playback.

I could not build iOS or tvOS locally, as CoreStore fails to compile with the Xcode version I have, so I am relying on CI for those. I would appreciate other contributors testing locally and reporting any issues.

## Notes for reviewers

This is a first attempt to bring Swiftfin to macOS. I think it is better to get feedback from other contributors and implement fixes for the various problems, so I would not consider this final yet. I also need to ask for iOS and tvOS testing just to be sure everything works as intended on all platforms.


ComicBit:
@LePips could you check this PR when you can? Thank you!!

LePips (changes requested):
Thank you for wanting to build something so substantive as your first contribution, but I cannot accept this work.

There are many prerequisites prior to a macOS target being ready, including but not exhaustive:
- multi-scene/window management
- collection view performance and macOS support
	- You recreate `Collection_Stack` with scroll views and stacks, which degrades performance
- PiP
	- proper PiP is not just making an overlay window
- code organization and backports
	- in some places you make proper platform dud functions (good) and not others
	- still need to hone down platform conditionals so we don't need conditionals everywhere

Those don't even go into many of the other issues, changes you're introducing, and breaking out from project code styles and practices.

Without an AI disclosure, this surely seems like it was. Can you please clarify your usage in the future?

ComicBit:
@LePips that is absolutely not a problem. Honestly I don't expect a +5K LOC PR to be approved in place. That was not the point. I would ask you kindly to provide all points/material to read in order to bring the PR to be mergeable. Also I'd ask you to give me more details about the AI Disclosure.

Best

LePips:
Please refer to Jellyfin's AI Policy.

https://jellyfin.org/docs/general/contributing/llm-policies/

ComicBit:
Why did you just close the PR? We can't work on it?

LePips:
While we state to open a PR even for draft work, as stated above there are many architectural prerequisites that need to be completed, which influence the entire project. This is also your first contribution and respectfully I'm going to request a different approach to start contributing to the project rather than something so large without understanding overall project context.

I've thought about a more progressive approach of first having macOS compilation and then allowing work to build on top off it from there, rather than one broad attempt.

Issue #2285, opened 2026-09-12
Translation Error


tobylibo:
As shown in the image, a red "alert" appears when accessing the translation website. This issue has persisted for at least a week. It is unclear whether this is an undetected issue or something planned.

<img width="1229" height="1101" alt="Image" src="https://github.com/user-attachments/assets/2a51e149-0157-49d1-a711-2bb9a4700034" />

LePips:
Thanks for pointing this out, looks like a merge conflict.

LePips:
Fixed by reverting the changes to the localization files in #2287, also added checks to prevent this in the future.

PR #2274, opened 2026-09-05
mpv player

LePips:
Closes #221

Implements experimental mpv player using [MPVUI](https://github.com/LePips/MPVUI). This is currently experimental as the package is being developed and tested, we fill out documentation (deliberately did not update here), and video player settings are finalized. Currently reuses the VLC device profiles and as we determine mpv's support we should have them consolidated as much as possible.

Also:
- renames `Swiftfin` to `VLC`
- moves video player settings to `Advanced`

### HDR
mpv brings improved HDR support compared to SwiftVLC. Unable to share screenshots here that effectively portray the differences. One may notice with their own media that the low ends may not go fully black and brights may be blown out, which I will be adjusting on the MPVUI side.

### Subtitles

MPUI can expose the current text subtitles, which allows using the device's caption styling.

<details>
<summary>Subtitles</summary>

| SwiftVLC | MPVUI |
| :---: | :---: |
| sub-swiftvlc | sub-mpv |

</details>

## AI Disclosure

Utilized 6 Astra for the media track and subtitle presentation layer.

PR #2287, opened 2026-09-12
Localized file validation on PR

LePips:
Accidentally leaked changing localization files directly in #2274, causing the weblate to break #2285. Would also help against PRs wanting to change strings directly.


LePips:
For testing the validation, I deliberately reverted the changes from #2274 to cause a failure. The changes should still be let through here to probably get the Weblate fixed.

JPKribs:
The code portion looks correct and I understand the UTF-16 check. I just want to confirm the build errors here are intentional as a revision of the MPV label changes to UTF-8?

LePips:
There isn't a utf16 check, what are you referring to for utf8? If it's about the `.gitattributes`, that is to help view `.strings` file changes in a git diff and doesn't make changes.

JPKribs:
I guess I misunderstand why we're doing the git check. I assumed from the comment we were trying to look for re-encoding.

I'm just thinking about the last time it desynced I believe was a 16 -> 8 change.

Sorry, I'm out and about, I can take a look at this more in the morning with a keener eye. Primarily I just want to confirm the current build error is intentional as a resolution from MPV?

LePips:
The `gitattributes` are not a part of the check, that's only for viewing git diffs locally on your machine.

There's more to check than just the encoding. We also need to ensure that only the English source file could ever get new strings from PRs, as all other changes come through weblate.

Yes, the build error here is intentional.

JPKribs (approved):

PR #2288, opened 2026-09-12
Nested Edit Glass

JPKribs:
### Summary

Fixes the Edit Buttons having nested button inside of glass. iOS 26+ puts buttons in glass then we are putting it inside another glass. Prominent doesn't seem to have this same issue so that was left untouched.

| Before | After |
| :---: | :---: |
| Simulator Screenshot - iPhone 17 Pro - 2026-09-12 at 12 22 42 | Simulator Screenshot - iPhone 17 Pro - 2026-09-12 at 12 24 35 |

---

tvOS seems to handle this differently so tvOS was left intact as well:

Screenshot 2026-09-12 at 12 31 08

Screenshot 2026-09-12 at 12 31 01

### DIFF
--- a/Shared/Components/ButtonStyles/BackportGlassButtonStyles.swift
+++ b/Shared/Components/ButtonStyles/BackportGlassButtonStyles.swift
@@ -92,7 +92,9 @@ private struct BackportGlassButtonStyleBody: View {
         switch prominence {
         case .standard:
             nativeButton
-                .buttonStyle(.glass)
+                #if !os(iOS)
+                    .buttonStyle(.glass)
+                #endif
         case .prominent:
             nativeButton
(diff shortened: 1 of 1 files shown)


LePips (approved):
This is a fine fix for the toolbar buttons, but technically would be a problem if we did apply `.buttonStyle(.glass)` to a button on iOS 26 (we don't).

The true fix would be to cleanup how toolbar button styling is handled, but we can do that later.

JPKribs:
Makes sense. I was also originally looking at the Apple Photos app and wanted to look at whatever these button placements were:

image

Not for most cases, but like we have some scenario where we have like an select all, cancel, & delete buttons which I think is where these placements are useful.

PR #2290, opened 2026-09-13
Separate App About Localization

oskarkasprzyk:
The same about key was used for both the app and item contexts. English can use “About” for both, but some languages need different translations, so this adds a separate aboutApp key.


LePips (approved):
Thanks, I'll follow this up with comments able to be preserved as typically we try to have the key be the camel-case of the contents of the string.

PR #2291, opened 2026-09-13
Show profile picker on tvOS cold start

Ralfino26:
### Summary

On tvOS cold start, Swiftfin currently restores the last signed-in user and goes straight to home. Switching profiles requires Settings → Switch user.

This change always clears `lastSignedInUserID` during `UserSessionManager.start()` on tvOS only, so cold start lands on the existing `SelectUserView`. Selecting a profile still opens that user’s home via the current sign-in flow.

Warm resume is unchanged (`start()` only runs when `state == .initial`). iOS launch behavior is unchanged.

### Test plan

- [ ] Cold start on tvOS with one or more saved users → profile picker appears
- [ ] Select a profile → that user’s home opens (PIN/local auth still works if required)
- [ ] Background the app, then return without killing → stay signed in
- [ ] Settings → Switch user → picker → select another profile
- [ ] Confirm iOS still auto-restores the last user (unless `signOutOnClose` is enabled)


Ralfino26:
Closing to reopen with a cleaner description.

PR #2292, opened 2026-09-13
Show profile picker on tvOS cold start

Ralfino26:
On tvOS cold start, Swiftfin restores the last signed-in user and goes straight to home, so switching profiles requires going through Settings.

This clears the last signed-in user during launch on tvOS so the existing profile picker is shown instead. Selecting a profile still opens that user's home. iOS launch behavior and warm resume are unchanged.


JPKribs:
This setting already exists for both iOS and tvOS. In fact, your change is one line above the code that already does this.

On the user selection screen, select the gear icon. Inside, select the advanced option.

Inside of advance, there are two settings. One of them logs out the user on first boot, allowing you to select the user. The other, logged the user out if the session has been in the background for a selected period of time.

I'm not near my computer or my TV, but this is what the setting name is:

image

Found from the gear on the user selection screen. The toggle name and description will be identical on tvOS.

Ralfino26:
I found it! Thank you!

Issue #886, opened 2023-10-20
Software Decoding and Nearest Neighbor Scaling


24fpsDaVinci:
**Describe the feature you'd like**
1.Toggle to enable/disable hardware acceleration
2.Option for Nearest Neighbour Scaling

**Additional context**
1.In the regular vlc app theres a toggle to enable/disable hardware acceleration, software decode may offer better visual quality.
2.Some users may prefer nearest neighbour for certain content for better visual quality.

PR #2301, opened 2026-09-17
feat(player): support client side deinterlace

hlhc:
## Summary 

Adds user-selectable **deinterlacing for the VLC player**, so interlaced content (e.g. 1080i
broadcast/TS) can be deinterlaced on-device. Previously Swiftfin passed no deinterlace option to
libVLC, so interlaced sources on Direct Play often showed combing.

By default, deinterlace is **off** to prevent older device having problem with auto. it also prevent any behavioural change of the app.

##  Known issue / follow-up (not introduced by this PR, probably, still investigating)

When testing on iOS, a weird re-render is observed when the deinterlace menu in the player actions is open, it will scroll back up (probably a re-render) after some time. Suspect it is related to `UserDefaults`.


LePips (changes requested):
Apologies, but I will not be accepting player settings like this quite yet. But as I'm narrowing down mpv, I have realized that some player settings would be desirable and I also haven't closed out #886 as we define what settings to provide. Frankly, it isn't that sensible to provide every possible setting with every possible value. I also won't have it implemented as individual settings like done here.

hlhc:
> Frankly, it isn’t that sensible to provide every possible setting with every possible value.

I agree. Exposing every possible VLC option would probably make the settings unnecessarily complicated.

> I also haven’t closed out #886 as we define what settings to provide.

For deinterlacing specifically, Jellyfin already handles it when transcoding, so this mainly matters for direct play of interlaced content. Perhaps VLC could just handle it automatically or select yadif like the server did instead of exposing individual modes for now? The only concern is changing it to auto/yadif might affect some lower end Apple TV / iPhones (which I don't have and cannot verify if it lower the performance)

> I also won’t have it implemented as individual settings like done here.

I’ll follow #886 and can adjust the implementation once the direction for player settings is clearer.

PR #2303, opened 2026-09-17
Glass button color labels and increased contrast

### LINKED ISSUE #1735: [iOS] Accent color doesn't respond to high-contrast mode
The app's accent color should change based on the high-contrast mode. It may also be good to vary it slightly between regular light and dark mode.

LePips:
- Have prominent glass buttons use the Defaults accent color and label with overlay color
- Do basic mixing based on appearance and color contrast accessibility setting
	- Just a basic mixing as recommended by Astra. Not going to look much more as I consider full accent color selection the remedy instead.
	- Closes #1735


JPKribs (approved):
Testing and IMO seems to work well.

PR #2306, opened 2026-09-18
Adjust capsule and selection button styles

LePips:
Replaces #2244

- Consolidate some of the capsule styling
- `SelectionTrack` for re-usage with supplement and season selections
- Fix navigation toolbar capsule button styling


JPKribs (approved):
Looks good. Tested and found some unrelated regression items for 1.7 I will open issues for. Tested with `supportsLiquidGlass` manually set to false and this looks as expected there as well.

Main item, I found https://github.com/jellyfin/Swiftfin/issues/2307 but this also seems to be the case on Main.

PR #2315, opened 2026-09-20
Fix double touch

llirik0:
the fix:
- pass `numberOfTouchesRequired` instead
- the normal recognizer reports `1` the two finger recognizer reports its configured value of `2`
- no gesture precedence, recognizers, settings, or player actions were otherwise changed

verification:
- reproduced the runtime difference between active and required touch counts first
- **SwiftFormat**: clean
- **SwiftLint**: 0 violations


LePips:
Frankly, I'm unable to replicate the original issue that double touch did not work at all. This change doesn't really do much, as these values should be the same for tap recognizers. You mention "runtime difference", can you describe how you saw this difference?

